PR #404, opened 2019-09-06
CI

dnfield:
pub get, build (if applicable), analysis (using tuneup), and test.

Should we throw dartfmt in?

We should also extend this to do deploys on commits to master, but that can be a follow up.

We should also probably do CI on the mobile/ directory, but that one had a ton of issues that I wasn't sure where to start with for analysis.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -4,9 +4,10 @@ container:
 app_task:
   env:
     matrix:
-      - CWD: app
+      # TODO(dnfield): Enable this.
+      # - CWD: app
       - CWD: app_dart
   pub_script: cd $CWD && pub get
-  analyze_script: cd $CWD && dartanalyzer --options analysis_options.yaml .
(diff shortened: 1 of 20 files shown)


dnfield:
Hm.  I spoke too early.  How do we run the dart2js tests?

jonahwilliams:
you have to run them in a browser `--platform=chrome`, otherwise it spawns the standalone vm

dnfield:
Hm.  They are, unfortunately, not working locally.  Do they work for you and I'm just missing something?

It's also unlikely that the image I'm using has chrome.  I'm going to skip testing these for now.

jonahwilliams:
I test in production

jonahwilliams:
One advantage about moving everything to flutter is that we could unit test most of the app using flutter-tester and skip the browser

dnfield:
Ok, excluding web tests for now.

PR #406, opened 2019-09-09
Adapt to manifest.yaml change: flaky -> new

liyuqian:
For https://github.com/flutter/flutter/pull/39856


liyuqian:
Close this for now as it seems that our default action is to
1. not use "flaky" field, 
2. revert the new test if it's not passing.

It is a little annoying (see, e.g., the 5 commits revert/reland chain https://github.com/flutter/flutter/pull/41578), but acceptable.

PR #409, opened 2019-09-11
Skia Gold support

Piinks:
### Skia Gold Support
This is part of the changes laid out in the design doc: [flutter.dev/go/golden-workflow](https://docs.google.com/document/d/1MuIUz9pyE_bBZPbtMCj3pYgkdgG4s4Egh6FMknTngKw/edit?usp=sharing)
- Support for https://github.com/flutter/flutter/pull/39853
  - incorporates new signal for detecting golden file changes
  - reminds contributors to triage results in Skia Gold after merging

`GoldenTriageMessage`:
>Nice merge! 🎉
It looks like this PR made changes to golden files. Be sure to visit [Flutter Gold](https://flutter-gold.skia.org/?query=source_type%3Dflutter) to triage the results when post-submit testing has completed. The status of these tests can be seen on the [Flutter Dashboard](https://flutter-dashboard.appspot.com/build.html).
>
>For more information about working with golden files, see the wiki page [Writing a Golden File Test for package:flutter/flutter](https://github.com/flutter/flutter/wiki/Writing-a-golden-file-test-for-package:flutter).

__TODO__
- [x] Update wiki to provide guidance on Skia Gold triage


app_dart/test/request_handlers/github_webhook_test.dart
```diff
@@ -325,6 +328,160 @@ void main() {
...
+        )
+      );
+
+      when(issuesService.listCommentsByIssue(slug, issueNumber))
+        .thenAnswer((_) => Stream<IssueComment>.value(
+          IssueComment()..body = 'some other comment',
+        ),
+      );
+
+      httpClient.request.response.body = '[{"note" : "124"}]';
```
Piinks: I need to mock this for when the request is made to skia gold for ignores. I've tried a number of different ways, but haven't been able to successfully do it. `FakeHttpClient` isn't a `Mock`, so I'm no sure how to manipulate this single response.
@dnfield, do you have any suggestions?

app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -270,6 +278,57 @@ class GithubWebhook extends RequestHandler<Body> {
     await buildBucketClient.batch(BatchRequest(requests: requests));
   }
 
+  Future<bool> _isIgnoredForGold(PullRequestEvent event) async {
```
Piinks: This is the request I need to fake for the test.

Piinks:
All green now @dnfield PTAL. :)

app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -270,6 +285,49 @@ class GithubWebhook extends RequestHandler<Body> {
...
+      final String rawResponse = await utf8.decodeStream(response);
+      final List<dynamic> ignores = jsonDecode(rawResponse);
+      for (Map<String, dynamic> ignore in ignores) {
+        if (ignore['note'].isNotEmpty &&
+          event.number.toString() == ignore['note'].split('/').last) {
+          ignored = true;
+          break;
+        }
+      }
+    } catch(_) {}
```
dnfield: What are we actually trying to catch here?  We should generally avoid catch blocks that catch all exceptions, and also avoid catch blocks that swallow exceptions.

Piinks: I figured that if there is an error making the request it should just return false. If (let us hope never) Skia Gold is down and the request fails, I wouldn't want it to have a domino effect and break the bot as a result.

dnfield: I think in that case we should just try to catch `SocketException` and maybe `HttpException`, and just print something to the log that we couldn't reach Skia Gold.

We may also want to catch `FormatException` and print out the raw data we got and rethrow in case the response format changes.

Piinks: SGTM! I'll put it together! :)

dnfield: Or maybe just catch `IOException` instead of `Socket/Http` separately.

Piinks:
> LGTM!
> 
> Will you be deploying this?

Thanks! Can do! I haven't done that before - should be fun!

PR #410, opened 2019-09-11
Flutter create project for build dashboard v2

CaseyHillers:
All of this is from the output when running flutter create with version 1.9

Updated the README to include a link to the design doc.


jonahwilliams:
The source files in this repository will need a copyright header like the other projects. These should be something like:

```
// Copyright (c) 2019 The Chromium Authors. All rights reserved.
// found in the LICENSE file.
```

jonahwilliams:
cc @dnfield , how do we hook up the unit tests from this project to CI?

app_flutter/test/widget_test.dart
```diff
@@ -0,0 +1,30 @@
+// This is a basic Flutter widget test.
```
jonahwilliams: The test file needs one too

app_flutter/web/index.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
```
jonahwilliams: You need one here in an html comment, see https://github.com/flutter/flutter/blob/master/examples/hello_world/web/index.html#L1

app_flutter/test/widget_test.dart
CaseyHillers: Added! Thank you!

app_flutter/web/index.html
CaseyHillers: Added! Thank you!

dnfield:
We should add another task for testing Flutter - with a Flutter enabled image.  Right now, the .cirrus.yml image we use only has Dart.

Alternatively, we could just switch entirely to the Flutter image and use dart from that image, which is probably just easier to think about anyway.

PR #421, opened 2019-09-17
Basic Status Grid

CaseyHillers:
Scrolls both horizontally and vertically. However, the vertical scroll is much slower than the horizontal scroll (SingleChildScrollView is quicker than the GridView scroll).


app_flutter/lib/status_grid.dart
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) 2019 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter/material.dart';
+
+import 'result_box.dart';
+
```
jonahwilliams: Add a short sentence describing what this is for

app_flutter/lib/status_grid.dart
```diff
@@ -0,0 +1,32 @@
...
+class StatusGrid extends StatelessWidget {
+  static const int taskCount = 80; // rough estimate based on existing dashboard
```
jonahwilliams: Generally good practice to add a const constructor that takes a key:

`const StatusGrid({Key key}) : super(key: key);`

app_flutter/lib/status_grid.dart
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) 2019 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter/material.dart';
+
+import 'result_box.dart';
+
+class StatusGrid extends StatelessWidget {
```
jonahwilliams: Good to add a basic smoke test. It lowers the barrier for new tests in the future.

app_flutter/lib/status_grid.dart

app_flutter/lib/status_grid.dart

app_flutter/lib/status_grid.dart
CaseyHillers: Added! Left a TODO to remove the smoke test when useful tests are added.

Issue #426, opened 2019-09-17
Commit Info


CaseyHillers:
Show information regarding a commit (author, avatar, link to commit, message).

CaseyHillers:
Since v1 does not show commit message, I am marking this as resolved.

#435 is the follow up that this issue was waiting on.

PR #427, opened 2019-09-17
Commit Widget

CaseyHillers:
This is a widget that is intended to display information from a git commit.

Added mockito dependency to get Image.network widget to work in tests.


CaseyHillers:
Adds a parent widget for #426

app_flutter/lib/commit_box.dart
```diff
@@ -0,0 +1,33 @@
...
+  const CommitBox({
+    Key key,
+    @required this.message,
+    @required this.avatarUrl,
+  }) : super(key: key);
+
+  final String message;
+  final String avatarUrl;
+
+  Widget build(BuildContext context) {
```
jonahwilliams:
```suggestion
  @override
  Widget build(BuildContext context) {
```

app_flutter/lib/commit_box.dart
```diff
@@ -0,0 +1,33 @@
...
+/// Displays Git commit information.
+class CommitBox extends StatelessWidget {
+  // TODO(chillers): convert to use commit model
+  const CommitBox({
+    Key key,
+    @required this.message,
+    @required this.avatarUrl,
+  }) : super(key: key);
+
+  final String message;
```
jonahwilliams: Nit: brieft description of what each of these does.

app_flutter/test/commit_box_test.dart
```diff
@@ -0,0 +1,86 @@
...
+
+import 'package:app_flutter/commit_box.dart';
+import 'package:mockito/mockito.dart';
+
+void main() {
+  group('CommitBox', () {
+    setUpAll(() async {
+      // Only needs to be done once since the HttpClient generated by this
+      // override is cached as a static singleton.
+      io.HttpOverrides.global = new TestHttpOverrides();
```
jonahwilliams: I think this is starting to stray into the territory of testing `Image.network` instead of testing your widget.

A good guideline would be to look at the additional code coverage that code added to a test case gives you (use flutter test --coverage). If its not giving you more lines, you might be testing someone else's code.

CaseyHillers: I wasn't happy with this approach, but otherwise the tests will fail as `Image.Network` will throw an exception for not being able to load the image.

I found [a StackOverflow on widgets tests with Image.Network](https://stackoverflow.com/a/52585057), but I figured we couldn't import that library to the project. So, I used [flutter/flutter-markdown](https://github.com/flutter/flutter_markdown/blob/a2fa53b7a29dc58b6f26b53e23a6f5a8d283dd66/test/flutter_markdown_test.dart#L506-L548) as the example on testing a widget with a network image.

Is there a better approach I should try?

jonahwilliams: You can expect the failure, that should allow the test to pass, add this line to the end of the test:
```
tester.takeException();
```

CaseyHillers: That worked. Thank you!

Issue #422, opened 2019-09-17
Status Grid


CaseyHillers:
Flutter widget that takes in results for commits, and shows them in a matrix. Should be able to scroll horizontally and vertically to show overflow.

## Milestones
- [X] Widget that can show a grid
- [X] Widget takes in List<CommitStatus> and shows it in the grid (CommitStatus imported from `app_dart/`)
- [ ] Widget is stateful and passes updates down to child widgets

CaseyHillers:
#421 creates the StatusGrid widget with the ability to scroll on both axes when there is overflow.

Work still needs to be done on passing result data to the widgete.

PR #428, opened 2019-09-18
Fit commit widget into status grid

CaseyHillers:
Removed displaying the message. This allows commit box to fit easily into the grid with no extra changes.

Added a TODO for work on the overlay that will show more information for a commit.

commit box inlined


CaseyHillers:
Inlines commit box to the status grid (for work on #422)

PR #431, opened 2019-09-19
StatusGrid using builder

CaseyHillers:
While working on improving the scroll, I realized switching the GridView to the builder format improved scrolling (not completely). Before implementing a custom scroll physics so the vertical scroll matches the horizontal scroll on the StatusGrid, I wanted to make this improvement.

Additionally, I made the grid take the rest of the screen space (full screen).

full_screen_grid

### DIFF
--- a/app_flutter/lib/status_grid.dart
+++ b/app_flutter/lib/status_grid.dart
@@ -27,7 +27,6 @@ class StatusGrid extends StatelessWidget {
         scrollDirection: Axis.horizontal,
         child: Container(
           width: columnCount * 50.0,
-          // height: MediaQuery.of(context).size.height,
           child: GridView.builder(
             // TODO(chillers): implement custom scroll physics to match horizontal scroll
             itemCount: columnCount * commitCount,
--- a/app_flutter/lib/status_grid.dart
+++ b/app_flutter/lib/status_grid.dart
@@ -28,10 +28,12 @@ class StatusGrid extends StatelessWidget {
         width: columnCount * 50.0,
         height: MediaQuery.of(context).size.height - 56,
         child: GridView.builder(
+          // TODO(chillers): implement custom scroll physics to match horizontal scroll
           itemCount: columnCount * commitCount,
           gridDelegate: SliverGridDelegateWithFixedCrossAxisCount(
               crossAxisCount: columnCount),
           itemBuilder: (BuildContext context, int index) {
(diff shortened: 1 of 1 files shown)


app_flutter/lib/status_grid.dart
```diff
@@ -24,14 +25,16 @@ class StatusGrid extends StatelessWidget {
     return SingleChildScrollView(
       scrollDirection: Axis.horizontal,
       child: Container(
-        width: taskCount * 50.0,
-        height: 500.0,
-        child: GridView.count(
+        width: columnCount * 50.0,
+        height: MediaQuery.of(context).size.height - 56,
```
jonahwilliams: I think you should be able to get the constrain system to lay this our properly with hard coding any numbers here. What happens if you place (in the parent widget, a column with a fix sized container? Like:
```
Column(children: [
  SizedBox(height: 56),
  <<ThisWidget>>
]);
```

Or does it need a specific size to work?

app_flutter/lib/status_grid.dart
```diff
@@ -24,14 +25,16 @@ class StatusGrid extends StatelessWidget {
     return SingleChildScrollView(
```
jonahwilliams: You could test that the height of this in the widget tester is the window height less 56, or similar.

app_flutter/lib/status_grid.dart
CaseyHillers: It needed a specific size to work.

Implementing the SizedBox just added 56 pixels to the overflow. Are there any other layout widgets I should try, as I don't like the idea of changing this magic number to prevent overflow.

jonahwilliams: I think the solution will likely involve a Viewport of some kind. That would essentially remove the overflow problem by allowing the "inside to be bigger", so to speak

CaseyHillers: Expanded did the trick, and made it so the Grid didn't need to explicitly define height

jonahwilliams: No overflow issues?

app_flutter/lib/status_grid.dart
CaseyHillers: Since I just used the Expanded widget, should this still be tested?

jonahwilliams: Should be good.

PR #432, opened 2019-09-19
Populate Task endTimestamp

keyonghan:
Existing cocoon fails to populate task endTimestamp accordingly, resulting 0 value for all tasks. Now we are trying to calculate task execution time, and here is the fix.

### DIFF
--- a/app_dart/lib/src/request_handlers/update_task_status.dart
+++ b/app_dart/lib/src/request_handlers/update_task_status.dart
@@ -64,7 +64,7 @@ class UpdateTaskStatus extends ApiRequestHandler<UpdateTaskStatusResponse> {
       if (task.attempts >= maxRetries) {
         task.status = Task.statusFailed;
         task.reason = 'Task failed on agent';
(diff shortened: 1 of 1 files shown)


app_dart/lib/src/request_handlers/update_task_status.dart
```diff
@@ -58,19 +58,22 @@ class UpdateTaskStatus extends ApiRequestHandler<UpdateTaskStatusResponse> {
...
         task.reason = 'Task failed on agent';
+        task.endTimestamp = endTimestamp;
       } else {
         // This will cause this task to be picked up by an agent again.
         task.status = Task.statusNew;
         task.startTimestamp = 0;
       }
     } else {
       task.status = newStatus;
+      task.endTimestamp = endTimestamp;
```
digiter: I'd prefer making the call inline: `task.endTimestamp = DateTime.now().millisecondsSinceEpoch;`.
Also applies to line 68, and remove the `endTimestamp` variable.

keyonghan: Updated.

PR #433, opened 2019-09-20
Provide proto representations of some model classes

tvolkert:


tvolkert:
This was extracted out of a work-in-progress change to add a caching layer -- it sounds like @CaseyHillers may have use for these sooner rather than later.

CaseyHillers:
LGTM

Thank you for the quick follow up!

Issue #437, opened 2019-09-20
GitHub/LUCI integration isn't cancelling builds spawned by recipe


dnfield:
The user agent and created by fields are different (as well as some of the buildset).

PR #438, opened 2019-09-20
Cancel all builds scheduled by a tryjob

dnfield:
With the new setup for LUCI builds, we schedule builds from the recipe itself - which means we can no longer expect the same user agent and created by values in all builds for a tryjob.

This updates the logic so that it will cancel the "drone" builders.

Fixes https://github.com/flutter/cocoon/issues/437


dnfield:
Yes.  I think Todd and I had the line length set to 100 or 120.  We'd just have to add a `dartfmt` to the .cirrus.yml in the root of the repo

PR #443, opened 2019-09-24
Flutter Cocoon Service

CaseyHillers:
This service allows the Flutter application to talk to the cocoon backend to get data.

It offers both a production service (talks to AppEngine via HTTP) and a mock service (generate dummy data) based on whether the application was built for production or not.

Most of the code in AppEngineCocoonService will be deleted once protos are being used in the backend.

### DIFF
--- a/app_flutter/lib/service/appengine_cocoon.dart
+++ b/app_flutter/lib/service/appengine_cocoon.dart
@@ -2,6 +2,8 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'dart:io';
+
 import 'package:cocoon_service/protos.dart' show CommitStatus;
 import 'package:http/http.dart' as http;
 
@@ -23,9 +25,14 @@ class AppEngineCocoonService implements CocoonService {
   @override
   Future<List<CommitStatus>> getStats() async {
     /// This endpoint returns a JSON [List<Agent>, List<CommitStatus>]
-    var response = await client.get('$baseApiUrl/public/get-status');
+    http.Response response = await client.get('$baseApiUrl/public/get-status');
 
-    return _jsonDecodeCommitStatuses(jsonDecode(response.body)[1]);
+    if (response.statusCode != 200) {
+      throw new HttpException(
+          '$baseApiUrl/public/get-status returned ${response.statusCode}');
+    }
+
+    return _jsonDecodeCommitStatuses(jsonDecode(response.body));
   }
 
   List<CommitStatus> _jsonDecodeCommitStatuses(List<String> pieces) {
--- a/app_flutter/lib/service/appengine_cocoon.dart
+++ b/app_flutter/lib/service/appengine_cocoon.dart
@@ -30,7 +30,7 @@ class AppEngineCocoonService implements CocoonService {
     http.Response response = await client.get('$baseApiUrl/public/get-status');
 
     if (response.statusCode != 200) {
-      throw new HttpException(
+      throw HttpException(
           '$baseApiUrl/public/get-status returned ${response.statusCode}');
     }
 
--- a/app_flutter/lib/service/appengine_cocoon.dart
+++ b/app_flutter/lib/service/appengine_cocoon.dart
@@ -8,7 +8,8 @@ import 'dart:convert';
 import 'package:http/http.dart' as http;
 import 'package:fixnum/fixnum.dart';
 
-import 'package:cocoon_service/protos.dart' show Commit, CommitStatus, Stage;
+import 'package:cocoon_service/protos.dart'
+    show Commit, CommitStatus, Stage, Task;
 
 import 'cocoon.dart';
 
@@ -36,36 +37,80 @@ class AppEngineCocoonService implements CocoonService {
     Map<String, dynamic> jsonResponse = jsonDecode(response.body);
     assert(jsonResponse != null);
 
-    return _jsonDecodeCommitStatuses(jsonResponse['Statuses']);
+    return _commitStatusesFromJson(jsonResponse['Statuses']);
   }
 
-  List<CommitStatus> _jsonDecodeCommitStatuses(
-      List<dynamic> jsonCommitStatuses) {
+  List<CommitStatus> _commitStatusesFromJson(List<dynamic> jsonCommitStatuses) {
     assert(jsonCommitStatuses != null);
     // TODO(chillers): Remove adapter code to just use proto fromJson method. https://github.com/flutter/cocoon/issues/441
 
     List<CommitStatus> statuses = List();
 
     jsonCommitStatuses.forEach((jsonCommitStatus) {
-      var jsonCommit = jsonCommitStatus['Checklist']['Checklist'];
-      assert(jsonCommit != null);
-
-      Commit commit = Commit()
-        ..timestamp = Int64() + jsonCommit['CreateTimestamp']
-        ..sha = jsonCommit['Commit']['Sha']
-        ..author = jsonCommit['Commit']['Author']['Login']
-        ..authorAvatarUrl = jsonCommit['Commit']['Author']['avatar_url']
-        ..repository = jsonCommit['FlutterRepositoryPath'];
-
-      List<dynamic> stagePieces = jsonCommitStatus['Stages'];
-      List<Stage> stages = List();
-      stagePieces.map((piece) => stages.add(Stage.fromJson(piece)));
-
       statuses.add(CommitStatus()
-        ..commit = commit
-        ..stages.addAll(stages));
+        ..commit = _commitFromJson(jsonCommitStatus['Checklist']['Checklist'])
+        ..stages.addAll(_stagesFromJson(jsonCommitStatus['Stages'])));
     });
 
     return statuses;
+  }
+
+  Commit _commitFromJson(Map<String, dynamic> jsonCommit) {
+    assert(jsonCommit != null);
+
+    return Commit()
+      ..timestamp = Int64() + jsonCommit['CreateTimestamp']
+      ..sha = jsonCommit['Commit']['Sha']
+      ..author = jsonCommit['Commit']['Author']['Login']
+      ..authorAvatarUrl = jsonCommit['Commit']['Author']['avatar_url']
+      ..repository = jsonCommit['FlutterRepositoryPath'];
+  }
+
+  List<Stage> _stagesFromJson(List<dynamic> json) {
+    List<Stage> stages = List();
+
+    json.forEach((jsonStage) => stages.add(_stageFromJson(jsonStage)));
+
+    return stages;
+  }
+
+  Stage _stageFromJson(Map<String, dynamic> json) {
+    assert(json != null);
+
+    return Stage()
+      ..name = json['Name']
+      ..tasks.addAll(_tasksFromJson(json['Tasks']))
+      ..taskStatus = json['Status'];
+  }
+
+  List<Task> _tasksFromJson(List<dynamic> json) {
+    List<Task> tasks = List();
+
+    json.forEach((jsonTask) => tasks.add(_taskFromJson(jsonTask['Task'])));
+
+    return tasks;
+  }
+
+  Task _taskFromJson(Map<String, dynamic> json) {
+    assert(json != null);
+
+    List<String> requiredCapabilities = List();
+    List<dynamic> dynamicRequiredCapabilities = json['RequiredCapabilities'];
+    dynamicRequiredCapabilities.forEach((dynamicCapability) =>
+        requiredCapabilities.add(dynamicRequiredCapabilities.toString()));
+
+    return Task()
+      ..createTimestamp = Int64(json['CreateTimestamp'])
+      ..startTimestamp = Int64(json['StartTimestamp'])
+      ..endTimestamp = Int64(json['EndTimestamp'])
+      ..name = json['Name']
+      ..attempts = json['Attempts']
+      ..isFlaky = json['Flaky']
+      ..timeoutInMinutes = json['TimeoutInMinutes']
+      ..reason = json['Reason']
+      ..requiredCapabilities.addAll(requiredCapabilities)
+      ..reservedForAgentId = json['ReservedForAgentID']
+      ..stageName = json['StageName']
+      ..status = json['Status'];
   }
 }
--- a/app_flutter/lib/service/appengine_cocoon.dart
+++ b/app_flutter/lib/service/appengine_cocoon.dart
@@ -22,12 +22,18 @@ class AppEngineCocoonService implements CocoonService {
   /// This is the base for all API requests to cocoon
   static const _baseApiUrl = 'https://flutter-dashboard.appspot.com/api';
 
-  http.Client client = http.Client();
+  http.Client _client;
+
+  /// Creates a new [AppEngineCocoonService].
+  ///
+  /// If a [client] is not specified, a new [http.Client] instance is created.
+  AppEngineCocoonService({http.Client client}) :
+    _client = client ?? http.Client();
 
   @override
   Future<List<CommitStatus>> fetchCommitStatuses() async {
     /// This endpoint returns JSON [List<Agent>, List<CommitStatus>]
-    http.Response response = await client.get('$_baseApiUrl/public/get-status');
+    http.Response response = await _client.get('$_baseApiUrl/public/get-status');
 
     if (response.statusCode != HttpStatus.ok) {
       throw HttpException(
--- a/app_flutter/lib/service/appengine_cocoon.dart
+++ b/app_flutter/lib/service/appengine_cocoon.dart
@@ -9,18 +9,27 @@ import 'dart:convert';
 
 import 'cocoon.dart';
 
+/// CocoonService for interacting with flutter/flutter production build data.
+///
+/// This queries API endpoints that are hosted on AppEngine.
 class AppEngineCocoonService implements CocoonService {
   /// The Cocoon API endpoint to query
+  ///
+  /// This is the base for all API requests to cocoon
   static const baseApiUrl = 'https://flutter-dashboard.appspot.com/api';
 
   @override
   Future<List<CommitStatus>> getStats() async {
+    /// This endpoint returns a JSON [List<Agent>, List<CommitStatus>]
     var response = await http.get('$baseApiUrl/public/get-status');
 
+    return _jsonDecodeCommitStatuses(jsonDecode(response.body)[1]);
+  }
+
+  List<CommitStatus> _jsonDecodeCommitStatuses(List<String> pieces) {
     List<CommitStatus> statuses = List();
 
-    List<dynamic> responsePiece = jsonDecode(response.body);
-    responsePiece.map((piece) => statuses.add(CommitStatus.fromJson(piece)));
+    pieces.map((piece) => statuses.add(CommitStatus.fromJson(piece)));
 
     return statuses;
   }
--- a/app_flutter/lib/service/appengine_cocoon.dart
+++ b/app_flutter/lib/service/appengine_cocoon.dart
@@ -18,10 +18,12 @@ class AppEngineCocoonService implements CocoonService {
   /// This is the base for all API requests to cocoon
   static const baseApiUrl = 'https://flutter-dashboard.appspot.com/api';
 
+  http.Client client = http.Client();
+
   @override
   Future<List<CommitStatus>> getStats() async {
     /// This endpoint returns a JSON [List<Agent>, List<CommitStatus>]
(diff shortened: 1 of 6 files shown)


app_flutter/lib/service/cocoon.dart
```diff
@@ -0,0 +1,33 @@
...
+/// Service class for interacting with flutter/flutter build data.
+///
+/// This service exists as a common interface for getting build data from a data source.
+abstract class CocoonService {
+  /// Creates a new [CocoonService] based on if the Flutter app is in production.
+  ///
+  /// Production uses the Cocoon backend running on AppEngine.
+  /// Otherwise, it uses fake data populated from a mock service.
+  factory CocoonService() {
+    const bool isProduction = bool.fromEnvironment('dart.vm.product');
```
jonahwilliams: I would prefer to use https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/foundation/constants.dart#L13

app_flutter/lib/service/mock_cocoon.dart
```diff
@@ -0,0 +1,70 @@
+import 'package:fixnum/fixnum.dart';
```
jonahwilliams: Missing license header

app_flutter/lib/service/mock_cocoon.dart
```diff
@@ -0,0 +1,70 @@
...
+
+import 'package:cocoon_service/protos.dart'
+    show Commit, CommitStatus, Stage, Task;
+
+import 'cocoon.dart';
+
+class MockCocoonService implements CocoonService {
```
jonahwilliams: This class doesn't use Mockito, so it is a Fake and not a Mock - it should be named appropriately.

app_flutter/pubspec.yaml
```diff
@@ -34,6 +34,8 @@ dependencies:
 dev_dependencies:
   flutter_test:
     sdk: flutter
+
```
jonahwilliams: You can use the test methods exported from flutter_test, no need to add a package:test dependency.

app_flutter/test/service/appengine_cocoon_test.dart
```diff
@@ -0,0 +1,127 @@
...
+
+import 'package:http/http.dart' show Response;
+import 'package:http/testing.dart';
+import 'package:fixnum/fixnum.dart';
+
+void main() {
+  // This is based off data the Cocoon backend sends out from v1.
+  // It doesn't map directly to protos since the backend does
+  // not use protos yet.
+  final String jsonGetStatsResponse = """
```
jonahwilliams: I would move this up to a top level const in the file

app_flutter/test/service/appengine_cocoon_test.dart
```diff
@@ -0,0 +1,127 @@
...
+        }
+  """;
+
+  group('AppEngine CocoonService', () {
+    AppEngineCocoonService service;
+
+    setUp(() async {
+      service = AppEngineCocoonService();
+      service.client = MockClient((request) {
+        return Future<Response>.delayed(Duration(microseconds: 100),
```
jonahwilliams: Don't add artificial delays to test scipts.You can make this return a future by making the method body `async`.

app_flutter/lib/service/appengine_cocoon.dart
```diff
@@ -0,0 +1,118 @@
...
+
+  http.Client client = http.Client();
+
+  @override
+  Future<List<CommitStatus>> getStats() async {
+    /// This endpoint returns JSON [List<Agent>, List<CommitStatus>]
+    http.Response response = await client.get('$baseApiUrl/public/get-status');
+
+    if (response.statusCode != 200) {
+      throw new HttpException(
```
jonahwilliams:
```suggestion
      throw HttpException(
```

app_flutter/lib/service/appengine_cocoon.dart
```diff
@@ -0,0 +1,118 @@
...
+  static const baseApiUrl = 'https://flutter-dashboard.appspot.com/api';
+
+  http.Client client = http.Client();
+
+  @override
+  Future<List<CommitStatus>> getStats() async {
+    /// This endpoint returns JSON [List<Agent>, List<CommitStatus>]
+    http.Response response = await client.get('$baseApiUrl/public/get-status');
+
+    if (response.statusCode != 200) {
```
jonahwilliams: There should be a const `HttpStatus.ok` in `dart:io`

app_flutter/lib/service/appengine_cocoon.dart
```diff
@@ -0,0 +1,118 @@
...
+      throw new HttpException(
+          '$baseApiUrl/public/get-status returned ${response.statusCode}');
+    }
+
+    Map<String, dynamic> jsonResponse = jsonDecode(response.body);
```
jonahwilliams: Use `Map<String, Object>` to avoid accidental dynamic

app_flutter/lib/service/appengine_cocoon.dart
```diff
@@ -0,0 +1,118 @@
...
+    assert(jsonResponse != null);
```
jonahwilliams: I'm not sure if `jsonDecode` can actually return null. If it encounters an error I believe it throws a FormatException

app_flutter/lib/service/appengine_cocoon.dart
```diff
@@ -0,0 +1,118 @@
...
+
+    return _commitStatusesFromJson(jsonResponse['Statuses']);
+  }
+
+  List<CommitStatus> _commitStatusesFromJson(List<dynamic> jsonCommitStatuses) {
```
jonahwilliams:
```suggestion
  List<CommitStatus> _commitStatusesFromJson(List<Object> jsonCommitStatuses) {
```

app_flutter/lib/service/appengine_cocoon.dart
```diff
@@ -0,0 +1,118 @@
...
+    assert(jsonCommitStatuses != null);
+    // TODO(chillers): Remove adapter code to just use proto fromJson method. https://github.com/flutter/cocoon/issues/441
+
+    List<CommitStatus> statuses = List();
```
jonahwilliams: Use literals:

```suggestion
    List<CommitStatus> statuses = <CommitStatus>[];
```

app_flutter/lib/service/appengine_cocoon.dart
```diff
@@ -0,0 +1,118 @@
...
+
+    jsonCommitStatuses.forEach((jsonCommitStatus) {
```
jonahwilliams: Generally prefer using a `for` loop instead of forEach:

```suggestion
    for (var jsonCommitStatus in jsonCommitStatuses)
```

app_flutter/lib/service/appengine_cocoon.dart
```diff
@@ -0,0 +1,118 @@
...
+      statuses.add(CommitStatus()
+        ..commit = _commitFromJson(jsonCommitStatus['Checklist']['Checklist'])
+        ..stages.addAll(_stagesFromJson(jsonCommitStatus['Stages'])));
+    });
+
+    return statuses;
+  }
+
+  Commit _commitFromJson(Map<String, dynamic> jsonCommit) {
```
jonahwilliams:
```suggestion
  Commit _commitFromJson(Map<String, Object> jsonCommit) {
```

app_flutter/lib/service/appengine_cocoon.dart
```diff
@@ -0,0 +1,118 @@
...
+
+    return Commit()
+      ..timestamp = Int64() + jsonCommit['CreateTimestamp']
+      ..sha = jsonCommit['Commit']['Sha']
+      ..author = jsonCommit['Commit']['Author']['Login']
+      ..authorAvatarUrl = jsonCommit['Commit']['Author']['avatar_url']
+      ..repository = jsonCommit['FlutterRepositoryPath'];
+  }
+
+  List<Stage> _stagesFromJson(List<dynamic> json) {
```
jonahwilliams:
```suggestion
  List<Stage> _stagesFromJson(List<Object> json) {
```

app_flutter/lib/service/fake_cocoon.dart
```diff
@@ -0,0 +1,74 @@
...
+// found in the LICENSE file.
+
+import 'package:fixnum/fixnum.dart';
+
+import 'package:cocoon_service/protos.dart'
+    show Commit, CommitStatus, Stage, Task;
+
+import 'cocoon.dart';
+
+class FakeCocoonService implements CocoonService {
```
jonahwilliams: Class should have documentation

app_flutter/lib/service/fake_cocoon.dart
```diff
@@ -0,0 +1,74 @@
...
+  @override
+  Future<List<CommitStatus>> getStats() {
+    return Future.delayed(Duration(seconds: 1), () => _getFakeStats());
+  }
+
+  List<CommitStatus> _getFakeStats() {
+    List<CommitStatus> stats = List();
```
jonahwilliams: Use a literal: 

```suggestion
    List<CommitStatus> stats = <CommitStatus>[];
```

app_flutter/lib/service/cocoon.dart
CaseyHillers: Thanks for sharing!

app_flutter/lib/service/mock_cocoon.dart
CaseyHillers: Good catch!

app_flutter/lib/service/mock_cocoon.dart
CaseyHillers: Thank you! I just got a good idea on the difference between the two :)

app_flutter/lib/service/fake_cocoon.dart
```diff
@@ -0,0 +1,74 @@
...
+
+import 'package:cocoon_service/protos.dart'
+    show Commit, CommitStatus, Stage, Task;
+
+import 'cocoon.dart';
+
+class FakeCocoonService implements CocoonService {
+  @override
+  Future<List<CommitStatus>> getStats() {
+    return Future.delayed(Duration(seconds: 1), () => _getFakeStats());
```
jonahwilliams: Don't induce artificial delays in the code. Try using the `Future.value` constructor:

```suggestion
    return Future.value( _getFakeStats());
```

app_flutter/pubspec.yaml
CaseyHillers: Switching to flutter_test caused the TypeMatcher test to fail.

```
    test('should return List<CommitStatus>', () {
      expect(service.getStats(), TypeMatcher<Future<List<CommitStatus>>>());
    });
```

Fails with:

```
  Expected: <Instance of 'TypeMatcher<Future<List<CommitStatus>>>'>
    Actual: <Instance of 'Future<List<CommitStatus>>'>
```

Do you have any idea why?

app_flutter/lib/service/cocoon.dart
```diff
@@ -0,0 +1,34 @@
...
+
+    // TODO(chillers): LocalCocoonService. https://github.com/flutter/cocoon/issues/442
+
+    return FakeCocoonService();
+  }
+
+  /// Gets build information from the last 200 commits.
+  ///
+  /// TODO(chillers): Make configurable to get range of commits
+  Future<List<CommitStatus>> getStats();
```
jonahwilliams: A general nit on API design - use `get` as a verb of last resort. This also downloads a commit status but as named as returning stats - these should probably match

```suggestion
  Future<List<CommitStatus>> fetchCommitStatuses();
```

fetch is nearly as generic as get, so that isn't a huge improvement either

app_flutter/lib/service/fake_cocoon.dart
```diff
@@ -0,0 +1,74 @@
...
+        ..commit = commit
+        ..stages.addAll(_getFakeStages(i, commit));
+
+      stats.add(status);
+    }
+
+    return stats;
+  }
+
+  Commit _getFakeCommit(int index, int baseTimestamp) {
```
jonahwilliams: I wouldn't name these methods with `get` because they are actually instantiating the Commit object:

```suggestion
  Commit _createFakeCommit(int index, int baseTimestamp) {
```

app_flutter/test/service/appengine_cocoon_test.dart
CaseyHillers: Done

app_flutter/test/service/appengine_cocoon_test.dart
CaseyHillers: This is the best thing I learned today :+1:

app_flutter/pubspec.yaml
jonahwilliams: Ahh, TypeMatcher was already a type in Flutter when it was added to package test, so that import is hidden in flutter_test. To use it you have to import package:test directly as you were doing - carry on!

app_flutter/lib/service/appengine_cocoon.dart
CaseyHillers: Done

app_flutter/lib/service/appengine_cocoon.dart
CaseyHillers: Thank you!

app_flutter/lib/service/fake_cocoon.dart
CaseyHillers: Done

app_flutter/lib/service/fake_cocoon.dart
CaseyHillers: Done

app_flutter/lib/service/fake_cocoon.dart
CaseyHillers: That's a good point, thank you for sharing. Done!

app_flutter/lib/service/fake_cocoon.dart
CaseyHillers: Added

app_flutter/lib/service/appengine_cocoon.dart
CaseyHillers: Is this a part of the style guide?

app_flutter/lib/service/appengine_cocoon.dart
CaseyHillers: Done

app_flutter/lib/service/appengine_cocoon.dart
CaseyHillers: Since these are all `Map<String, Object>` would it be better to expand `List<Object>` to be `List<Map<String, Object>>`?

app_flutter/lib/service/appengine_cocoon.dart
CaseyHillers: Okay, removed.

app_flutter/lib/service/cocoon.dart
CaseyHillers: Thank you!

app_flutter/lib/service/appengine_cocoon.dart
jonahwilliams: You would think so, but this is a place where Dart's reified generics make life a little trickier.

The JSON decoder which instantiates this List/map can't know or enforce the type of all the objects that get place inside of it, so it creates a List<Object> (these are the same as List<dynamic> or List<void> technically). This actually inserts a marker into the object which encodes the values of the generic. So at runtime, this is more like `ListObject` and not `List<Object>`, if you follow me.

This can cause unexpected problems with runtime type checks. A simpler demonstration:

```
List<Object> xs = <Object>[];
xs.add('a'); // OK
xs.add(1); // OK

List<int> ys = <int>[];
ys.add(1); // OK
List<Object> zs = ys; // OK
zs.add('a'); // FAILURE!
```

app_flutter/lib/service/appengine_cocoon.dart
jonahwilliams: https://dart.dev/guides/language/effective-dart/usage#avoid-using-iterableforeach-with-a-function-literal

app_flutter/lib/service/appengine_cocoon.dart
```diff
@@ -0,0 +1,123 @@
...
+import 'cocoon.dart';
+
+/// CocoonService for interacting with flutter/flutter production build data.
+///
+/// This queries API endpoints that are hosted on AppEngine.
+class AppEngineCocoonService implements CocoonService {
+  /// The Cocoon API endpoint to query
+  ///
+  /// This is the base for all API requests to cocoon
+  static const baseApiUrl = 'https://flutter-dashboard.appspot.com/api';
```
jonahwilliams: These fields could probably be private

app_flutter/test/service/appengine_cocoon_test.dart
```diff
@@ -0,0 +1,125 @@
...
+            ..isFlaky = false
+            ..timeoutInMinutes = 0
+            ..reason = ''
+            ..requiredCapabilities.add('[linux/android]')
+            ..reservedForAgentId = 'linux2'
+            ..stageName = 'devicelab'
+            ..status = 'Succeeded'));
+
+      expect(statuses.length, 1);
+      expect(statuses.elementAt(0), expectedStatus);
```
jonahwilliams:
```suggestion
      expect(statuses.first, expectedStatus);
```

app_flutter/test/service/appengine_cocoon_test.dart
```diff
@@ -0,0 +1,125 @@
...
+          TypeMatcher<Future<List<CommitStatus>>>());
+    });
+
+    test('should return expected List<CommitStatus>', () async {
+      List<CommitStatus> statuses = await service.fetchCommitStatuses();
+
+      CommitStatus expectedStatus = CommitStatus()
+        ..commit = (Commit()
+          ..timestamp = Int64() + 123456789
+          ..sha = 'ShaShankHash'
```
jonahwilliams: Ha!

app_flutter/lib/service/appengine_cocoon.dart
CaseyHillers: Thank you for the thoughtful explanation! I also read https://dart.dev/guides/language/sound-dart#type-inference which gave some more insight to what's technically going on.

app_flutter/lib/service/appengine_cocoon.dart
CaseyHillers: baseApiUrl can, but http.Client needs to be accessible for tests to inject a mock

app_flutter/test/service/appengine_cocoon_test.dart
CaseyHillers: Thanks!

app_flutter/lib/service/appengine_cocoon.dart
jonahwilliams: I also edited my comment to fix `<>` from being formatted out...

app_flutter/lib/service/appengine_cocoon.dart
jonahwilliams: Rather than having a setter, I would make the constructor take an optional `http.Client` argument, defaulting to the regular client if one isn't provided. Since you're testing the implementation, you should still be able to inject the right value:

```
class AppEngineCocoonService {

  /// Creates a new [AppEngineCocoonService].
  ///
  /// If a [client] is not specified, a new [http.Client] instances is created.
  AppEngineCocoonService({http.Client client})
    _client = client ?? http.Client();
}

```

jonahwilliams: This reduces the API surface, but still lets you test with a fake implementation.

CaseyHillers: Thank you! I like the simple solution :)

app_flutter/lib/service/appengine_cocoon.dart
```diff
@@ -0,0 +1,129 @@
...
+/// CocoonService for interacting with flutter/flutter production build data.
+///
+/// This queries API endpoints that are hosted on AppEngine.
+class AppEngineCocoonService implements CocoonService {
+  /// The Cocoon API endpoint to query
+  ///
+  /// This is the base for all API requests to cocoon
+  static const _baseApiUrl = 'https://flutter-dashboard.appspot.com/api';
+
+  http.Client _client;
```
jonahwilliams:
```suggestion
  final http.Client _client;
```

app_flutter/lib/service/mock_cocoon.dart
digiter: See more at https://martinfowler.com/bliki/TestDouble.html

PR #447, opened 2019-09-26
Commit overlay able to redirect to commit on Github

CaseyHillers:


app_flutter/test/commit_box_test.dart
```diff
@@ -71,5 +72,50 @@ void main() {
       // Image.Network throws a 400 exception in tests
       tester.takeException();
     });
+
+    testWidgets('tapping redirect button should redirect to Github',
```
digiter: Is it possible to use a fake or mock url_launcher in this test? @jonahwilliams

jonahwilliams: That would be the setMockMethodCallHandler below

goderbauer: Isn't that what the setMockMethodCallHandler is for?

CaseyHillers: If it helps, I just did what the plugin does :)
https://github.com/flutter/plugins/blob/master/packages/url_launcher/test/url_launcher_test.dart

app_flutter/test/commit_box_test.dart
```diff
@@ -71,5 +72,50 @@ void main() {
...
+        log.add(methodCall);
+      });
+
+      await tester.pumpWidget(MaterialApp(
+        home: CommitBox(
+          commit: expectedCommit,
+        ),
+      ));
+
+      tester.takeException();
```
goderbauer: Why are we ignoring exceptions here and below? Maybe add a comment with a quick explanation or (better:) add an expect to assert that the exception returned is actually what you expect it to be.

app_flutter/test/commit_box_test.dart
digiter: They are not the same. Using MockMethodCallHandler is testing the implementation of url_handler, which is worse than testing the interface. With a fake url_handler, the test can be independent of the implementation.

app_flutter/test/commit_box_test.dart
CaseyHillers: Removed, it was left over :)

app_flutter/test/commit_box_test.dart
CaseyHillers: I agree on this testing the implementation too much, but this is a limitation of the url_launcher plugin and is out of scope for this project. Right now this test ensures the functionality and may need updates if url_launcher updates. Other than that, I don't have a solution that can ensure the functionality is still working.

Should I file a feature request to the plugin for better testing support @digiter ?

digiter: I agree this is out of scope. I just want to learn the typical approach from others.

PR #449, opened 2019-09-27
StatusGrid refreshes to show the latest information

### LINKED ISSUE #422: Status Grid

CaseyHillers:
Implemented the Provider package to handle the state management for the new build dashboard. Created a state class for the build dashboard. Once feature parity is reached with the current dashboard, it will contain more variables (e.g. tree build status, agent statuses).

Created a new wrapper widget for the StatusGrid that handles the state management section for now. If we find this creates performance issues, we can refactor it.

This fixes #422 since the feature parity is met. The dashboard loads production data in, and updates it with the latest information on a set interval.

## Tested
I did a manual test by building the app for release on my phone, and seeing if tasks that were running went to a different state.

### DIFF
--- a/app_flutter/lib/state/flutter_build.dart
+++ b/app_flutter/lib/state/flutter_build.dart
@@ -18,17 +18,32 @@ class FlutterBuildState extends ChangeNotifier {
   /// How often to query the Cocoon backend for the current build state.
   final Duration _refreshRate = Duration(seconds: 10);
 
+  /// Timer that calls [_fetchBuildStatusUpdate] on a set interval.
+  Timer _refreshTimer;
+
   /// The current status of the commits loaded.
   List<CommitStatus> statuses = [];
 
   /// Start a fixed interval loop that fetches build state updates based on [_refreshRate].
   void startFetchingBuildStateUpdates() async {
-    Timer.periodic(_refreshRate, (t) => _fetchBuildStatusUpdate());
+    if (_refreshTimer != null) {
+      throw 'already fetching build state updates';
+    }
+
+    _refreshTimer =
+        Timer.periodic(_refreshRate, (t) => _fetchBuildStatusUpdate());
   }
 
   /// Request the latest [statuses] from [CocoonService].
   void _fetchBuildStatusUpdate() async {
     statuses = await _cocoonService.fetchCommitStatuses();
     notifyListeners();
+  }
+
+  @override
+  void dispose() {
+    super.dispose();
+
+    _refreshTimer?.cancel();
   }
 }
--- a/app_flutter/lib/state/flutter_build.dart
+++ b/app_flutter/lib/state/flutter_build.dart
@@ -1,0 +1,18 @@
+// Copyright 2019 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:app_flutter/service/cocoon.dart';
+import 'package:flutter/widgets.dart';
+
+import 'package:cocoon_service/protos.dart' show CommitStatus;
+
+class FlutterBuildState extends ChangeNotifier {
+  final CocoonService _cocoonService = CocoonService();
+
+  List<CommitStatus> statuses;
+
+  void fetchBuildStatusUpdate() async {
+    _cocoonService.fetchCommitStatuses();
+  }
+}
--- a/app_flutter/lib/state/flutter_build.dart
+++ b/app_flutter/lib/state/flutter_build.dart
@@ -13,10 +13,11 @@ import '../service/cocoon.dart';
 /// State for the Flutter Build Dashboard
 class FlutterBuildState extends ChangeNotifier {
   /// Cocoon backend service that retrieves the data needed for this state.
-  final CocoonService _cocoonService = CocoonService();
+  final CocoonService _cocoonService;
 
   /// How often to query the Cocoon backend for the current build state.
-  final Duration _refreshRate = Duration(seconds: 10);
+  @visibleForTesting
+  final Duration refreshRate = Duration(seconds: 10);
 
   /// Timer that calls [_fetchBuildStatusUpdate] on a set interval.
   @visibleForTesting
@@ -25,7 +26,13 @@ class FlutterBuildState extends ChangeNotifier {
   /// The current status of the commits loaded.
   List<CommitStatus> statuses = [];
 
-  /// Start a fixed interval loop that fetches build state updates based on [_refreshRate].
+  /// Creates a new [FlutterBuildState].
+  ///
+  /// If [CocoonService] is not specified, a new [CocoonService] instance is created.
+  FlutterBuildState({CocoonService cocoonService})
+      : _cocoonService = cocoonService ?? CocoonService();
+
+  /// Start a fixed interval loop that fetches build state updates based on [refreshRate].
   void startFetchingBuildStateUpdates() async {
     if (refreshTimer != null) {
       // There's already an update loop, no need to make another.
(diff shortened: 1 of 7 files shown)


app_flutter/lib/state/flutter_build.dart
```diff
@@ -0,0 +1,34 @@
...
+import 'dart:async';
+
+import 'package:flutter/foundation.dart';
+
+import 'package:cocoon_service/protos.dart' show CommitStatus;
+
+import '../service/cocoon.dart';
+
+/// State for the Flutter Build Dashboard
+class FlutterBuildState extends ChangeNotifier {
```
jonahwilliams: This class should have a `dispose` method that cancels the periodic timer. Then in the _BuildStatusPageState, you can call that method in the widget's `dispose`

app_flutter/lib/status_grid.dart
```diff
@@ -3,43 +3,61 @@
...
 
+import 'state/flutter_build.dart';
 import 'commit_box.dart';
 import 'task_box.dart';
 
+/// Container that manages the layout and data handling for [StatusGrid].
+///
+/// If there's no data for [StatusGrid], it shows [CircularProgressIndicator].
+class StatusGridContainer extends StatelessWidget {
+  @override
```
jonahwilliams: This class can have a `const` constructor

app_flutter/lib/status_grid.dart
```diff
@@ -3,43 +3,61 @@
...
     show Commit, CommitStatus, Stage, Task;
 
+import 'state/flutter_build.dart';
 import 'commit_box.dart';
 import 'task_box.dart';
 
+/// Container that manages the layout and data handling for [StatusGrid].
+///
+/// If there's no data for [StatusGrid], it shows [CircularProgressIndicator].
+class StatusGridContainer extends StatelessWidget {
```
jonahwilliams: Are there existing tests that cover this class? Would like to have a test that asserts that the progress indicator display when status is empty is correct

app_flutter/lib/state/flutter_build.dart
CaseyHillers: Added

app_flutter/lib/status_grid.dart
CaseyHillers: Yes, they're in `status_grid_test.dart` and cover that case. Should they be moved to `status_grid_container_test.dart`?

app_flutter/lib/state/flutter_build.dart
```diff
@@ -0,0 +1,49 @@
...
+  /// Timer that calls [_fetchBuildStatusUpdate] on a set interval.
+  Timer _refreshTimer;
+
+  /// The current status of the commits loaded.
+  List<CommitStatus> statuses = [];
+
+  /// Start a fixed interval loop that fetches build state updates based on [_refreshRate].
+  void startFetchingBuildStateUpdates() async {
+    if (_refreshTimer != null) {
+      throw 'already fetching build state updates';
```
jonahwilliams: While it's not a compile error to throw a String, generally you should only throw `Exception` or `Error`. That said, I don't think this should throw at all, just return if it is already set.

app_flutter/lib/status_grid.dart

app_flutter/lib/state/flutter_build.dart
CaseyHillers: Done. I modified the test to check if the `refreshTimer` is the same across multiple start calls.

Thanks for sharing the use about `Exception` and `Error`. I saw the backend doing it that way so I assumed that was the better practice :)

app_flutter/lib/status_grid.dart
jonahwilliams: Nah, that is fine

app_flutter/lib/state/flutter_build.dart
```diff
@@ -0,0 +1,51 @@
...
+      return;
+    }
+
+    refreshTimer =
+        Timer.periodic(_refreshRate, (t) => _fetchBuildStatusUpdate());
+  }
+
+  /// Request the latest [statuses] from [CocoonService].
+  void _fetchBuildStatusUpdate() async {
+    statuses = await _cocoonService.fetchCommitStatuses();
```
jonahwilliams: Would be nice to test that this gets called without waiting 10 seconds

app_flutter/lib/state/flutter_build.dart
```diff
@@ -0,0 +1,50 @@
...
+    notifyListeners();
+  }
```
digiter: What happens when `fetchCommitStatuses` returns a HttpException?

app_flutter/lib/state/flutter_build.dart
CaseyHillers: Would `WidgetTester.pump(Duration)` work for this?

app_flutter/lib/state/flutter_build.dart
CaseyHillers: That's a good point. Do you have any recommendations on how to forward this error to the widget?

My current approach would be `Stream`, but that seems to over engineer what we want to do here. A naive approach would be to just assume this was a one off incident, and don't update statuses.

app_flutter/lib/state/flutter_build.dart
jonahwilliams: Maybe...

app_flutter/test/service/mock_cocoon.dart
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2019 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:app_flutter/service/fake_cocoon.dart';
+import 'package:mockito/mockito.dart';
+
+/// CocoonService for checking interactions.
+class MockCocoonService extends Mock implements FakeCocoonService {}
```
jonahwilliams: Its fairly common to leave mocks as a single line declaration at the bottom of the test they are used in

app_flutter/lib/state/flutter_build.dart
CaseyHillers: https://github.com/flutter/cocoon/blob/411ae1a428db2c445da6a2dfc7ee5de5ceb5d195/app_flutter/test/state/flutter_build_test.dart#L24

I tried a basic test here, but it throws an exception since the refreshTimer is still running at the end of the test. On tear down, I call FlutterBuildState.dispose() that cancels the timer.

app_flutter/test/service/mock_cocoon.dart
CaseyHillers: If this gets used in another file (maybe in another PR), would I move it to its own file?

jonahwilliams: Not generally, since it so small

app_flutter/lib/state/flutter_build.dart
jonahwilliams: I would tear down the buildState in the test and not tearDown so the test can pass then

app_flutter/lib/state/flutter_build.dart
CaseyHillers: We talked in person and I created https://github.com/flutter/cocoon/issues/451 that sums up our solution.

app_flutter/lib/state/flutter_build.dart
CaseyHillers: Okay that worked. That's an interesting issue.

The test I implemented takes less than a few seconds to run, so I think this is done.

jonahwilliams: The widget tests should be running in a fake async zone that controls the flow of time

PR #452, opened 2019-09-30
Tree Build Status changes the AppBar color

CaseyHillers:
- Cocoon service fetches tree build status
- FlutterBuildState stores the tree build status
- AppBar changes color based on the current tree build status

tree_failing

### DIFF
--- a/app_flutter/lib/main.dart
+++ b/app_flutter/lib/main.dart
@@ -40,12 +40,12 @@ class _BuildDashboardPageState extends State<BuildDashboardPage> {
 
   @override
   Widget build(BuildContext context) {
+    final ThemeData theme = Theme.of(context);
     return Scaffold(
       appBar: AppBar(
         title: Text('Flutter Build Dashboard v2'),
-        backgroundColor: buildState.isTreeBuilding
-            ? Theme.of(context).primaryColor
-            : Theme.of(context).errorColor,
+        backgroundColor:
+            buildState.isTreeBuilding ? theme.primaryColor : theme.errorColor,
       ),
       body: Column(
         children: [
--- a/app_flutter/lib/main.dart
+++ b/app_flutter/lib/main.dart
@@ -43,6 +43,9 @@ class _BuildDashboardPageState extends State<BuildDashboardPage> {
     return Scaffold(
       appBar: AppBar(
         title: Text('Flutter Build Dashboard v2'),
+        backgroundColor: buildState.isTreeBuilding
+            ? Theme.of(context).primaryColor
+            : Theme.of(context).errorColor,
       ),
       body: Column(
         children: [
--- a/app_flutter/lib/service/appengine_cocoon.dart
+++ b/app_flutter/lib/service/appengine_cocoon.dart
@@ -69,6 +69,7 @@ class AppEngineCocoonService implements CocoonService {
 
   /// Check if [Map<String,Object>] follows the format for build-status.
   ///
+  /// Template of the response it should receive:
   /// ```json
   /// {
   ///   "AnticipatedBuildStatus": "Succeeded"|"Failed"
--- a/app_flutter/lib/service/appengine_cocoon.dart
+++ b/app_flutter/lib/service/appengine_cocoon.dart
@@ -46,6 +46,47 @@ class AppEngineCocoonService implements CocoonService {
     return _commitStatusesFromJson(jsonResponse['Statuses']);
   }
 
+  @override
+  Future<bool> fetchTreeBuildStatus() async {
+    /// This endpoint returns JSON {AnticipatedBuildStatus: [BuildStatus]}
+    http.Response response =
+        await _client.get('$_baseApiUrl/public/build-status');
+
+    if (response.statusCode != HttpStatus.ok) {
+      throw HttpException(
+          '$_baseApiUrl/public/build-status returned ${response.statusCode}');
+    }
+
+    Map<String, Object> jsonResponse = jsonDecode(response.body);
+
+    if (!_isBuildStatusResponseValid(jsonResponse)) {
+      throw HttpException(
+          '$_baseApiUrl/public/build-status had a malformed response');
+    }
+
+    return jsonResponse['AnticipatedBuildStatus'] == "Succeeded";
+  }
+
+  /// Check if [Map<String,Object>] follows the format for build-status.
+  ///
+  /// ```json
+  /// {
+  ///   "AnticipatedBuildStatus": "Succeeded"|"Failed"
+  /// }
+  /// ```
+  bool _isBuildStatusResponseValid(Map<String, Object> response) {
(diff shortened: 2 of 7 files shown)


CaseyHillers:
This finishes #425

app_flutter/lib/main.dart
```diff
@@ -43,6 +43,9 @@ class _BuildDashboardPageState extends State<BuildDashboardPage> {
     return Scaffold(
       appBar: AppBar(
         title: Text('Flutter Build Dashboard v2'),
+        backgroundColor: buildState.isTreeBuilding
+            ? Theme.of(context).primaryColor
```
jonahwilliams: Nit: look up the theme once and reuse it:

app_flutter/lib/state/flutter_build.dart
```diff
@@ -26,6 +26,9 @@ class FlutterBuildState extends ChangeNotifier {
   /// The current status of the commits loaded.
   List<CommitStatus> statuses = [];
 
+  /// Whether or not flutter/flutter currently passes tests.
+  bool isTreeBuilding = true;
```
jonahwilliams: This should probably be a getter for a private field. That will prevent it from being accidentally set by consuming code:

```
bool get isTreeBuilding => _isTreeBuilding;
bool _isTreeBuilding = true;
```

app_flutter/lib/state/flutter_build.dart
```diff
@@ -46,6 +49,8 @@ class FlutterBuildState extends ChangeNotifier {
   /// Request the latest [statuses] from [CocoonService].
   void _fetchBuildStatusUpdate() async {
     statuses = await _cocoonService.fetchCommitStatuses();
```
jonahwilliams: These requests don't need to be made serially:

```
await Future.wait([
  _cocoonService.fetchCommitStatuses().then((commitStatus) => statuses = commitStatus),
  _cocoonService.fetchTreeBuildStatus().then((buildStatus) => statuses = buildStatus),
]);
```

jonahwilliams: modulo types and formatting

app_flutter/test/service/appengine_cocoon_test.dart
```diff
@@ -63,8 +63,14 @@ const String jsonGetStatsResponse = """
       }
 """;
 
+const String jsonBuildStatusResponse = """
```
jonahwilliams: Include a status: failure case

CaseyHillers: Added

app_flutter/lib/state/flutter_build.dart
CaseyHillers: Thank you for sharing! This is a nice way of handling it.

app_flutter/lib/state/flutter_build.dart
CaseyHillers: Good point, added

app_flutter/lib/main.dart
CaseyHillers: Done

app_flutter/lib/service/appengine_cocoon.dart
```diff
@@ -46,6 +46,48 @@ class AppEngineCocoonService implements CocoonService {
...
+        await _client.get('$_baseApiUrl/public/build-status');
+
+    if (response.statusCode != HttpStatus.ok) {
+      throw HttpException(
+          '$_baseApiUrl/public/build-status returned ${response.statusCode}');
+    }
+
+    Map<String, Object> jsonResponse = jsonDecode(response.body);
+
+    if (!_isBuildStatusResponseValid(jsonResponse)) {
```
digiter: Better log this response, otherwise when we see the exception, we don't know why/how it's malformed.

app_flutter/lib/service/appengine_cocoon.dart
```diff
@@ -46,6 +46,48 @@ class AppEngineCocoonService implements CocoonService {
...
+  /// ```
+  bool _isBuildStatusResponseValid(Map<String, Object> response) {
+    if (!response.containsKey('AnticipatedBuildStatus')) {
+      return false;
+    }
+
+    String treeBuildStatus = response['AnticipatedBuildStatus'];
+    if (treeBuildStatus != 'Failed' && treeBuildStatus != 'Succeeded') {
+      return false;
+    }
```
digiter: Nit: it's more readable to write this in a positive way:
```
if (treeBuildStatus == "Succeeded" || treeBuildStatus == "Failed") {
  return true;
}

return false;
```

"In general checking the presence or absence of a positive is simpler for readers to understand than checking the presence or absence of a negative."

app_flutter/lib/main.dart
```diff
@@ -40,9 +40,12 @@ class _BuildDashboardPageState extends State<BuildDashboardPage> {
 
   @override
   Widget build(BuildContext context) {
+    final ThemeData theme = Theme.of(context);
     return Scaffold(
       appBar: AppBar(
         title: Text('Flutter Build Dashboard v2'),
+        backgroundColor:
+            buildState.isTreeBuilding ? theme.primaryColor : theme.errorColor,
```
digiter: In addition to "Succeeded" and "Failed", it's useful to know how outdated is the information.

app_flutter/lib/service/appengine_cocoon.dart
CaseyHillers: This will print the error to a local console. Should we think about implementing a logging system such as Firebase Crashlytics so users don't have to report these errors?

app_flutter/lib/service/appengine_cocoon.dart
CaseyHillers: For future reference: Testing on the Toilet #530: Improve Readability with Positive Booleans

The reason I chose this way was to have cases that returned false next to each other, instead of case 1 returns false, case 2 returns true, final case returns false. I wanted to try and reduce the cognitive switching on the return value.

app_flutter/lib/service/appengine_cocoon.dart
jonahwilliams: The issue with logging to console is that the #1 usecase for this will be the TVs in the immediate aread.

digiter: Not only the TV, I always have the build dashboard open on my chrome, so it would be nice to know the errors when they happen.

CaseyHillers: When errors happen on the TV, we just need to know an error is occurring. We do not need to know what the error is. The only time we would want to know what the error is on the TV is when we can't reproduce it on any other device. However, in that case the issue is most likely with the TV and not the application.

Thus, showing a generic error UI would indicate that someone needs to check what is going on by opening the dashboard on their computer to investigate further.

PR #453, opened 2019-10-01
In Progress Tasks show work is being done

CaseyHillers:
- Added a special case for In Progress tasks to show a loading indicator in the box
- Updated old references of ResultBox to TaskBox
- Added test for this new case

Here's a GIF (the performance issue comes from the GIF, not the app)
task_running

### DIFF
--- a/app_flutter/lib/task_box.dart
+++ b/app_flutter/lib/task_box.dart
@@ -17,24 +17,32 @@ class TaskBox extends StatelessWidget {
   /// [Task] to show information from.
   final Task task;
 
+  /// Status messages that map to TaskStatus enums.
+  /// TODO(chillers): Remove these and use TaskStatus enum when available. https://github.com/flutter/cocoon/issues/441
+  static const String statusFailed = 'Failed';
+  static const String statusNew = 'New';
+  static const String statusSkipped = 'Skipped';
+  static const String statusSucceeded = 'Succeeded';
+  static const String statusUnderperformed = 'Underperformed';
+  static const String statusInProgress = 'In Progress';
+
   /// A lookup table to define the background color for this TaskBox.
   ///
   /// The status messages are based on the messages the backend sends.
-  /// TODO(chillers): Update these to TaskStatus enum when backend switches to protos. https://github.com/flutter/cocoon/issues/441
   static const statusColor = <String, Color>{
-    'Failed': Colors.red,
-    'New': Colors.blue,
-    'Skipped': Colors.transparent,
-    'Succeeded': Colors.green,
-    'Underperformed': Colors.orange,
+    statusFailed: Colors.red,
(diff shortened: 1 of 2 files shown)


app_flutter/lib/task_box.dart
```diff
@@ -29,10 +30,26 @@ class TaskBox extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    if (task.status == 'In Progress') {
```
jonahwilliams: I would pull these string values into constants

app_flutter/lib/task_box.dart
```diff
@@ -29,10 +30,26 @@ class TaskBox extends StatelessWidget {
...
+      return Container(
+        margin: const EdgeInsets.all(1.0),
+        color: statusColor['New'],
```
jonahwilliams: With some careful planning this whole widget could be const

app_flutter/lib/task_box.dart
CaseyHillers: Done. I added a TODO to convert the usage to enums when the backend starts sending out protos.

app_flutter/lib/task_box.dart
CaseyHillers: I simplified some of the code reuse, but right now TaskBox can't be made const because of the mapping function to get a Task in StatusGrid isn't const. 
Do you have any recommendations on what to try to make the widget const?

app_flutter/lib/task_box.dart
```diff
@@ -8,32 +8,52 @@ import 'package:cocoon_service/protos.dart' show Task;
...
     return Container(
       margin: const EdgeInsets.all(1.0),
-      color: resultColor.containsKey(task.status)
-          ? resultColor[task.status]
+      color: statusColor.containsKey(task.status)
+          ? statusColor[task.status]
           : Colors.black,
+      child: (task.status == statusInProgress)
+          ? const Padding(
+              padding: const EdgeInsets.all(15.0),
```
jonahwilliams: Since you're already in a const context here the `const` is redundant. I think there is a lint that will catch this?

```suggestion
              padding: EdgeInsets.all(15.0),
```

CaseyHillers: Flutter format didn't change anything, and flutter analyze doesn't throw errors. Is there another tool for stuff like this?

jonahwilliams: Check out https://dart.dev/guides/language/analysis-options

CaseyHillers: Should this project use the same analysis_options.yaml that the Dart backend project is using?

jonahwilliams: They should probably be consistent

Issue #455, opened 2019-10-01
Match analysis_options for app_dart and app_flutter


CaseyHillers:
The analysis_options should be the same for both projects so there is less friction for developers to contribute to both code bases.

CaseyHillers:
Followup, app_dart/[analysis_options](https://github.com/flutter/cocoon/blob/master/app_dart/analysis_options.yaml) closely resembles flutter/flutter/[analysis_options](https://github.com/flutter/flutter/blob/master/analysis_options.yaml). It may be better to just match flutter/flutter.

PR #457, opened 2019-10-02
Fix tree status updates not reflecting in UI

### LINKED ISSUE #454: AppBar doesn't change color when TreeBuildStatus changes
#452 introduced this feature.

Since the AppBar isn't using a state variable, it doesn't update.

CaseyHillers:
Fixes #454 

Issue was there was no state for the AppBar. This creates a parent widget for the build dashboard that maintains the state, and wraps its children in a ChangeNotifierProvider.

# Tested
I updated the fake cocoon service to give a random value for the tree status, and verified that the AppBar switched colors randomly every 10 seconds.

I think a regression test for this bug would involve writing an integration test, and that is out of the scope of this project.

### DIFF
--- a/app_flutter/lib/build_dashboard.dart
+++ b/app_flutter/lib/build_dashboard.dart
@@ -38,7 +38,7 @@ class _BuildDashboardPageState extends State<BuildDashboardPage> {
 }
 
 /// Shows information about the current build status of flutter/flutter.
-/// 
+///
 /// The tree's current build status is reflected in the color of [AppBar].
 /// The results from tasks run on individual commits is shown in [StatusGrid].
 class BuildDashboard extends StatelessWidget {
--- a/app_flutter/lib/build_dashboard.dart
+++ b/app_flutter/lib/build_dashboard.dart
@@ -38,6 +38,9 @@ class _BuildDashboardPageState extends State<BuildDashboardPage> {
 }
 
 /// Shows information about the current build status of flutter/flutter.
+/// 
+/// The tree's current build status is reflected in the color of [AppBar].
+/// The results from tasks run on individual commits is shown in [StatusGrid].
 class BuildDashboard extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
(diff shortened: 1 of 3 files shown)


app_flutter/lib/service/fake_cocoon.dart
```diff
@@ -20,7 +22,7 @@ class FakeCocoonService implements CocoonService {
 
   @override
   Future<bool> fetchTreeBuildStatus() async {
-    return true;
+    return Random().nextBool();
```
jonahwilliams: Instead  of making this random  -  and therefore untestable, maybe make the value configurable for the fake?

app_flutter/lib/build_dashboard.dart
```diff
@@ -0,0 +1,63 @@
...
+/// The tree's current build status is reflected in the color of [AppBar].
+/// The results from tasks run on individual commits is shown in [StatusGrid].
+class BuildDashboard extends StatelessWidget {
+  @override
+  Widget build(BuildContext context) {
+    final ThemeData theme = Theme.of(context);
+    return Consumer<FlutterBuildState>(
+      builder: (_, buildState, child) => Scaffold(
+        appBar: AppBar(
+          title: Text('Flutter Build Dashboard v2'),
```
jonahwilliams: Did you turn on any of the lints that catch these?

```suggestion
          title: const Text('Flutter Build Dashboard v2'),
```

CaseyHillers: https://github.com/flutter/cocoon/issues/455

Not yet, but that can get added today. Do you think this project should just use the flutter/flutter analysis options?

jonahwilliams: Yeah, since it is a flutter project that makes sense. no need to do it in this PR

app_flutter/lib/service/fake_cocoon.dart
CaseyHillers: I added a way  to inject it, so tests can inject their own "random" generator.

PR #459, opened 2019-10-02
Update analysis_options to match flutter/flutter for app_flutter

### LINKED ISSUE #455: Match analysis_options for app_dart and app_flutter

CaseyHillers:
Mostly formatting stuff.

Removed unnecessary forEach for converting required capabilities json to the proto. Could just be done with a toString() in AppEngineCocoonService.

Fixes #455

### DIFF
--- a/app_flutter/analysis_options.yaml
+++ b/app_flutter/analysis_options.yaml
@@ -1,0 +1,192 @@
+# Specify analysis options.
+#
+# Until there are meta linter rules, each desired lint must be explicitly enabled.
+# See: https://github.com/dart-lang/linter/issues/288
+#
+# For a list of lints, see: http://dart-lang.github.io/linter/lints/
+# See the configuration guide for more
+# https://github.com/dart-lang/sdk/tree/master/pkg/analyzer#configuring-the-analyzer
+#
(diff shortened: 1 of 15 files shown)


app_flutter/analysis_options.yaml
```diff
@@ -0,0 +1,192 @@
+# Specify analysis options.
+#
+# Until there are meta linter rules, each desired lint must be explicitly enabled.
+# See: https://github.com/dart-lang/linter/issues/288
+#
+# For a list of lints, see: http://dart-lang.github.io/linter/lints/
+# See the configuration guide for more
+# https://github.com/dart-lang/sdk/tree/master/pkg/analyzer#configuring-the-analyzer
+#
+# There are other similar analysis options files in the flutter repos,
```
digiter: I guess you copied this file from somewhere, is this section still relevant?

CaseyHillers: Good catch! I removed this header and left a comment saying to stay in sync with the flutter/flutter lint rules.

PR #460, opened 2019-10-02
TaskBox reflects v1 task colors

### LINKED ISSUE #456: Tasks show as running even if they have already failed once
It would be nice to differentiate the tasks that are running for the first time vs tasks that are doing a repeat run.

Could be as simple as checking if it's In Progress, then looking at how many attempts have been made on the Task.

CaseyHillers:
- If tasks have been attempted, update their status to the correct status (New->Underperformed, Succeeded->SucceededButFlaky, InProgress->UnderpformedInProgress)
- Removed unncessary tester.takeException()s in StatusGrid tests

Fixes #456 , a feature request that asked for tasks that are in progress but have run before to be differentiated from first time running tasks.

image

image

### DIFF
--- a/app_flutter/lib/task_box.dart
+++ b/app_flutter/lib/task_box.dart
@@ -23,7 +23,10 @@ class TaskBox extends StatelessWidget {
   static const String statusNew = 'New';
   static const String statusSkipped = 'Skipped';
   static const String statusSucceeded = 'Succeeded';
+  static const String statusSucceededButFlaky = 'Succeeded Flaky';
   static const String statusUnderperformed = 'Underperformed';
+  static const String statusUnderperformedInProgress =
+      'Underperfomed In Progress';
   static const String statusInProgress = 'In Progress';
 
   /// A lookup table to define the background color for this TaskBox.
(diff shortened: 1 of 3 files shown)


app_flutter/lib/task_box.dart
```diff
@@ -35,17 +38,31 @@ class TaskBox extends StatelessWidget {
...
     statusUnderperformed: Colors.orange,
+    statusUnderperformedInProgress: Colors.orange,
   };
 
   @override
   Widget build(BuildContext context) {
+    final bool attempted = task.attempts > 1;
+    if (attempted) {
+      if (task.status == statusSucceeded) {
+        task.status = statusSucceededButFlaky;
```
jonahwilliams: I don't think you should be mutating the values of the task you are provided. Conceptually, some parent widget already "owns" this data, and might not be expecting it to change. Another way to think of this is that `build` methods should be side-effect free.

Instead I would keep some local that you update with the task status

CaseyHillers: Updated to have a local status var. The original intention was to match how v1 was doing it, but I looked and v1 doesn't seem to advantage of mutating the status beyond what this widget does.

Issue #450, opened 2019-09-30
Flutter app fails to build for web because url_launcher does not support web


CaseyHillers:
#447 introduced the dependency on the `url_launcher` plugin which does not current support the web.

https://github.com/flutter/plugins/pull/2119 is a pull request for adding support for this. This issue will probably get resolved on its own :)

CaseyHillers:
Also `dart:io` doesn't build for web. Luckily, it's only used in `AppEngineCocoonService` for some trivial pieces.

PR #462, opened 2019-10-03
Task icon row

CaseyHillers:
When you long tap the icon, it will show the task name in a Tooltip.

There was some left over formatting from #459 in the tests folder.

image

I tried adding tests to ensure Images were loading for the correct stage name, but couldn't find a way to differentiate Images from each other.


jonahwilliams:
Does this work with a mouse?

app_flutter/lib/status_grid.dart
```diff
@@ -69,14 +70,33 @@ class StatusGrid extends StatelessWidget {
             gridDelegate: SliverGridDelegateWithFixedCrossAxisCount(
                 crossAxisCount: columnCount),
             itemBuilder: (BuildContext context, int gridIndex) {
-              final int statusIndex = gridIndex ~/ columnCount;
+              if (gridIndex == 0) {
+                /// The top left corner of the grid is nothing since
+                /// the left column is for [CommitBox] and the top
+                /// row is for [TaskIcon].
+                return Container();
```
jonahwilliams: use a `const SizedBox()` instead

app_flutter/lib/task_icon.dart
```diff
@@ -0,0 +1,53 @@
...
+  static const String stageCirrus = 'cirrus';
+  static const String stageLuci = 'chromebot';
+  static const String stageDevicelab = 'devicelab';
+  static const String stageDevicelabWin = 'devicelab_win';
+  static const String stageDevicelabIOs = 'devicelab_ios';
+
+  /// A lookup table for matching [stageName] to [Image].
+  ///
+  /// [stageName] is based on the backend.
+  static Map<String, Image> stageIcons = <String, Image>{
```
jonahwilliams: Should be final

app_flutter/lib/task_icon.dart
```diff
@@ -0,0 +1,53 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    return Tooltip(
+      message: task.name,
+      child: Container(
+        margin: const EdgeInsets.all(7.5),
+        child: stageIcons.containsKey(task.stageName)
+            ? stageIcons[task.stageName]
+            : Icon(Icons.help),
```
jonahwilliams: `Icon(Icons.help)` should be const, right? I'm surprised that the lints didn't catch this one

app_flutter/lib/status_grid.dart
CaseyHillers: Done

app_flutter/lib/task_icon.dart
CaseyHillers: Yes, linter didn't catch it. Added.

CaseyHillers:
@jonahwilliams it works with a mouse on Android emulator. I assumed using Tooltip from the framework would work the same on the web, but right now the project does not build for web.

See #450

app_flutter/lib/task_icon.dart
CaseyHillers: Can it be made final? I get issues from the constructor saying it can't be final since Image.asset is not const.

jonahwilliams: `static final Map<String, Image> stageIcons = ...` should be fine

CaseyHillers: Thanks, that fixed it

jonahwilliams:
The web builds should work again once we roll build_runner

PR #463, opened 2019-10-03
Remove dart:io dep to build app_flutter for web

### LINKED ISSUE #450: Flutter app fails to build for web because url_launcher does not support web

CaseyHillers:
Fixes #450 

# Tested

I was able to run a release build on chrome, but ran into a CORS issue. I was able to see an indefinite spinner for the StatusGrid, which is the current behavior.

### DIFF
--- a/app_flutter/lib/service/appengine_cocoon.dart
+++ b/app_flutter/lib/service/appengine_cocoon.dart
@@ -3,7 +3,6 @@
 // found in the LICENSE file.
 
 import 'dart:convert';
-import 'dart:io';
 
 import 'package:http/http.dart' as http;
 import 'package:fixnum/fixnum.dart';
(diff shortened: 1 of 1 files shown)


CaseyHillers:
Will `dart:io` run on web? I was reading on the Flutter Web documentation not to use dart:io since the web does not have a file system.

CaseyHillers:
See https://flutter.dev/docs/development/platform-integration/web#can-i-use-dartio-with-a-web-app

jonahwilliams:
On the latest flutter master this should work fine in dev and release mode

CaseyHillers:
## Master branch
**Dev**: Works!
**Release**: Works!

## This Web fix branch
**Dev**: Works!
**Release**: Works!

*NOTE:* For release mode I force the Fake CocoonService since the AppEngine service throws CORS network issues since I am running it on localhost.

Thanks @jonahwilliams !

Issue #451, opened 2019-09-30
StatusGrid UI does not notify user of issues with Cocoon


CaseyHillers:
Currently, `FlutterBuildState` does not forward errors up to the widgets that use the data. One approach is to have additional state variables for errors like `commitStatusError`, and when it is set an error occurred and the widget should notify the user.

Somehow users should know if there was an issue when fetching data from Cocoon.

PR #466, opened 2019-10-07
FlutterBuildState error catching

CaseyHillers:
This sets up the data necessary for #451 

With errors recorded in the state, we can add error widgets that notify users of these issues. Such as letting users know that the Cocoon backend is having issues.


app_flutter/lib/state/flutter_build.dart
```diff
@@ -50,14 +56,22 @@ class FlutterBuildState extends ChangeNotifier {
...
+  /// Request the latest [statuses] and [isTreeBuilding] from [CocoonService].
   Future<void> _fetchBuildStatusUpdate() async {
-    await Future.wait<void>(<Future<void>>[
-      _cocoonService.fetchCommitStatuses().then<List<CommitStatus>>(
-          (List<CommitStatus> commitStatuses) => statuses = commitStatuses),
-      _cocoonService
-          .fetchTreeBuildStatus()
-          .then<bool>((bool treeStatus) => _isTreeBuilding = treeStatus),
+    await Future.wait(<Future<void>>[
+      _cocoonService.fetchCommitStatuses().then(
```
jonahwilliams: Should the _cocoonServer be responsible for determining what is and isn't an error?

CaseyHillers:
@digiter and I had an offline discussion following up from https://github.com/flutter/cocoon/pull/452#discussion_r330170811

We deemed a logging service would be nice (for example, Crashlytics), but is out of the scope of the project. This is due to the user base of the dashboard being Flutter contributors, they can get the stdout from the application to share bugs.

With that said, would you recommend I switch the print statements to use the [logging package](https://pub.dev/packages/logging) instead?

app_flutter/lib/state/flutter_build.dart
CaseyHillers: Each implementation of CocoonService has to decide what is an error and what isn't an error. I don't think that can be done in FlutterBuildState where it can take in different implementations of CocoonService. This might lead to over generalizing what an error is.

Since the logic behind handling errors is the same, I think we should handle this in FlutterBuildState to just write this error handling code once instead of in each CocoonService implementation.

jonahwilliams: From the example code here, the _cocoonService is already deciding what is and isn't an error by throwing vs returning normally.

CaseyHillers: Yes, the CocoonService should determine what is an isn't an error so FlutterBuildState doesn't need any extra logic handling errors.

CaseyHillers:
I came across https://flutter.dev/docs/testing/code-debugging#logging and there is a `dart:developer` package (supports web) that has log support. That is another alternative to doing print statements.

app_flutter/lib/state/flutter_build.dart
jonahwilliams: Why not return a CocoonResponse object instead?

CaseyHillers: Done.

This makes more sense since it simplifies FlutterBuildState to just hold data. No extra responsibilities :)

Thanks for the suggestion!

PR #467, opened 2019-10-08
Task Overlay UI

CaseyHillers:
This adds UI that allows users to view more information from a Task. New information shown is the duration of the task (how long it took to complete in seconds) and what agent it ran on. 

Currently, the backend is not  populating the end timestamp, so it defaults to 0. When the backend gets switched over to Dart, the duration will be correct.

*Images are from a phone running a release build. The overlay fits much better on desktop.*

image

image

image

### DIFF
--- a/app_flutter/lib/task_box.dart
+++ b/app_flutter/lib/task_box.dart
@@ -5,7 +5,6 @@
 import 'package:flutter/material.dart';
 
 import 'package:cocoon_service/protos.dart' show Task;
-import 'package:url_launcher/url_launcher.dart';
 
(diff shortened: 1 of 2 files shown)


jonahwilliams:
In the images presented above its not really clear what box is active. Maybe add a border/box decoration on however?

CaseyHillers:
@jonahwilliams thanks for the feedback. I updated to overlay a container that highlights the specific TaskBox.

task_overlay_selected

jonahwilliams:
That looks great!

Issue #420, opened 2019-09-16
GitHub commits can be added in the wrong order


tvolkert:
If multiple commits have landed in one refresh cycle, they can be added to the Cocoon datastore in the wrong order.

goderbauer:
I just ran into this as well: https://github.com/flutter/flutter/commit/81e693a7fe3b808cc9ae2bb3a2cbe404e67ec773 and https://github.com/flutter/flutter/commit/43932aa2ca0b16bcdda1efef7368877c8a8ca7e1 are showing up in the wrong order on the dashboard. It is slightly amusing because it makes it look like a test ran before the test was actually committed.

goderbauer:
This is more serious than I thought: The commits also show up in the wrong order in the benchmarks dashboard, which makes it hard to identify the culprit of a performance regression: Imagine two commit landing at roughly the same time. The first one introduces a regression. If they show up in inverse order in the benchmarks, the second one is blamed as the culprit.

Hixie:
This is really really bad.

tvolkert:
@godofredoc @keyonghan

keyonghan:
What I found: 
* commit order on build/benchmark pages are same as that in datastore. So the problem lies in inserting data to datastore, rather than retrieving data from datastore (then show up on web).
* https://github.com/flutter/cocoon/blob/master/app_dart/lib/src/request_handlers/refresh_github_commits.dart#L65
gives us the commit order from latest to oldest. However when inserting data to datastore, we need to insert oldest first. This way, it will give the correct order together with the first bullet above.
* https://github.com/flutter/cocoon/blob/master/app_dart/lib/src/request_handlers/refresh_github_commits.dart#L66
is using one single same timestamp for all commits within the cronjob period (1 min). We need to distinguish them, since benchmark/build pages get the commit list ordered by timestamp.

Will address above issues and validate.

keyonghan:
PR [465](https://github.com/flutter/cocoon/pull/465) was merged moment ago. Before I re-deploy to app engine, I need to fix issue [448](https://github.com/flutter/cocoon/issues/448)

PR #468, opened 2019-10-09
Implement /api/public/get-timeseries-history

keyonghan:
This PR implements /api/public/get-timeseries-history, and provides test cases.


app_dart/lib/src/request_handlers/get_timeseries_history.dart
```diff
@@ -0,0 +1,117 @@
...
+  final DatastoreServiceProvider datastoreProvider;
+  static const String timeSeriesKeyParam = 'TimeSeriesKey';
+  static const String startFromParam = 'StartFrom';
+
+  @override
+  Future<GetTimeSeriesHistoryResponse> post() async {
+    checkRequiredParameters(<String>[timeSeriesKeyParam]);
+    const int maxRecords = 1500;
+    final DatastoreService datastore = datastoreProvider();
+    final KeyHelper keyHelper = KeyHelper(applicationContext: AppEngineContext(false,'','','','','',Uri()));
```
dnfield: I don't think we should be passing in the `applicationContext` like this here.  It seems like this was done for tests - we should just make the test be able to pass in an application context to the ctor, like we're doing with `datastoreProvider`.

app_dart/lib/src/request_handlers/get_timeseries_history.dart
```diff
@@ -0,0 +1,117 @@
...
+    final Set<Commit> commits =
+        await datastore.queryRecentCommits(limit: maxRecords).toSet();
+
+    Key timeSeriesKey;
+    try {
+      timeSeriesKey = keyHelper.decode(requestData[timeSeriesKeyParam]);
+    } catch (error) {
```
dnfield: Please catch a specific exception/error here.  I suspect you want `FormatException`?

app_dart/lib/src/request_handlers/get_timeseries_history.dart
```diff
@@ -0,0 +1,117 @@
...
+  final TimeSeries timeSeries;
+  final List<TimeSeriesValue> timeSeriesValues;
+
+  @override
+  Map<String, dynamic> toJson() {
+    return <String, dynamic>{
+      'BenchmarkData': BenchmarkData(
+          timeSeriesEntity: TimeseriesEntity(timeSeries: timeSeries),
+          values: timeSeriesValues),
+      //LastPosition to do @Keyong
```
dnfield: Format should be: `// TODO: implement last position https://github.com/link/to/issue`

app_dart/lib/src/request_handlers/get_timeseries_history.dart
```diff
@@ -0,0 +1,117 @@
...
+  for (Commit commit in commits) {
+    TimeSeriesValue value;
+
+    if (valuesByCommit.containsKey(commit.sha)) {
+      value = valuesByCommit[commit.sha];
+      if (value.value < 0) {
+        value.value = 0;
+      }
+    } else {
+      // Insert placeholder entries for missing values.
```
dnfield: When do we expect missing values?  Maybe that could be added to the comment here.

app_dart/lib/src/request_handling/regular_request_handler.dart
```diff
@@ -0,0 +1,119 @@
...
+import 'dart:typed_data';
+
+import 'package:meta/meta.dart';
+
+import '../datastore/cocoon_config.dart';
+import 'body.dart';
+import 'exceptions.dart';
+import 'request_handler.dart';
+
+/// A [RequestHandler] that handles regular requests.
```
dnfield: What's a "regular" request?

How is it different from the api handler we have already?

app_dart/lib/src/service/datastore.dart
```diff
@@ -133,4 +147,4 @@ class DatastoreService {
       return previousStatusUpdates.single;
     }
   }
-}
+}
```
dnfield: nit: new line at EOF

app_dart/lib/src/request_handlers/get_timeseries_history.dart
```diff
@@ -0,0 +1,117 @@
...
+    final Set<Commit> commits =
+        await datastore.queryRecentCommits(limit: maxRecords).toSet();
+
+    Key timeSeriesKey;
+    try {
+      timeSeriesKey = keyHelper.decode(requestData[timeSeriesKeyParam]);
+    } catch (error) {
+      throw BadRequestException(
+          'Bad timeSeries key: ${requestData[timeSeriesKeyParam]}');
+    }
```
digiter: Would the exception here make the backend crash?

app_dart/lib/src/request_handlers/get_timeseries_history.dart
digiter: What does this comment mean?

app_dart/lib/src/request_handlers/get_timeseries_history.dart
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  for (Commit commit in commits) {
+    TimeSeriesValue value;
+
+    if (valuesByCommit.containsKey(commit.sha)) {
+      value = valuesByCommit[commit.sha];
+      if (value.value < 0) {
+        value.value = 0;
+      }
```
digiter: Why negative value should be regulated to zero?

app_dart/lib/src/request_handling/regular_request_handler.dart
```diff
@@ -0,0 +1,119 @@
...
+///
+/// Regular requests extend by enabling request data for non-API requests
```
digiter: What are non-API requests? I thought the endpoints of the backend are all APIs.

app_dart/lib/src/request_handlers/get_timeseries_history.dart
keyonghan: The different part from other existing APIs is: 
for this API: we do not want to have authentication, namely no login needed to view on the benchmark page
for other APIs: we need authenticated client to move forward.

So, not only for test, even in the code, we do not need any authenticated application context. That's why I simply put it like this.

dnfield: Doesn't applicationContext handle more than just authentication though?

app_dart/lib/src/request_handling/regular_request_handler.dart
```diff
@@ -0,0 +1,119 @@
...
+  /// This is guaranteed to be non-null. If the request body was empty, or if
+  /// it contained non-JSON or binary (non-UTF-8) data, this will be an empty
+  /// map.
+  ///
+  /// See also:
+  ///
+  ///  * [requestBody], which specifies the raw bytes of the HTTP request body.
+  @protected
+  Map<String, dynamic> get requestData =>
+      getValue<Map<String, dynamic>>(RegularKey.requestData);
```
digiter: Nit: I'd name the two functions above to "requestBodyRaw" and "requestBody". As the raw one is used less frequently, it could have a longer name.

app_dart/lib/src/request_handling/regular_request_handler.dart
keyonghan: regular request differs from API handler on needing no authentication.

app_dart/test/request_handlers/get_timeseries_history_test.dart
```diff
@@ -0,0 +1,108 @@
...
+    GetTimeSeriesHistory handler;
+    RegularRequestHandlerTester tester;
+
+    setUp(() {
+      config = FakeConfig();
+      db = FakeDatastoreDB();
+      tester = RegularRequestHandlerTester();
+      tester.requestData = <String, dynamic>{
+        'TimeSeriesKey':
+            'ahNzfmZsdXR0ZXItZGFzaGJvYXJkcjULEgpUaW1lc2VyaWVzIiVhbmFseXplcl9iZW5jaG1hcmsuZmx1dHRlcl9yZXBvX2JhdGNoDA',
```
digiter: Use a shorter string?

app_dart/lib/src/request_handlers/get_timeseries_history.dart
keyonghan: Added comments like:
      // Insert placeholder entries for missing values. 
      // Missing values happen when there is any queried timeSeriesValue 
      // whose commit does not appear in the queried top [maxRecords] commits

keyonghan: This may due to [issue 420](https://github.com/flutter/cocoon/issues/420).

app_dart/lib/src/service/datastore.dart
keyonghan: added.

app_dart/lib/src/request_handlers/get_timeseries_history.dart
keyonghan: Added.

app_dart/lib/src/request_handlers/get_timeseries_history.dart
keyonghan: Updated to:
// TODO(keyonghan): implemement last position 
 // https://github.com/flutter/flutter/issues/42362

app_dart/lib/src/request_handlers/get_timeseries_history.dart
keyonghan: Existing other APIs are quoting keyHelper like below:

    final ClientContext clientContext = authContext.clientContext;
    final KeyHelper keyHelper =
        KeyHelper(applicationContext: clientContext.applicationContext);

authContext is defined in api_request_hander.dart

    AuthenticatedContext get authContext => getValue<AuthenticatedContext>(ApiKey.authContext);
    context = await authenticationProvider.authenticate(request);

Then it goes to the authentication.authenticate(request).
Since this API is public, we do not pass authProvider when called.. What we need here is simply an empty AppEngineContext to move forward.

app_dart/lib/src/request_handlers/get_timeseries_history.dart
keyonghan: It shouldn't.

app_dart/lib/src/request_handlers/get_timeseries_history.dart
keyonghan: Updated to
// TODO(keyonghan): implemement last position
// flutter/flutter#42362

app_dart/lib/src/request_handlers/get_timeseries_history.dart
keyonghan: Same from go.

app_dart/lib/src/request_handling/regular_request_handler.dart
keyonghan: non-API requests deal with public ones.

app_dart/lib/src/request_handling/regular_request_handler.dart
keyonghan: These two are conventional names from api_request_hander and we have been using requestData for couple of scripts,, so we may want to keep them consistent.

app_dart/test/request_handlers/get_timeseries_history_test.dart
keyonghan: This is a real key, and we need to test that.

app_dart/lib/src/request_handling/regular_request_handler.dart
digiter: Then the name should reflect the fact, consider something like "NoAuthRequestHandler".

I wonder why this request doesn't need auth? Time series data is public?

app_dart/lib/src/request_handlers/get_timeseries_history.dart
digiter: Please leave a comment saying this behavior is copied from the Go code.

app_dart/lib/src/request_handling/regular_request_handler.dart
digiter: "non-API" sounds too weird. Consider changing the name of "api_request_hander" to "auth_request_handler", and name this one "request_handler".

app_dart/test/request_handlers/get_timeseries_history_test.dart
digiter: Why? You are talking to a fake datastore, you can use a fake key.

app_dart/lib/src/request_handlers/get_timeseries_history.dart
keyonghan: Now that everything on dart side is exactly a direct mirror from GO side. Actually there are a lot of other similar places: such as the maxNum for commits/taskvalues of get-benchmark API/get-timeseries-history API. Another thing to mention is that our existing cocoon backend currently has quite a few places to improve, and our next step after the migration finishes is to focus on improvement. Since we assume everything we are doing now is a direct copy from GO, it maybe not necessary to add that comment? Please let me know your thought based on these info. I can add that if you prefer.

app_dart/lib/src/request_handling/regular_request_handler.dart
keyonghan: Uhh, api_request_hander has been widely used by dozens' of different files; request_handler is an existing class based on which api_request_handler and regular_request_handler are extended from. Here I use non-API is mainly for comparison purpose to distinguish from existing API request.

app_dart/test/request_handlers/get_timeseries_history_test.dart
keyonghan: This key is url safe, and it is visible to every one even you are not logged in.
This key is exactly the one our API uses
In the API, we need to decode that key to the real timeseries, which is also pre loaded in our test. Together with the above two facts, it does not hurt to use this real key:-)

app_dart/lib/src/request_handling/regular_request_handler.dart
```diff
@@ -0,0 +1,119 @@
+// Copyright 2019 The Chromium Authors. All rights reserved.
```
digiter: This class copies functionalities from `api_request_handler` to handle JSON payload, which could be refactored into a `json_request_handler`. Then refactor to a cleaner hierarchy: `api_request_handler` extends `json_request_handler` extends `request_handler`. And process "no auth" requests via `json_request_handler`.

app_dart/lib/src/request_handling/regular_request_handler.dart
keyonghan: Changed to NoAuthRequestHandler as suggested to make it clear.

Yes, they are public.

app_dart/lib/src/request_handlers/get_timeseries_history.dart
keyonghan: Added.

app_dart/lib/src/request_handling/regular_request_handler.dart
keyonghan: Changed to NoAuth instead as suggested above.

app_dart/lib/src/request_handling/regular_request_handler.dart
keyonghan: Have changed 'regular' to 'noauth' to make it clear. Yeah, this is one option as well. Since api request involves lots of files and existing APIs, I would prefer to avoid touching it for now. Together with the 'noauth' name change, it is should be clear now.

app_dart/lib/src/request_handlers/get_timeseries_history.dart
dnfield: This should just make the backend return a 4xx error to the client

app_dart/lib/src/request_handlers/update_timeseries.dart
```diff
@@ -52,8 +52,8 @@ class UpdateTimeSeries extends ApiRequestHandler<UpdateTimeSeriesResponse> {
     final ClientContext clientContext = authContext.clientContext;
     final KeyHelper keyHelper =
         KeyHelper(applicationContext: clientContext.applicationContext);
-    double goal = requestData[goalParam];
-    double baseline = requestData[baselineParam];
+    double goal = requestData[goalParam].toDouble();
+    double baseline = requestData[baselineParam].toDouble();
```
dnfield: Can these be null?  Should it be `requestData[goalParam]?.toDouble()`?  The current code would not throw if they were null, but it's not clear to me if we expect them to always be nonnull.

digiter:
LGTM, leave approval to Dan.

dnfield:
I'm down to just the question about nullability on the doubles.  My inclination is to assume they can be null.

app_dart/lib/src/request_handlers/update_timeseries.dart
keyonghan: They are expected to be non-null, since the ui doesn't allow to call api when any value is null. To be double safe, I updated accordingly.

keyonghan:
Thank you both for your detailed and quick comments!

PR #471, opened 2019-10-12
[WIP] Deploy app_flutter to AppEngine under /v2/

CaseyHillers:
- Added code to the Dart server to serve the Flutter project build files
  - Added mime package for easy lookup of content type for a file
- Updated `app.yaml` to upload the Flutter project build files and make accessible to the Dart server
- Added a deployment script that handles building app_flutter and copying it over to app_dart.

Open questions:
- Is there a better way of handling deployments?
- Do I need to add tests for `bin/server.dart`?


app_dart/bin/server.dart
```diff
@@ -55,7 +56,60 @@ Future<void> main() async {
...
+    /// 
+    /// Currently the Flutter application will run at
+    /// https://flutter-dashboard.appspot.com/v2/
+    bool isRequestForFlutterApplicationBeta(HttpRequest request) {
+      return request.uri.path.contains('/v2');
+    }
+
     return await runAppEngine((HttpRequest request) async {
+      if (isRequestForFlutterApplicationBeta(request)) {
+        await sendFlutterApplicationAssets(request);
```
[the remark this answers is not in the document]
CaseyHillers: My understanding of the custom runtime for dart is that everything is redirected to `bin/server.dart`. I started off by adding handlers for this endpoint, but they are never run. I tested this by adding the same handler code to the go project and deploying it, and that worked.

I tried forking `app_flutter` into its own AppEngine service to serve the content from a separate runtime, but that was a more invasive change as those runtimes end up on different sub-domains.

app_dart/bin/server.dart
```diff
@@ -4,11 +4,12 @@
...
 
-import 'package:cocoon_service/cocoon_service.dart';
-
 import 'package:appengine/appengine.dart';
-import 'package:cocoon_service/src/service/access_token_provider.dart';
 import 'package:gcloud/db.dart';
+import 'package:mime/mime.dart';
+
+import 'package:cocoon_service/cocoon_service.dart';
+import 'package:cocoon_service/src/service/access_token_provider.dart';
```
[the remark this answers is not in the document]
CaseyHillers: Done

app_dart/app.yaml
```diff
@@ -8,3 +8,12 @@ service: default
...
   memory_gb: 4.0
+
+handlers:
+# The Dart server handles all requests. However, this is used to ensure the 
+# assets the dart server needs are uploaded to AppEngine.
+- url: /v2/(.*\.(html|css|js|ico|svg|png|jpg|map))$
+  application_readable: true # So the dart server can read the files
+  static_files: build/web/\1
+  upload: build/web/.*\.(html|css|js|ico|svg|png|jpg|map)$ # This cannot use the parent directory
+  secure: always
```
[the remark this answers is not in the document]
CaseyHillers: Added

app_dart/app.yaml
```diff
@@ -8,3 +8,12 @@ service: default
 
 resources:
   memory_gb: 4.0
+
+handlers:
+# The Dart server handles all requests. However, this is used to ensure the 
+# assets the dart server needs are uploaded to AppEngine.
+- url: /v2/(.*\.(html|css|js|ico|svg|png|jpg|map))$
+  application_readable: true # So the dart server can read the files
+  static_files: build/web/\1
```
tvolkert: If there's no built-in static file handling for custom runtimes in GAE, then are the `static_files` and `secure` lines needed?

app_dart/bin/server.dart
```diff
@@ -55,7 +55,60 @@ Future<void> main() async {
...
+      response.statusCode = HttpStatus.internalServerError;
+      await response.close();
+    }
+
+    /// Send static assets for the Flutter application to the client.
+    /// 
+    /// If the asset cannot be found, send a 404 Not Found response.
+    /// If there is an error sending the asset, send a 500 Internal Server Error
+    /// response.
+    Future<void> sendFlutterApplicationAssets(HttpRequest request) async {
```
tvolkert: Rather than serving these directly in `bin/server.dart`, I'd rather make a `StaticFileHandler` class in https://github.com/flutter/cocoon/tree/master/app_dart/lib/src/request_handling and register a new `regexHandlers` map to live alongside the `handlers` map.

app_dart/app.yaml
CaseyHillers: Removed `secure`.

`static_files` cannot be removed as handler requires some sort of url handling logic. I simplified so that it just returns `index.html` just in case this changes in the future. That way we know about the change and can remove the dart handler code.

app_dart/bin/server.dart
CaseyHillers: That makes more sense. I started work here https://github.com/flutter/cocoon/blob/5b65714384081a3a4a03023e02f5207e7b45eff6/app_dart/lib/src/request_handling/static_file_handler.dart but do you think it is necessary for this PR to do the RegExp handling? Right now we just want to serve out of `/v2/*`.

tvolkert: Yeah, it's fine to just wire up the static file handler straight to `/v2/*`

app_dart/README.md
```diff
@@ -91,14 +91,14 @@ $ docker images|grep local|tr -s ' '|cut -d' ' -f3|xargs docker rmi -f
...
 https://console.cloud.google.com/appengine/versions?project=flutter-dashboard
 for the list of current versions.
 
 ```sh
-$ gcloud app deploy \
-    --project flutter-dashboard \
-    --no-promote \
-    --no-stop-previous-version \
-    --version $VERSION
+$ deploy.sh
```
tvolkert: `dart dev/deploy.dart`

app_dart/lib/src/request_handling/static_file_handler.dart
```diff
@@ -0,0 +1,145 @@
...
+
+import 'package:appengine/appengine.dart';
+import 'package:meta/meta.dart';
+import 'package:mime/mime.dart';
+
+import 'exceptions.dart';
+
+/// A class based on [RequestHandler] for serving static files.
+@immutable
+class StaticFileHandler {
```
tvolkert: This should extend [`RequestHandler<Body>`](https://github.com/flutter/cocoon/blob/master/app_dart/lib/src/request_handling/request_handler.dart) and override `get()`.  It can probably return `Body.forStream()` on the stream of bytes you get when you open the file.

app_dart/lib/src/request_handling/static_file_handler.dart
```diff
@@ -0,0 +1,145 @@
...
+  /// Creates a new [StaticFileHandler].
+  const StaticFileHandler();
+
+  /// The location for the Flutter application
+  // TODO(chillers): Remove this when deployed for production use. https://github.com/flutter/cocoon/issues/472
+  static const String flutterBetaUrlPrefix = '/v2';
```
tvolkert: This shouldn't be hard-coded in the handler -- the handler should take the info it needs to convert from request path to static file in its constructor, then that info should be passed to the handler from `bin/server.dart`

app_dart/dev/deploy.dart
```diff
@@ -0,0 +1,115 @@
...
+      '--no-promote',
+      '--no-stop-previous-version',
+    ],
+  );
+
+  await process.stderr.pipe(stdout);
+  await process.stdout.pipe(stdout);
+
+  // So this is probably not the best way to do it
+  await Future<void>.delayed(const Duration(seconds: 15), () => process.stdin.write('y\n'));
```
digiter: How about `echo 'y' | gcloud app deploy ...`?

app_dart/README.md
CaseyHillers: Good catch! Fixed.

app_dart/dev/deploy.dart
CaseyHillers: That's a much better way of thinking about it. I updated it so it pipes the stdin of the user to this process. That way  the user can confirm/decline the deployment themselves.

CaseyHillers: However, do you know if there's any way to get around this lint ignore? https://github.com/flutter/cocoon/pull/471/commits/ed09516aa654446eafb8d37204f56b478daa6cea#diff-48a39cddbe6cd5ba24449d014cd75722R79

When I await, it will stall (I assume because stdin never ever finishes piping information, which makes sense).

app_dart/lib/src/request_handling/static_file_handler.dart
CaseyHillers: Added

app_dart/dev/deploy.dart
CaseyHillers: Jonah recommended `unawaited` from pedantic, and that worked.

app_dart/lib/src/request_handling/static_file_handler.dart
CaseyHillers: Updated. Thank you for the explanation

app_dart/test/request_handling/static_file_handler_test.dart
```diff
@@ -0,0 +1,37 @@
...
+    RequestHandlerTester tester;
+    StaticFileHandler<Body> staticFileHandler;
+
+    final FakeConfig config = FakeConfig();
+
+    setUp(() {
+      tester = RequestHandlerTester();
+    });
+
+    test('returns 404 response when file does not exist', () async {
```
CaseyHillers: @tvolkert should I add tests for mocking the filesystem? I would like to be able to verify a file response is correctly sent, but I am having a hard time finding resources on that topic in Dart 2.

jonahwilliams: It might be a bit complicated for this, but there is always IOOverrides, see

https://github.com/flutter/flutter/blob/c8b3c9b7270bf6d768899fd78bd5b40466514334/packages/flutter_tools/test/src/io.dart#L5

app_dart/dev/deploy.dart
```diff
@@ -0,0 +1,115 @@
...
+  return await process.exitCode == 0;
+}
+
+Future<int> main(List<String> arguments) async {
+  final ArgParser argParser = ArgParser()
+    ..addOption(gcloudProjectIdFlag, abbr: gcloudProjectIdAbbrFlag)
+    ..addOption(gcloudProjectVersionFlag, abbr: gcloudProjectVersionAbbrFlag);
+
+  if (!_getArgs(argParser, arguments)) {
+    return 1;
```
tvolkert: Does the return value from `main()` yield the exit code?  If so, til...

app_dart/lib/src/request_handling/exceptions.dart
```diff
@@ -28,6 +28,12 @@ class BadRequestException extends HttpStatusException {
       : super(HttpStatus.badRequest, message);
 }
 
+/// Exception that will trigger an HTTP 404 not found
+class NotFoundException extends HttpStatusException {
+  const NotFoundException(String missing)
+      : super(HttpStatus.notFound, 'Not found: $missing');
```
tvolkert: `assert(missing != null)`

app_dart/lib/src/request_handling/static_file_handler.dart
```diff
@@ -0,0 +1,48 @@
...
+import 'package:mime/mime.dart';
+
+import 'body.dart';
+import 'exceptions.dart';
+
+/// A class based on [RequestHandler] for serving static files.
+@immutable
+class StaticFileHandler<T extends Body> extends RequestHandler<Body> {
+  /// Creates a new [StaticFileHandler].
+  const StaticFileHandler(this.filePath, {Config config}) : super(config: config);
```
tvolkert: `@required this.config`

app_dart/lib/src/request_handling/static_file_handler.dart
```diff
@@ -0,0 +1,48 @@
...
+
+    /// The file path in app_dart to the files to serve
+    const String basePath = 'build/web';
+    final File file = File('$basePath$resultPath');
+
+    if (file.existsSync()) {
+      try {
+        final String mimeType = lookupMimeType(resultPath);
+        response.headers.contentType = ContentType.parse(mimeType);
+        return Body.forStream(Stream<Uint8List>.fromFuture(file.readAsBytes()));
```
tvolkert: `File.openRead()` will give you the byte stream directly.

app_dart/lib/src/request_handling/static_file_handler.dart
```diff
@@ -0,0 +1,48 @@
...
+      } catch (error, stackTrace) {
```
tvolkert: `RequestHandler` will catch errors and do this for you.

app_dart/lib/src/request_handling/static_file_handler.dart
tvolkert: If you made this class inject a `FileSystem`, defaulting to `const LocalFileSystem()`, then you could mock out the file access in the test.

(see package:file)

app_dart/test/request_handling/static_file_handler_test.dart
tvolkert: See comment above about injecting a `FileSystem` into the handler.

app_dart/app.yaml
```diff
@@ -8,3 +8,15 @@ service: default
...
+# The Dart server handles all requests. However, this is used to ensure the 
+# assets the dart server needs are uploaded to AppEngine.
+- url: /v2/(.*\.(html|css|js|ico|svg|png|jpg|map))$
+  application_readable: true # So the dart server can read the files
+  # If the Dart custom runtime changes and starts using this handler,
+  # we want to know. This will have the app only serve HTML and we
+  # will know we can remove the Dart handling code. Just swap
+  # index.html to \1
+  static_files: build/web/index.html
+  upload: build/web/.*\.(html|css|js|ico|svg|png|jpg|map)$ # This cannot use the parent directory
```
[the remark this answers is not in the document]
CaseyHillers: Added

app_dart/bin/server.dart
```diff
@@ -55,7 +58,23 @@ Future<void> main() async {
       port: await config.forwardPort,
     );
 
+    /// Check if the requested URI is for the Flutter Application
+    /// 
+    /// Currently the Flutter application will run at
+    /// https://flutter-dashboard.appspot.com/v2/
+    bool isRequestForFlutterApplicationBeta(HttpRequest request) {
+      return request.uri.path.contains(flutterBetaUrlPrefix);
```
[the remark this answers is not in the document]
CaseyHillers: Added. I didn't know that existed :)

app_dart/bin/server.dart
```diff
@@ -55,7 +58,23 @@ Future<void> main() async {
...
+    }
+
     return await runAppEngine((HttpRequest request) async {
+      if (isRequestForFlutterApplicationBeta(request)) {
+        String filePath = request.uri.toFilePath();
+        // TODO(chillers): Remove this when deployed for production use. https://github.com/flutter/cocoon/issues/472
+        filePath = filePath.replaceFirst(flutterBetaUrlPrefix, '');
+        
+        await const StaticFileHandler().service(request, filePath);
+      }
```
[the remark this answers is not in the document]
CaseyHillers: That's a good catch! RequestHandler closes the response. I added a return statement to ensure nothing else gets run.

app_dart/dev/deploy.dart
CaseyHillers: It does not. I just did a test with `dart dev/deploy.dart && echo 'I should not be shown'` and it shows the message to not be shown. Which indicates deploy.dart returns 0.

I switched these statements to `exit(1)` which works.

app_dart/dev/deploy.dart
```diff
@@ -0,0 +1,115 @@
...
+
+import 'package:args/args.dart';
+
+const String gcloudProjectIdFlag = 'project';
+const String gcloudProjectIdAbbrFlag = 'p';
+
+const String gcloudProjectVersionFlag = 'version';
+const String gcloudProjectVersionAbbrFlag = 'v';
+
+String gcloudProjectId;
```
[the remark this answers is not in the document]
CaseyHillers: Added

app_dart/lib/src/request_handling/exceptions.dart
CaseyHillers: Added

app_dart/lib/src/request_handling/static_file_handler.dart
CaseyHillers: `Body.forStream(file.openRead())` throws `_FileStream is not a substype of type Stream`.  Is there another approach we should try?

app_dart/lib/src/request_handling/static_file_handler.dart
CaseyHillers: Added, thank you for pointing that out!

app_dart/lib/src/request_handling/static_file_handler.dart
tvolkert: Try `Body.forStream(file.openRead().cast<Uint8List>())`

app_dart/lib/src/request_handling/static_file_handler.dart
CaseyHillers: Thank you! Added

app_dart/lib/src/request_handling/static_file_handler.dart
CaseyHillers: Added

app_dart/lib/src/request_handling/static_file_handler.dart
```diff
@@ -0,0 +1,53 @@
...
+import 'package:file/local.dart';
+import 'package:meta/meta.dart';
+import 'package:mime/mime.dart';
+
+import 'body.dart';
+import 'exceptions.dart';
+
+/// A class based on [RequestHandler] for serving static files.
+@immutable
+class StaticFileHandler<T extends Body> extends RequestHandler<Body> {
```
tvolkert: This doesn't need to be generic -- it can just extend `RequestHandler<Body>`.  Then you can just instantiate `StaticFileHandler` in server.dart (not `StaticFileHandler<Body>`)

app_dart/lib/src/request_handling/static_file_handler.dart
```diff
@@ -0,0 +1,53 @@
...
+  /// Creates a new [StaticFileHandler].
+  const StaticFileHandler(this.filePath,
+      {@required Config config, FileSystem fs})
```
tvolkert: You can use a default value here -- `this.fs = const LocalFileSystem()` (and no harm in making the instance variable public).

app_dart/test/request_handling/static_file_handler_test.dart
```diff
@@ -0,0 +1,61 @@
...
+
+      final Body body = await tester.get(staticFileHandler);
+      expect(body, isNotNull);
+      final String response = await _decodeHandlerBody(body);
+      expect(response, indexFileContent);
+    });
+  });
+}
+
+class MockHttpRequest extends Mock implements HttpRequest {}
```
tvolkert: This is no longer needed.

app_dart/lib/src/request_handling/static_file_handler.dart
CaseyHillers: Added, thank you!

app_dart/test/request_handling/static_file_handler_test.dart
CaseyHillers: Done

PR #473, opened 2019-10-14
/api/authorize-agent Implementation (GO->Dart)

keyonghan:
This PR implements API /api/authorize-agent, and creates corresponding test.


app_dart/lib/src/service/datastore.dart
```diff
@@ -147,4 +151,25 @@ class DatastoreService {
       return previousStatusUpdates.single;
     }
   }
+
+  /// Generate new authorization token for [agent]
+  ///
+  /// The hashed code of token will be returned as a list
+  List<int> refreshAgentAuthToken(Agent agent){
```
tvolkert: Nit: space in between close parens and open curly brace.  Does `dartfmt` not do this automatically?

app_dart/lib/src/service/datastore.dart
```diff
@@ -147,4 +151,25 @@ class DatastoreService {
...
+    const int length = 16;
+    const String urlSafeChars = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
+
+    final Random random = Random();
+    final List<String> token = <String>[];
```
tvolkert: You can use `StringBuffer` here.

app_dart/lib/src/service/datastore.dart
```diff
@@ -147,4 +151,25 @@ class DatastoreService {
...
+
+    for (int i = 0; i < length; i++){
```
tvolkert: Nit: space in between close parens and open curly brace

app_dart/lib/src/service/datastore.dart
```diff
@@ -147,4 +151,25 @@ class DatastoreService {
...
+      token.add(urlSafeChars[random.nextInt(urlSafeChars.length)]);
+    }
+
+    final String hashToken = DBCrypt().hashpw(token.toString(), DBCrypt().gensalt());
+    final List<int> encodeToken = ascii.encode(hashToken);
```
tvolkert: Nit: no need for the `encodeToken` local variable - you can return the value directly here.

app_dart/lib/src/service/datastore.dart
tvolkert: This doesn't do any datastore access, so it's a little out of place in `DatastoreService`.  Maybe create a new `AgentService`?

keyonghan: Running darfmt to have that space.

app_dart/lib/src/service/datastore.dart
keyonghan: Updated.

app_dart/lib/src/service/datastore.dart
keyonghan: Done.

app_dart/lib/src/service/datastore.dart
keyonghan: Updated as suggested.

app_dart/lib/src/service/datastore.dart
keyonghan: Put it here only because this function was defined in db.go from GO side. Yeah, it makes sense to have a separate service for agent. Just created new AgentService as suggested.

app_dart/lib/src/request_handlers/authorize_agent.dart
```diff
@@ -0,0 +1,69 @@
...
+
+  static const String agentIdParam = 'AgentID';
+
+  @override
+  Future<AuthorizeAgentResponse> post() async {
+    checkRequiredParameters(<String>[agentIdParam]);
+
+    final String agentId = requestData[agentIdParam];
+    final DatastoreService datastore = datastoreProvider();
+    final AgentService agentService = AgentService();
```
tvolkert: This should be injected so that it can be mocked out in tests.

app_dart/lib/src/service/agent_service.dart
```diff
@@ -0,0 +1,41 @@
...
+
+import 'package:dbcrypt/dbcrypt.dart';
+import 'package:meta/meta.dart';
+
+/// Service class for Agent.
+///
+/// This service exists to provide common agent queries made by
+/// the Cocoon backend.
+@immutable
+class AgentService {
```
tvolkert: You should probably add a const constructor (since you can).

app_dart/lib/src/service/agent_service.dart
```diff
@@ -0,0 +1,41 @@
...
+
+import 'dart:convert';
+import 'dart:math';
+
+import 'package:dbcrypt/dbcrypt.dart';
+import 'package:meta/meta.dart';
+
+/// Service class for Agent.
+///
+/// This service exists to provide common agent queries made by
```
tvolkert: Nit: "queries" isn't quite right.  Maybe "This service provides functionality for interacting with [Agent] instances."

app_dart/lib/src/service/agent_service.dart
```diff
@@ -0,0 +1,41 @@
...
+/// the Cocoon backend.
+@immutable
+class AgentService {
+  /// Generate new authorization token for [agent]
+  ///
+  /// The hashed code of token will be returned as a list
+  Map<String, dynamic> refreshAgentAuthToken() {
```
tvolkert: Let's create an `AgentAuthToken` class that has a `value` getter for the raw String and a `hash` getter for the ascii-encoded hash - then return that instead of a map.

app_dart/lib/src/request_handlers/authorize_agent.dart
keyonghan: Revised following style of datastore.

app_dart/lib/src/service/agent_service.dart
keyonghan: Added.

app_dart/lib/src/service/agent_service.dart
keyonghan: Updated.

app_dart/lib/src/service/agent_service.dart
keyonghan: Updated.

PR #474, opened 2019-10-16
[cocoon] Update the net interface we are getting the ip from.

godofredoc:
With the linux host migration to new hardware and Debian, ip addresses
are not reported anymore. This is because the network interface name
changed and needs to be updated.


godofredoc:
This is to fix [Linux host not updating the ip address](https://github.com/flutter/flutter/issues/42859)

PR #475, opened 2019-10-17
[Agent] Add helper scripts

digiter:
* `ci.sh` starts the agent.
* `smoke_test.sh` validates the runtime environment setup.

### DIFF
--- a/agent/bin/ci.sh
+++ b/agent/bin/ci.sh
@@ -1,32 +1,38 @@
 #!/bin/bash
+# Starts a Cocoon agent to run in continuous integration (CI) mode.
+#
+# Expectes current directory to be cocoon/agent.
+# Requires $HOME environment variable.
 set -e
-
-# Starts a Cocoon agent.
 
 USER_NAME="flutter"
 CONFIG_FILE="config.yaml"
+LOG_FILE="$HOME/agent.log"
 
 if [ "$(whoami)" != "$USER_NAME" ]; then
-  echo "Should run under user '$USER_NAME'."
+  echo "Should run under user: $USER_NAME."
   exit 1
 fi
-
-pushd "$HOME/cocoon/agent"
 
 if [ ! -f "$CONFIG_FILE" ]; then
   echo "Unable to find $CONFIG_FILE."
   exit 1
 fi
 
-# Unlocks the login keychain, otherwise the error of "Your session has expired" occurs shortly.
+# Unlocks the login keychain, otherwise Xcode build reports the error of
+# "Your session has expired" shortly.
 if [ "$(uname)" == "Darwin" ]; then
   security unlock-keychain login.keychain
 fi
 
 pub get
 
-dart bin/agent.dart ci -c "$CONFIG_FILE" 2>&1 >> "$HOME/agent.log" & disown
+dart bin/agent.dart ci -c "$CONFIG_FILE" 2>&1 >> "$LOG_FILE" &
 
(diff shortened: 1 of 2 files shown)


agent/bin/ci.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
godofredoc: Please add docs explaining what this script does and the expected environment it should run from. E.g it expects env a and b to be set, or if it has to be run from a certain location.

agent/bin/ci.sh
```diff
@@ -0,0 +1,32 @@
...
+set -e
+
+# Starts a Cocoon agent.
+
+USER_NAME="flutter"
+CONFIG_FILE="config.yaml"
+
+if [ "$(whoami)" != "$USER_NAME" ]; then
```
godofredoc: I think this script will only work for mac and linux how are we going to handle windows?

agent/bin/ci.sh
```diff
@@ -0,0 +1,32 @@
...
+# Unlocks the login keychain, otherwise the error of "Your session has expired" occurs shortly.
+if [ "$(uname)" == "Darwin" ]; then
+  security unlock-keychain login.keychain
+fi
+
+pub get
+
+dart bin/agent.dart ci -c "$CONFIG_FILE" 2>&1 >> "$HOME/agent.log" & disown
+
+tail -f "$HOME/agent.log"
```
godofredoc: rm? this will block forever. Also are we expecting this script to run on a user terminal or as a daemon?

agent/bin/smoke_test.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
godofredoc: Please add docs for this script.

agent/bin/smoke_test.sh
```diff
@@ -0,0 +1,29 @@
...
+set -e
+
+# Runs a smoke test in order to validate the runtime environment for Cocoon agent.
+
+LOG=$(mktemp)
+
+case "$1" in
+  ios)
+    security unlock-keychain login.keychain
+    dart bin/agent.dart run -r HEAD -t flutter_gallery_ios__start_up | tee $LOG
```
godofredoc: Can we document this script expects current directory as $HOME/cocoon/agent?

agent/bin/smoke_test.sh
godofredoc: Is this expected to run manually during the provisioning of the scripts or as part of the health checks?

agent/bin/ci.sh
digiter: Done.

agent/bin/ci.sh
digiter: We will add windows script later.

agent/bin/smoke_test.sh
digiter: Added, manual run.

agent/bin/smoke_test.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+set -e
+
+# Runs a smoke test in order to validate the runtime environment for Cocoon agent.
+
+LOG=$(mktemp)
+
+case "$1" in
+  ios)
+    security unlock-keychain login.keychain
```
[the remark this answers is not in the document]
digiter: Done.

PR #477, opened 2019-10-17
Task matrix

### LINKED ISSUE #461: Task Matrix
Build Dashboard v2 naively shows tasks based on the structure given in the first CommitStatus (the latest one). This means CommitStatuses after the first one may not align with the grid.

Instead, it should map the results of List<CommitStatus> to a matrix of Tasks where a column matches the same Task.stageName and Task.name. Rows would continue to map to the same CommitStatus.

This would improve the current lookup time of Tasks in the grid from O(Task.stages) to O(1). Additionally, it would make sure that columns in StatusGrid map to the same Task and better show Tasks that are skipped.

CaseyHillers:
Fixes #461 

This helps with:
1. Easier ordering of the columns of the StatusGrid (for sorting purposes, like showing most recent failures first)
2. Faster prototyping experimental versions of the StatusGrid to improve the performance of it on Web

This fixes the issue where the StatusGrid was naively using the first CommitStatus as the base for constructing the StatusGrid. However, extra code had to be added to make the TaskIcon row at the top show the correct icons.

### DIFF
--- a/app_flutter/lib/status_grid.dart
+++ b/app_flutter/lib/status_grid.dart
@@ -44,37 +44,60 @@ class StatusGridContainer extends StatelessWidget {
           );
         }
 
-        final Map<String, int> taskColumnKeyIndex =
-            createTaskColumnKeyIndex(statuses);
-        final List<List<Task>> taskMatrix =
-            createTaskMatrix(statuses, taskColumnKeyIndex);
-        final List<Task> taskIconRow = createTaskIconRow(statuses, taskColumnKeyIndex);
-
+        final StatusGridHelper statusGridHelper =
+            StatusGridHelper(statuses: statuses);
         return StatusGrid(
           statuses: statuses,
-          taskMatrix: taskMatrix,
-          taskIconRow: taskIconRow,
+          taskMatrix: statusGridHelper.taskMatrix,
+          taskIconRow: statusGridHelper.taskIconRow,
         );
       },
     );
   }
+}
+
+/// Class to handle data operations on [List<CommitStatus>].
+/// 
+/// Flattens the mapping of one [CommitStatus] from many [Stage] objects,
+/// where each [Stage] object maps to many [Task] objects, to a 2D matrix.
+/// 
+// TODO(chillers): Support special ordering of taskMatrix. https://github.com/flutter/cocoon/issues/478
+class StatusGridHelper {
+  StatusGridHelper({@required this.statuses}) {
+    _columnKeyIndex = _createTaskColumnKeyIndex(statuses);
+
+    _taskIconRow = _createTaskIconRow(statuses, _columnKeyIndex);
+    _taskMatrix = _createTaskMatrix(statuses, _columnKeyIndex);
+  }
+
+  final List<CommitStatus> statuses;
+
+  /// A key, value table to find what column a [Task] is in.
+  Map<String, int> _columnKeyIndex;
+
+  /// Computed 2D array of [Task] to make it easy to retrieve and sort tasks.
+  List<List<Task>> get taskMatrix => _taskMatrix;
+  List<List<Task>> _taskMatrix;
+
+  /// A list of [Task] objects that can be used for the row of [TaskIcon].
+  List<Task> get taskIconRow => _taskIconRow;
+  List<Task> _taskIconRow;
 
   /// A unique index for grouping [Task] from separate [CommitStatus].
   ///
   /// [task.stageName] and [task.name] are not unique. However, they
   /// are unique when combined together.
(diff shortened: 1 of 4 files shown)


app_flutter/lib/status_grid.dart
```diff
@@ -45,29 +44,128 @@ class StatusGridContainer extends StatelessWidget {
...
+          taskMatrix: taskMatrix,
+          taskIconRow: taskIconRow,
         );
       },
     );
   }
+
+  /// A unique index for grouping [Task] from separate [CommitStatus].
+  ///
+  /// [task.stageName] and [task.name] are not unique. However, they
```
jonahwilliams: Good idea!

app_flutter/lib/status_grid.dart
```diff
@@ -45,29 +44,128 @@ class StatusGridContainer extends StatelessWidget {
...
+      : super(key: key);
 
   /// The build status data to display in the grid.
   final List<CommitStatus> statuses;
 
+  /// Computed 2D array of [Task] to make it easy to retrieve and sort tasks.
+  final List<List<Task>> taskMatrix;
+
+  /// [Task] that can be used for the row of [TaskIcon].
+  final List<Task> taskIconRow;
```
jonahwilliams: Nit, there is more than one  task here: `A list of [Task] objects that ...`

app_flutter/lib/status_grid.dart
```diff
@@ -89,71 +187,34 @@ class StatusGrid extends StatelessWidget {
...
+              final int statusIndex = index ~/ columnCount;
               if (index % columnCount == 0) {
-                final int statusIndex = index ~/ columnCount;
                 return CommitBox(commit: statuses[statusIndex].commit);
               }
 
+              final int taskIndex = (index % columnCount) - 1;
+              if (taskMatrix[statusIndex][taskIndex] == null) {
+                /// [Task] was skipped so don't show anything.
+                return Container();
```
jonahwilliams: Nit:
```suggestion
                return const SizedBox();
```

app_flutter/lib/status_grid.dart
```diff
@@ -45,29 +44,128 @@ class StatusGridContainer extends StatelessWidget {
...
+  }
+
+  /// A map of all [taskColumnKey] to a unique index from [List<CommitStatus>].
+  ///
+  /// Scans through all [Task] in [List<CommitStatus>] to find all unique [taskColumnKey].
+  /// This ensures a column is allocated in [createTaskMatrix].
+  ///
+  /// When a new [taskColumnKey] is found, it is inserted and the index is incremeted.
+  @visibleForTesting
+  static Map<String, int> createTaskColumnKeyIndex(List<CommitStatus> statuses) {
```
jonahwilliams: Rather than creating a series of staticMethods that are visibleForTesting, I would create a separate `TaskMatrix` or similar class and collect all of the data/methods into that. Otherwise, this is a great improvement!

app_flutter/lib/status_grid.dart
CaseyHillers: I just want to point out Github does not support reacting with the smiley emoji. What a tragedy :P

app_flutter/lib/status_grid.dart
CaseyHillers: Good catch! Fixed

app_flutter/lib/status_grid.dart
CaseyHillers: Thank you!

app_flutter/lib/status_grid.dart
CaseyHillers: I created `StatusGridHelper` that can manage this. Down the line we can add functions for ordering the Tasks in the matrix there. Can you take a look?

app_flutter/lib/status_grid.dart
```diff
@@ -45,29 +44,151 @@ class StatusGridContainer extends StatelessWidget {
...
+            continue;
+          }
+
+          taskColumnKeyIndex[key] = currentIndex++;
+        }
+      }
+    }
+
+    return taskColumnKeyIndex;
+  }
```
digiter: Just want to mention another lazy style to calculate taskColumnKeyIndex, which returns the index on the fly: 
```
Map<String, int> taskColumnKeyIndex = <String, int>{};
int getColumnKeyIndex(Task task) {
  final String key = _taskColumnKey(task);
  if (taskColumnKeyIndex.containsKey(key)) {
    return taskColumnKeyIndex[key];
  }
  int id = taskColumnKeyIndex.keys.length.
  taskColumnKeyIndex[key] = id;
  return id;
}
```

PR #479, opened 2019-10-18
url launching on the web

CaseyHillers:


app_flutter/lib/commit_box.dart
```diff
@@ -132,7 +132,7 @@ class CommitOverlayContents extends StatelessWidget {
                       onPressed: () async {
                         final String githubUrl =
                             'https://github.com/${commit.repository}/commit/${commit.sha}';
-                        await launch(githubUrl);
+                        launch(githubUrl);
```
jonahwilliams: Did you need to await this future for whatever reason?

CaseyHillers: `url_launcher` example did it, but here we really don't care about when the url opens.

PR #480, opened 2019-10-18
/api/create-agent Implementation (GO->dart)

keyonghan:
This PR implements /api/create-agent, and creates corresponding tests.


app_dart/lib/src/request_handlers/create_agent.dart
```diff
@@ -0,0 +1,86 @@
...
+    final List<String> capabilities =
+        requestData[capabilitiesParam].cast<String>().toList();
+    final DatastoreService datastore = datastoreProvider();
+    final AgentService agentService = agentServiceProvider();
+    final Key key = datastore.db.emptyKey.append(Agent, id: agentId);
+
+    Agent agent;
+    AgentAuthToken agentAuthToken;
+
+    if (await datastore.db.lookupValue<Agent>(key, orElse: () => null) ==
```
tvolkert: Nit: if you structured this as `if (...) { throw ... }`, then you could set the `agent` and `agentAuthToken` variables when you declare them.

app_dart/lib/src/request_handlers/create_agent.dart
```diff
@@ -0,0 +1,86 @@
...
+        healthCheckTimestamp: DateTime.now().millisecondsSinceEpoch,
+        authToken: agentAuthToken.hash,
+        isHealthy: false,
+        key: key,
+      );
+    } else {
+      throw BadRequestException('Agent ID: $agentId already exists');
+    }
+
+    if (agent == null) {
```
tvolkert: ... and then you can get rid of this null check

app_dart/lib/src/request_handlers/create_agent.dart
keyonghan: You are right! Just updated accordingly.

app_dart/lib/src/request_handlers/create_agent.dart
keyonghan: Yes. Removed.

PR #481, opened 2019-10-18
Revert url launching on web

CaseyHillers:
Tests not run on that PR, and now tests are failing.

/cc @jonahwilliams @tvolkert Do you know why the cirrus tests were not run on #479 ?

### DIFF
--- a/app_flutter/pubspec.yaml
+++ b/app_flutter/pubspec.yaml
@@ -32,10 +32,6 @@ dependencies:
     path: ../app_dart
   provider: ^3.0.0
   url_launcher: ^5.1.0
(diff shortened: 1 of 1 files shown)


keyonghan:
For your reference, here is the info. for [failed test](https://cirrus-ci.com/task/5357304894980096?command=pub#L864)

FINE: Resolving dependencies finished (1.1s).
ERR : Because every version of url_launcher_web from git depends on flutter_web_plugins any from sdk which is forbidden, url_launcher_web from git is forbidden.
    | So, because app_flutter depends on url_launcher_web from git, version solving failed.

PR #482, opened 2019-10-21
Order StatusGrid by most recently failed

CaseyHillers:
Fixes https://github.com/flutter/flutter/issues/43093

I simplified the formula for ordering most recently failed tasks from the [flutter.dev/go/build-dashboard-v2](https://docs.google.com/document/d/1kwh48xYuoVtadYFb2I2t6BJMGEW07K7T5WYcieAZZiY/edit#heading=h.op0579dy52k7) design doc. Instead of sorting in descending order with decimals, we can sort in ascending order using integers. This makes it so we can just use a linear function of when the last occurrence was. The downside to this is it is harder to account for flaky tasks, so they are ignored for now.

Screenshot (Oct 21, 2019 10_57_57 AM)

## Open questions

- ~Is there a better way to sort columns of a matrix in Dart?~

*After this, I feel good about putting the Flutter project on the TV :)*


app_flutter/lib/status_grid.dart
```diff
@@ -143,24 +146,94 @@ class StatusGridHelper {
...
+    // Map the current index to its given weight
+    final Map<int, int> weightIndex = <int, int>{};
+    for (int i = 0; i < weights.length; i++) {
+      weightIndex[i] = weights[i];
+    }
+
+    // Sort by weight in [weightIndex]
+    final List<int> sortedWeightKeys = weightIndex.keys.toList()
+      ..sort((int k1, int k2) => weightIndex[k1].compareTo(weightIndex[k2]));
+    final LinkedHashMap<int, int> sortedWeights =
```
jonahwilliams: nit: add a note that you are using LinkedHashMap to preserve the key order

app_flutter/lib/status_grid.dart
```diff
@@ -143,24 +146,94 @@ class StatusGridHelper {
...
+        LinkedHashMap<int, int>.fromIterable(sortedWeightKeys,
+            key: (dynamic k) => k, value: (dynamic k) => weightIndex[k]);
```
jonahwilliams: Why are you using dynamic when you know the types?

app_flutter/lib/status_grid.dart
```diff
@@ -143,24 +146,94 @@ class StatusGridHelper {
...
+
+    // Reassign indices based on the order
+    int newIndex = 0;
+    sortedWeights.forEach((int key, int value) {
```
jonahwilliams: I'm a bit confused by this part. You have the sorted map of indexes/weights, you should be able to use that to look things instead of traversing the entire columnKeyIndex

app_flutter/lib/status_grid.dart
```diff
@@ -57,17 +59,18 @@ class StatusGridContainer extends StatelessWidget {
...
 /// where each [Stage] object maps to many [Task] objects, to a 2D matrix.
-/// 
+///
 // TODO(chillers): Support special ordering of taskMatrix. https://github.com/flutter/cocoon/issues/478
 class StatusGridHelper {
   StatusGridHelper({@required this.statuses}) {
     _columnKeyIndex = _createTaskColumnKeyIndex(statuses);
 
-    _taskIconRow = _createTaskIconRow(statuses, _columnKeyIndex);
     _taskMatrix = _createTaskMatrix(statuses, _columnKeyIndex);
```
jonahwilliams: You're doing a lot of different passes over the data. It would be easy to follow this algorithm if it were more compact, instead of being spread out across multiple methods.

app_flutter/lib/status_grid.dart
CaseyHillers: That's what I thought, but `The function expression type 'int Function(int)' isn't of type 'int Function(dynamic)'`.

Is there a better way of doing this, as I agree these should just be `int`.

app_flutter/lib/status_grid.dart
CaseyHillers: Added

app_flutter/lib/status_grid.dart
CaseyHillers: I realized creating a "reverse" map allowed for the quick lookup. I was able to simplify this further, thank you for pointing it out!

app_flutter/lib/status_grid.dart
CaseyHillers: Do you think it would be better to add another layer of indirection and just use a `Map<int, int>` in StatusGrid that remaps the columns of the TaskMatrix to the order they should be in? Thus we never recreate the TaskMatrix unless new data comes in.

app_flutter/lib/status_grid.dart
jonahwilliams: fromIterable is meant to be run over dynamic things. I would instead create the map and populate it yourself.

app_flutter/lib/status_grid.dart
jonahwilliams: I'm not quite sure - it's fairly difficult for me to follow along with how the task matrix is being created. One solution is to better document each step so there is a clear "1., 2. 3."

digiter: Encapsulation could help this, create classes to encapsulate data manipulations.

For example, a TaskMatrix class encapsulates a `taskIconRow` and a list of `Columns`. It has a static factory method to create itself from a list of CommitStatus. It has a sort method to sort the columns.

A Column class encapsulates a list of `Tasks` and a `Weight`. The weight is used by TaskMatrix in sorting.

app_flutter/lib/task_matrix.dart
```diff
@@ -0,0 +1,152 @@
...
+  List<Column> _matrix;
+
+  /// A map for taking in the corresponding column in [StatusGrid] and
+  /// where it maps to in [_taskMatrix].
+  Map<int, int> _columnMap;
+
+  /// A key, value table to find what column a [Task] is in.
+  Map<String, int> _columnKeyIndex;
+
+  /// Return [Task] in [_matrix] for a row and col in [StatusGrid].
```
jonahwilliams: dartdoc identifiers have to be in scope - _matrix is private so it shouldn't be mentioned

CaseyHillers: Thank you for pointing that out. Is it okay to phrase it as `"Return [Task] in this matrix for a given row and column in [StatusGrid]"`. This mentions the matrix, but doesn't give the identifier.

app_flutter/lib/status_grid.dart
CaseyHillers: Thank you both for the help! I was able to greatly simplify this using the feedback. I added `Column` as an encapsulation layer that made sorting much more concise, and I hope clearer to follow. I also renamed `StatusGridHelper` to `TaskMatrix`, which was discussed in the last PR, but makes much more sense to name it now.

app_flutter/lib/status_grid.dart
CaseyHillers: Thank you! That worked, but we went another route that doesn't require the `LinkedHashMap`.

app_flutter/lib/status_grid.dart
```diff
@@ -44,122 +45,35 @@ class StatusGridContainer extends StatelessWidget {
...
-          final int taskIndex = taskColumnKeyIndex[_taskColumnKey(task)];
-          if (taskIconRow[taskIndex] == null) {
-            taskIconRow[taskIndex] = task;
-          }
-        }
-      }
-    }
-
-    return taskIconRow;
+    return a.tasks.length;
```
digiter: This method is called O(nlogn) times during the sorting and there's only O(n) columns,  so a small optimization is to save the answer without re-calculating.

Even better, I'd make Column be immutable after creation, and let `lastFailed` be a constant property of Column. The spirit is to encapsulate low level details (lastFailed) in the low level class (Column), and let high level class (StatusGrid) decided how to utilize it.

CaseyHillers: With this performance improvement, I think we should hold off until the team has explored the v2 dashboard. This generalization makes it easy to easily implement [the different ordering techniques we discussed in the design doc](https://docs.google.com/document/d/1kwh48xYuoVtadYFb2I2t6BJMGEW07K7T5WYcieAZZiY/edit#heading=h.7a1ri2tf3g6a). If the team finds some ordering techniques to be much more useful than others, we can optimize for them by implementing these const properties as needed.

Do you have any resources I can look into for making an object immutable after it has been created?

digiter: SG.

Sorry, not immutable, my idea was to make Column.tasks read only. That's not a big deal though.

CaseyHillers:
Friendly ping

app_flutter/lib/task_matrix.dart
```diff
@@ -0,0 +1,160 @@
...
+/// Class to handle data operations on [List<CommitStatus>].
+///
+/// Flattens the mapping of one [CommitStatus] from many [Stage] objects,
+/// where each [Stage] object maps to many [Task] objects, to a 2D matrix.
+/// Simplifies logic further by moving related [Task] into a [Column], so the
+/// matrix is composed of just a list of columns.
+///
+/// After construction, the underlying matrix and statuses do not get modified.
+/// Only the column mapping is changed with additional operations.
+class TaskMatrix {
```
jonahwilliams: Spoke offline about this - the details and documentation are very solid, but I would like a high-level overview of this algorithm to make it easier for people unfamiliar with it to get up to speed quickly

app_flutter/lib/status_grid.dart
```diff
@@ -217,23 +127,22 @@ class StatusGrid extends StatelessWidget {
...
+              final Task task = taskMatrix.task(row, column);
+              if (task == null) {
                 /// [Task] was skipped so don't show anything.
                 return const SizedBox();
               }
 
-              return TaskBox(
-                task: taskMatrix[statusIndex][taskIndex],
-              );
+              return TaskBox(task: task);
```
digiter: To reduce the magic in this function, we can create another 2D matrix (say `TaskView`) that has the same dimensions as this GridView. And populate `TaskView` with simple for loops to traverse the TaskMatrix, and add CommitBox.

This approach would use more memory, but not bad on time complexity as the TestMatrix needs to be traversed anyway.

CaseyHillers: I plan to refactor `StatusGrid` after v1 parity to move out the `TaskIcon` row and `CommitBox` column from the `GridView` so both of those show no matter where you scroll in the `TaskMatrix`. Such as if you scroll far to the right, you should still see the `CommitBox` column. If you scroll to very old commits, you should still be able to see the `TaskIcon` row.

`TaskView` is a good name for what would be left of the `GridView`, and I will leave a TODO for now for https://github.com/flutter/flutter/issues/43376

app_flutter/lib/status_grid.dart
```diff
@@ -44,122 +45,35 @@ class StatusGridContainer extends StatelessWidget {
...
-            continue;
-          }
-
-          taskColumnKeyIndex[key] = currentIndex++;
-        }
-      }
-    }
 
-    return taskColumnKeyIndex;
+  /// Order columns by showing those that have failed reecently first.
```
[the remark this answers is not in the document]
CaseyHillers: Fixed! Good catch

app_flutter/lib/task_matrix.dart
CaseyHillers: I updated the overview and expanded the discussion on the sorting algorithm. Let me know if there's other parts that should be covered

CaseyHillers: Updated again using Jonah's reference of [the Wikipedia Path-based strong component algorithm page](https://en.wikipedia.org/wiki/Path-based_strong_component_algorithm).

app_flutter/lib/task_matrix.dart
```diff
@@ -0,0 +1,172 @@
...
+import 'package:cocoon_service/protos.dart' show CommitStatus, Stage, Task;
+
+/// Simplify [List<CommitStatus>] for easy retrieval of tasks and ordering of columns.
+///
+/// Construction performs the following steps:
+///   1. Scan through every [Task] in [List<CommitStatus>] to find all unique
+///      columns in the matrix. Uniqueness is defined by the key "[Task.stageName]:[Task.name]".
+///   2. Construct an empty [List<Column>] based on the size given from (1).
+///   3. Scan through [List<CommitStatus>] again, but insert every [Task] into
+///      their associated [Column] at the row they appear in.
```
jonahwilliams: nit: what is "at the row they appear in."

app_flutter/lib/task_matrix.dart
```diff
@@ -0,0 +1,172 @@
...
+///      A. If a [Column.sampleTask] is null, set it to this [Task].
```
jonahwilliams: nit: what is "this [Task]"

CaseyHillers: Reworded

app_flutter/lib/task_matrix.dart
CaseyHillers: Reworded. Let me know if I need to clarify further.

PR #484, opened 2019-10-23
/api/public/get-status serves cached responses

CaseyHillers:
[Faster Cocoon Design Doc](https://flutter.dev/go/faster-cocoon)

- Create `CacheRequestHandler` that returns a cache response, or falls back to the old handler
- Create a cron job to update the cached response every minute (moved the current `/api/public/get-status` handler to be found at `/api/public/update-status` and added code to store the response in redis)
- `GetStatus` handler was moved to `UpdateStatusCache`

## Impact
- `/api/public/get-status` latency dropped from `6.16s` to `0.74s` using CURL for benchmarking, an `87.93%` speedup

## Future Work
- Instead of a cron job updating the cache, writers should schedule the next cache refresh (To be done in the next week)
- Switching the JSON response to JSPB will reduce the size, and will further reduce the response time (albeit not another 80%+ speedup)
- Extend this to the other high trafficked endpoints as listed in the design doc

Demo: http://testchillers.flutter-dashboard.appspot.com/build.html
Currently, the cron job doesn't run on this service. If you query this when no one has checked in the past hour, you can see the fallback handler working.


app_dart/lib/src/request_handling/cache_request_handler.dart
```diff
@@ -0,0 +1,53 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:convert';
+
+import 'package:meta/meta.dart';
+import 'package:neat_cache/neat_cache.dart';
+
+import 'package:cocoon_service/cocoon_service.dart';
```
tvolkert: Our convention has been to use relative imports for files you need.

app_dart/lib/src/request_handling/cache_request_handler.dart
```diff
@@ -0,0 +1,53 @@
...
+/// This should be [fallbackHandler], but does not need to be.
+@immutable
+class CachedRequestHandler extends RequestHandler<Body> {
+  /// Creates a new [CachedRequestHandler].
+  const CachedRequestHandler(this.responseKey, this.fallbackHandler,
+      {@required Config config, @required this.cache})
+      : super(config: config);
+
+  /// The key in the subcache for responses that stores this response.
+  final String responseKey;
```
tvolkert: Can we just use the request URI as the cache key?

app_dart/lib/src/request_handling/cache_request_handler.dart
```diff
@@ -0,0 +1,53 @@
...
+
+  /// [RequestHandler] that queries Datastore for 
+  final RequestHandler<Body> fallbackHandler;
```
tvolkert: Nit: I'd call this `delegate`

app_dart/lib/src/request_handling/cache_request_handler.dart
```diff
@@ -0,0 +1,53 @@
...
+class CachedRequestHandler extends RequestHandler<Body> {
+  /// Creates a new [CachedRequestHandler].
+  const CachedRequestHandler(this.responseKey, this.fallbackHandler,
+      {@required Config config, @required this.cache})
+      : super(config: config);
+
+  /// The key in the subcache for responses that stores this response.
+  final String responseKey;
+
+  /// [RequestHandler] that queries Datastore for
```
tvolkert: Looks like these docs need to be updated

app_dart/lib/src/request_handling/cache_request_handler.dart
```diff
@@ -0,0 +1,53 @@
...
+/// A [RequestHandler] for serving cached responses.
+/// 
+/// High trafficked endpoints that have responses that do not change
+/// based on request are good for caching. Additionally, saves
+/// reading from Datastore which is expensive both timewise and monetarily.
+/// 
+/// Implementing requires a writer that will keep [responseKey] in the cache updated.
+/// This should be [fallbackHandler], but does not need to be.
+@immutable
+class CachedRequestHandler extends RequestHandler<Body> {
```
tvolkert: This should probably be `CachedRequestHandler<T extends Body> extends RequestHandler<T>`

app_dart/lib/src/request_handling/cache_request_handler.dart
tvolkert: Then this becomes `RequestHandler<T>`

app_dart/lib/src/request_handling/cache_request_handler.dart
```diff
@@ -0,0 +1,53 @@
...
+  /// Creates a new [CachedRequestHandler].
+  const CachedRequestHandler(this.responseKey, this.fallbackHandler,
```
tvolkert: Nit: make these all named `@required`

app_dart/lib/src/request_handling/cache_request_handler.dart
```diff
@@ -0,0 +1,53 @@
...
+    final Cache<String> responseCache =
+        cache.withPrefix(await config.redisResponseSubcache).withCodec(utf8);
+
+    final String cachedResponse = await responseCache[responseKey].get();
+
+    if (cachedResponse != null) {
+      final Map<String, dynamic> jsonResponse = jsonDecode(cachedResponse);
+      return Body.forJson(jsonResponse);
+    } else {
+      return fallbackHandler.get();
```
tvolkert: Why not have this populate the cache?  You can serialize the body to get the bytes, and stick that directly in the cache, without worrying about utf8.  Then when you get a cache hit, you use `Body.forStream()` to reconstitute the body.

app_dart/lib/src/request_handlers/update_status_cache.dart
```diff
@@ -0,0 +1,97 @@
...
+import '../request_handling/request_handler.dart';
+import '../service/build_status_provider.dart';
+import '../service/datastore.dart';
+
+/// Cache writer for /api/public/get-status.
+/// 
+/// Queries datastore to get the latest information on the commits
+/// to flutter/flutter and the agents in the devicelab.
+@immutable
+class UpdateStatusCache extends RequestHandler<Body> {
```
tvolkert: We shouldn't need this handler -- if the caching handler populates the cache entries whenever it has to delegate to the underlying handler, then you get free cache population by virtue of wrapping a handler in a caching handler.

CaseyHillers: Currently it does not do that. However, this is a much more elegant solution and I have implemented it that way.

app_dart/lib/src/request_handling/cache_request_handler.dart
CaseyHillers: Good catch! VS Code would care to disagree :)
Fixed!

app_dart/lib/src/request_handling/cache_request_handler.dart
CaseyHillers: Yes. This is left over from prototyping. Updated!

app_dart/lib/src/request_handling/cache_request_handler.dart
CaseyHillers: Added

app_dart/lib/src/request_handling/cache_request_handler.dart
CaseyHillers: Fixed

app_dart/lib/src/request_handling/cache_request_handler.dart
CaseyHillers: Added

app_dart/lib/src/request_handling/cache_request_handler.dart
CaseyHillers: 👍

app_dart/lib/src/request_handling/cache_request_handler.dart
CaseyHillers: Added

app_dart/lib/src/request_handling/cache_request_handler.dart
CaseyHillers: Added, but do you know if there's a better way of serializing?

Right now I do `      final Stream<Uint8List> response =
          Stream<Uint8List>.fromIterable(cachedResponse.cast<Uint8List>());`
which works, but I feel like there is a much cleaner way.

digiter:
This doesn't need to be address in this PR. But think about adding cache to other API endpoints. Some may want a different timeout, therefore the cache duration needs to be a parameter. Some may have already extended a subclass of RequestHandler, therefore the cache class needs to be a mixin.

app_dart/test/request_handling/cache_request_handler_test.dart
```diff
@@ -0,0 +1,88 @@
...
+          CachedRequestHandler<Body>(
+              fallbackDelegate: fallbackHandlerMock,
+              cache: cache,
+              config: config);
+
+      final Body body = await tester.get(cacheRequestHandler);
+      final String response = await _decodeHandlerBody(body);
+      final Map<String, dynamic> jsonResponse = jsonDecode(response);
+
+      expect(jsonResponse, expectedJsonResponse);
```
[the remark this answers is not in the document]
CaseyHillers: Added

app_dart/lib/src/request_handling/cache_request_handler.dart
```diff
@@ -0,0 +1,68 @@
...
+      return body;
+    }
+  }
+
+  /// Update cache with the latest response.
+  ///
+  /// This response will be served for the next minute of requests.
+  Future<void> updateCache(Body body) async {
+    final Cache<List<int>> responseCache =
+        cache.withPrefix(await config.redisResponseSubcache);
```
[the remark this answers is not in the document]
CaseyHillers: Good point. Added

app_dart/lib/src/request_handling/cache_request_handler.dart
```diff
@@ -0,0 +1,68 @@
...
+    final List<int> cachedResponse =
+        await responseCache[request.uri.path].get();
+
+    if (cachedResponse != null) {
+      final Stream<Uint8List> response =
+          Stream<Uint8List>.fromIterable(cachedResponse.cast<Uint8List>());
+      return Body.forStream(response);
+    } else {
+      final Body body = await fallbackDelegate.get();
+      await updateCache(body);
```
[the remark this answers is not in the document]
CaseyHillers: Good catch. Added

app_dart/bin/server.dart
```diff
@@ -20,6 +22,9 @@ Future<void> main() async {
       accessTokenProvider: AccessTokenProvider(config),
     );
 
+    final CacheProvider<List<int>> redisCacheProvider = Cache.redisCacheProvider(await config.redisUrl);
```
tvolkert: We should probably refactor this into a `CacheService` (`lib/src/service/cache_service.dart`).  Then `CachedRequestHandler` can take a `CacheService` in its constructor rather than a `Cache`.  This will let us abstract away the type of cache and provide it as a `Cache<Uint8List>` out of the box:

```dart
@immutable
class CacheService {
  const CacheService(this. config);

  final Config config;

  Future<Cache<Uint8List>> redisCache() async {
    final CacheProvider<List<int>> provider = Cache.redisCacheProvider(await config.redisUrl);
    return Cache<List<int>>(provider).withCodec<Uint8List>(const _CacheCodec());
  }
}

class _CacheCodec extends Codec<Uint8List, List<int>> {
  const _CacheCodec();

  @override
  Converter<Uint8List, List<int>> get encoder => const _ListIntConverter();

  @override
  Converter<List<int>, Uint8List> get decoder => const _Uint8ListConverter();
}

class _ListIntConverter extends Converter<Uint8List, List<int>> {
  const _ListIntConverter();

  List<int> convert(Uint8List input) => input;
}

class _Uint8ListConverter extends Converter<List<int>, Uint8List> {
  const _Uint8ListConverter();

  Uint8List convert(List<int> input) => Uint8List.fromList(input);
}
```

app_dart/lib/src/request_handlers/update_status_cache.dart
tvolkert: That's how you're gonna create the stream from the list, but you could create a `Body.forBytes(Uint8List bytes)` factory constructor in `Body` to abstract this away, so your call here would become `return Body.forBytes(cachedResponse)`.  (`cachedResponse` will be a `Uint8List` after you make the other suggested changes)

app_dart/cron.yaml
```diff
@@ -19,3 +19,4 @@ cron:
 - description: refresh github CI status (cirrus)
   url: /api/refresh-cirrus-status
   schedule: every 2 minutes
+
```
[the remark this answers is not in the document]
CaseyHillers: Untouched!

app_dart/lib/src/request_handling/cache_request_handler.dart
```diff
@@ -0,0 +1,69 @@
...
+import 'package:neat_cache/neat_cache.dart';
+import 'package:pedantic/pedantic.dart';
+
+import '../datastore/cocoon_config.dart';
+import '../request_handling/request_handler.dart';
+import 'body.dart';
+
+/// A [RequestHandler] for serving cached responses.
+///
+/// High trafficked endpoints that have responses that do not change
```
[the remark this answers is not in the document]
CaseyHillers: Fixed

app_dart/lib/src/request_handling/cache_request_handler.dart
```diff
@@ -0,0 +1,69 @@
...
+/// based on request are good for caching. Additionally, saves
+/// reading from Datastore which is expensive both timewise and monetarily.
+@immutable
+class CachedRequestHandler<T extends Body> extends RequestHandler<T> {
+  /// Creates a new [CachedRequestHandler].
+  const CachedRequestHandler(
+      {@required this.fallbackDelegate,
+      @required Config config,
+      @required this.cache,
+      Duration timeToLiveValue})
```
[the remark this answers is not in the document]
CaseyHillers: Added

app_dart/lib/src/request_handling/cache_request_handler.dart
```diff
@@ -0,0 +1,69 @@
...
+      : timeToLive = timeToLiveValue ?? const Duration(minutes: 1),
+        super(config: config);
+
+  /// [RequestHandler] that queries Datastore for the response.
+  final RequestHandler<T> fallbackDelegate;
+
+  final Cache<List<int>> cache;
+
+  final Duration timeToLive;
```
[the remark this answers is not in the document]
CaseyHillers: Changed

app_dart/lib/src/request_handling/cache_request_handler.dart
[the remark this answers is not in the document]
CaseyHillers: Changed

app_dart/lib/src/request_handling/cache_request_handler.dart
```diff
@@ -0,0 +1,69 @@
...
+/// A [RequestHandler] for serving cached responses.
+///
+/// High trafficked endpoints that have responses that do not change
+/// based on request are good for caching. Additionally, saves
+/// reading from Datastore which is expensive both timewise and monetarily.
+@immutable
+class CachedRequestHandler<T extends Body> extends RequestHandler<T> {
+  /// Creates a new [CachedRequestHandler].
+  const CachedRequestHandler(
+      {@required this.fallbackDelegate,
```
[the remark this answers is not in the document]
CaseyHillers: Updated

app_dart/lib/src/request_handling/cache_request_handler.dart
```diff
@@ -0,0 +1,69 @@
...
+      @required Config config,
+      @required this.cache,
+      Duration timeToLiveValue})
+      : timeToLive = timeToLiveValue ?? const Duration(minutes: 1),
+        super(config: config);
+
+  /// [RequestHandler] that queries Datastore for the response.
```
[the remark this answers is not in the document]
CaseyHillers: Updated

app_dart/lib/src/request_handling/cache_request_handler.dart
```diff
@@ -0,0 +1,69 @@
...
+  final Duration timeToLive;
+
+  /// Services a cached request.
+  @override
+  Future<T> get() async {
+    final Cache<List<int>> responseCache =
+        cache.withPrefix(await config.redisResponseSubcache);
+
+    final List<int> cachedResponse =
+        await responseCache[request.uri.path].get();
```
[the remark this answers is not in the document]
CaseyHillers: Added

app_dart/lib/src/request_handling/cache_request_handler.dart
```diff
@@ -0,0 +1,70 @@
...
+import '../request_handling/request_handler.dart';
+import 'body.dart';
+
+/// A [RequestHandler] for serving cached responses.
+///
+/// High traffic endpoints that have responses that do not change
+/// based on request are good for caching. Additionally, saves
+/// reading from Datastore which is expensive both timewise and monetarily.
+@immutable
+class CachedRequestHandler<T extends Body> extends RequestHandler<T> {
```
tvolkert: _Super-nit_: The class name is `CachedRequestHandler`, but the file name is `cache_request_handler.dart`.  The two should match.

app_dart/lib/src/request_handling/cache_request_handler.dart
```diff
@@ -0,0 +1,70 @@
...
+  /// Creates a new [CachedRequestHandler].
+  const CachedRequestHandler({
+    @required this.delegate,
+    @required Config config,
+    @required this.cache,
+    Duration ttlValue,
```
tvolkert: You can use a default value and set the field inline here (no need for the constructor initializer):

```dart
    this.ttl = const Duration(minutes: 1),
```

app_dart/lib/src/request_handling/cache_request_handler.dart
```diff
@@ -0,0 +1,70 @@
...
+  @override
+  Future<T> get() async {
+    final Cache<Uint8List> responseCache =
+        cache.withPrefix(await config.redisResponseSubcache);
+
+    final String responseKey = '${request.uri.path}:${request.uri.query}';
+    final Uint8List cachedResponse = await responseCache[responseKey].get();
+
+    if (cachedResponse != null) {
+      final Stream<Uint8List> response =
```
tvolkert: Nit: no need for the `response` variable here -- you can just inline:

```dart
      return Body.forStream(Stream<Uint8List>.value(cachedResponse));
```

app_dart/lib/src/request_handling/cache_request_handler.dart
```diff
@@ -0,0 +1,70 @@
...
+      unawaited(updateCache(responseCache, body));
+
+      return body;
+    }
+  }
+
+  /// Update cache with the latest response.
+  ///
+  /// This response will be served for the next [ttl] duration of requests.
+  Future<void> updateCache(Cache<List<int>> responseCache, Body body) async {
```
tvolkert: It's not clear that factoring this into a separate method buys us anything.  The method's only called once, and now we have to construct the cache key in two places, which is fragile.  I'd just inline this method back into the body of `get()`.  You can still use unawaited:

```dart
      final Uint8List serializedBody = ...
      unawaited(() => responseCache[responseKey].set(serializedBody, ttl));
```

app_dart/lib/src/request_handling/cache_request_handler.dart
```diff
@@ -0,0 +1,70 @@
...
+    final Uint8List serializedBody = await body.serialize().first;
```
tvolkert: The stream isn't guaranteed to only contain one chunk of bytes.  This should be:

```dart
    final List<int> rawBytes = await body
        .serialize()
        .expand<int>((Uint8List chunk) => chunk)
        .toList();
    final Uint8List bytes = Uint8List.fromList(rawBytes);
```

app_dart/test/request_handling/cache_request_handler_test.dart
```diff
@@ -0,0 +1,85 @@
...
+
+      await responseCache['$testHttpPath:'].set(serializedBody);
+
+      final CachedRequestHandler<Body> cacheRequestHandler =
+          CachedRequestHandler<Body>(
+              delegate: fallbackHandlerMock, cache: cache, config: config);
+
+      final Body body = await tester.get(cacheRequestHandler);
+      final Uint8List response = await body.serialize().first;
+      final String strResponse = String.fromCharCodes(response);
```
tvolkert: We should use `utf8.decode()` here instead of `String.fromCharCodes()`

app_dart/lib/src/request_handling/cache_request_handler.dart
CaseyHillers: Good catch! Renamed class to `CacheRequestHandler`.

app_dart/lib/src/request_handling/cache_request_handler.dart
CaseyHillers: Thank you!

app_dart/test/request_handling/cache_request_handler_test.dart
CaseyHillers: Added. Is this because `String.fromCharCodes()` does UTF16?

app_dart/lib/src/request_handling/cache_request_handler.dart
CaseyHillers: Thank you! I just noticed this in manual testing.

app_dart/lib/src/request_handling/cache_request_handler.dart
CaseyHillers: Done

app_dart/lib/src/request_handling/cache_request_handler.dart
CaseyHillers: Done

app_dart/test/request_handling/cache_request_handler_test.dart
tvolkert: Yep.

PR #486, opened 2019-10-24
Reland url launcher for web

CaseyHillers:


CaseyHillers:
Used matrix ordering branch on accident

PR #489, opened 2019-10-25
[WIP] TV refreshes itself once a day

CaseyHillers:
image

This will keep the TV relatively up to date without any human intervention. @jmagman did this for the repository dashboard, so this just seemed like a good idea to port over.

## Tested

I set the duration to 5 seconds, and it refreshed.

### DIFF
--- a/app_flutter/lib/main.dart
+++ b/app_flutter/lib/main.dart
@@ -10,18 +10,21 @@ import 'package:flutter/material.dart';
 
 import 'build_dashboard.dart';
 
-void main() => runApp(MyApp());
+void main() {
+  if (kIsWeb) {
+    /// This application is displayed on a TV that is hard to reach.
+    /// Refreshing this page ensures the TV stays relatively up to date without
+    /// anyone having to get on a ladder to use the TV as a computer.
+    Timer.periodic(
+        const Duration(days: 1), (_) => html.window.location.reload());
+  }
+  
+  runApp(MyApp());
+}
 
 class MyApp extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
-    if (kIsWeb) {
-      /// This application is displayed on a TV that is hard to reach.
-      /// Refreshing this page ensures the TV stays relatively up to date without
-      /// anyone having to get on a ladder to use the TV as a computer.
-      Timer.periodic(const Duration(days: 1), (_) => html.window.location.reload());
-    }
-
     return MaterialApp(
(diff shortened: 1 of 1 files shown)


CaseyHillers:
Moving to WIP as Flutter builds for more than web :)

app_flutter/lib/main.dart
```diff
@@ -11,6 +15,13 @@ void main() => runApp(MyApp());
 class MyApp extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
+    if (kIsWeb) {
+      /// This application is displayed on a TV that is hard to reach.
+      /// Refreshing this page ensures the TV stays relatively up to date without
+      /// anyone having to get on a ladder to use the TV as a computer.
+      Timer.periodic(const Duration(days: 1), (_) => html.window.location.reload());
```
jonahwilliams: I would put this in main, before runapp. @nturgut @mdebbar is there a way to refresh the page using the regular navigator controls?

mdebbar: @jonahwilliams not that I know of. We could implement this as a platform message that the engine receives and refreshes the page. What would this be equivalent to on mobile? Close the app and reopen it? Is that even possible?

jonahwilliams: No, I don't think so. I wouldn't worry about adding the functionality, just curious.

CaseyHillers: If that's the case, is there a way to have a conditional import have `dart:html` only on the Web platform.

jonahwilliams:
Well, you could create a `main_tv.dart` with this reset logic in it - that would also require you to host multiple versions of this app. I'm not sure you can do any TV based feature detection from a browser

app_flutter/lib/main.dart
```diff
@@ -2,11 +2,25 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'dart:async';
+import 'dart:html' as html;
```
jonahwilliams: FYI, if you import dart:html at all you can no longer build for android or iOS. You should attempt to put this behind a plugin interface

CaseyHillers:
https://github.com/flutter/cocoon/pull/487 fixed the Cirrus issue with this, however, it still breaks building for Android and iOS, so I am putting this on hold for now.

CaseyHillers:
Closing this since I implemented it just for web at https://github.com/flutter/cocoon/pull/491

PR #493, opened 2019-10-25
Update authentication.dart overview documentation

CaseyHillers:
Looks like a new authentication method was added


keyonghan:
which authentication are you referring to?

CaseyHillers:
This is an artifact from #370

PR #494, opened 2019-10-28
Cache get-benchmarks

CaseyHillers:
Following #484, this caches the endpoint that returns all the time series history for the different benchmarks. Currently, it takes ~37 seconds to return, and this will bring it to under a second when returning from cache.

## Future Work
- Caching the `get-timeseries-history` endpoint, but we need to move the POST parameter to GET params. `CacheRequestHandler` stores responses by the URI and query parameters, and does not use the POST body.

### DIFF
--- a/app_dart/bin/server.dart
+++ b/app_dart/bin/server.dart
@@ -46,11 +46,7 @@ Future<void> main() async {
       '/api/debug/get-task-by-id': DebugGetTaskById(config, authProvider),
       '/api/debug/reset-pending-tasks': DebugResetPendingTasks(config, authProvider),
 
-      '/api/public/build-status': CacheRequestHandler<Body>(
-        cache: await cacheService.redisCache(),
-        config: config,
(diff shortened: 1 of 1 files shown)


app_dart/bin/server.dart
```diff
@@ -47,7 +47,12 @@ Future<void> main() async {
       '/api/debug/reset-pending-tasks': DebugResetPendingTasks(config, authProvider),
 
       '/api/public/build-status': GetBuildStatus(config),
-      '/api/public/get-benchmarks': GetBenchmarks(config),
+      '/api/public/get-benchmarks': CacheRequestHandler<Body>(
+        cache: await cacheService.redisCache(),
```
tvolkert: You're calling `await cacheService.redisCache()` twice -- can you store it as a local variable and reuse it?

CaseyHillers: Added

PR #495, opened 2019-10-28
Google Sign In Service

CaseyHillers:
Added Google Sign In service to `FlutterBuildState`. This gives easy access to authentication from the build dashboard.

Since Flutter's [google_sign_in](https://pub.dev/packages/google_sign_in) does not support web, this currently uses an open sourced fork called [google_sign_in_all](https://pub.dev/packages/google_sign_in_all).

This fixes https://github.com/flutter/flutter/issues/43108 bringing us one pull request closer to feature parity with v1.

## Preview
sign_in_button
user_avatar

## Open Issues
- The user avatar image loaded has its colors inverted
  - I think this is a bug in the `google_sign_in_all`, but because of the low priority I have not investigated further

## Future Work
- Pass `accessToken` to `CocoonService` for authenticated requests
  - Necessary for rerunning tasks and viewing logs
- Convert use of `google_sign_in_all` plugin to `google_sign_in` when web support is added
  - Convert the user image widget to the `google_sign_in` `GoogleUserCircleAvatar` widget
- Move `GoogleSignInService` to a more global scope when new pages are added


jonahwilliams:
> The user avatar image loaded has its colors inverted

IIRC this is the case if you don't specify what size image you want.

app_flutter/lib/service/google_authentication.dart
```diff
@@ -0,0 +1,45 @@
...
+
+import 'package:google_sign_in_all/google_sign_in_all.dart';
+
+/// Service class for interacting with Google Sign In authentication for Cocoon backend.
+class GoogleSignInService {
+  /// Creates a new [GoogleSignIn].
+  GoogleSignInService({GoogleSignIn googleSignIn})
+      : _googleSignIn = googleSignIn ??
+            setupGoogleSignIn(
+              scopes: <String>[
```
jonahwilliams: Maybe hoist these into private consts

CaseyHillers:
> > The user avatar image loaded has its colors inverted
> 
> IIRC this is the case if you don't specify what size image you want.

https://stackoverflow.com/questions/42893664/firebase-photourl-from-a-google-auth-provider-returns-a-jpg-with-colors-inverted

So this is a Googler only bug?

jonahwilliams:
Ahh weird, probably fine though :)

app_flutter/lib/service/google_authentication.dart
CaseyHillers: Added

ditman:
So close! https://github.com/flutter/plugins/pull/2244

PR #496, opened 2019-10-29
[WIP] TaskHelper for opening source configuration from TaskIcon

CaseyHillers:
This makes it so when you click `TaskIcon`, you are redirected to its configuration.

I used this PR as an opportunity to refactor some of the common `Task` logic into a `TaskHelper` class. Most of the logic created in this PR will want to be reused by `TaskBox` when we implement viewing logs.

Fixes https://github.com/flutter/flutter/issues/43094

## Future Work
- Add `logUrl` getter in `TaskHelper` to return the URL to open for viewing the log of a `Task`

### DIFF
--- a/app_flutter/lib/task_helper.dart
+++ b/app_flutter/lib/task_helper.dart
@@ -1,0 +1,66 @@
+// Copyright (c) 2019 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:cocoon_service/protos.dart' show Task;
+
+/// A helper class for common utilities done with a [Task].
+class TaskHelper {
+  /// Base URLs for various endpoints that can relate to a [Task].
+  static const String flutterGithubSourceUrl =
+      'https://github.com/flutter/flutter/blob/master';
+  static const String cirrusUrl =
+      'https://cirrus-ci.com/github/flutter/flutter';
+  static const String luciUrl = 'https://ci.chromium.org/p/flutter';
+
+  /// [Task.stageName] that maps to StageName enums.
+  // TODO(chillers): Remove these and use StageName enum when available. https://github.com/flutter/cocoon/issues/441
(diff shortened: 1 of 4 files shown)


app_flutter/lib/task_helper.dart
```diff
@@ -0,0 +1,66 @@
+// Copyright (c) 2019 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:cocoon_service/protos.dart' show Task;
+
+/// A helper class for common utilities done with a [Task].
+class TaskHelper {
```
jonahwilliams: See https://dart.dev/guides/language/effective-dart/design#avoid-defining-a-class-that-contains-only-static-members

Its okay for the strings, but I would suggest leaving the methods as top levels

CaseyHillers: Added. Thanks for sharing the resource!

digiter: I thought Jonah's comment is about moving the methods out of the class.

CaseyHillers: `... you don’t need a class just to define something.` This is a cool feature!

I moved the helper functions to the top level, and left the strings in a StageName class. This should be easy to swap out when the enums are created in the protobuf

app_flutter/lib/task_helper.dart
```diff
@@ -0,0 +1,68 @@
...
+  static const String devicelab = 'devicelab';
+  static const String devicelabWin = 'devicelab_win';
+  static const String devicelabIOs = 'devicelab_ios';
+}
+
+/// Get the URL for [Task] that shows its configuration.
+///
+/// Devicelab tasks are stored in the flutter/flutter Github repository.
+/// Luci tasks are stored on Luci.
+/// Cirrus tasks are stored on Cirrus.
```
jonahwilliams: Nit: document what/when it throws exceptions.

```
/// Throws [Exception] if ...
```

PR #500, opened 2019-10-30
Migrate Checklist from datastore to bigquery - /api/migrate-checklist

keyonghan:
This PR implements /api/migrate-checklist to migrate Checklist data from datastore to bigquery. Each time it migrates 10 latest unexported commits, and marks them as exported.

This API will be called using cronjob after merged.


app_dart/lib/src/request_handlers/migrate_checklist.dart
```diff
@@ -0,0 +1,86 @@
...
+  })  : assert(datastoreProvider != null),
+        tabledataResourceApi = tabledataResourceApi ?? TabledataResourceApi,
+        super(config: config, authenticationProvider: authenticationProvider);
+
+  final DatastoreServiceProvider datastoreProvider;
+  final TabledataResourceApi tabledataResourceApi;
+
+  @override
+  Future<MigrateChecklistResponse> get() async {
+    const int maxRecords = 10;
```
dnfield: Why 10?

keyonghan: Based on our daily commits number, 10 seems to be a reasonably enough number (or use a smaller). I am thinking about running cronjob hourly. What do you think?

dnfield: We should add a comment explaining the rationale for this number. At some point it may need adjustment and it will be helpful to understand what the right count should be.

10 also just seems small to me. Why not 100 or 1000?

keyonghan: Based on discussion with @dnfield , increased the number to 100.

Another thing to clarify, this API only copies data from Datastore to Bigquery, rather than moves data. 

The main purpose is for data analysis based on BigQuery.

keyonghan:
An alternative PR [505](https://github.com/flutter/cocoon/pull/505) has been merged. Closing this one.

PR #501, opened 2019-10-30
[Agent] Disown the agent process.

digiter:
Also add `pub get` before a smoke test.

### DIFF
--- a/agent/bin/ci.sh
+++ b/agent/bin/ci.sh
@@ -34,9 +34,7 @@ fi
 
 pub get
 
-dart bin/agent.dart ci -c "$CONFIG_FILE" >> "$LOG_FILE" 2>&1 &
-
-disown
+dart bin/agent.dart ci -c "$CONFIG_FILE" >> "$LOG_FILE" 2>&1 & disown
 
 pid="$(pgrep -f bin/agent.dart)" || true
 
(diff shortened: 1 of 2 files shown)


digiter:
@dnfield Do you mind taking a look at the full content of the scripts? I'd like to get advices on them.

dnfield:
They seem sane enough to me - what are you looking for in particular?

/cc @gspencergoog who's more of a bash expert than I

agent/bin/ci.sh
```diff
@@ -36,6 +36,8 @@ pub get
 
 dart bin/agent.dart ci -c "$CONFIG_FILE" 2>&1 >> "$LOG_FILE" &
 
+disown
```
gspencergoog: I'd put this at the end of the previous line, after the "&", so it's clear what it applies to.

agent/bin/ci.sh
```diff
@@ -36,6 +36,8 @@ pub get
 
 dart bin/agent.dart ci -c "$CONFIG_FILE" 2>&1 >> "$LOG_FILE" &
```
gspencergoog: I think this should be:
```suggestion
dart bin/agent.dart ci -c "$CONFIG_FILE" >> "$LOG_FILE" 2>&1 &
```
In order to redirect the error output that you want.

digiter:
Nothing particular, just advices from a bash expert.

PR #502, opened 2019-10-31
Cocoon authentication accepts access tokens as a method of authentication

CaseyHillers:
This allows authenticating in the Cocoon backend via a Google Auth access token sent through the `X-Flutter-AccessToken` header.

The `google_sign_in_all` plugin the Flutter app is using for handling Google Auth returns a null id token, but does return a valid access token. Debugging lead to a rabbit hole of where the issue could be occurring, so I decided to just add authenticating via access tokens. The process is very similar to authenticating for id tokens which the backend already does.

## Tested
- Added unit tests for the different authentication cases (based heavily off the id token's tests)
- Deployed to my test version on the AppEngine project and verified it worked with a working access token. Tested with an expired one too. (via curl)

## Future Work
- Use this authentication method for the Flutter App's actions to the backend that require authentication
- When Flutter's `google_sign_in` has web support, we can switch back to id token.

### DIFF
--- a/app_dart/lib/src/request_handling/authentication.dart
+++ b/app_dart/lib/src/request_handling/authentication.dart
@@ -54,7 +54,7 @@ import 'exceptions.dart';
 ///     account or is a whitelisted account in the Cocoon backend.
 /// 
 ///  4. If the request has the `'X-Flutter-IdToken'` HTTP header set to a valid
-///     JWT token, then this will follow the same logic as #3.
+///     access token, then this will follow the same logic as #3.
 ///
 /// If none of the above authentication methods yield an authenticated
 /// request, then the request is unauthenticated, and any call to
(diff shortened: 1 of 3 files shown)


app_dart/lib/src/request_handling/authentication.dart
```diff
@@ -180,11 +185,81 @@ class AuthenticationProvider {
...
+          'access_token': accessToken,
+        },
+      );
+      final HttpClientRequest verifyAccessTokenRequest = await client.getUrl(googleOauthTokenInfoUri);
+      final HttpClientResponse verifyAccessTokenResponse = await verifyAccessTokenRequest.close();
+
+      /// This also checks for when an access token is expired.
+      if (verifyAccessTokenResponse.statusCode != HttpStatus.ok) {
+        final String body = await utf8.decodeStream(verifyAccessTokenResponse);
+        log.warning('Access token verification failed: ${verifyAccessTokenResponse.statusCode}; $body');
```
dnfield: I suspect this will be really noisy. Will this log be helpful for us?

app_dart/lib/src/request_handling/authentication.dart
```diff
@@ -180,11 +185,81 @@ class AuthenticationProvider {
...
+      try {
+        accessTokenData = json.decode(accessTokenJson);
+      } on FormatException {
+        throw InternalServerError('Invalid JSON: "$accessTokenJson"');
+      }
+
+      final String clientId = await _config.oauthClientId;
+      assert(clientId != null);
+      if (accessTokenData['audience'] != clientId) {
+        log.warning('Possible forged token: "${accessTokenData['audience']}" (expected "$clientId")');
```
dnfield: Ditto on the log here.

If these are helpful for debugging auth flows, mayeb just log.debug?

app_dart/lib/src/request_handling/authentication.dart
CaseyHillers: This is mirrored from the id token authentication to maintain consistency

app_dart/lib/src/request_handling/authentication.dart
CaseyHillers: This is mirrored from the id token authentication to maintain consistency

jonahwilliams:
This should be good to submit, if there are any issues we can always roll back :)

CaseyHillers:
@tvolkert feel free to let me know if there any changes and I will gladly follow up :)

PR #503, opened 2019-10-31
Rerun tasks

CaseyHillers:
Waiting on https://github.com/flutter/cocoon/pull/502 for integration testing.

Fixes https://github.com/flutter/flutter/issues/43109 leaving just viewing logs for feature parity with the v1 dashboard.

Added dependency on `flutter_progress_button` for pretty async buttons (MIT license) and added UI for rerunning a `Task` for devicelab tasks in `TaskOverlay`.

Added to `FlutterBuildState` the ability to rerun tasks (forwards to `CocoonService`). `CocoonService` has code for making authenticated requests.

When an error occurs, it will show a `SnackBar` saying an error occurred rerunning the task. Otherwise, it will show a `SnackBar` saying the task is rerunning and it may take a minute to propagate. This is because `/api/public/get-status` cache TTL is 1 minute, and we do not want to mutate the local state to have it pop back to failed.

task_rerunning

### DIFF
--- a/app_flutter/lib/service/appengine_cocoon.dart
+++ b/app_flutter/lib/service/appengine_cocoon.dart
@@ -26,7 +26,8 @@ class AppEngineCocoonService implements CocoonService {
   /// The Cocoon API endpoint to query
   ///
   /// This is the base for all API requests to cocoon
-  static const String _baseApiUrl = 'https://flutter-dashboard.appspot.com/api';
+  static const String _cocoonHost = 'https://flutter-dashboard.appspot.com';
+  static const String _baseApiUrl = '$_cocoonHost/api';
 
   final http.Client _client;
 
@@ -85,10 +86,10 @@ class AppEngineCocoonService implements CocoonService {
   @override
   Future<bool> rerunTask(Task task, String accessToken) async {
     final http.Request request = http.Request(
-        'POST', Uri(host: '$_baseApiUrl', path: '/reset-devicelab-task'));
-    request.bodyFields = <String, String>{
-      'Key': task.key.toString(),
-    };
+        'POST', Uri(host: _cocoonHost, path: '/api/reset-devicelab-task'))
+      ..bodyFields = <String, String>{
+        'Key': task.key.toString(),
+      };
 
     // TODO(chillers): add auth token
 
--- a/app_flutter/lib/service/appengine_cocoon.dart
+++ b/app_flutter/lib/service/appengine_cocoon.dart
@@ -84,9 +84,15 @@ class AppEngineCocoonService implements CocoonService {
 
   @override
   Future<bool> rerunTask(Task task, String accessToken) async {
+    final http.Request request = http.Request(
+        'POST', Uri(host: '$_baseApiUrl', path: '/reset-devicelab-task'));
+    request.bodyFields = <String, String>{
+      'Key': task.key.toString(),
+    };
+
     // TODO(chillers): add auth token
-    final http.Response response =
-        await _client.get('$_baseApiUrl/reset-devicelab-task');
+
+    final http.StreamedResponse response = await _client.send(request);
 
     return response.statusCode == 200;
   }
--- a/app_flutter/lib/service/appengine_cocoon.dart
+++ b/app_flutter/lib/service/appengine_cocoon.dart
@@ -26,8 +26,8 @@ class AppEngineCocoonService implements CocoonService {
   /// The Cocoon API endpoint to query
   ///
   /// This is the base for all API requests to cocoon
-  static const String _cocoonHost = 'https://flutter-dashboard.appspot.com';
-  static const String _baseApiUrl = '$_cocoonHost/api';
+  static const String _appengineAuthority = 'flutter-dashboard.appspot.com';
+  static const String _baseApiUrl = 'https://$_appengineAuthority/api';
 
   final http.Client _client;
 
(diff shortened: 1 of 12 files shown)


app_flutter/lib/service/appengine_cocoon.dart
```diff
@@ -82,6 +83,25 @@ class AppEngineCocoonService implements CocoonService {
...
+      ..bodyFields = <String, String>{
+        'Key': task.key.toString(),
+      }
+      ..headers.addAll(<String, String>{
+        'X-Flutter-AccessToken': accessToken,
+      });
+
+    final http.StreamedResponse response = await _client.send(request);
+
+    return response.statusCode == 200;
```
jonahwilliams: nit: use HttpStatus.Ok

CaseyHillers: Ok!

app_flutter/lib/task_box.dart
```diff
@@ -234,12 +254,28 @@ class TaskOverlayContents extends StatelessWidget {
...
                       },
                     ),
-                    IconButton(
-                      icon: const Icon(Icons.redo),
-                      onPressed: () {
-                        // TODO(chillers): Rerun task. https://github.com/flutter/cocoon/issues/424
-                      },
-                    ),
+                    if (isDevicelab(task))
+                      ProgressButton(
```
jonahwilliams: I think this widget has reached a point of complexity where it would make sense to split up the buttons into separate classes

CaseyHillers: I think we should rename `TaskOverlayContents` to `TaskOverlayEntry` and fork the actual content into `TaskOverlayContents`. Most of the overhead is from the stuff needed for the `OverlayEntry`.

SG?

jonahwilliams: SGTM!

CaseyHillers: Done.

app_flutter/lib/task_box.dart
```diff
@@ -215,38 +212,119 @@ class TaskOverlayContents extends StatelessWidget {
...
+    @required this.buildState,
+    @required this.task,
+    @required this.taskStatus,
+  })  : assert(parentContext != null),
+        assert(buildState != null),
+        assert(task != null),
+        super(key: key);
+
+  /// The parent context that has the size of the whole screen
+  final BuildContext parentContext;
```
jonahwilliams: It looks like you're only using this to show a snackbar, can you use the child context instead?

app_flutter/lib/task_box.dart
```diff
@@ -215,38 +212,119 @@ class TaskOverlayContents extends StatelessWidget {
...
+                  // TODO(chillers): Open log in new window. https://github.com/flutter/cocoon/issues/436
+                },
+              ),
+              if (isDevicelab(task))
+                ProgressButton(
+                  defaultWidget: const Text('Rerun task'),
+                  progressWidget: const CircularProgressIndicator(),
+                  width: 120,
+                  height: 50,
+                  onPressed: () async {
```
jonahwilliams: Lift this into a private method

app_flutter/lib/task_box.dart
```diff
@@ -215,38 +212,119 @@ class TaskOverlayContents extends StatelessWidget {
...
+                    final bool success = await buildState.rerunTask(task);
+                    return () {
```
jonahwilliams: Why are you returning a callback here?

app_flutter/lib/task_box.dart
CaseyHillers: I tried that initially and the `SnackBar` wouldn't appear, but using the parent context worked.

app_flutter/lib/task_box.dart
CaseyHillers: Done.

app_flutter/lib/task_box.dart
CaseyHillers: It was in the example, and I have never done it before :)

Removed since it didn't go anywhere.

app_flutter/lib/task_box.dart
jonahwilliams: Oh right, this is in an overlay, so it doesn't actually have access to the context with a Scaffold parent. In this case, I would instead pass down a callback that takes a Snackbar and internally invokes parentContext.showSnackbar - this cuts down the API surface/coupling dramatically.

CaseyHillers: Like
`  final ScaffoldFeatureController<SnackBar, SnackBarClosedReason> Function(
      SnackBar) showSnackbarCallback;`?

app_flutter/lib/task_box.dart
```diff
@@ -215,38 +212,119 @@ class TaskOverlayContents extends StatelessWidget {
...
+    return Card(
+      child: Column(
+        mainAxisSize: MainAxisSize.max,
+        children: <Widget>[
+          ListTile(
+            leading:
+                Tooltip(message: taskStatus, child: statusIcon[taskStatus]),
+            title: Text(task.name),
+            subtitle: Text(
+                'Attempts: ${task.attempts}\nDuration: ${task.endTimestamp - task.startTimestamp} seconds\nAgent: ${task.reservedForAgentId}'),
```
jonahwilliams: This is getting a bit long - you can split a string onto multiple lines like so:

```
var x = 'This is all '
  'part of one String';
```

CaseyHillers: Thanks for sharing! This is a cool feature.

Done.

PR #507, opened 2019-11-02
Check for PRs waiting for green

dnfield:
This needs tests - putting it up to get some initial feedback on the approach.

Basic idea is run a cron job every 5 minutes, get the top 100 PRs that have the waiting for green label, check that they:

- Have passed CI on their latest commit
- Have at least one "approve" review
- Have no change request reviews
- Have no merge conflicts

If so, squash and merge. If not and the issue is anything other than green CI, post a message and remove the label.

I'm not sure if we should make the message configurable - we did that for other things like this, but if we do make it configurable we should probably just list all the conditions in every message rather than listing only the ones that apply to this PR.

Using GraphQL because it lets me make one focused request for the PRs we're interested in.


app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
```diff
@@ -0,0 +1,309 @@
...
+        nodes {
+          state
+          author {
+            login
+          }
+        }
+      }
+    }
+  }
+}}''';
```
Hixie: what dark magic is this? :-o

Hixie:
I propose that we only merge the oldest PR with the label set each five minutes, so that we avoid any risk of the tree being red for a day and then the script checking in 100 PRs all at once.

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
dnfield: GraphQL query.  This lets us get reviews for specifically labeled PRs in a single request, instead of making a request for PR, separate request for its reviews, and then joining/filtering all of those client side.

dnfield:
That SGTM.

app_dart/lib/src/datastore/cocoon_config.dart
```diff
@@ -51,35 +52,43 @@ class Config {
...
   /// projects.
   Future<String> get redisUrl => _getSingleValue('RedisConnectionSpec');
 
   /// The name of the subcache in the Redis instance that stores responses.
-  Future<String> get redisResponseSubcache => _getSingleValue('RedisResponseSubcache');
+  Future<String> get redisResponseSubcache =>
+      _getSingleValue('RedisResponseSubcache');
+
+  Future<String> get waitingForTreeToGoGreenLabelName =>
+      _getSingleValue('WaitingForTreeToGreenLabelName');
```
Hixie: (trivial and optional: i'm guessing the changes above this line to this file are dartfmt changes? they sure make the file harder to read. maybe we should run it with a wider line width?)

dnfield: I think we actually were using a wider width and I just lost it.  I'll try again using wider.

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
```diff
@@ -0,0 +1,352 @@
...
+                commit {
+                  abbreviatedOid
+                  oid
+                  status {
+                    state
+                  }
+                }
+              }
+            }
+            reviews(first: 100, states: [CHANGES_REQUESTED, APPROVED]) {
```
liyuqian: Can we do 
```
changeRequestedReviews: reviews(first: 1, states: [CHANGES_REQUESTED]) {
...
}
approvedReviews: reviews(first: 1, states: [APPROVED]) {
...
}
```
just in case that there are more than 100 reviews?

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
```diff
@@ -0,0 +1,352 @@
...
+  }
+}''';
+
+const String _mergePullRequestMutation = r'''
+mutation MergePR($id: ID!, $oid: GitObjectID!) {
+  mergePullRequest(input: {
+    pullRequestId: $id,
+    expectedHeadOid: $oid,
+    mergeMethod: SQUASH,
+    commitBody: ""
```
liyuqian: Can we by default copy the PR description into the merge commit description?

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
```diff
@@ -0,0 +1,352 @@
...
+            break;
+          case 'CHANGES_REQUESTED':
+            hasChangesRequested = true;
+            break;
+        }
+      }
+      final Map<String, dynamic> commit =
+          pullRequest['commits']['nodes'].single['commit'];
+      final String sha = commit['oid'];
+      final bool ciSuccessful = commit['status']['state'] == 'SUCCESS';
```
liyuqian: This may not be sufficient. We still need to check that at least a fixed number of tests are passed just in case that some presubmit tests are not started yet: https://github.com/flutter/flutter/issues/43672

dnfield:
We don't currently enforce formatting consistently in this repo.  We should fix that in a separate PR.  For now I've formatted the files I'm changing with line length 100.

This has tests now and is ready for a real review.

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
dnfield: 100 is the max the API allows AFAIK.

I don't believe the queries you're proposing are valid, but if they are I'd be happy to change it.

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
dnfield: That's usually not what you want in the commit description - it's very rarely formatted as a proper commit message and just dirties the log.  Thsi should result in just the PR title as the commit message.

app_dart/bin/gh.dart
```diff
@@ -0,0 +1,306 @@
...
+  );
+
+  final Link _link = _authLink.concat(_httpLink);
+
+  final GraphQLClient _client = GraphQLClient(
+    cache: InMemoryCache(),
+    link: _link,
+  );
+
+  const String _labeledPullRequestsWithReviewsQuery = r'''
```
liyuqian: Is this part of the test? Why is it not using `labeledPullRequestsWithReviewsQuery` from check_for_waiting_pull_requests_queries.dart which I think should be the subject of the test?

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
liyuqian: I was trying to use the [graphql aliases](https://graphql.org/learn/queries/#aliases) which seems to allow you to query the same thing multiple times with different conditions.

dnfield: Ahh very nice, this works.  Updated.

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
liyuqian: My habit is to always write a nice PR description that would be nice for merge commit message. When I merge manually, I'll copy it to the merge commit message.

It seems that others are doing similar things. Take the last 3 PRs as examples.
- https://github.com/flutter/flutter/pull/44151 has the related issue in the PR description.
- https://github.com/flutter/flutter/pull/44150 has the corresponding engine commits.
- https://github.com/flutter/flutter/pull/44149 has relevant information until the "checklist" section.

I think if auto-merged commits have an empty body, it will be very hard to check the `git log` to figure out what the commit is for (e.g., what issue does it fix).

app_dart/bin/gh.dart
dnfield: This was mistakenly checked in.  It's not used anywhere - was using it for testing queries against actual github.  You'll notice it even had an auto code from my account in there, which I have revoked  :)

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
dnfield: I see.  Do we know what that number is?

I wonder if this one should be ok because we expect people to only add this label _after_ all other checks have passed.  but if we want to check/enforce that, we'd need to know the number...

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
dnfield: The first two are fine, but the third one is not something I'd want to see in a git log - and it's the only one of the three using our actual PR template :\

app_dart/bin/gh.dart
liyuqian: Hmm... then where's the test? Did you forget to add some files?

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
liyuqian: You'll have the number automatically once https://github.com/flutter/flutter/pull/43681 lands. For now, that number should be 99.

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
liyuqian: I was using that template quite often until recently that I figured out that most people don't follow it anyway. In case that someone is using the template, it should be relatively simple to remove the "checklist" section automatically?

app_dart/bin/gh.dart
dnfield: I did.  It should be present now.

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
Hixie: You could also explicitly wait an hour after the last commit was uploaded before considering a PR. That would have the advantage of giving time for all the checks to begin, and for people to rereview the PR before it landed. (I would probably do this in addition to the test of how many checks we want.)

If you want something like https://github.com/flutter/flutter/pull/43681 then we should talk about what that should look like. I assume you don't want to check out the entire PR locally to run code from it...

Hixie: We could require one particular shard to always run, and for that shard to include the number you care about in its output? Would reading the data from a shard be possible here?

liyuqian: I believe the script here can either clone the PR branch and run the commit locally, or check the shard output through either Cirrus or Github GraphQL API.

FWIW, waiting 1 hour alone may still be insufficient. Cirrus has run into scheduling problems before so it's possible that some tests are still not started after 1 hour.

dnfield: I think giving it an hour is probably good enough for now.  Anything else is going to be more complicated/have similar problems to what we have in flutter/flutter#43681.

Added the check for older than an hour + a new test.

dnfield:
@rmistry - once this lands, we could update the auto-roller(s) to apply this label when they create their PR(s).  This would then do the work of merging them once they're ready to merge.

The delay of an hour here is not ideal, but it should be good enough to avoid the case of merging a PR that hasn't actually run all (or any) checks yet, and give us some room to figure out how to correctly handle the number of checks we want in a robust way.

rmistry:
> @rmistry - once this lands, we could update the auto-roller(s) to apply this label when they create their PR(s). This would then do the work of merging them once they're ready to merge.

What is the label name? "waiting for tree to go green" ?

> 
> The delay of an hour here is not ideal, but it should be good enough to avoid the case of merging a PR that hasn't actually run all (or any) checks yet, and give us some room to figure out how to correctly handle the number of checks we want in a robust way.

See my other comment.

dnfield:
Yes, waiting for tree to go green

Perhaps what we really want is something that generates all the checks for us from a single file

rmistry:
FYI, Corresponding change for the autorollers is here: https://skia-review.googlesource.com/c/buildbot/+/252477/

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
```diff
@@ -0,0 +1,280 @@
...
+
+  final LoggingProvider loggingProvider;
+
+  @override
+  Future<Body> get() async {
+    final Logging log = loggingProvider();
+    final GraphQLClient client = await config.createGitHubGraphQLClient();
+
+    await _checkPRs('flutter', 'flutter', log, client);
+    await _checkPRs('flutter', 'engine', log, client);
```
digiter: There's no need to wait and these two functions can run concurrently.

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
```diff
@@ -0,0 +1,280 @@
...
+    bool hasMerged = false;
+    final Map<String, dynamic> data = await _queryGraphQL(
+      owner,
+      name,
+      log,
+      client,
+    );
+    for (_AutoMergeQueryResult queryResult in _parseQueryData(data)) {
+      if (!hasMerged && queryResult.shouldMerge) {
+        hasMerged = await _mergePullRequest(
```
digiter: hasMerged is not used.

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
```diff
@@ -0,0 +1,280 @@
...
+    }
+  }
+
+  Future<Map<String, dynamic>> _queryGraphQL(
+    String owner,
+    String name,
+    Logging log,
+    GraphQLClient client,
+  ) async {
+    final String labelName = await config.waitingForTreeToGoGreenLabelName;
```
digiter: Move this to parent function would save it from waiting every time.

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
```diff
@@ -0,0 +1,280 @@
...
+    return true;
+  }
+
+  /// Parses a GraphQL query to a list of [_AutoMergeQueryResult]s.
+  ///
+  /// This method will not return null, but may return an empty list.
+  List<_AutoMergeQueryResult> _parseQueryData(Map<String, dynamic> data) {
+    final Map<String, dynamic> repository = data['repository'];
+    if (repository == null || repository.isEmpty) {
+      return <_AutoMergeQueryResult>[];
```
digiter: This sounds like an exception.

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests_queries.dart
```diff
@@ -0,0 +1,66 @@
+// Copyright 2019 The Chromium Authors. All rights reserved.
```
digiter: Just curious, how did you debug the GraphQL in this file?

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
dnfield: Ah thanks - at one point I was returning whether the merge succeeded.

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
dnfield: Hm.  Probably a good idea rather than returning an empty list - I believe the schema for this allows null, but that would indicate we asked for a the wrong repo (or we don't have access to it I guess?).

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
dnfield: Actually it is used - we stop looking once we merged.

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests_queries.dart
dnfield: https://developer.github.com/v4/explorer/

I also had a local dart program that was basically the `get` method o fthe request handler

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
dnfield: We should wait to avoid returning before the functions finish - we also have a lint on for this repo to avoid unawaited functions. The queries we're using don't take very long though.

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
dnfield: This should be fast after the first time - datastore does some caching I believe, right?

I could move it up the call stack but then we're passing the label around and these parameter lists are already fairly long.  That could probably be improved though by making this class not be immutable..  I'll do that instead.

dnfield: Actually I can't do that because the parent class is immutable.

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
digiter: Oh, right.

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
digiter: Yeah, null or empty repository feels like an error in our configuration.

app_dart/test/request_handlers/check_for_waiting_pull_requests_test.dart
```diff
@@ -0,0 +1,402 @@
...
+  Stream<FetchResult> subscribe(Operation operation) {
+    throw UnimplementedError();
+  }
+
+  @override
+  ObservableQuery watchQuery(WatchQueryOptions options) {
+    throw UnimplementedError();
+  }
+
+  void verifyQueries(List<QueryOptions> expectedQueries) {
```
liyuqian: `verifyQueries` and `verifyMutations` have a lot of common code. Maybe we can create a function `_verifyOptions` as follows to remove the duplicate code:
```
void _verifyOptions(List<BaseOptions> actual, List<BaseOptions> expected) {
    expect(actual.length, expected.length);
    for (int i = 0; i < actual.length; i++) {
      expect(actual[i].toKey(), expected[i].toKey());
      expect(actual[i].fetchPolicy, expected[i].fetchPolicy);
    }
}

void verifyQueries(List<QueryOptions> expectedQueries) => _verifyOptions(queries, expectedQueries);
void verifyMutations(List<MutationOptions> expectedMutations)  => _verifyOptions(mutations, expectedMutations);
```

BTW, maybe also add a comment like: `toKey` encapsulates document, variables, ..., of the GraphQL query so we're verifying those through `toKey`.

app_dart/test/request_handlers/check_for_waiting_pull_requests_test.dart
```diff
@@ -0,0 +1,402 @@
...
+              'sOwner': 'flutter',
+              'sName': 'engine',
+              'sLabelName': config.waitingForTreeToGoGreenLabelNameValue,
+            },
+          ),
+        ],
+      );
+    }
+
+    test('Merges first PR in list, all successful', () async {
```
liyuqian: Can we also add a test with a lot of pull requests (e.g., 200) and make sure that we won't crash (because we set some limit to 100 in our code)?

app_dart/test/request_handlers/check_for_waiting_pull_requests_test.dart
```diff
@@ -0,0 +1,402 @@
...
+            variables: <String, dynamic>{
+              'id': engineRepoPRs.first.id,
+              'oid': 'deadbeef',
+            },
+          ),
+        ],
+      );
+    });
+
+    test('Ignores PRs that are too new', () async {
```
liyuqian: Can we please also have a test around `now - 50 minutes` and `now - 70 minutes` to verify our threshold of `now - 60 minutes`?

app_dart/test/request_handlers/check_for_waiting_pull_requests_test.dart
```diff
@@ -0,0 +1,402 @@
...
+              'oid': 'deadbeef',
+            },
+          ),
+        ],
+      );
+    });
+
+    test('Remove labels', () async {
+      final PullRequestHelper prOneBadReview = PullRequestHelper(
+        hasApprovedReview: false,
```
liyuqian: Is there any test that covers how we parsed the reviews and set the `hasApprovedReview`, `hasChangeRequestReview`, ...,  flags?

app_dart/test/request_handlers/check_for_waiting_pull_requests_test.dart
dnfield: The API has a max of 100 and we specify we only want 100 in our query. Do you still think such a test is useful?

app_dart/test/request_handlers/check_for_waiting_pull_requests_test.dart
dnfield: Done.

app_dart/test/request_handlers/check_for_waiting_pull_requests_test.dart
dnfield: The `Remove Labels` test is doing that - it's basically exercising a whole bunch of scenarios we care about in those branches.

app_dart/test/request_handlers/check_for_waiting_pull_requests_test.dart
liyuqian: Ok, if the limit is on the GIthub side, maybe it's not very useful to test on our side.

app_dart/test/request_handlers/check_for_waiting_pull_requests_test.dart
liyuqian: Oh, now I see that `PullRequestHelper` is a test class that generates the corresponding string using those flags.

dnfield:
Going to land this to avoid more merge conflicts while I'm out.  It should work for now, but definitely has room for improvement. Specifcially, we want a better method than just letting a PR exist for an hour before auto-merging.

PR #509, opened 2019-11-04
Force the phone's screen on before running a test.

godofredoc:
We are currently sending the phone to sleep after running a task by
sending a power key press intent. This is making a few tests fail and
some other tests to not give reliable signals. We are forcing the phone
to wake up before running tests.

Bug: 44004
Test: Manual execution of the change locally.

### DIFF
--- a/agent/lib/src/commands/ci.dart
+++ b/agent/lib/src/commands/ci.dart
@@ -114,6 +114,9 @@ class ContinuousIntegrationCommand extends Command {
               // the task timeout.
               await getFlutterAt(task.revision).timeout(_kInstallationTimeout);
               await _cleanBuildDirectories(agent, task);
+              // Ensure the phone's screen is on before running a task.
+              await _screensOn();
+
               await _runTask(task);
             } else {
               logger.info('No tasks available for this agent.');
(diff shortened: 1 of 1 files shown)


godofredoc:
This is to fix: https://github.com/flutter/flutter/issues/44004

xster:
cc @liyuqian

agent/lib/src/commands/ci.dart
```diff
@@ -201,6 +204,17 @@ class ContinuousIntegrationCommand extends Command {
     }
   }
 
+  Future<Null> _screensOn() async {
+    try {
+      for (Device device in await devices.discoverDevices()) {
+        await device.wakeUp();
```
xster: We don't have to fix it right now, but the fact that we just toggle without detecting the current screen on state seems a bit fragile in the long run.

xster: Oh sorry there is a _getWakefulness currently. Never mind.

godofredoc: adb.wakeUp does that validation: https://github.com/flutter/cocoon/blob/master/agent/lib/src/adb.dart#L188

yjbanov:
IIRC the reason why I did not automatically wake up the device was because not all tests needed a device, so it was up to the test to wake it up. With this change we may end up with tests attempting to wake up the device even though it is already awake. Not sure if that's a big deal :man_shrugging:

xster:
Having the tests call the wake up function seems ok as well. I don't have a strong preference.

PR #512, opened 2019-11-05
Add Agent status data to BigQuery

keyonghan:
This PR inserts Agent status data to BigQuery whenever updating Agent health status to Datastore.


app_dart/lib/src/request_handlers/update_agent_health.dart
```diff
@@ -45,13 +50,38 @@ class UpdateAgentHealth extends ApiRequestHandler<UpdateAgentHealthResponse> {
...
+
+    try {
+      await tabledataResourceApi.insertAll(rows, projectId, dataset, table);
+    } catch(ApiRequestError){
+      log.warning('Failed to add $agentId status to BigQuery: $ApiRequestError');
+    }
+
     return UpdateAgentHealthResponse(agent);
   }
 }
```
digiter: I would put the new changes into a sub-function to avoid interleave the the old logic.

Even better, put the old logic into a sub-function too. Become something like
```
post() {
  _updateDatastore();
  _updateBigQuery();
}
```

app_dart/lib/src/request_handlers/update_agent_health.dart
```diff
@@ -45,13 +50,38 @@ class UpdateAgentHealth extends ApiRequestHandler<UpdateAgentHealthResponse> {
         throw BadRequestException('Invalid agent ID: $agentId');
       },
     );
+    final TabledataResourceApi tabledataResourceApi = await config.createTabledataResourceApi();
+    final List<Map<String, Object>> tableDataInsertAllRequestRows = <Map<String, Object>>[];
```
digiter: Nit: This var name is too long.

app_dart/lib/src/request_handlers/update_agent_health.dart
```diff
@@ -45,13 +50,38 @@ class UpdateAgentHealth extends ApiRequestHandler<UpdateAgentHealthResponse> {
...
 
     agent.isHealthy = isHealthy;
     agent.healthDetails = healthDetails;
     agent.healthCheckTimestamp = DateTime.now().millisecondsSinceEpoch;
 
     await datastore.db.commit(inserts: <Agent>[agent]);
 
+    /// Insert data to [BigQuery] whenever updating data in [Datastore] 
+    tableDataInsertAllRequestRows.add(<String, Object>{
```
digiter: Inline the variable declaration here.

app_dart/lib/src/request_handlers/update_agent_health.dart
```diff
@@ -45,13 +50,38 @@ class UpdateAgentHealth extends ApiRequestHandler<UpdateAgentHealthResponse> {
...
+      'json': <String, Object>{
+        'Timestamp': agent.healthCheckTimestamp,
+        'AgentID': agentId,
+        // TODO(keyonghan): add more detailed states https://github.com/flutter/flutter/issues/44213
+        'Status': isHealthy?'Healthy':'Unhealthy',
+        'Detail': healthDetails,
+      },
+    });
+
+    /// Final [rows] to be inserted to [BigQuery]
```
digiter: "Final" in the comment is redundant.

app_dart/test/request_handlers/update_agent_health_test.dart
```diff
@@ -49,3 +57,6 @@ void main() {
     });
   });
 }
+
+class MockTabledataResourceApi extends Mock implements TabledataResourceApi {} 
+
```
digiter: This PR needs tests.

app_dart/lib/src/request_handlers/update_agent_health.dart
keyonghan: Added _updateBigquery() as recommended.
As for _updateDatastore(), since we have only one datastore insert script, so just leave it there.

app_dart/lib/src/request_handlers/update_agent_health.dart
keyonghan: Changed [tableDataInsertAllRequestRows] to [requestRows].

app_dart/lib/src/request_handlers/update_agent_health.dart
keyonghan: Removed.

app_dart/lib/src/request_handlers/update_agent_health.dart
keyonghan: updated.

app_dart/test/request_handlers/update_agent_health_test.dart
keyonghan: Added.

app_dart/lib/src/request_handlers/update_agent_health.dart
```diff
@@ -45,22 +49,55 @@ class UpdateAgentHealth extends ApiRequestHandler<UpdateAgentHealthResponse> {
         throw BadRequestException('Invalid agent ID: $agentId');
       },
     );
+    final TabledataResourceApi tabledataResourceApi = await config.createTabledataResourceApi();
```
digiter: This line only makes sense with _insertBigQuery(), therefore it should be encapsulate within that function. Otherwise it's unexpected to see a big query api occurs among data store updating logic.

app_dart/test/request_handlers/update_agent_health_test.dart
```diff
@@ -46,6 +49,8 @@ void main() {
       expect(agent.agentId, 'test');
       expect(response.agent.isHealthy, true);
       expect(agent.healthDetails, 'bar detail');
+      expect(int.parse(response.totalRows), 1);
```
digiter: Since this test has access to FakeTabledataResourceApi, it could read the result from there instead of a response.

app_dart/lib/src/request_handlers/update_agent_health.dart
```diff
@@ -45,22 +49,55 @@ class UpdateAgentHealth extends ApiRequestHandler<UpdateAgentHealthResponse> {
...
 }
 
 @immutable
 class UpdateAgentHealthResponse extends JsonBody {
-  const UpdateAgentHealthResponse(this.agent) : assert(agent != null);
+  const UpdateAgentHealthResponse(this.agent, this.totalRows) : assert(agent != null);
 
   final Agent agent;
+  /// [totalRows] is for test purpose only.
+  final String totalRows;
```
digiter: By removing totalRows, a response avoids carrying an unused field in production.

app_dart/lib/src/request_handlers/update_agent_health.dart
keyonghan: Put it inside.

keyonghan: Removed this line.

app_dart/lib/src/request_handlers/update_agent_health.dart
keyonghan: Removed.

app_dart/test/request_handlers/update_agent_health_test.dart
keyonghan: updated.

app_dart/lib/src/request_handlers/update_agent_health.dart
keyonghan: moved.

app_dart/test/src/bigquery/fake_tabledata_resource.dart
```diff
@@ -0,0 +1,37 @@
...
+      {int maxResults,
+      String selectedFields,
+      String startIndex,
+      String pageToken,
+      String $fields}) async {
+    return TableDataList.fromJson(
+        <String, String>{'totalRows': rows.toString()});
+  }
+}
+
```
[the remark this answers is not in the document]
keyonghan: Line 22-23 and line 34 are the implementations, though they are simple. You can refer to the case of datastore. They are only used to mimic real API to let test continue. As for how complex our implement should be, then it depends on our real purpose of targeted features.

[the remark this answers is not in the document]
digiter: Then `list` can be applied to avoid exposing the implementation details.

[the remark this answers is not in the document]
keyonghan: Not sure what you mean. Can you elaborate?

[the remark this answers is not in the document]
digiter: As `insertAll` writes data and `list` reads data, a test could utilize these original API surfaces without interacting with an implementation detail (`row`).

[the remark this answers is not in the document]
keyonghan: What you are suggesting is good if we can directly use the original APIs. However, they are not applicable in the test case since they communicate with real BigQuery APIs. And that's why we have to implement the fake, which mimics functionalities of real APIs. And here comes the 'row'..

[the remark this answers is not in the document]
keyonghan: What you are suggesting is good if we can directly use the original APIs. However, they are not applicable in the test case since they communicate with real BigQuery APIs. And that's why we have to implement the fake, which mimics functionalities of real APIs. And here comes the 'row'..

app_dart/test/request_handlers/update_agent_health_test.dart
```diff
@@ -46,6 +48,8 @@ void main() {
       expect(agent.agentId, 'test');
       expect(response.agent.isHealthy, true);
       expect(agent.healthDetails, 'bar detail');
+      expect(tabledataResourceApi.rows, 1);
```
digiter: Is it possible to use `list` on this line?

keyonghan: Changed to use list. But not sure the superiority over directly using 'rows'?

digiter: "Prefer to test via the public interface of a component." https://abseil.io/tips/135

PR #514, opened 2019-11-06
[devicelab][windows] Install TigerVNC server instead of TightVNC

digiter:
And a few improvements:
* Moves the provision script to a new directory.
* Clone cocoon repository directly in the script.
* Removes starting a Cocoon agent on startup.

Fix https://github.com/flutter/flutter/issues/44275


devicelab/windows_android.ps1
```diff
@@ -0,0 +1,42 @@
+# Copyright (c) 2016, the Flutter project authors.  Please see the AUTHORS file
```
godofredoc: 2019?

devicelab/windows_android.ps1
```diff
@@ -0,0 +1,42 @@
...
+choco install -y dart-sdk
+choco install -y android-sdk
+choco install -y sysinternals
+
+# Installs a VNC server and a SSH server for remote management
+choco install -y tigervnc
+choco install -y openssh
+
+# Installs convenient tools
+choco install -y vscode
```
godofredoc: Do we actually need this? I'd prefer something that can be used over ssh, e.g. vim

devicelab/windows_android.ps1
```diff
@@ -0,0 +1,42 @@
...
+choco install -y GoogleChrome
+
```
godofredoc: Can we also add "Set-ExecutionPolicy Unrestricted"

devicelab/windows_android.ps1
digiter: Done.

devicelab/windows_android.ps1
digiter: No problem, added vim too.

devicelab/windows_android.ps1
digiter: What's the benefit of doing that?

godofredoc: When we are running powershell scripts over ssh some of the will fail is the execution policy is not set correctly. I didn't try any of the other more restrictive policies just went with the most permissive one.

digiter: SG, done.

PR #519, opened 2019-11-07
[devicelab][windows] agent manager for windows.

godofredoc:
This is a script the will be running on a cron job every 5 minutes to
validate the agent is running on windows machines and restart it if
necessary. We will be implementing similar scripts for Linux and Mac
machines.

### DIFF
--- a/devicelab/agent_manager.ps1
+++ b/devicelab/agent_manager.ps1
@@ -1,0 +1,69 @@
+#requires -version 2
+
+# Copyright 2019 The Chromium Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+<#
+.SYNOPSIS
+  Script to perform several operations on the cocoon agent. It will be run
+  from a cron job at a 5 mins interval.
+.DESCRIPTION
+  The script is intended for the following operations:
+    * Start agent if it is not running.
+    * Check if there is a new version of the agent then update and restart.
+    * Cleanup host state. Killing stuck processes or rebooting if required.
+
+  Only the first operation will be implemented at this time.
+.INPUTS
+  None
+.OUTPUTS
+  Log file stored in C:\Users\flutter\agent_manager.log
+  Log file stored in C:\Users\flutter\agent_stderr.log
+  Log file stored in C:\Users\flutter\agent_stdout.log
+.NOTES
+  Version:        0.1
+  Creation Date:  11/05/2019
+  Purpose/Change: Initial agent manager development.
+
+.EXAMPLE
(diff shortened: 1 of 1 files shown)


godofredoc:
This PR fixes: 

https://github.com/flutter/flutter/issues/44209

devicelab/agent_manager.ps1
```diff
@@ -0,0 +1,54 @@
...
+  Schtasks /create /tn "Agent Manager" /sc MINUTE /mo 5 /tr "PowerShell c:\Users\flutter\cocoon\devicelab\agent_manager.ps1"
+#>
+
+#---------------------------------------------------------[Initialisations]--------------------------------------------------------
+
+#----------------------------------------------------------[Declarations]----------------------------------------------------------
+$AgentFolder = "C:\Users\flutter\cocoon\agent"
+$LogFile = "C:\Users\flutter\agent_manager.log"
+$StdErr = "C:\Users\flutter\agent_stderr.log"
+$StdOut = "C:\Users\flutter\agent_stdout.log"
```
digiter: Combining stdout and stderr into one log file makes it easier to check errors. Only one file needs to be checked.

devicelab/agent_manager.ps1
```diff
@@ -0,0 +1,54 @@
...
+#-----------------------------------------------------------[Functions]------------------------------------------------------------
+function ReStart-Agent {
```
digiter: Nit: Restart-Agent

devicelab/agent_manager.ps1
```diff
@@ -0,0 +1,54 @@
...
+  $Process = Get-Process | Where-Object ProcessName -Match "^dart*"
+  if (!$process) {
+    echo "Restarting Agent"
+    cd $AgentFolder
+    Start-Process C:\tools\dart-sdk\bin\dart -ArgumentList "bin\agent.dart ci" -RedirectStandardOutput $StdOut -RedirectStandardError $StdErr -Verbose
```
digiter: It would be nice to see when the process started in logs.

devicelab/agent_manager.ps1
```diff
@@ -0,0 +1,54 @@
...
+  } else {
+    echo "Agent already running. Skipping."
```
digiter: There will be lots of this in the logs, I think it's ok to remove it. With the above comment, the logs will only records when the manager tries to start an agent.

devicelab/agent_manager.ps1
```diff
@@ -0,0 +1,54 @@
+#requires -version 2
```
digiter: Is there a style guide? It's better to keep the style consistent.

digiter: Looks unnecessarily verbose, ¯\_(ツ)_/¯

PR #522, opened 2019-11-07
CocoonService uses relative domain for API on web

CaseyHillers:
Closes https://github.com/flutter/flutter/issues/43103

Allows for using production data on AppEngine versions that are not in production. Additionally, allows for use of local dart server.

## Open questions
- ~Is there a better way of building the URL? I think it is unnecessary to include the endpoint twice, but my only idea is to move it to a separate variable.~

### DIFF
--- a/app_flutter/lib/service/appengine_cocoon.dart
+++ b/app_flutter/lib/service/appengine_cocoon.dart
@@ -23,25 +23,17 @@ class AppEngineCocoonService implements CocoonService {
   AppEngineCocoonService({http.Client client})
       : _client = client ?? http.Client();
 
-  /// The Cocoon API endpoint to query
-  ///
-  /// This is the base for all API requests to cocoon
-  static const String _appengineAuthority = 'flutter-dashboard.appspot.com';
-  static const String _baseApiUrl = 'https://$_appengineAuthority/api';
-
   final http.Client _client;
 
   @override
   Future<CocoonResponse<List<CommitStatus>>> fetchCommitStatuses() async {
     /// This endpoint returns JSON [List<Agent>, List<CommitStatus>]
-    final http.Response response =
-        await _client.get('$_baseApiUrl/public/get-status');
+    final http.Response response = await _client.get('/api/public/get-status');
 
     if (response.statusCode != HttpStatus.ok) {
       print(response.body);
       return CocoonResponse<List<CommitStatus>>()
-        ..error =
-            '$_baseApiUrl/public/get-status returned ${response.statusCode}';
+        ..error = '/api/public/get-status returned ${response.statusCode}';
     }
 
     try {
@@ -57,13 +49,12 @@ class AppEngineCocoonService implements CocoonService {
   Future<CocoonResponse<bool>> fetchTreeBuildStatus() async {
     /// This endpoint returns JSON {AnticipatedBuildStatus: [BuildStatus]}
     final http.Response response =
-        await _client.get('$_baseApiUrl/public/build-status');
+        await _client.get('/api/public/build-status');
 
     if (response.statusCode != HttpStatus.ok) {
       print(response.body);
       return CocoonResponse<bool>()
-        ..error =
-            '$_baseApiUrl/public/build-status returned ${response.statusCode}';
+        ..error = '/api/public/build-status returned ${response.statusCode}';
     }
 
     Map<String, Object> jsonResponse;
@@ -71,12 +62,12 @@ class AppEngineCocoonService implements CocoonService {
       jsonResponse = jsonDecode(response.body);
     } catch (error) {
       return CocoonResponse<bool>()
-        ..error = '$_baseApiUrl/public/build-status had a malformed response';
+        ..error = '/api/public/build-status had a malformed response';
     }
 
     if (!_isBuildStatusResponseValid(jsonResponse)) {
       return CocoonResponse<bool>()
-        ..error = '$_baseApiUrl/public/build-status had a malformed response';
+        ..error = '/api/public/build-status had a malformed response';
     }
 
     return CocoonResponse<bool>()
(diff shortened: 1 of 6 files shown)


CaseyHillers:
@jonahwilliams is there a way to configure the `kIsWeb` constant in tests? Right now my approach would be to test for whatever the test framework defaults to, but that doesn't test the other case.

jonahwilliams:
You can also run tests with `flutter test --platform=chrome`. Then in the test case, use kIsWeb to determine what you're looking for

app_flutter/lib/service/appengine_cocoon.dart
```diff
@@ -26,22 +27,23 @@ class AppEngineCocoonService implements CocoonService {
...
+  static const String _baseApiUrl = 'https://flutter-dashboard.appspot.com';
 
   final http.Client _client;
 
   @override
   Future<CocoonResponse<List<CommitStatus>>> fetchCommitStatuses() async {
+    final String getStatusUrl = kIsWeb
+        ? '/api/public/get-status'
+        : '$_baseApiUrl/api/public/get-status';
+
```
digiter: Consider encapsulate this pattern into a function, so that this pattern isn't scattered in 3 functions. For example,
```
// This function helps the Web app talks to a local Cocoon AppEngine instance when running locally, or talks to 
// the production instance when deployed into production.
// The urlSuffix begins with a slash, e.g. "/api/public/get-status".
String getEndpoint(String urlSuffix) {
  return kIsWeb ? urlSuffix : 'https://flutter-dashboard.appspot.com$urlSuffix;
}
```

CaseyHillers:
Should I report differences in the platform tests?

`--platform=chrome` does not throw `NetworkImageLoadException` for a `NetworkImage` widget when it is expected to. Instead it throws `null`.

For now, I can just have Cirrus run this platform test on this particular test.

app_flutter/lib/service/appengine_cocoon.dart
CaseyHillers: Thank you! Added.

CaseyHillers:
@jonahwilliams I added tests and added the chrome test for Cirrus. Currently only runs this group of tests as there are some failures.

jonahwilliams:
Can you file an issue for the lack of the network exception on the web?

app_flutter/lib/service/appengine_cocoon.dart
```diff
@@ -86,22 +86,31 @@ class AppEngineCocoonService implements CocoonService {
...
+    });
 
     return response.statusCode == HttpStatus.ok;
   }
 
+  /// Construct the API endpoint based on the priority of using a local endpoint
+  /// before falling back to the production endpoint.
+  ///
+  /// This function helps the Web app talks to the Cocoon service relative to
+  /// where it's running. On mobile, default to use the production
```
digiter: Explain the magic: "On Web, a relative url automatically resolves to a local Cocoon instance when available, otherwise it resolvs to the production one. Therefore this function returns a relative url for Web.

Mobile apps don't support relative url."

CaseyHillers:
Of course! Do you know if there is a best practice for installing `google-chrome` on Cirrus?

I noticed flutter/flutter just installs it in the Dockerfile https://github.com/flutter/flutter/blob/master/dev/ci/docker_linux/Dockerfile

app_flutter/lib/service/appengine_cocoon.dart
```diff
@@ -26,22 +27,21 @@ class AppEngineCocoonService implements CocoonService {
...
-    final http.Response response =
-        await _client.get('$_baseApiUrl/public/get-status');
+    final http.Response response = await _client.get(getStatusUrl);
 
     if (response.statusCode != HttpStatus.ok) {
       print(response.body);
       return CocoonResponse<List<CommitStatus>>()
-        ..error =
-            '$_baseApiUrl/public/get-status returned ${response.statusCode}';
+        ..error = '/api/public/get-status returned ${response.statusCode}';
```
digiter: Why this line doesn't call _apiEndpoint?

app_flutter/test/service/appengine_cocoon_test.dart
```diff
@@ -227,6 +263,34 @@ void main() {
...
+            headers: captureAnyNamed('headers'),
+            body: captureAnyNamed('body'),
+          ));
+        }
+      });
     });
   });
 }
+
+class MockHttpClient extends Mock implements Client {}
```
digiter: Looks like a good use case to put the new tests into a group: https://flutter.dev/docs/cookbook/testing/unit/introduction#5-combine-multiple-tests-in-a-group

And a group can avoid the redundant setup code. Is that right? @jonahwilliams

jonahwilliams:
There might be a better approach, but that's what I came up with in 5 minutes. I would sync with the rest of the infra team.

CaseyHillers:
Filed difference in test bug here: https://github.com/flutter/flutter/issues/44370

app_flutter/lib/service/appengine_cocoon.dart
CaseyHillers: It's just an error message, so I think just giving the endpoint is fine. I was trying to keep the error messages concise so they could be shown in the UI.

app_flutter/lib/service/appengine_cocoon.dart
CaseyHillers: Okay. I expanded to explain more on the relative url, and why we do this.

app_flutter/test/service/appengine_cocoon_test.dart
CaseyHillers: Right now the tests are grouped by the endpoint they relate to. Should we swap it so we group by the functionality under test?

CaseyHillers:
@jonahwilliams is it okay to table running the tests on chrome on Cirrus? We can't add Google chrome via the cirrus yml, so that is going to be a bigger change that requires syncing with the infra team and more research.

I can comment the line out on `cirrus.yml` and link to a Github issue on it to enable web tests.

app_flutter/test/service/appengine_cocoon_test.dart
digiter: Either way is fine.

PR #524, opened 2019-11-07
CocoonService error UI

CaseyHillers:
This adds notifications to the Build Dashboard when an error occurs fetching data. Anywhere that has access to the build state has access to using this functionality to show an error snackbar.

## Issues
- Fixes https://github.com/flutter/flutter/issues/43099

## Preview
error_snackbar_new


app_flutter/test/status_grid_test.dart
```diff
@@ -158,5 +161,134 @@ void main() {
...
+class TestFlutterBuildState extends ChangeNotifier
+    implements FlutterBuildState {
+  TestFlutterBuildState(
+      {@required this.isTreeBuildingValue, @required this.statusesValue});
+
+  // Only return an error on the first call
+  int hasErrorCallCount = 1;
+
+  @override
+  bool get hasError => hasErrorCallCount++ == 1;
```
digiter: This isn't flexible and straightforward, how about adding a method to set `_error` to true/false, and `hasError` simply returns that value.

app_flutter/lib/status_grid.dart
```diff
@@ -29,11 +36,28 @@ class StatusGridContainer extends StatelessWidget {
         final List<CommitStatus> statuses = buildState.statuses.data;
 
         if (buildState.hasError) {
-          print('FlutterBuildState has an error');
           print('isTreeBuilding: ${buildState.isTreeBuilding.error}');
           print('statuses: ${buildState.statuses.error}');
 
-          // TODO(chillers): Display the error
+          SchedulerBinding.instance.endOfFrame.then((_) {
```
jonahwilliams: I don't think you should trigger a side-effect like showing a snackbar as a side effect of a build method. An alternative way to structure this would be for the buildStatus to include a change notifier that you can subscribe to in initState/dispose. Then trigger the snackbar there, which removes the need for the post frame callback

CaseyHillers: Would this require switching the `hasErrors` member in `FlutterBuildState` to its own `ChangeNotifier`? Or should I make it a `Stream` that can be subscribed to?

app_flutter/test/status_grid_test.dart
CaseyHillers: Added. Switched type to `TestFlutterBuildState` since the type does not matter when being passed in, this allowed manually setting values.

app_flutter/lib/status_grid.dart
jonahwilliams: I think ChangeNotifiers are generally preferable to Streams, but both work. Streams don't have a "current" value, which makes timing issues easier to hit

CaseyHillers:
Updated to use a `ChangeNotifier` for errors. This will be a nice change as anywhere can take advantage of this `ChangeNotifer` to send error snackbars.

CaseyHillers:
Friendly ping

app_flutter/lib/build_dashboard.dart
```diff
@@ -11,29 +11,60 @@ import 'status_grid.dart';
...
-  final FlutterBuildState buildState = FlutterBuildState();
+  final GlobalKey<ScaffoldState> _scaffoldKey = GlobalKey<ScaffoldState>();
+
+  FlutterBuildState buildState;
 
   @override
   void initState() {
     super.initState();
 
+    buildState = widget.buildState;
```
jonahwilliams: Why do you capture the buildState here? If the widget rebuilds with a new BuildState object you will hold on to and old reference

CaseyHillers: Thanks for pointing this out! I moved this into the build method since it's just updating a reference.

PR #525, opened 2019-11-08
Switch to federated google sign in plugin

CaseyHillers:
The Flutter Google Sign In plugin should be getting official web support very soon. The open source fork the Flutter app was using did not get id tokens which required adding a new authentication method to the backend. This updates all the places in the Flutter app that reference the `GoogleSignInService` to use the new methods (using `GoogleSignInAccount` instead of separate fields).

Right now the web version is in the review stage, so this uses a forked version of it by the @ditman the PR owner.

Added ability to sign out by clicking the user avatar, and then clicking log out in the drop down.

Waiting on ~https://github.com/flutter/plugins/pull/2266~ and preferably https://github.com/flutter/plugins/pull/2280

## Preview

Demo available at http://testchillers.flutter-dashboard.appspot.com/v2/

logout screenshot

### DIFF
--- a/app_flutter/lib/service/google_authentication.dart
+++ b/app_flutter/lib/service/google_authentication.dart
@@ -41,9 +41,9 @@ class GoogleSignInService {
   /// Whether or not the application has been signed in to.
   bool get isAuthenticated => user != null;
 
-  /// The Google Account for the signed in user,
+  /// The Google Account for the signed in user, null if no user is signed in.
   ///
-  /// Null if no user is signed in.
+  /// Read only object with only access to clear client auth tokens.
   GoogleSignInAccount user;
 
   /// Authentication token to be sent to Cocoon Backend to verify API calls.
--- a/app_flutter/lib/service/google_authentication.dart
+++ b/app_flutter/lib/service/google_authentication.dart
@@ -2,18 +2,20 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-import 'package:google_sign_in_all/google_sign_in_all.dart';
+import 'package:google_sign_in/google_sign_in.dart';
 
 /// Service class for interacting with Google Sign In authentication for Cocoon backend.
 class GoogleSignInService {
   /// Creates a new [GoogleSignIn].
   GoogleSignInService({GoogleSignIn googleSignIn})
       : _googleSignIn = googleSignIn ??
-            setupGoogleSignIn(
+            GoogleSignIn(
               scopes: _googleScopes,
-              webClientId:
-                  '308150028417-vlj9mqlm3gk1d03fb0efif1fu5nagdtt.apps.googleusercontent.com',
-            );
+            ) {
+    _googleSignIn
+        .signInSilently()
+        .then((GoogleSignInAccount accountValue) => user = accountValue);
+  }
 
   /// A list of Google API OAuth Scopes this project needs access to.
   ///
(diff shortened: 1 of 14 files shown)


CaseyHillers:
All the plumbing is fixed, now I have to sort out how to get this dependency :)

app_flutter/lib/service/google_authentication.dart
```diff
@@ -24,30 +33,30 @@ class GoogleSignInService {
...
 
   /// Initiate the Google Sign In process.
   Future<void> signIn() async {
-    _credentials = await _googleSignIn.signIn();
-    _user = await _googleSignIn.getCurrentUser();
+    user = await _googleSignIn.signIn();
+  }
+
+  Future<void> signOut() async {
+    await _googleSignIn.signOut();
```
digiter: No action required, just want to mention that it's amusing to see "SignIn.signOut" :)

app_flutter/pubspec.yaml
```diff
@@ -31,7 +31,14 @@ dependencies:
...
   flutter_progress_button: ^1.0.0
-  google_sign_in_all: ^0.0.3
+  google_sign_in: ^4.0.14
+  google_sign_in_web:
+    git:
+      # TODO(chillers): https://github.com/flutter/plugins/pull/2280 and the Google Sign In plugin
+      # publishes the update switch to the official version.
+      url: git://github.com/ditman/plugins.git
+      ref: federated_google_sign_in_web
+      path: packages/google_sign_in/google_sign_in_web
```
ditman: I had no idea you could point to a repo+branch when using the git source. Nice!

app_flutter/lib/sign_in_button.dart
```diff
@@ -0,0 +1,47 @@
...
+class SignInButton extends StatelessWidget {
+  const SignInButton({@required this.authService, Key key}) : super(key: key);
+
+  final GoogleSignInService authService;
+
+  @override
+  Widget build(BuildContext context) {
+    if (authService.isAuthenticated) {
+      return PopupMenuButton<String>(
+        child: Image.network(authService.user.photoUrl),
```
ditman: There's a `GoogleUserCircleAvatar` widget provided by the `google_sign_in` plugin that you may be able to use. It's used in the [example](https://github.com/flutter/plugins/blob/master/packages/google_sign_in/google_sign_in/example/lib/main.dart#L115-L117) app.

CaseyHillers: Using `GoogleUserCircleAvatar` would throw 2 exceptions in the tests, but I could only catch the last one with `WidgetTester.takeException()` and it would fail because the first was not caught. So for now, I am just going to leave a TODO :)

CaseyHillers:
> The demo doesn't show avatar currently, is it expected?
> The rest LGTM.

@digiter that was to be expected at that time. I have redeployed with latest code from David's fork, and that seems to be working. I noticed an issue where it isn't working in incognito mode (works in release mode on my chrome, dev mode, but not incognito mode release mode), and need to follow up with David before submitting.

app_flutter/lib/service/google_authentication.dart
```diff
@@ -39,7 +39,7 @@ class GoogleSignInService {
   final GoogleSignIn _googleSignIn;
 
   /// Whether or not the application has been signed in to.
-  bool get isAuthenticated => user != null;
+  Future<bool> get isAuthenticated => _googleSignIn.isSignedIn();
```
ditman: If you don't want to change the getter signature you may be able to do:

```
bool get isAuthenticated => await _googleSignIn.isSignedIn();
```

ditman: (You might need to change it to an async fn, so maybe you still need to change the signature after all, hmmm)

CaseyHillers:
```
The await expression can only be used in an async function.
Try marking the function body with either 'async' or 'async*'
```

PR #527, opened 2019-11-08
Use Task Key

CaseyHillers:
The `Task` key has been null for the flutter app. This makes it non null and uses it correctly in the places it needs to be.

## Tested

- Updated test for building a CommitStatus to include the task key

## Future Work
- Fix authentication and v1 parity is met


CaseyHillers:
Friendly ping

PR #528, opened 2019-11-09
X-Flutter-IdToken header authentication

CaseyHillers:
This makes it easier for the Flutter application to authenticate with the Cocoon backend. It takes the current id token authentication method and allows it be run on the header `X-Flutter-IdToken` or the cookie `X-Flutter-IdToken`.

This is follow up from https://github.com/flutter/cocoon/pull/526 since the Google Sign In plugin is in the review phase for web support, and that works with id tokens.

## Tested

Added checks if the cookie is unauthenticated but the header is (in the case of just using the Flutter application), and if the cookie is authenticated but the header is not (in the case of using the performance dashboard)

### DIFF
--- a/app_dart/lib/src/request_handling/authentication.dart
+++ b/app_dart/lib/src/request_handling/authentication.dart
@@ -149,7 +149,7 @@ class AuthenticationProvider {
       /// request is authenticated.
       if (idTokenFromCookie != null) {
         /// The case where [idTokenFromCookie] is not valid but [idTokenFromHeader]
-        /// is requires us to catch the throw [Unauthenticated] exception.
+        /// is requires us to catch the thrown [Unauthenticated] exception.
         try {
           return await authenticateIdToken(idTokenFromCookie,
               clientContext: clientContext, log: log);
--- a/app_dart/lib/src/request_handling/authentication.dart
+++ b/app_dart/lib/src/request_handling/authentication.dart
@@ -7,6 +7,7 @@ import 'dart:convert';
 import 'dart:io';
 
 import 'package:appengine/appengine.dart';
+import 'package:cocoon_service/cocoon_service.dart';
 import 'package:dbcrypt/dbcrypt.dart';
 import 'package:gcloud/db.dart';
 import 'package:meta/meta.dart';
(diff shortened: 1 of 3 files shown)


app_dart/lib/src/request_handling/authentication.dart
```diff
@@ -135,68 +139,99 @@ class AuthenticationProvider {
...
-        TokenInfo token;
+    } else if (idTokenFromCookie != null || idTokenFromHeader != null) {
+      /// There are two possible sources for an id token:
+      ///
+      /// 1. Angular Dart app sends it as a Cookie
+      /// 2. Flutter app sends it as an HTTP header
+      ///
+      /// The Flutter application sends both since the client cookies are forwarded
+      /// in the request. One of these may be unauthenticated, so we need to try
+      /// authenticating both.
```
digiter: The criterion is unclear, is it "the user is authenticated as long as one id token is valid"?

app_dart/lib/src/request_handling/authentication.dart
```diff
@@ -135,68 +139,99 @@ class AuthenticationProvider {
...
-        final String clientId = await _config.oauthClientId;
-        assert(clientId != null);
-        if (token.audience != clientId) {
-          log.warning('Possible forged token: "${token.audience}" (expected "$clientId")');
-          throw const Unauthenticated('Invalid ID token');
-        }
+      if (idTokenFromHeader != null) {
+        return authenticateIdToken(idTokenFromHeader,
+            clientContext: clientContext, log: log);
+      }
```
digiter: Why the above one has `await`, `try catch` but not this one?

app_dart/lib/src/request_handling/authentication.dart
```diff
@@ -135,68 +139,99 @@ class AuthenticationProvider {
...
+        '/tokeninfo',
+        <String, String>{
+          'id_token': idToken,
+        },
+      ));
+      final HttpClientResponse verifyTokenResponse =
+          await verifyTokenRequest.close();
+
+      if (verifyTokenResponse.statusCode != HttpStatus.ok) {
+        final String body = await utf8.decodeStream(verifyTokenResponse);
```
digiter: I'm skeptical, when the response isn't ok, should we decode stream?

app_dart/lib/src/request_handling/authentication.dart
CaseyHillers: How does "As long as one of these two id tokens are authenticated, the request is authenticated." sound?

app_dart/lib/src/request_handling/authentication.dart
CaseyHillers: We wait to see if the user can be authenticated with the cookie id token first. This is to prevent making duplicate requests to the Google Auth API where we might be flagged for always sending two requests.

The try catch is to allow us fallback and try this method. If this last method doesn't authenticate, the user is not authenticated anyways.

Should I add a comment expanding on this more in the code?

app_dart/lib/src/request_handling/authentication.dart
CaseyHillers: This code just shows added since it was moved to its own method.

The reason this decodes it is because the Google Auth API can give some reasoning to why the user was unauthenticated. Usually this is just "invalid token", but it can be other things like "token expired." This lets us better understand in the logs why authentication failures are happening.

app_dart/lib/src/request_handling/authentication.dart
digiter: Yeah, a comment would be very helpful to clarify the differences.

PR #529, opened 2019-11-09
Run flutter clean at the end of every task.

godofredoc:
This is to remove build caches and make the tests more hermetic.


godofredoc:
Related bug:

https://github.com/flutter/flutter/issues/41303

jonahwilliams:
I think many of the tests run in temp directories, making this unnecessary. Another approach would be to fix them ones that don't

godofredoc:
Is not needed because tests are running in their own temp folders.

PR #523, opened 2019-11-07
Reland Dart project builds `app` project and serves it

CaseyHillers:
Original PR: https://github.com/flutter/cocoon/pull/513

Updates the Dart project to serve all the web assets out of it.
  - Deploy script builds the `app` project assets and uploads them to AppEngine
  - `bin/server.dart` instead of falling back to the Go code tries to serve static files

Closes https://github.com/flutter/flutter/issues/43095

Demo available at http://testchillers.flutter-dashboard.appspot.com/

~This is waiting on https://github.com/flutter/cocoon/pull/528, https://github.com/flutter/cocoon/pull/527, and https://github.com/flutter/cocoon/pull/525 before landing.~

~Also waiting on https://github.com/flutter/cocoon/pull/533 and https://github.com/flutter/cocoon/pull/531~

PR #531, opened 2019-11-13
Download devicelab logs

CaseyHillers:
Downloads the log from the Cocoon backend, encodes the log data to be uri safe, and downloads it in the browser. If it fails to download, a `SnackBar` with an error message will be shown.

If the task is not run in the devicelab, it will default to opening the source configuration (the logic v1 had).

Adds the dependency on `universal_html` to make web only features easy to add.

## Issues
Fixes https://github.com/flutter/flutter/issues/43102

Fixes https://github.com/flutter/flutter/issues/44770

## Tested
- Updated `TaskBox` tests to check for cases where we have devicelab tasks and non-devicelab tasks and whether they open the url or call `FlutterBuildState` to download the log.
- Updated `AppEngineCocoonService` tests to check for cases with downloading logs (request failures, web only, auth not given).
  - Web platform specific tests added that require manually running, but EOW there will be a PR that adds the custom dockerfile that makes it possible to run these web platform specific tests (https://github.com/flutter/flutter/issues/44373 for tracking).

## Open Questions

1. Currently this downloads the log to a file named ~`$taskName_$taskEndTimeStamp.log`~ `$taskName_$attemptNumber_$commitSha` where `$commitSha` is shortened to the first 7 characters like v1 does for shas. Does anyone have thoughts on a better naming structure? I am trying to keep it from becoming too verbose while maintaining the uniqueness of the log name.

Demo at https://testchillers-dot-flutter-dashboard.appspot.com/2/
  - Make sure you are signed in


app_flutter/lib/service/appengine_cocoon.dart
```diff
@@ -100,6 +101,50 @@ class AppEngineCocoonService implements CocoonService {
...
+        'X-Flutter-IdToken': idToken,
+      },
+    );
+
+    if (response.statusCode != HttpStatus.ok) {
+      print(response.body);
+      return false;
+    }
+
+    final String logEncodedContent = Uri.encodeComponent(response.body);
```
jonahwilliams: Is this slow on larger log messages? Is there a way we could use built-in browser functionality instead?

CaseyHillers: Good catch! This actually crashes on larger log messages. See [microbenchmark_ios](https://flutter-dashboard.appspot.com/api/get-log?ownerKey=ahFmbHV0dGVyLWRhc2hib2FyZHJYCxIJQ2hlY2tsaXN0IjhmbHV0dGVyL2ZsdXR0ZXIvNDM5ZDc5ODcxMjU1MDc2NTkwYjQ2YWUyMjA5OWEwZTIwZjA2YzA5YwwLEgRUYXNrGICA0PDOkekJDA). The browser returns `Failed - Network error` when trying to store that much data in `AnchorElement`. Granted, that link still crashes my browser when I open it. I haven't been able to locate a size limit for this current approach.

An alternative is to just write the idToken to a cookie for now, and open the log url. This would be the same as v1. The original reason for not doing this was because we didn't have storing the devicelab logs in GCS prioritized until now. With https://github.com/flutter/flutter/issues/43124 being worked on, we can just redirect users to the Google Cloud Log Viewer link with the GCS url passed in which is an ideal solution when that work is done.

Do you know of any other browser built in way for triggering downloads? There isn't much documentation on it, so I was only able to find the trick with using an anchor element to download the downloaded data.

CaseyHillers: For reference: The largest log I was able to download (just randomly trying) was 300kB.

jonahwilliams: I think you can put a `download` attribute on an anchor element and that will trigger a download when you click on it. i don't know if that will fix the crashing.

CaseyHillers: That's what the current approach did :)

I have rewritten it to write a cookie first, then just download via the log url. Before it was downloading the log data, then encoding it for the anchor element.

This approach works with the logs that crash chrome in v1 (1.9 mb log referenced above works like a charm).

Just need to fix the imports so it can run the tests :)

CaseyHillers:
@digiter thanks that's a good point! How does the format `$taskName_$taskAttemptNumber_$commitSha` sound? I think it's useful to be able to differentiate the logs when reruns are done, so that's what attempts is included.

digiter:
That sounds good to me.

app_flutter/lib/service/appengine_cocoon.dart
jonahwilliams: Sounds good!

app_flutter/lib/service/appengine_cocoon.dart
```diff
@@ -100,6 +101,47 @@ class AppEngineCocoonService implements CocoonService {
...
+
+    // This line is dangerous as it fails silently. Be careful.
+    html.document.cookie = 'X-Flutter-IdToken=$idToken;path=/';
+    // This wait is a hack as the above line is not synchronous. It takes time
+    // to write the cookie back to the browser. This is required before making
+    // the download request.
+    await Future<void>.delayed(const Duration(milliseconds: 500));
+
+    // Only show the first 7 characters of the commit sha.
+    final String shortSha = commitSha.substring(0, 6);
```
digiter: "Returns the substring of this string that extends from startIndex, inclusive, to endIndex, exclusive."

app_flutter/lib/service/appengine_cocoon.dart
```diff
@@ -100,6 +101,47 @@ class AppEngineCocoonService implements CocoonService {
...
+
+    final String getTaskLogUrl =
+        _apiEndpoint('/api/get-log?ownerKey=${task.key.child.name}');
+
+    // This line is dangerous as it fails silently. Be careful.
+    html.document.cookie = 'X-Flutter-IdToken=$idToken;path=/';
+    // This wait is a hack as the above line is not synchronous. It takes time
+    // to write the cookie back to the browser. This is required before making
+    // the download request.
+    await Future<void>.delayed(const Duration(milliseconds: 500));
```
digiter: Any way to avoid the hack?

app_flutter/lib/service/appengine_cocoon.dart
CaseyHillers: Thanks for pointing this out! I did a refactor so this can be testable too!

app_flutter/lib/service/appengine_cocoon.dart
CaseyHillers: One design I have in mind is to poll the cookies for some duration until the cookie has been written. This polling would wait until it reads the current id token for `X-Flutter-IdToken` or fail on some timeout. However, this design assumes that the `dart:html` interface for reading reads from the browser instead of what dart stores in memory.

CaseyHillers: I need more research on my part to see if the above is a feasible approach.

CaseyHillers:
@jonahwilliams to make the downloading code more testable, I refactored it into a separate downloading service. Do you know if there's a better way of doing this? Without this, the tests fail to build since `dart:html` can't be included in the normal `flutter test`.

https://github.com/flutter/cocoon/blob/b3db8c0ae482357b21d4402434b6645f79ae7a80/app_flutter/lib/service/downloader.dart

CaseyHillers:
I am happy with this now, so I am going to merge it in so we can focus on https://github.com/flutter/cocoon/pull/523 and the PR with UI tweaks for release.

PR #518, opened 2019-11-06
Implement StatusGrid performance improvements

CaseyHillers:
Follow up from the experimenting at https://github.com/flutter/cocoon/pull/511. Ideally, we want to use Skia, but the support is not there yet.

## Improvements
- Convert `StatusGrid` to be built with a `ListView` of `ListView`
- Convert Task in progress indicator to a static icon to reduce perceived jank
- Made fake data similar to production size

## Tested

Should be the same as the previous StatusGrid. Removed last task test because of weirdness with ListView ListView

Demo: https://testchillers-dot-flutter-dashboard.appspot.com/v2/#/

PR #533, opened 2019-11-16
Fix skipped tasks being collated in ListView's TaskBoxes

CaseyHillers:
A bug that started in https://github.com/flutter/cocoon/pull/518 where tasks that should be skipped or collated. This lead to skipped tasks not showing correctly and the grid becoming messed up.

The cause was becaused `SizedBox()` was not given a size for skipped tasks. With `GridView` this was okay as everything was constrained, however, this lead to them becoming hidden when switching to the `ListView` implementation.

## Tested

The only way I could think of regression testing this would be tapping at certain points on the screen and checking the task overlay to be correct. However, I did not know if this would become more so E2E testing of the application which is out of scope of the project.

Demo: https://testskipfix-dot-flutter-dashboard.appspot.com/v2/

### DIFF
--- a/app_flutter/test/status_grid_test.dart
+++ b/app_flutter/test/status_grid_test.dart
@@ -96,42 +96,43 @@ void main() {
       expect(firstTask.task, taskMatrix.task(0, 0));
     });
 
+    /// Matrix Diagram:
+    /// ✓☐☐
+    /// ☐✓☐
+    /// ☐☐✓
+    /// To construct the matrix from this diagram, each [CommitStatus] must have a unique [Task]
+    /// that does not share its name with any other [Task]. This will make that [CommitStatus] have
+    /// its task on its own unique row and column.
+    final List<CommitStatus> statusesWithSkips = <CommitStatus>[
+      CommitStatus()
+        ..stages.add(Stage()
+          ..name = 'A'
+          ..tasks.addAll(<Task>[
+            Task()
+              ..name = '1'
+              ..status = TaskBox.statusSucceeded
+          ])),
+      CommitStatus()
+        ..stages.add(Stage()
+          ..name = 'A'
+          ..tasks.addAll(<Task>[
+            Task()
+              ..name = '2'
+              ..status = TaskBox.statusSucceeded
+          ])),
+      CommitStatus()
+        ..stages.add(Stage()
+          ..name = 'A'
+          ..tasks.addAll(<Task>[
+            Task()
+              ..name = '3'
+              ..status = TaskBox.statusSucceeded
+          ]))
+    ];
+
     testWidgets('skipped tasks do not break the grid',
         (WidgetTester tester) async {
-      /// Matrix Diagram:
-      /// ✓☐☐
-      /// ☐✓☐
-      /// ☐☐✓
-      /// To construct the matrix from this diagram, each [CommitStatus] must have a unique [Task]
-      /// that does not share its name with any other [Task]. This will make that [CommitStatus] have
-      /// its task on its own unique row and column.
-      final List<CommitStatus> statuses = <CommitStatus>[
-        CommitStatus()
-          ..stages.add(Stage()
-            ..name = 'A'
-            ..tasks.addAll(<Task>[
-              Task()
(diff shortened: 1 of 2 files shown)


jonahwilliams:
Consider placing a value key on the empty sized box - then you could would write a test that rendered the screen with missing tasks and searched for the valuekey + type, making an assertion on the size

CaseyHillers:
Thanks for the suggestion! I was thinking at the bigger picture level (maybe because I noticed the bug by comparing the big TV version to my monitor :P ).

I added the ability to insert keys to the `StatusGrid` cells based on their location in the grid. This makes it easier to lookup the widget corresponding to a cell for comparison. Enhanced one of the previous tests to ensure skipped tasks existed, and followed up with a regression test to ensure all the cells in the grid have the same size.

app_flutter/test/status_grid_test.dart
```diff
@@ -96,48 +96,50 @@ void main() {
...
+          ])),
+      CommitStatus()
+        ..stages.add(Stage()
+          ..name = 'A'
+          ..tasks.addAll(<Task>[
+            Task()
+              ..name = '3'
+              ..status = TaskBox.statusSucceeded
+          ]))
+    ];
```
digiter: I have difficulty in understanding how this commit status list be converted to the matrix diagram above.

CaseyHillers: Would it be helpful if I added the comment below the diagram stating:
```
/// To construct the matrix from this diagram, each [CommitStatus] must have a unique [Task] that does not share its name with any other [Task].
```

CaseyHillers:
I accidentally overwrote this test version with a demo for downloading logs this morning. The new demo link is located at https://testskipfix-dot-flutter-dashboard.appspot.com/v2/

Updated demo link at top to reflect that.

app_flutter/test/status_grid_test.dart
digiter: Yes, that would be helpful.

CaseyHillers: Done.

CaseyHillers:
Uh oh, the tests got removed when I did a rebase. Let me fix that.

CaseyHillers:
Tests readded. Must have been I forgot to fetch origin and didn't think twice on doing a force push.

app_flutter/test/status_grid_test.dart
```diff
@@ -148,15 +153,78 @@ void main() {
...
+      expect(skippedTaskRow3Col2, isNotNull);
+
+      final TaskBox lastTask =
+          find.byKey(const Key('cell-2-2')).evaluate().first.widget;
+      expect(lastTask.task, statusesWithSkips[2].stages[0].tasks[0]);
+    });
+
+    testWidgets('all cells in the grid have the same size',
+        (WidgetTester tester) async {
+      /// Use [statusesWithSkips] to check edge cases with skipped tasks.
```
digiter: This comment is simply repeating the code. Consider removing it or merging it into the test description.

app_flutter/test/status_grid_test.dart
```diff
@@ -148,15 +153,78 @@ void main() {
...
+          ),
+        ),
+      );
+
+      // Compare all the cells to the first cell to check they all have
+      // the same size
+      final Element taskBox =
+          find.byKey(const Key('cell-0-0')).evaluate().first;
+      for (int rowIndex = 0; rowIndex < statusesWithSkips.length; rowIndex++) {
+        for (int colIndex = 0; colIndex < 3; colIndex++) {
```
digiter: Avoid hard coded 3?

app_flutter/test/status_grid_test.dart
CaseyHillers: Thanks for pointing that out! I updated the test description to reflect the edge case the comment mentioned.

app_flutter/test/status_grid_test.dart
CaseyHillers: Good catch! Switched these to use the `TaskMatrix` values.

PR #534, opened 2019-11-19
Explicit tree status in AppBar

CaseyHillers:
Replaces the "Build Dashboard v2" title in `AppBar` to be `Tree is Open` or `Tree is Closed`. Additionally, when the tree is open it is set to match the color of a successful task box (green).

Recommended by @digiter 

## Tested

Added tests to check the text and color change of the AppBar based on tree status.

## Preview

Demo: https://testsummary-dot-flutter-dashboard.appspot.com

Tree is Open

Tree is Closed

### DIFF
--- a/app_flutter/lib/main.dart
+++ b/app_flutter/lib/main.dart
@@ -8,7 +8,7 @@ import 'build_dashboard.dart';
 
 void main() => runApp(MyApp());
 
-ThemeData theme = ThemeData(
+final ThemeData theme = ThemeData(
   appBarTheme: AppBarTheme(color: Colors.green),
   primarySwatch: Colors.blue,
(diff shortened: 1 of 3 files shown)


app_flutter/lib/main.dart
```diff
@@ -8,14 +8,17 @@ import 'build_dashboard.dart';
 
 void main() => runApp(MyApp());
 
+ThemeData theme = ThemeData(
```
jonahwilliams:
```suggestion
final ThemeData theme = ThemeData(
```

CaseyHillers: Thanks! It's weird that the linter didn't complain about this.

PR #537, opened 2019-11-20
Commit Overlay and Task Overlay UI Cleanup

CaseyHillers:
These changes are to help make the app more explicit and concise. This was feedback given from the team with the v1 dashboard.

Removed the place holder button for rerunning all failed tasks for a commit: Requires a day of work to handle the errors that occur, as right now worst case you will get 10 minutes of error snackbars if the application is not signed in.

Expanded the open in GitHub icon button a button that just says GitHub.

Shortened commit sha in to just show the first 7 characters which is unique enough to lookup a flutter/flutter commit.

Renamed the "Rerun task" button to "Rerun".

Fixed the duration in the task overlay to be in minutes

## Open Questions

- Is there a way to make text copyable on web?

## Preview

Demo: https://testui-dot-flutter-dashboard.appspot.com/v2/

image

image

image

### DIFF
--- a/app_flutter/lib/commit_box.dart
+++ b/app_flutter/lib/commit_box.dart
@@ -3,6 +3,7 @@
 // found in the LICENSE file.
 
 import 'package:flutter/material.dart';
+import 'package:flutter_progress_button/flutter_progress_button.dart';
 import 'package:url_launcher/url_launcher.dart';
 
 import 'package:cocoon_service/protos.dart' show Commit;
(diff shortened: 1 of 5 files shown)


jonahwilliams:
Maybe https://github.com/flutter/flutter/blob/f0553ba58e6455aa63fafcdca16100b81ff5c3ce/packages/flutter/lib/src/material/selectable_text.dart#L181 ?

CaseyHillers:
> Maybe https://github.com/flutter/flutter/blob/f0553ba58e6455aa63fafcdca16100b81ff5c3ce/packages/flutter/lib/src/material/selectable_text.dart#L181 ?

@jonahwilliams `SelectableText` wasn't able to be copied on web (works on mobile).

The Clipboard service in Flutter was not working on web either.

CaseyHillers:
padding_fix

@digiter Good catch! I pushed a change to improve it. Does this look better?

CaseyHillers:
I noticed that we only call `notifyListener()` when both the build status and statuses are returned. I updated to do them individually as the build status has been taking ~6 seconds. It might be a good request to cache.

PR #539, opened 2019-11-21
Pipe Commit through TaskBox for Cirrus logs

CaseyHillers:
Cirrus needs the Commit sha appended to the end of the Cirrus URL to end up on the log specific for that Cirrus task.

Plumbed `Commit` variable to this location to make it easier to expand the `TaskOverlay` with more information regarding the task in a later commit.

## Tested

Added tests for checking the cirrus log redirects to the sha url when Commit is given instead of master.

## Preview

Demo: https://testchillers-dot-flutter-dashboard.appspot.com/build.html


CaseyHillers:
Thank you @jmagman for pointing this issue out!

PR #540, opened 2019-11-21
Do not use the MERGEABLE flag

dnfield:
Fixes https://github.com/flutter/flutter/issues/45357

The API does not return the flag in a useable way for this.  If the PR isn't actually mergeable we'll know when we try to merge it anyway.


liyuqian:
Wow, Dan you're fixing the bug more quickly than I can read through my emails... I'm just scanning my emails, filing a issue when I see the merge conflict email, and then I see an email saying that the bug is fixed... Thank you so much @dnfield !

PR #545, opened 2019-11-25
Tree status for loading

CaseyHillers:
Update the tree status to use the null value. This is used on initialization to let users know that we're loading the value. During this state, it shows a grey `AppBar` that says loading.

Originally, we just showed the failed status since that's the worst case. However, it was marginally noticeable as `/api/public/build-status` would return in a few seconds. Lately, it has been averaging at ~7 seconds.

## Issues

Fixes https://github.com/flutter/flutter/issues/45285

## Preview

loading_appbar

Demo at https://testchillers-dot-flutter-dashboard.appspot.com/build.html

### DIFF
--- a/app_flutter/lib/build_dashboard.dart
+++ b/app_flutter/lib/build_dashboard.dart
@@ -83,12 +83,14 @@ class BuildDashboard extends StatelessWidget {
   Widget build(BuildContext context) {
     final ThemeData theme = Theme.of(context);
 
+    /// Color of [AppBar] based on [buildState.isTreeBuilding].
     final Map<bool, Color> colorTable = <bool, Color>{
       null: Colors.grey,
       false: theme.errorColor,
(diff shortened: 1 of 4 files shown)


app_flutter/lib/build_dashboard.dart
```diff
@@ -82,16 +82,27 @@ class BuildDashboard extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     final ThemeData theme = Theme.of(context);
+
+    /// Color of [AppBar] based on [buildState.isTreeBuilding].
+    final Map<bool, Color> colorTable = <bool, Color>{
+      null: Colors.grey,
```
jonahwilliams: Good ol' tri-state booleans :)

CaseyHillers: Is this quantum computing? :P

PR #547, opened 2019-11-25
Web tests for Cocoon

CaseyHillers:
This is left overs from trying to get web tests running on cirrus for Cocoon (https://github.com/flutter/flutter/issues/44373). We ended up not being able to do the web tests as they require too many resources that are not available on the community tier.

This updates the tests to be compatible when running `flutter test --platform=chrome`


CaseyHillers:
The custom dockerfile looks complete. Just need to update the tests to work with web :)

CaseyHillers:
Running into issues with the Cirrus tests for the web tests:

`Connection closed before test suite loaded.`

` - did not complete [E] `

Running the same set up locally (flutter dev channel) does not produce these errors.

CaseyHillers:
@jonahwilliams I noticed adding this change where we start running web tests is flaky. It takes a few runs to get a successful run. Is there a way to give more time for the web tests?

This is the error as it seems Cirrus starts to timeout (but for different files each time).
```
  Failed to load "/tmp/cirrus-ci-build/app_flutter/test/task_box_test.dart": Connection closed before test suite loaded.
```

jonahwilliams:
Adding @yjbanov as the expert at getting web tests working on cirrus

CaseyHillers:
@yjbanov I tried dealing with this timeout issue on the CI side (splitting it so web tests are run as a separate task), but the tests still timeout. Do you have any ideas on how to get this to not timeout on the tool side?

CaseyHillers:
Adding more memory seems to be the trick, but now Cirrus runs into this when running the web tests which has never happened before.

```
Unable to find modules for some sources, this is usually the result of either a
bad import, a missing dependency in a package (or possibly a dev_dependency
needs to move to a real dependency), or a build failure (if importing a
generated file).
Please check the following imports:
`import 'generated_plugin_registrant.dart';` from app_flutter|lib/main_web_entrypoint.dart at 5:1
```

PR #548, opened 2019-11-26
Runner scripts to run cocoon tests on luci.

godofredoc:
We need new scripts that encapsulates the commands cirrus was running
for tests in this repo.

https://github.com/flutter/flutter/issues/43348

### DIFF
--- a/test_utilities/bin/flutter_test_runner.sh
+++ b/test_utilities/bin/flutter_test_runner.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+# Copyright 2019 The Chromium Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+# Runner for flutter tests. It expects a single parameter with the full
+# path to the flutter project where tests will be run.
+
+set -e
+
+echo "Running flutter tests from $1"
+pushd $1 > /dev/null
+
+flutter packages get
+# Remove the " || true" portion when https://github.com/flutter/flutter/issues/44370
+# is fixed.
+flutter analyze || true
+flutter config --enable-web
+flutter build web
+flutter test
+
+popd > /dev/null
+
--- a/test_utilities/bin/flutter_test_runner.sh
+++ b/test_utilities/bin/flutter_test_runner.sh
@@ -15,6 +15,7 @@ flutter packages get
 # Remove the " || true" portion when https://github.com/flutter/flutter/issues/44370
 # is fixed.
 flutter analyze || true
+flutter format --set-exit-if-changed lib/ test/
 flutter config --enable-web
 flutter build web
 flutter test
--- a/test_utilities/pubspec.yaml
+++ b/test_utilities/pubspec.yaml
@@ -1,0 +1,12 @@
+name: global_test_runner
+author: Flutter Authors <flutter-dev@googlegroups.com>
+description: Flutter continuous integration agent (Cocoon agent)
+homepage: https://github.com/flutter/cocoon
+
+environment:
+  sdk: '>=2.2.0'
+
+dev_dependencies:
+  args: ^1.5.2
+  yaml: ^2.2.0
+  path: any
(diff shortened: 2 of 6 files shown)


test_utilities/pubspec.yaml
```diff
@@ -0,0 +1,12 @@
+name: global_test_runner
```
CaseyHillers: Does this file need a copyright header?

tests.yaml
```diff
@@ -0,0 +1,19 @@
+# Tests configuration file.
```
CaseyHillers: Does this file need a copyright header too?

test_utilities/bin/flutter_test_runner.sh
```diff
@@ -0,0 +1,23 @@
...
+
+set -e
+
+echo "Running flutter tests from $1"
+pushd $1 > /dev/null
+
+flutter packages get
+# Remove the " || true" portion when https://github.com/flutter/flutter/issues/44370
+# is fixed.
+flutter analyze || true
```
CaseyHillers: Flutter analyze doesn't throw an issue with the linked issue. Is there another issue luci runs into?

test_utilities/bin/flutter_test_runner.sh
```diff
@@ -0,0 +1,23 @@
...
+flutter config --enable-web
```
CaseyHillers: Missing `flutter format --set-exit-if-changed lib/ test/` which ensures the code being committed followers the formatter

test_utilities/bin/flutter_test_runner.sh
godofredoc: info • Unused import: 'dart:ui' • lib/generated_plugin_registrant.dart:4:8 • unused_import and it finishes with a status code 1

CaseyHillers: Is this doing a fresh install of the flutter app? We dealt with this on cirrus by running `flutter analyze` before any of the flutter scripts had run that would generate `lib/generated_plugin_registrant.dart`.

test_utilities/bin/flutter_test_runner.sh
godofredoc: Added!

test_utilities/bin/flutter_test_runner.sh
godofredoc: This is just after running "flutter channel dev" and "flutter upgrade". Ideally from luci we will checkout the source code once and run tests in the different paths using a single bot.

CaseyHillers: With the case of checking out the code once, we may just want to `rm lib/generated_plugin_registrant.dart` before running analyze. That was the alternative when we faced this issue on Cirrus.

godofredoc: Isn't that file needed for anything in the flutter app?

CaseyHillers: It's necessary for using plugins in the Flutter app, but for testing it's not necessary since all the plugins are mocked in testing. It's a side effect from building the app, and we build the app as another check to ensure master can be deployed at any time.

test_utilities/pubspec.yaml
```diff
@@ -5,6 +5,9 @@ homepage: https://github.com/flutter/cocoon
 
 environment:
   sdk: '>=2.2.0'
+# Copyright 2019 The Chromium Authors. All rights reserved.
```
CaseyHillers: I think this got pasted on the wrong line :)

godofredoc: Fixed!

test_utilities/bin/flutter_test_runner.sh
godofredoc: "flutter analyze" fails with the same line even in a fresh checkout of flutter/cocoon is there a bug where we can keep the discussion? I still think this needs to get fixed rather than workarounded in infra.

CaseyHillers: I think web plugins are still relatively WIP so this will change, but I filed an issue at https://github.com/flutter/flutter/issues/45711 for keeping track of this.

godofredoc: Thanks, I found many other small details we need to address before starting to use these scripts. I'll open separate bugs.

PR #553, opened 2019-11-28
Add linux VM support to devicelab.

godofredoc:
Adds a NoOp class for device discovery to bypass all the
device checks on VMs. This will be used to create several VMs and start
moving tests.

https://github.com/flutter/flutter/issues/27917

### DIFF
--- a/agent/lib/src/adb.dart
+++ b/agent/lib/src/adb.dart
@@ -14,7 +14,10 @@ import 'utils.dart';
 DeviceDiscovery get devices => DeviceDiscovery();
 
 /// Operating system on the devices that this agent is configured to test.
-enum DeviceOperatingSystem { android, ios }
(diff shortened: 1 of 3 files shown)


agent/lib/src/adb.dart
```diff
@@ -316,6 +321,34 @@ class IosDeviceDiscovery implements DeviceDiscovery {
   }
 }
 
+class NoOpDeviceDiscovery implements DeviceDiscovery {
```
keyonghan: How about naming it as 'vmDeviceDiscovery', and adding comments explaining to bypass device discovery?

keyonghan:
@godofredoc you may want to rerun pre-checks. The failure of 'app_flutter' is due to issue solved in [PR 552](https://github.com/flutter/cocoon/pull/552).

PR #555, opened 2019-11-28
Update test scripts to be able to run all tests with a command.

godofredoc:


test_utilities/bin/global_test_runner.dart
```diff
@@ -23,20 +22,24 @@ Future<Null> main(List<String> rawArgs) async {
...
+  doc['tasks'].forEach((task) async {
     String scriptPath = join(baseDir, task['script']);
     String taskPath = join(baseDir, task['task']);
-    Process.run('sh', <String>[scriptPath, taskPath]).then((result) {
+    await runShellCommand(<String>[scriptPath, taskPath], task['task']);
+  });
+}
+
+void runShellCommand(List<String> args, String taskName) async {
+  Process.run('sh', args).then((result) {
```
digiter: Nit: I'm slightly concerned about using `sh` might not be consistent across different platforms.

godofredoc: Cocoon tests are currently running only on Linux.

PR #556, opened 2019-12-03
Increase history max record for benchmark

keyonghan:
This PR increases the max record from 1500 to 6000 for benchmark history chart.


liyuqian:
Nit: the diff looks good, but the commit history seems to be very strange. You might need to rebase (it will happen anyway when you merge).

liyuqian:
https://github.com/flutter/flutter/issues/29177#issuecomment-556736348 is the context of why we started this PR. CC @CaseyHillers to double check if 6000 Ok with the Redis memcache.

keyonghan:
Yeah, the commit history is due to merges.. I need to deal with that later soon. Thanks.

CaseyHillers:
This endpoint isn't cached in redis, but we can add it if there's interest.

digiter:
There's messy commit history on this PR, here's a suggestion: always create a new branch from master for a new feature, and always pull/rebase upstream master. Ian's repo has examples, see `new` and `pull` in https://github.com/Hixie/usable-git/.

PR #564, opened 2019-12-06
Upload devicelab logs to GCP Stackdriver logging

CaseyHillers:
This updates `/api/append-log` to forward the `LogChunk` data to [Google Cloud Stackdriver](https://cloud.google.com/stackdriver/). Stackdriver offers great functionality for viewing and understanding logs. The downside is its not UTF16 compatible so some of the pretty Flutter log symbols get simplified.

Stackdriver logs are an improvement for those with access to the Google Cloud project (i.e. `*@google.com`). Whitelisted accounts will still be able to download the log from Datastore.

## Issues

Fixes https://github.com/flutter/flutter/issues/46902

Possible fix to https://github.com/flutter/flutter/issues/43124
  - Flutter app needs to have UI added to redirect to these urls

## Tested

- Check to see if the split lines functionality was working as expected

## Preview

https://console.cloud.google.com/logs/viewer?project=flutter-dashboard&resource=global&minLogLevel=0&expandAll=false&interval=NO_LIMIT&dateRangeUnbound=backwardInTime&logName=projects%2Fflutter-dashboard%2Flogs%2FahFmbHV0dGVyLWRhc2hib2FyZHJYCxIJQ2hlY2tsaXN0IjhmbHV0dGVyL2ZsdXR0ZXIvYTQ5MTY5Nzg5M2IzY2IyMGY0NWI4Y2UxNGQ3MDI0N2M4MjcxNTNjZAwLEgRUYXNrGICA0Oa93pkJDA&timestamp=2019-12-11T00:06:30.736000000Z&customFacets=&limitCustomFacetWidth=true&scrollTimestamp=2019-12-10T01:38:28.323638714Z

## Future Work

Add UI to `app_flutter` to offer the ability to view these logs from the build dashboard.


CaseyHillers:
@keyonghan I figured I would share this. The Dart AppEngine has access to the storage service already due to the service scopes from the appengine library.

app_dart/lib/src/request_handlers/append_log.dart
```diff
@@ -3,41 +3,56 @@
...
 import '../model/appengine/key_helper.dart';
 import '../model/appengine/log_chunk.dart';
+import '../model/appengine/task.dart';
 import '../request_handling/api_request_handler.dart';
 import '../request_handling/authentication.dart';
 import '../request_handling/body.dart';
 import '../request_handling/exceptions.dart';
 
-@immutable
+// ignore: must_be_immutable
```
digiter: Why this becomes ignored?

app_dart/lib/src/request_handlers/append_log.dart
```diff
@@ -51,6 +66,13 @@ class AppendLog extends ApiRequestHandler<Body> {
       await transaction.commit();
     });
 
+    await writeToStackdriver('${encodedOwnerKey}_${task.attempts}');
```
digiter: It would be more readable / convenient to have an example logName in comment.

app_dart/lib/src/service/stackdriver_logger.dart
```diff
@@ -0,0 +1,70 @@
...
+import 'package:googleapis_auth/auth_io.dart';
+import 'package:http/http.dart';
+import 'package:meta/meta.dart';
+
+import '../datastore/cocoon_config.dart';
+
+/// Service class for interacting with the AppEngine project's stackdriver logging service.
+///
+/// Stackdriver is a Google Cloud service for monitoring and managing services.
+/// https://cloud.google.com/stackdriver/ for more details.
```
digiter: Nit: make it a complete sentence: "See https://...".

app_dart/lib/src/request_handlers/append_log.dart
```diff
@@ -51,6 +66,13 @@ class AppendLog extends ApiRequestHandler<Body> {
...
+
     return Body.empty;
   }
+
+  Future<void> writeToStackdriver(String logName) async {
+    final List<String> lines = String.fromCharCodes(requestBody).split('\n');
+    await stackdriverLogger.writeLines(logName, lines);
+  }
```
digiter: Please file a bug and add a TODO here to keep track of the UTF16 issue.

app_dart/lib/src/request_handlers/append_log.dart
CaseyHillers: Updated to follow the style guide: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#comment-all--ignores

We need to be able to inject `requestBody` for the `AppendLog` tests. Ideally, I would have updated `ApiRequestHandler` (the parent class), but doing so makes it no longer immutable, and removes immutability from all the children. I don't know of a good solution yet, so I have just filed an issue to track progress on improving this: https://github.com/flutter/flutter/issues/46902

Figuring this out means we can get more test coverage on the backend code that uses requestBody as it seems no where else in the code tests when requestBody is used.

app_dart/lib/src/request_handlers/append_log.dart
CaseyHillers: Added!

app_dart/lib/src/service/stackdriver_logger.dart
CaseyHillers: Good catch

app_dart/lib/src/request_handlers/append_log.dart
CaseyHillers: Added! Stackdriver seems to have this at a won't fix, but I filed the issue on our tracker so we know it's a known limitation.

app_dart/lib/src/request_handlers/append_log.dart
digiter: Instead of injecting `requestBody`, how about using a local http client to send the request with body? Similar to https://github.com/flutter/cocoon/blob/master/app_dart/test/request_handling/api_request_handler_test.dart#L49

CaseyHillers: I made it a constructor injection to get this working. Originally, I was trying to do it with parameter injection. This ignore is no longer needed as I was able to keep everything immutable.

CaseyHillers:
Force pushed to update with the dartfmt PR

app_dart/lib/src/datastore/cocoon_config.dart
```diff
@@ -99,6 +100,11 @@ class Config {
     return ServiceAccountInfo.fromJson(json.decode(rawValue));
   }
 
+  Future<ServiceAccountCredentials> get taskLogServiceAccount async {
+    final String rawValue = await _getSingleValue('TaskLogServiceAccount');
```
keyonghan: Why not use existing DevicelabServiceAccount, and just grant logging.write to it?

CaseyHillers: My understanding of service accounts is that in a perfect world they should have the minimum permissions needed. I didn't want to start a pattern of adding a bunch of responsibilities to one service account which would make it hard to track if there are any unnecessary permissions granted to the account.

If we remove this logging functionality down the line, it makes it easy to just delete this service account with this feature than to figure out if the permissions are used elsewhere.

app_dart/lib/src/request_handlers/append_log.dart
```diff
@@ -3,25 +3,37 @@
...
+    StackdriverLoggerService stackdriverLogger,
+    @visibleForTesting Uint8List requestBodyValue,
+  })  : stackdriverLogger =
+            stackdriverLogger ?? StackdriverLoggerService(config: config),
+        super(
+            config: config,
+            authenticationProvider: authenticationProvider,
+            requestBodyValue: requestBodyValue);
+
+  final StackdriverLoggerService stackdriverLogger;
```
keyonghan: I have a couple of questions/concerns regarding moving log to Stackdriver:

1. Now it is easy to search with Stackdriver, which is good. However our existing logic is also easy to search.
2. With Stackdriver, I just tried searching and loading log data
2.1 It is not convenient to loading log data, is there any way to have a one-time load for all logs? Currently I have to keep scrolling down or up to load next chunk of log. Once loaded, and I have to load it again when scroll back. 
2.2 Once I searched some key-words, and after I clicked one record in the middle, it is difficult for me to view neighborhood logs.
Both of these would make it not easy to debug/look for specific information.
3. Is Stackdriver mainly for log view of executing tasks, or both? Considering above two points, have you tried GCS? I see there is a bucket named 'flutter-task-logs' in flutter-dashboard. Could we at least point to log file in GCS for finished tasks?

Please let me know if I am wrong.

app_dart/lib/src/request_handlers/append_log.dart
```diff
@@ -54,6 +66,19 @@ class AppendLog extends ApiRequestHandler<Body> {
...
+
     return Body.empty;
   }
+
+  /// Write the log data from this request to Stackdriver under [logName].
+  ///
+  /// [logName] must follow the format `[encodedOwnerKey]_[task.attempt]` so each
+  /// attempt for a task can be located.
+  ///
+  /// This will write the log to the global path `projects/flutter-dashboard/logs/[logName]` on Google Cloud.
```
keyonghan: Is there a GUI entrance to view these path and logs?

app_dart/test/request_handlers/append_log_test.dart
```diff
@@ -0,0 +1,72 @@
...
+import 'package:test/test.dart';
+
+import 'package:cocoon_service/src/request_handlers/append_log.dart';
+import 'package:cocoon_service/src/service/stackdriver_logger.dart';
+
+import '../src/datastore/fake_cocoon_config.dart';
+import '../src/request_handling/fake_authentication.dart';
+
+void main() {
+  group('AppendLog', () {
```
keyonghan: I believe append_log_test.dart was missing before this PR. It would be good to add tests for existing features in append_log.dart. Thanks.

app_dart/test/request_handlers/append_log_test.dart
```diff
@@ -0,0 +1,72 @@
...
+    FakeConfig config;
+    AppendLog handler;
+
```
keyonghan: Nit: remove empty line.

app_dart/lib/src/service/stackdriver_logger.dart
```diff
@@ -0,0 +1,71 @@
...
+class StackdriverLoggerService {
+  /// Creates a new [StackdriverLoggerService].
+  ///
+  /// The [config] argument must not be null.
+  StackdriverLoggerService({@required this.config}) : assert(config != null);
+
+  /// The Cocoon configuration. Guaranteed to be non-null.
+  final Config config;
+
+  /// Interface to interacting with Stackdriver.
```
keyonghan: Nit: rephrase comments.

app_dart/lib/src/request_handlers/append_log.dart
CaseyHillers: 1. This is just to let the team evaluate Stackdriver. From https://flutter.dev/go/build-dashboard-v2 the team requested a GUI for viewing logs, and this is a possible solution.
2. Scrolling to see the whole log is a pain. However, a few contributors pointed out that they only really check the end of the log as that's where the failures tend to be. Can you clarify what you mean by neighborhood logs?
3. This PR implemented GCS initially, but there was no direct advantage to it. Initially I investigated GCS as a solution as I thought Stackdriver would support that, but on implementation I was unable to load a GCS file in the Stackdriver viewer.

app_dart/lib/src/request_handlers/append_log.dart
CaseyHillers: The follow up PR to this has more info on the GUI aspect of this change: https://github.com/flutter/cocoon/pull/575

app_dart/test/request_handlers/append_log_test.dart
CaseyHillers: Done

app_dart/lib/src/service/stackdriver_logger.dart
CaseyHillers: Done

app_dart/lib/src/request_handlers/append_log.dart
keyonghan: 2.2 Suppose we have 1000 lines (line 1 to line 1000) in the log, and searching 'error' returns me 4 lines (line 100, line 200, line 300, line 400). If one wants to see the pre or post logs around line 300, he has to clear the 'error' keyword, and refresh the page, and then scroll to the lines around 200.

Oh, there is one option to 'pin this log' at the end of each line. This seems to help.

3. GCS at least gives the quick downloading option for the whole log, if people wants to search around/move around/scroll the whole log file. 

Anyway, this is a temporary solution for log view. We can keep Stackdriver to see how it goes.

app_dart/test/request_handlers/append_log_test.dart
CaseyHillers: Added!

CaseyHillers:
@keyonghan I believe I have addressed all your comments.

PR #567, opened 2019-12-10
Disable index for Agent HealthDetails property

keyonghan:
This PR solves https://github.com/flutter/flutter/issues/46239

### DIFF
--- a/app_dart/bin/server.dart
+++ b/app_dart/bin/server.dart
@@ -10,6 +10,7 @@ import 'package:cocoon_service/cocoon_service.dart';
 import 'package:gcloud/db.dart';
 import 'package:neat_cache/neat_cache.dart';
 
+
 Future<void> main() async {
   await withAppEngineServices(() async {
     final Config config = Config(dbService);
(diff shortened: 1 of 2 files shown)


keyonghan:
Check here: https://cloud.google.com/appengine/docs/standard/python/config/indexconfig
> Every Datastore query made by an application needs a corresponding index. Indexes for simple queries, such as queries over a single property, are created automatically. Indexes for complex queries must be defined in a configuration file named index.yaml

keyonghan:
> LGTM with nit
> 
> Is it possible to add a regression test to check that this is set?

A test for this may be skipped. From gcloud-0.6.3 lib/src/db/annotations.dart, it has comments below. So that is a direct mapping from dart end to datastore end.

> /// Annotation used to mark dart classes which can be stored into datastore.
> ///
> /// The `Kind` annotation on a class as well as other `Property` annotations on
> /// fields or getters of the class itself (and any of it's superclasses) up to
> /// the [Model] class describe the *mapping* of *dart objects* to datastore
> /// *entities*.

CaseyHillers:
I was thinking checking
https://pub.dev/documentation/analyzer/latest/dart_element_element/Element/metadata.html
to ensure the annotation stays to help prevent a regression on this.

PR #568, opened 2019-12-11
Unlock keychain before running a task.

godofredoc:
This is a prerequisite for auto-restart agents when they become
unhealthy. Also we will add functionality for nightly reboots of
devicelab host machines to improve testbed stability.

https://github.com/flutter/flutter/issues/43125

### DIFF
--- a/agent/lib/src/commands/ci.dart
+++ b/agent/lib/src/commands/ci.dart
@@ -39,6 +39,8 @@ class ContinuousIntegrationCommand extends Command {
     // is wrong.
     AgentHealth health = await performHealthChecks(agent);
     section('Pre-flight checks:');
+    // Ensure keychain is unlocked before running health checks.
+    await _unlockKeyChain();
     health
         .toString()
         .split('\n')
@@ -66,6 +68,7 @@ class ContinuousIntegrationCommand extends Command {
         // console.
         try {
           section('Preflight checks');
+          
           await devices.performPreflightTasks();
 
           // Check health before requesting a new task.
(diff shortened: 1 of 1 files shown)


agent/lib/src/commands/ci.dart
```diff
@@ -228,6 +233,15 @@ class ContinuousIntegrationCommand extends Command {
     }
   }
 
+  Future<Null> _unlockKeyChain() async {
+    if (Platform.isMacOS) {
+      await exec(
+        'security',
+        <String>['unlock-keychain', '-p', '\$SECRET', 'login.keychain'],
```
digiter: Consider rename "SECRET" to something like "FLUTTER_PASS" because there will be more secrets in the future, e.g. fastlane code signing will add a secret to decode encrypted provisioning profiles.

godofredoc: Done! updated the env variable name.

godofredoc:
Yes, we need to add recovery mechanisms. The way I've thinking about is running health checks and keep a key that we can reference from the recovery methods. E.g. we run too_many_files health check and it fails, the on the recovery script we already know we need to kill a couple of processes. I think with this change we can benefit from automating killing and restarting the agent process.

agent/lib/src/commands/ci.dart
```diff
@@ -228,6 +233,16 @@ class ContinuousIntegrationCommand extends Command {
     }
   }
 
+  Future<Null> _unlockKeyChain() async {
```
digiter: Explain why in comments:
1. Xcode build requires the login keychain be unlocked in order to access the certificate for code signing.
2. This seems mitigates the issue of "Your session has expired", see https://github.com/flutter/flutter/issues/17860.

godofredoc: Done!

PR #569, opened 2019-12-11
Cache CocoonConfig

CaseyHillers:
Most of Cocoon's requests wait on Datastore operations to get config values that usually never change. This updates `CocoonConfig` to take advantage of caching. To make this simple in `CocoonConfig` I refactored `CacheService` to handle more of the cache logic.

After some discussion, this also moves some of the config values to be hard coded. This helps us better track changes made to the config. Any config values that do not need to be private have been pushed from Datastore into the code.

All Cocoon requests will see a speedup based on how many config variables they used (each variable use was a separate datastore query).

## Open Questions

1. ~The current `CacheService` used the Datastore config to get values. To not have a circular dependency on the two resources, I have made the config values hard coded instead of being stored in datastore. Is this okay, or is there a better way to store these config values? (specifically `memorystoreUrl`, `responseSubcacheName`, `configSubcacheName`).~
  - This is actually a better way of storing values that are okay to be public. We can track the changes to the config in version control.
2. Config values are stored in the cache for 12 hours. Should we store them indefinitely and just manually clear the Memorystore cache from GCP when a config value needs to be changed?

## Tested
- Added test for `createFn`
- Added tests for the race condition issue from https://github.com/flutter/flutter/issues/43762. A read to the cache when a write is in process throws an exception.

## Issues
- Fixes https://github.com/flutter/flutter/issues/43762

### DIFF
--- a/app_dart/lib/src/datastore/cocoon_config.dart
+++ b/app_dart/lib/src/datastore/cocoon_config.dart
@@ -107,10 +107,6 @@ class Config {
 
   Future<String> get cqLabelName => _getSingleValue('CqLabelName');
 
-  /// The name of the subcache in the Redis instance that stores responses.
-  Future<String> get redisResponseSubcache =>
-      _getSingleValue('RedisResponseSubcache');
-
   Future<String> get waitingForTreeToGoGreenLabelName =>
       _getSingleValue('WaitingForTreeToGreenLabelName');
 
--- a/app_dart/lib/src/datastore/cocoon_config.dart
+++ b/app_dart/lib/src/datastore/cocoon_config.dart
@@ -57,16 +57,6 @@ class Config {
     return Uint8List.fromList(result.value.codeUnits);
   }
 
-  Future<List<T>> _getJsonList<T>(String id) async {
-    final String rawValue = await _getSingleValue(id);
-    try {
-      return json.decode(rawValue).cast<T>();
-    } on FormatException {
-      loggingService.error('Invalid JSON format for property "$id": $rawValue');
-      throw InvalidConfigurationException(id);
-    }
-  }
-
   /// Per the docs in [DatastoreDB.withTransaction], only 5 entity groups can
   /// be touched in any given transaction, or the backing datastore will throw
   /// an error.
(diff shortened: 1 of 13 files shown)


CaseyHillers:
Offline discussion with the infra team: We would like to move a lot of these static config options to just be in the code. It will help us track changes (something that Datastore does not support easily). This further improves the loading of config values as they do not need to be `await`ed

CaseyHillers:
@digiter I updated this with the changes we discussed earlier. Can you take a look?

app_dart/lib/src/datastore/cocoon_config.dart
```diff
@@ -56,92 +67,151 @@ class Config {
 
   Future<String> get githubOAuthToken => _getSingleValue('GitHubPRToken');
 
-  Future<String> get nonMasterPullRequestMessage =>
-      _getSingleValue('NonMasterPullRequestMessage');
+  String get nonMasterPullRequestMessage =>
+      '''This pull request was opened against a branch other than _master_. Since Flutter pull requests should not normally be opened against branches other than master, I have changed the base to master. If this was intended, you may modify the base back to {{branch}}. See the [Release Process](https://github.com/flutter/flutter/wiki/Release-process) for information about how other branches get updated.
```
digiter: Nit: this line is too long, consider making it be multiple lines. Also applies to below.

app_dart/lib/src/request_handling/cache_request_handler.dart
```diff
@@ -30,32 +30,29 @@ class CacheRequestHandler<T extends Body> extends RequestHandler<T> {
...
+
+    return Body.forStream(Stream<Uint8List>.value(cachedResponse));
+  }
+
+  Future<Uint8List> getBodyBytesFromDelegate(RequestHandler<T> delegate) async {
+    final Body body = await delegate.get();
+    final List<int> rawBytes =
+        await body.serialize().expand<int>((Uint8List chunk) => chunk).toList();
+    return Uint8List.fromList(rawBytes);
   }
```
digiter: I don't understand why converting body to bytes isn't trivial. Is it due to the body has a different encoding?

app_dart/lib/src/service/cache_service.dart
```diff
@@ -5,28 +5,85 @@
...
-
-  CacheProvider<List<int>> _provider;
-
-  Future<Cache<Uint8List>> redisCache() async {
-    _provider = Cache.redisCacheProvider(await config.redisUrl);
-    return Cache<List<int>>(_provider)
-        .withCodec<Uint8List>(const _CacheCodec());
+  CacheService({
+    bool inMemory = false,
+    int inMemoryMaxSize = 256,
```
digiter: What's the unit? 256 Kb or Mb? It's generally a good idea to add unit to the name, e.g. `inMemoryMaxSizeMb`.

app_dart/lib/src/service/cache_service.dart
```diff
@@ -5,28 +5,85 @@
...
+
+  /// Get value of [key] from the subcache [subcacheName]. If the key has no
+  /// value, call [createFn] to create a value for it, set it, and return it.
+  ///
+  /// The underlying cache get function is inherently racy as if there is a
+  /// write operation while a read operation, getting the value can fail. To
+  /// handle this racy condition, this attempts to get the value [maxCacheGetAttempts]
+  /// times before giving up. This is because the cache is magnitudes faster
+  /// than the fallback operation (usually a Datastore query).
+  Future<Uint8List> get(
```
digiter: Nit: since this function is doing two things, I'd call it `getOrCreate()`.

app_dart/lib/src/service/cache_service.dart
CaseyHillers: I renamed it to `inMemoryMaxNumberEntries` and added class documentation further explaining it.

app_dart/lib/src/datastore/cocoon_config.dart
CaseyHillers: Due to the way multilines are handled in Dart, I don't think we can fit these values within a width of 120 characters without making it show up when the GitHub bot posts it. Adding newlines in a multiline string shows in the final string.

I have separated the lines by sentences instead which makes it easier to read, is this okay?

app_dart/lib/src/request_handling/cache_request_handler.dart
CaseyHillers: This isn't trivial because we can only create a Uint8List from a list of ints.

`Uint8List value = Uint8List.fromList(await body.serialize().toList());`
Type error: `The argument type 'List<Uint8List>' can't be assigned to the parameter type 'List<int>'.dart(argument_type_not_assignable)`

We discussed this more in https://github.com/flutter/cocoon/pull/484, specifically https://github.com/flutter/cocoon/pull/484#discussion_r338807711 after some offline discussion.

app_dart/lib/src/service/cache_service.dart
CaseyHillers: Good catch! Added.

app_dart/lib/src/datastore/cocoon_config.dart
digiter: 1. Logic is more important than format, so I would keep the lines in one place if they are logically one thing.
2.  There's another way to solve it:
```
var s2 = 'You can create ' + 
    'multi-line strings like this one.';
```

app_dart/lib/src/request_handling/cache_request_handler.dart
digiter: Non-trivial stuff deserves explanation in a comment.

app_dart/lib/src/datastore/cocoon_config.dart
jonahwilliams: You can do it a bit more simply as:
```
var s2 = 'foo bloo bar'
  'fizz flazz bazz';
```

CaseyHillers: Thank you both for this! I was able to clean it up and retain the original formatting.

app_dart/lib/src/request_handling/cache_request_handler.dart
CaseyHillers: Updated with the explaination, and added documentation for the function.

PR #575, opened 2019-12-12
Task Attempt Summary

CaseyHillers:
A list of the attempts on a Task with a link to each individual runs Stackdriver log.

## Tested
- Check that `TaskAttemptSummary` is only shown on devicelab tasks
- Check cases for attempts=0,1,2
- Check that url launcher launches expected URL

## Future Work

- Update `Task`entity to make start and endtimestamps repeatable so they can contain the start and end for each individual attempt. In general, make everything in `Task` repeatable so we can know the results of individuals runs. Once this data is available, show it in the UI.

## Preview

task_attempt_summary

task_attempt_summary

Demo: https://testchillers-dot-flutter-dashboard.appspot.com/#/build


CaseyHillers:
This currently has an issue where url encoding seems to break the URL at the log name. This will default it to show the logs for everything. Moving back to WIP.

CaseyHillers:
Fixed! Demo should be working now. There's some tweaking we can do to the parameters down the line as we let the team experiment and see what works best.

PR #580, opened 2019-12-16
Update agent health status history in BigQuery

keyonghan:
This PR creates an API: /api/update-agent-health-history to update Agent health status history in BigQuery.

The old strategy misses records for cases when Agents hang, or tasks are running. This will make it difficult to tell accurate Agent health status for a certain time period.

This new API will be called via a cronjob every 1 minute, and guarantees continuous records of Agents statuses.

### DIFF
--- a/app_dart/lib/src/request_handlers/update_agent_health_history.dart
+++ b/app_dart/lib/src/request_handlers/update_agent_health_history.dart
@@ -1,0 +1,92 @@
+// Copyright 2019 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:collection/collection.dart';
+import 'package:gcloud/db.dart';
+import 'package:googleapis/bigquery/v2.dart';
+import 'package:meta/meta.dart';
+
+import '../datastore/cocoon_config.dart';
+import '../model/appengine/agent.dart';
+import '../request_handling/api_request_handler.dart';
+import '../request_handling/authentication.dart';
+import '../request_handling/body.dart';
+import '../service/datastore.dart';
+
+const Duration maxHealthCheckAge = Duration(minutes: 10);
+
+@immutable
+class UpdateAgentHealthHistory extends ApiRequestHandler<Body> {
+  const UpdateAgentHealthHistory(
+    Config config,
+    AuthenticationProvider authenticationProvider, {
+    @visibleForTesting
+        this.datastoreProvider = DatastoreService.defaultProvider,
+  }) : super(config: config, authenticationProvider: authenticationProvider);
+
+  final DatastoreServiceProvider datastoreProvider;
+
+  @override
+  Future<Body> get() async {
+    const String projectId = 'flutter-dashboard';
+    const String dataset = 'cocoon';
+    const String table = 'Agent';
+
+    final TabledataResourceApi tabledataResourceApi =
+        await config.createTabledataResourceApi();
+    final DatastoreService datastore = datastoreProvider();
+    final Query<Agent> agentQuery = datastore.db.query<Agent>()
+      ..order('agentId');
+    final List<Agent> agents =
+        await agentQuery.run().where(_isVisible).toList();
+    agents.sort((Agent a, Agent b) =>
+        compareAsciiLowerCaseNatural(a.agentId, b.agentId));
+
+    final List<Map<String, Object>> tableDataInsertAllRequestRows =
+        <Map<String, Object>>[];
+
+    for (Agent agent in agents) {
+      final bool isHealthy = _isAgentHealthy(agent);
+
+      /// Consolidate [agents] together.
+      ///
+      /// Prepare for bigquery [insertAll].
+      tableDataInsertAllRequestRows.add(<String, Object>{
+        'json': <String, Object>{
+          'Timestamp': agent.healthCheckTimestamp,
+          'AgentID': agent.agentId,
+          'Status': isHealthy ? 'healthy' : 'unhealthy',
+          'Detail': isHealthy ? agent.healthDetails : 'out of date',
+        },
+      });
+    }
+
+    /// Prepare final [rows] to be inserted to `BigQuery`.
+    final TableDataInsertAllRequest rows = TableDataInsertAllRequest.fromJson(
+        <String, Object>{'rows': tableDataInsertAllRequestRows});
+
+    /// Insert [agents] to `BigQuery`.
+    try {
+      await tabledataResourceApi.insertAll(rows, projectId, dataset, table);
+    } catch (ApiRequestError) {
+      log.warning('Failed to add commits to BigQuery: $ApiRequestError');
+    }
+
+    return Body.forJson(<String, dynamic>{
+      'AgentStatuses': agents,
+    });
+  }
+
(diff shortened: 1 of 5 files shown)


app_dart/lib/src/request_handlers/update_agent_health_history.dart
```diff
@@ -0,0 +1,92 @@
...
+    AuthenticationProvider authenticationProvider, {
+    @visibleForTesting
+        this.datastoreProvider = DatastoreService.defaultProvider,
+  }) : super(config: config, authenticationProvider: authenticationProvider);
+
+  final DatastoreServiceProvider datastoreProvider;
+
+  @override
+  Future<Body> get() async {
+    const String projectId = 'flutter-dashboard';
```
CaseyHillers: Should these be class members instead so they include documentation on what they are for?

app_dart/lib/src/request_handlers/update_agent_health_history.dart
```diff
@@ -0,0 +1,92 @@
...
+        <Map<String, Object>>[];
+
+    for (Agent agent in agents) {
+      final bool isHealthy = _isAgentHealthy(agent);
+
+      /// Consolidate [agents] together.
+      ///
+      /// Prepare for bigquery [insertAll].
+      tableDataInsertAllRequestRows.add(<String, Object>{
+        'json': <String, Object>{
```
CaseyHillers: Is it possible to update `Agent.toJson()` to this instead?

app_dart/lib/src/request_handlers/update_agent_health_history.dart
```diff
@@ -0,0 +1,92 @@
...
+
+    /// Prepare final [rows] to be inserted to `BigQuery`.
+    final TableDataInsertAllRequest rows = TableDataInsertAllRequest.fromJson(
+        <String, Object>{'rows': tableDataInsertAllRequestRows});
+
+    /// Insert [agents] to `BigQuery`.
+    try {
+      await tabledataResourceApi.insertAll(rows, projectId, dataset, table);
+    } catch (ApiRequestError) {
+      log.warning('Failed to add commits to BigQuery: $ApiRequestError');
```
CaseyHillers: Should this be `log.error(...)` since the request failed?

app_dart/test/request_handlers/udpate_agent_health_history_test.dart
```diff
@@ -0,0 +1,90 @@
+// Copyright 2019 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:convert';
+
+import 'package:cocoon_service/cocoon_service.dart';
```
CaseyHillers: https://dart.dev/guides/language/effective-dart/style#prefer-placing-third-party-package-imports-before-other-imports

These `cocoon_service` imports should be placed after `package:test/test.dart`

app_dart/test/request_handlers/udpate_agent_health_history_test.dart
```diff
@@ -0,0 +1,90 @@
...
+import 'package:cocoon_service/src/service/datastore.dart';
+import 'package:googleapis/bigquery/v2.dart';
+import 'package:test/test.dart';
+
+import '../src/bigquery/fake_tabledata_resource.dart';
+import '../src/datastore/fake_cocoon_config.dart';
+import '../src/datastore/fake_datastore.dart';
+import '../src/request_handling/fake_authentication.dart';
+
+void main() {
```
CaseyHillers: There should be another test for when an error occurs uploading to bigtable (check to see if the error was logged). Also, a test for agents that are not visible.

app_dart/lib/src/request_handlers/update_agent_health_history.dart
keyonghan: I used to put those as class members, but got comments to include them in specific sub functions to make them more containable. I agreed to change them accordingly. After all, these variables only make sense for this sub function, considering if we have more other sub functions in the future.

app_dart/lib/src/request_handlers/update_agent_health_history.dart
keyonghan: Agent model have more other properties. Here these four properties are in accordance with those defined in BigQuery table.

app_dart/lib/src/request_handlers/update_agent_health_history.dart
keyonghan: Changed.

app_dart/test/request_handlers/udpate_agent_health_history_test.dart
keyonghan: Good catch! Also the googleapi one. 
BTW: most of our existing cocoon tests files seem to have such issue.. May need to update as a bunch later.

app_dart/lib/src/request_handlers/update_agent_health_history.dart
CaseyHillers: It seems like the only thing not in accordance is the `capabilities` field. If BigQuery does not support arrays, couldn't we still just do `agent.toJson()..remove('capabilities')` to make this more concise?

app_dart/lib/src/request_handlers/update_agent_health_history.dart
CaseyHillers: I still think they need some documentation. Project id refers to the Google Cloud project that is currently running. `dataset` and `table` should be documented as variables for BigQuery.

keyonghan: Added one comment: 
// Define const variables for [BigQuery] operations.

app_dart/lib/src/request_handlers/update_agent_health_history.dart
keyonghan: Yeah, that would make it more concise for sure.
However, another change is from bool `IsHealthy` to String `Status`.
Is it easy to do type cast here?

app_dart/test/request_handlers/udpate_agent_health_history_test.dart
keyonghan: Added test for log.info instead.
Existing test has considered the hidden case. Please refer to agent `windows1`.

CaseyHillers: That is correct. I misread `reportedAgents` for `expectedAgents`.

app_dart/lib/src/request_handlers/update_agent_health_history.dart
CaseyHillers: In that case, this is probably fine. Is there a reason for switching it to type String?

keyonghan: We may add more detailed categories, like maintenance, dead, quarantined, etc.

CaseyHillers: Oh okay, then this is probably best. Might be good to add a TODO to remind us why it was done this way.

keyonghan: Will add that when reverting earlier changes. Here is the TODO in my earlier PR:
https://github.com/flutter/cocoon/blob/master/app_dart/lib/src/request_handlers/update_agent_health.dart#L75

PR #581, opened 2019-12-17
[agent] Remove Xcode derived data directory each time.

digiter:
Fix flutter/flutter#45760.
Fix flutter/flutter#45729.

Tested manually and with unit tests.

### DIFF
--- a/agent/lib/src/health.dart
+++ b/agent/lib/src/health.dart
@@ -5,10 +5,15 @@
 import 'dart:async';
 import 'dart:io';
 
+import 'package:meta/meta.dart';
+import 'package:path/path.dart' as path;
+import 'package:file/local.dart' as local;
+import 'package:file/file.dart' as file;
+import 'package:platform/platform.dart' as platform;
+
 import 'adb.dart';
 import 'agent.dart';
 import 'utils.dart';
-import 'package:path/path.dart' as path;
 
 final RegExp _kLinuxIpAddrExp = RegExp(r'inet +(\d+\.\d+\.\d+.\d+)/\d+');
 final RegExp _kWindowsIpAddrExp =
@@ -68,6 +73,9 @@ Future<AgentHealth> performHealthChecks(Agent agent) async {
         }
       });
     }
+
+    results['remove-xcode-derived-data'] =
+        await _captureErrors(removeXcodeDerivedData);
   });
 
   return results;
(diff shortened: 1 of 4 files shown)


agent/lib/src/health.dart
```diff
@@ -5,15 +5,26 @@
...
 import 'utils.dart';
-import 'package:path/path.dart' as path;
 
 final RegExp _kLinuxIpAddrExp = RegExp(r'inet +(\d+\.\d+\.\d+.\d+)/\d+');
 final RegExp _kWindowsIpAddrExp =
     RegExp(r'IPv4 Address.*: +(\d+\.\d+\.\d+.\d+)\(Preferred\)');
 
+@visibleForTesting
+String Function() homeEnvVar = () {
+  String home = Platform.environment['HOME'];
```
dnfield: I know for this use case we only need to work with macOS, but maybe to avoid confusion in the future make this work for Windows as well?

agent/test/src/health_test.dart
```diff
@@ -0,0 +1,34 @@
...
+void main() {
+  testRemoveXcodeDerivedData();
+}
+
+void testRemoveXcodeDerivedData() {
+  MemoryFileSystem memfs;
+
+  setUp(() {
+    memfs = MemoryFileSystem();
+    homeEnvVar = () => '/foo';
```
dnfield: I'm not sure this is doing what it looks like it wants to do.  I think the way we would normally achieve this is to override the `Platform.environment` with a `Platform` abstraction.

agent/lib/src/health.dart
```diff
@@ -5,15 +5,26 @@
...
 import 'adb.dart';
 import 'agent.dart';
 import 'utils.dart';
-import 'package:path/path.dart' as path;
 
 final RegExp _kLinuxIpAddrExp = RegExp(r'inet +(\d+\.\d+\.\d+.\d+)/\d+');
 final RegExp _kWindowsIpAddrExp =
     RegExp(r'IPv4 Address.*: +(\d+\.\d+\.\d+.\d+)\(Preferred\)');
 
+@visibleForTesting
```
dnfield: See comment on test - I don't think this needs to be visible for testing, and should probably just be made private (and perhaps just a property).

agent/test/src/health_test.dart
digiter: How to override the `Platform.environment` with a `Platform` abstraction? Is there an example?

agent/lib/src/health.dart
dnfield: Rather than directly using `Platform` here, you can inject it - we do that in flutter_tools via a `Context` object that reads values from the current zone.  It makes it easier if you use https://pub.dev/packages/platform

agent/lib/src/health.dart
digiter: Added a check to ignore non-macOS.

agent/test/src/health_test.dart
digiter: Done.

agent/lib/src/health.dart
digiter: Thanks for the suggestion on platform package.

agent/lib/src/health.dart
```diff
@@ -131,3 +139,24 @@ Future<HealthCheckResult> _scrapeRemoteAccessInfo() async {
...
+    file.FileSystem fs = const local.LocalFileSystem()}) async {
+  if (!pf.isMacOS) {
+    return HealthCheckResult.success();
+  }
+  String home = pf.environment['HOME'];
+  if (home == null) {
+    return HealthCheckResult.failure('Missing \$HOME environment variable.');
+  }
+  String p = path.join(home, 'Library/Developer/Xcode/DerivedData');
+  rrm(fs.directory(p));
```
jmagman: Actually will this crash if the directory doesn't exist?  We wouldn't want two heath checks in a row to cause a crash.

digiter: The `rrm` function somehow doesn't crash in this case, I have tested manually and in a unit test.

PR #583, opened 2019-12-17
Index page rewritten in Flutter

CaseyHillers:
Moved the index page to be in `app_flutter`. 
Updated the performance dashboard to have the google sign in button since the index page was where you would previously sign in.
`build.html` has been moved to `/#/build`. `/build.html` will redirect there. This is because Flutter web handles route navigation via modifying the path after `/#`. Removing the hash to just use the HTML5 History API is being tracked at https://github.com/flutter/flutter/issues/33245

## Tested

- Tests for checking all the widgets are there
- Navigator tests

## Preview

Demo: https://testchillers-dot-flutter-dashboard.appspot.com

https://testchillers-dot-flutter-dashboard.appspot.com/build.html should redirect to /#/build


CaseyHillers:
Still working on tests for index page.

CaseyHillers:
Tests updated!

PR #585, opened 2019-12-17
Update with only the latest commit status when refreshing chromebot status

keyonghan:
Existing logic when refreshing chromebot status is keep updating task status for all executions of the same commit, from the latest to the oldest. This will keep the status on dashboard the same as  the earliest run's status. This is incorrect.

This PR stops updating task status whenever we find the latest status of one commit. This way the status on dashboard is always up-to-date.


keyonghan:
This PR fixes https://github.com/flutter/flutter/issues/47250

PR #572, opened 2019-12-12
Run presubmit tests for cocoon and engine automatically on luci.

godofredoc:
Both cocoon and engine are ready to start running luci presubmit tests
automatically. This add support for cocoon repo and removes the need for
CQ+1 on engine commits.

https://github.com/flutter/flutter/issues/43348

PR #586, opened 2019-12-17
Revert "Run presubmit tests for cocoon and engine automatically on luci."

godofredoc:
Reverts flutter/cocoon#572 we are getting 500s for cocoon and engine tasks. Reverting while we check what the problem is.


keyonghan:
Thanks!

PR #589, opened 2019-12-18
CacheService pass through ttl regression fix + flush cache

CaseyHillers:
When rolling out the CocoonConfig caching changes from https://github.com/flutter/cocoon/pull/569, there was issues with the cache not being cleared. I discovered it was because the TTLs never got set.

Additionally, memorystore GUI has no way to reset keys that have an indefinite ttl (created from this bug). I added `flushCache=true` to cached requests will reset the cache entry for the accessed response.

## Tested

- Added regression tests to ensure TTL is passed through to the final set from the different entry points
- Added test to ensure flush cache param in cache request handler works

## Preview

Broken example: ~https://v336-dot-flutter-dashboard.appspot.com/old_build.html~ - notice that all the agents have timed out

Patched version: https://v338-dot-flutter-dashboard.appspot.com/old_build.html - not all of them should be red now

### DIFF
--- a/app_dart/lib/src/service/cache_service.dart
+++ b/app_dart/lib/src/service/cache_service.dart
@@ -75,10 +75,7 @@ class CacheService {
       }
     }
 
-    // If given createFn, we can update the cache value if we are forcing a
-    // cache flush or if the value returned was null.
-    //
-    // Cache flush is a failover to manually fix cache issues in production.
+    // If given createFn, update the cache value if the value returned was null.
     if (createFn != null && value == null) {
       // Try creating the value
       value = await createFn();
--- a/app_dart/lib/src/service/cache_service.dart
+++ b/app_dart/lib/src/service/cache_service.dart
@@ -92,7 +92,7 @@ class CacheService {
     Duration ttl = const Duration(minutes: 1),
   }) async {
     final Cache<Uint8List> subcache = cache.withPrefix(subcacheName);
-    return subcache[key].set(value);
+    return subcache[key].set(value, ttl);
   }
 
   void dispose() {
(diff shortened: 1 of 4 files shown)


app_dart/lib/src/service/cache_service.dart
```diff
@@ -55,6 +55,7 @@ class CacheService {
     int attempt = 1,
     Future<Uint8List> Function() createFn,
     Duration ttl = const Duration(minutes: 1),
+    bool flushCache = false,
   }) async {
```
digiter: This function becomes too smart that it has three concerns now, indicators are the lengthy testing setups. 

It's better to separate the concerns into `get()`, `create()` and `flush()`, and when a smart business logic function is needed, compose that with these dump functions.

CaseyHillers: Good point. I moved this extra flush logic to a separate function. I made available the `purge` function with flushes a value from the cache, and had `CacheRequestHandler` use that instead.

app_dart/lib/src/service/cache_service.dart
```diff
@@ -75,7 +75,11 @@ class CacheService {
       }
     }
 
-    if (value == null && createFn != null) {
+    // If given createFn, we can update the cache value if we are forcing a
+    // cache flush or if the value returned was null.
+    //
+    // Cache flush is a failover to manually fix cache issues in production.
```
digiter: No more cache flush in this function.

app_dart/lib/src/service/cache_service.dart
digiter: Nit: I would split get() and create() as well. Doesn't need to address in this PR though.

app_dart/lib/src/service/cache_service.dart
CaseyHillers: Good catch! Fixed.

PR #591, opened 2019-12-19
Run presubmit tests for cocoon and engine automatically on luci.

godofredoc:
Both cocoon and engine are ready to start running luci presubmit tests
automatically. This add support for cocoon repo and removes the need for
CQ+1 on engine commits.

flutter/flutter#43348

### DIFF
--- a/app_dart/lib/src/request_handlers/github_webhook.dart
+++ b/app_dart/lib/src/request_handlers/github_webhook.dart
@@ -209,13 +209,8 @@ class GithubWebhook extends RequestHandler<Body> {
     ]));
     return batch.responses
         .map((Response response) => response.searchBuilds)
-        .expand((SearchBuildsResponse response) {
-      if (response.builds != null) {
-        return response.builds;
-      } else {
-        return <Build>[];
-      }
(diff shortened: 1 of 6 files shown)


godofredoc:
This is an updated version of https://github.com/flutter/cocoon/pull/572 which was reverted.

app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -186,8 +209,13 @@ class GithubWebhook extends RequestHandler<Body> {
...
     return batch.responses
         .map((Response response) => response.searchBuilds)
-        .expand((SearchBuildsResponse response) => response.builds)
-        .toList();
+        .expand((SearchBuildsResponse response) {
+      if (response.builds != null) {
+        return response.builds;
+      } else {
+        return <Build>[];
+      }
```
digiter: Could be simplified to `return response.builds ?? <Build>[];`

godofredoc: Done!

PR #592, opened 2019-12-19
append-log key only available in tests

CaseyHillers:
This moves the append-log code that impacts production so it only runs in tests.

## Issues

https://github.com/flutter/flutter/issues/47453

### DIFF
--- a/app_dart/lib/src/model/appengine/log_chunk.dart
+++ b/app_dart/lib/src/model/appengine/log_chunk.dart
@@ -10,10 +10,14 @@ import 'package:gcloud/db.dart';
 class LogChunk extends Model {
   /// Creates a new [LogChunk].
   LogChunk({
+    Key key,
     this.data,
     this.createTimestamp,
     this.ownerKey,
-  });
+  }) {
+    parentKey = key?.parent;
+    id = key?.id;
+  }
 
   /// The binary data of the log chunk.
   @BlobProperty(propertyName: 'Data', required: true)
--- a/app_dart/lib/src/model/appengine/log_chunk.dart
+++ b/app_dart/lib/src/model/appengine/log_chunk.dart
@@ -10,14 +10,10 @@ import 'package:gcloud/db.dart';
 class LogChunk extends Model {
   /// Creates a new [LogChunk].
   LogChunk({
-    Key key,
     this.data,
     this.createTimestamp,
     this.ownerKey,
-  }) {
-    parentKey = key?.parent;
-    id = key?.id;
-  }
+  });
 
   /// The binary data of the log chunk.
   @BlobProperty(propertyName: 'Data', required: true)
--- a/app_dart/lib/src/model/appengine/log_chunk.dart
+++ b/app_dart/lib/src/model/appengine/log_chunk.dart
@@ -10,13 +10,13 @@ import 'package:gcloud/db.dart';
 class LogChunk extends Model {
   /// Creates a new [LogChunk].
   LogChunk({
-    Key key,
+    Key parentKeyValue,
     this.data,
     this.createTimestamp,
     this.ownerKey,
   }) {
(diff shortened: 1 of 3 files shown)


app_dart/test/request_handlers/append_log_test.dart
```diff
@@ -42,19 +42,31 @@ Deleting build/ directories, if any.
...
+          commitKey: commit.key,
+          requiredCapabilities: <String>['ios']);
+
       mockStackdriverLoggerService = MockStackdriverLoggerService();
       handler = AppendLog(
         config,
         FakeAuthenticationProvider(),
         stackdriverLogger: mockStackdriverLoggerService,
         requestBodyValue: Uint8List.fromList(logData.codeUnits),
+        key: task.key,
```
keyonghan: This should be the logChunk Key, rather than task key.

app_dart/lib/src/request_handlers/append_log.dart
```diff
@@ -26,6 +26,7 @@ class AppendLog extends ApiRequestHandler<Body> {
     AuthenticationProvider authenticationProvider, {
     StackdriverLoggerService stackdriverLogger,
     @visibleForTesting Uint8List requestBodyValue,
+    @visibleForTesting this.key,
```
keyonghan: @dnfield may have a better idea how to deal with the key here.

app_dart/test/request_handlers/append_log_test.dart
dnfield: And in that case we should just move commit and task back out of setUp to the test they're for

app_dart/lib/src/request_handlers/append_log.dart
dnfield: I'm confused about where this comes from at runtime - it seems like it's not getting initialized to anything outside of tests.

CaseyHillers: Yeah that's what I was trying to do, only have this accessible in tests. I added the tests for this handler in https://github.com/flutter/cocoon/pull/564 and modified the original handler to make it more testable. However, that apparently caused a few Datastore issues :)

app_dart/test/request_handlers/append_log_test.dart
CaseyHillers: Done, this helps make the tests more damp :+1:

app_dart/lib/src/request_handlers/append_log.dart
dnfield: Making it only settable by tests is fine, but it looks like it never gets set at all in the implementation, which presumably is a problem.

CaseyHillers: Debugging this showed that `LogChunk` should have its parent set to a Checklist key. Instead it was being set to a Task key.

I set `LogChunk` to have a parameter for setting it's parent key so it can work in the tests. I'm debugging how it was getting it's parent key set in the first place.

keyonghan: In this case, this line can be removed.

app_dart/lib/src/model/appengine/log_chunk.dart
```diff
@@ -10,13 +10,13 @@ import 'package:gcloud/db.dart';
...
   /// Creates a new [LogChunk].
   LogChunk({
-    Key key,
+    Key parentKeyValue,
     this.data,
     this.createTimestamp,
     this.ownerKey,
   }) {
-    parentKey = key?.parent;
-    id = key?.id;
```
keyonghan: we may want to keep this line and line13?

CaseyHillers:
Closing in favor of https://github.com/flutter/cocoon/pull/595. Will come back to this after the new year.

PR #594, opened 2019-12-20
Infra Agent Dashboard

CaseyHillers:
Moves the Angular Dart Build Dashboard Agent Statuses and CLI commands into an independent dashboard
- Removes the Angular Dart build dashboard
- Lists agents and their statuses with ability to look at their current health checks (shows unhealthy agents first)
  - Uses Jonah's mobile project's agent health details
- UI to create agents (prints the token to the console so it can be copied)
 - Authorizing an agent (prints the token to the console so it can be copied)
  - Reserving a task for an agent (prints response to the console so it can be copied too)
- Agent button in Task Overlay on build dashboard redirects to the agent dashboard showing that agent

Adds `Agent` proto, updates proto generated files with latest protoc version
  - Latest protobuf updated appengine so updated fake logging in `app_dart` with a new field

Added a link to this GitHub repo in the sidebar to give more visibility to where the source code lives

## Open Questions

- Are there any more specific icons to show windows and mac agents? Currently it just shows a phone for iOS and a monitor for Windows
- Is there anything actionable from `/api/reserve-task` for humans? The response it returns seems tailored for Agents. Currently I just print the response to the console in case there is use for it.

## Issues

Fixes https://github.com/flutter/flutter/issues/43110

## Future Work

- UI to visualize the current running and queued tasks
  - Instead of just using the `/api/public/get-status` call I would like to update the `Task` entity to store the `Task` key it is currently running.

## Preview

https://testchillers-dot-flutter-dashboard.appspot.com/#/agents

FAB creates agents.
Popup menu on each agent offers raw health details, authorizing agent, and reserving a task.
Build dashboard has task overlay that will redirect to the agent dashboard showing that agent


app_flutter/lib/agent_dashboard_page.dart
```diff
@@ -0,0 +1,195 @@
...
+        appBar: AppBar(
+          title: const Text('Infra Agents'),
+          actions: <Widget>[
+            SignInButton(authService: agentState.authService),
+          ],
+        ),
+        body: Column(
+          mainAxisAlignment: MainAxisAlignment.center,
+          children: <Widget>[
+            // Without this container the ListView will not be centered.
```
jonahwilliams: Do you know why? If so could you comment here, or maybe do some digging - it seems odd at first glance

app_flutter/lib/agent_health_details.dart
```diff
@@ -0,0 +1,160 @@
...
+        canPerformHealthCheck &&
+        hasSshConnectivity &&
+        pingedRecently;
+  }
+
+  @visibleForTesting
+  static const int minutesUntilAgentIsUnresponsive = 10;
+
+  static final RegExp _hasSshConnectivity =
+      RegExp('ssh-connectivity: succeeded');
```
jonahwilliams: Do these need to be regexps, or could they just be Strings?

app_flutter/lib/agent_list.dart
```diff
@@ -0,0 +1,67 @@
...
+  @visibleForTesting
+  final bool insertKeys;
+
+  @override
+  Widget build(BuildContext context) {
+    if (agents.isEmpty) {
+      return const CircularProgressIndicator();
+    }
+
+    final List<FullAgent> fullAgents = agents
```
jonahwilliams: Could you sort the agents where you download them, to avoid re-sorting unnecessarily?

app_flutter/lib/agent_tile.dart
```diff
@@ -0,0 +1,138 @@
...
+          Text(rawHealthDetails),
+        ],
+      ),
+    );
+  }
+
+  Future<void> _authorizeAgent(BuildContext context, Agent agent) async {
+    final String token = await agentState.authorizeAgent(agent);
+    if (token != null) {
+      print(token);
```
jonahwilliams: Is this print important?

app_flutter/lib/agent_dashboard_page.dart
CaseyHillers: No, originally the container had a height to make it look pretty, but when I went to remove it I noticed the issue. I think it has to do with the expanded, but I'll try to do some more digging around.

app_flutter/lib/agent_health_details.dart
CaseyHillers: Good catch! Updated.

app_flutter/lib/agent_list.dart
CaseyHillers: I plan to update the backend in a followup PR for this (https://github.com/flutter/flutter/issues/48249).

app_flutter/lib/agent_tile.dart
CaseyHillers: It allows for easily copying the token. Since web does not have clipboard support yet (https://github.com/flutter/flutter/issues/46020), and copying text does not work yet, this is a workaround. Eventually, this should create UI to show the token so it can be copied without opening devtools.

Added a TODO.

app_flutter/lib/agent_list.dart
CaseyHillers: I plan to update the backend in a followup PR for this (https://github.com/flutter/flutter/issues/48249).

CaseyHillers:
@godofredoc @digiter @keyonghan this PR does not include the refresh github commits command. It is still accessible via querying `/api/refresh-github-commits`. Is this okay?

keyonghan:
Shall we push /api/refresh-github-commits to the build page? Also, it may be good to have a button on the agent page to create an agent.

CaseyHillers:
@keyonghan the floating action button in the bottom right should open a dialog for creating a new agent.

Also, I noticed that refresh github commits is on a 1 minute cronjob. Does it need to be manually triggered?

keyonghan:
Ha, sorry didn't notice it..  As for the github refresh, just from the aspect to make it consistent with old dashboard. Personally, I don't think people use it much.

CaseyHillers:
I updated the README documentation on the old browser CLI to how it works in the dashboard.

@jonahwilliams thanks for pointing out the coverage tool. I add some more tests to areas that were not covered. Is there an easy way to generate coverage reports for Cocoon like how flutter/flutter does?

CaseyHillers:
The demo is updated with the code that does the redirect from build dashboard and pulls up the agent in the agent dashboard. I need to write the tests and then rebase with master.

CaseyHillers:
@jonahwilliams I finished the final changes where users can go between the build dashboard to look at the agent that ran their task.

https://testchillers-dot-flutter-dashboard.appspot.com/#/agents

app_flutter/lib/agent_dashboard_page.dart
```diff
@@ -0,0 +1,221 @@
...
+
+  static const String routeName = '/agents';
+
+  final AgentState agentState;
+
+  final String agentFilter;
+
+  @visibleForTesting
+  static const Duration errorSnackbarDuration = Duration(seconds: 8);
+  @override
```
jonahwilliams:
```suggestion
  
  @override
```

app_flutter/lib/agent_dashboard_page.dart
```diff
@@ -0,0 +1,221 @@
...
+    AgentState agentState,
+    GoogleSignInService signInService,
+    @visibleForTesting this.agentFilter,
+  }) : agentState = agentState ?? AgentState(authServiceValue: signInService);
+
+  static const String routeName = '/agents';
+
+  final AgentState agentState;
+
+  final String agentFilter;
```
jonahwilliams: Document agent filter - it is kind of weird to have visibleForTesting members (though the way you have annotated it is only for the constructor argument)

app_flutter/lib/agent_dashboard_page.dart
```diff
@@ -0,0 +1,221 @@
...
+
+class _AgentDashboardPageState extends State<AgentDashboardPage> {
+  final GlobalKey<ScaffoldState> _scaffoldKey = GlobalKey<ScaffoldState>();
+
+  AgentState agentState;
+
+  @override
+  void initState() {
+    super.initState();
+
```
jonahwilliams:
```suggestion
```

app_flutter/lib/agent_dashboard_page.dart
```diff
@@ -0,0 +1,221 @@
...
+    widget.agentState.startFetchingStateUpdates();
+
```
jonahwilliams:
```suggestion
```

app_flutter/lib/agent_dashboard_page.dart
```diff
@@ -0,0 +1,221 @@
...
+    widget.agentState.errors.addListener(_showErrorSnackbar);
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    agentState = widget.agentState;
```
jonahwilliams: Why are you pulling it off of the widget here?

app_flutter/lib/agent_dashboard_page.dart
```diff
@@ -0,0 +1,221 @@
...
+          ),
+          color: Theme.of(context).primaryColor,
+          padding: const EdgeInsets.all(10.0),
+          onPressed: () => _showCreateAgentDialog(context, agentState),
+        ),
+      ),
+    );
+  }
+
+  void _showCreateAgentDialog(BuildContext context, AgentState agentState) {
```
jonahwilliams: This is complex enough to be its own widget

app_flutter/lib/agent_list.dart
```diff
@@ -0,0 +1,126 @@
...
+
+import 'package:flutter/material.dart';
+
+import 'package:cocoon_service/protos.dart' show Agent;
+
+import 'agent_health_details.dart';
+import 'agent_tile.dart';
+import 'state/agent.dart';
+
+class AgentList extends StatefulWidget {
```
jonahwilliams: Please document

app_flutter/lib/agent_list.dart
```diff
@@ -0,0 +1,126 @@
...
+    }
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    if (widget.agents.isEmpty) {
+      return const CircularProgressIndicator();
+    }
+
+    final List<FullAgent> fullAgents = widget.agents
```
jonahwilliams: Could you have the service filter agents instead of doing it in build?

app_flutter/lib/agent_list.dart
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    return agents
+        .where(
+            (FullAgent fullAgent) => fullAgent.agent.agentId.contains(filter))
+        .toList();
+  }
+}
+
+class FullAgent implements Comparable<FullAgent> {
```
jonahwilliams: Document

app_flutter/lib/agent_dashboard_page.dart
CaseyHillers: Thanks! Done.

app_flutter/lib/agent_dashboard_page.dart
CaseyHillers: Good catch, moved to `initState()`.

app_flutter/lib/agent_dashboard_page.dart
CaseyHillers: Thanks for pointing this out! Refactoring this out was able to simplify the rest of the widget into a stateless widget, only making the dialog part a stateful widget.

app_flutter/lib/agent_list.dart
CaseyHillers: Added

app_flutter/lib/agent_list.dart
CaseyHillers: Do you think there's an advantage to this? I left it in the widget since I made this into more of a search functionality as CanvasKit doesn't work with Ctrl+F. I thought it would be easier to follow if it was in the widget that uses it.

app_flutter/lib/agent_list.dart
CaseyHillers: Added

CaseyHillers:
@jonahwilliams I believe I have gone through and added the missing documentation (there's a lot, so I may have missed something).

The remaining question is for moving the filtering agents outside of the `AgentList` widget into `AgentState`: 

> Do you think there's an advantage to this? I left it in the widget since I made this into more of a search functionality as CanvasKit doesn't work with Ctrl+F. I thought it would be easier to follow if it was in the widget that uses it.

PR #599, opened 2019-12-21
Allow the use of a per repo token.

godofredoc:
Flutter and engine use the same token but we were not able to reuse the
token for cocoon project.


godofredoc:
This was not needed simply adding FlutterGitHubBot as a collaborator to the repo fixed the problem.

PR #600, opened 2019-12-27
Removing old code related to goldens.version file

Piinks:
There is no goldens.version file anymore. Just cleaning up. :)


dnfield:
Needs dartfmt

PR #609, opened 2020-01-07
Create Google Sign In Service only once

CaseyHillers:
This creates `GoogleSignInService` in `main.dart` and passes it around instead of recreating it on each page.

Updated sign in button to be a progress button to show if the sign in is in progress (good for issues when the plugin is being buggy). However, in CanvasKit this shows a weird color issue with the transparent color.

# Issues

Mitigates https://github.com/flutter/flutter/issues/47832

- Seems to be an issue with the web implementation of sign in silently where it causes it to hang.

## Tested

Checked for the states to have `notifyListener()` called when the sign in functions are called.

# Preview

https://testchillers1-dot-flutter-dashboard.appspot.com

### DIFF
--- a/app_flutter/test/state/flutter_build_test.dart
+++ b/app_flutter/test/state/flutter_build_test.dart
@@ -15,8 +15,6 @@ import 'package:app_flutter/service/cocoon.dart';
 import 'package:app_flutter/service/fake_cocoon.dart';
 import 'package:app_flutter/state/flutter_build.dart';
 
-import '../utils/fake_google_account.dart';
-
 void main() {
   group('FlutterBuildState', () {
     FlutterBuildState buildState;
--- a/app_flutter/test/state/flutter_build_test.dart
+++ b/app_flutter/test/state/flutter_build_test.dart
@@ -4,7 +4,9 @@
 
 import 'dart:async';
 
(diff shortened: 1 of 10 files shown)


CaseyHillers:
I was able to reproduce another issue with this when throttling the network connection.

app_flutter/test/state/flutter_build_test.dart
```diff
@@ -114,7 +116,35 @@ void main() {
       buildState.dispose();
     });
   });
+
+  testWidgets('sign in functions call notify listener',
+      (WidgetTester tester) async {
+    final MockGoogleSignInPlugin mockSignInPlugin = MockGoogleSignInPlugin();
+    when(mockSignInPlugin.onCurrentUserChanged)
+        .thenAnswer((_) => Stream<GoogleSignInAccount>.value(null));
+    when(mockSignInPlugin.signIn()).thenAnswer((_) async => null);
```
jonahwilliams: By default, non mocked methods will return null - which is both awaitable and assignable to future. Does this test pass without these?

CaseyHillers: Thanks! Removed. I started off by trying to test the stream logic of `onCurrentUserChanged` and realized there was a simpler way :)

PR #614, opened 2020-01-10
Check Flutter version to be from the past 3 weeks

CaseyHillers:
This is the only dependency on the Flutter side I check for when deploying. It shouldn't be an issue going forward, but it helps document a minimum Flutter version needed for deploying Cocoon.

The reason for Flutter 1.12+ is that the web plugins require it.

### DIFF
--- a/app_dart/dev/deploy.dart
+++ b/app_dart/dev/deploy.dart
@@ -46,8 +46,8 @@ bool _getArgs(ArgParser argParser, List<String> arguments) {
   return true;
 }
 
-/// Check the Flutter version installed and make sure it is a version with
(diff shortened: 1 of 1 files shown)


jonahwilliams:
I would recommend sticking with a later dev version over beta

CaseyHillers:
@jonahwilliams do you think a better check would be to see if it's a Flutter version from the past 2 weeks? We can check this as `flutter --version` outputs the line `Framework • revision 4f9b6cf017 (3 days ago) • 2020-01-07 16:32:04 +0100` which gives the published time of the currently installed version.

jonahwilliams:
SGTM

PR #615, opened 2020-01-10
Update to federated plugins and re-enable silent sign in

CaseyHillers:
Thank you @ditman for fixing the Google sign in web bug with silent sign in the dashboard was facing. This just updates the dependencies and re-enables the sign in code with their patch.

Reverts the sign in button back to a basic button so it doesn't have the CanvasKit issue.

## Issues

Fixes https://github.com/flutter/flutter/issues/47832

## Preview

https://testchillers1-dot-flutter-dashboard.appspot.com

### DIFF
--- a/app_flutter/lib/service/google_authentication.dart
+++ b/app_flutter/lib/service/google_authentication.dart
@@ -19,8 +19,7 @@ class GoogleSignInService {
       notifyListeners();
     });
 
-    // TODO(chillers): Uncomment following line when issue fixed. https://github.com/flutter/flutter/issues/47832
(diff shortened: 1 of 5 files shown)


ditman:
Apologies for not finding the bug in google_sign_in earlier!

CaseyHillers:
@ditman It's good that Cocoon can be a dogfood for these plugins :)

ditman:
> @ditman It's good that Cocoon can be a dogfood for these plugins :)

@CaseyHillers indeed it is!

**BTW, may I interest you in all the goodnesses of Firebase? :P**

PR #616, opened 2020-01-10
white list autorollers

dnfield:
fixes https://github.com/flutter/flutter/issues/46169

/cc @rmistry

### DIFF
--- a/app_dart/lib/src/datastore/cocoon_config.dart
+++ b/app_dart/lib/src/datastore/cocoon_config.dart
@@ -126,6 +126,11 @@ class Config {
     return ServiceAccountCredentials.fromJson(json.decode(rawValue));
   }
 
+  Set<String> get rollerAccounts => <String>{
+    'skia-flutter-autoroll',
+    'engine-flutter-autoroll',
+  };
+
   /// A List of builders for LUCI
   List<Map<String, dynamic>> get luciBuilders => <Map<String, String>>[
         <String, String>{
--- a/app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
+++ b/app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
@@ -179,9 +179,10 @@ class CheckForWaitingPullRequests extends ApiRequestHandler<Body> {
           .isAfter(DateTime.now().toUtc())) {
         continue;
       }
+      final String author = pullRequest['author']['login'];
       final String id = pullRequest['id'];
       final int number = pullRequest['number'];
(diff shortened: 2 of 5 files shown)


app_dart/lib/src/datastore/cocoon_config.dart
```diff
@@ -126,6 +126,14 @@ class Config {
...
   }
 
+  /// The names of autoroller accounts for the repositories.
+  ///
+  /// These accounts should not need reviews before merging.
+  Set<String> get rollerAccounts => <String>{
+    'skia-flutter-autoroll',
+    'engine-flutter-autoroll',
+  };
+
```
dnfield: Not sure if this should go into datastore - these rarely change and it seems like people would rather have values in cocoon.

Trivial to put it in datastore though, LMK.

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
```diff
@@ -179,9 +179,10 @@ class CheckForWaitingPullRequests extends ApiRequestHandler<Body> {
           .isAfter(DateTime.now().toUtc())) {
         continue;
       }
+      final String author = pullRequest['author']['login'];
       final String id = pullRequest['id'];
       final int number = pullRequest['number'];
-      final bool hasApproval =
+      final bool hasApproval = config.rollerAccounts.contains(author) ||
```
CaseyHillers: Can you add a test that checks that an author named `skia-flutter-autoroll-hackerman` does not get merged?

app_dart/lib/src/datastore/cocoon_config.dart
```diff
@@ -126,6 +126,14 @@ class Config {
     return ServiceAccountCredentials.fromJson(json.decode(rawValue));
   }
 
+  /// The names of autoroller accounts for the repositories.
+  ///
+  /// These accounts should not need reviews before merging.
```
CaseyHillers: Can we add a see also link to https://github.com/flutter/flutter/wiki/Autorollers?

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
dnfield: This should be covered by other tests, which are testing "some_rando", but I can add this too.

app_dart/lib/src/datastore/cocoon_config.dart
dnfield: Good call, done.

PR #617, opened 2020-01-10
Mark tree build status failed when tasks are rerunning

CaseyHillers:
This updates the build status algorithm to return failed if a task is having to rerun. 

Additionally, it reduces the amount of commits query from 100 to only the latest 20. Additionally, it caches this build status for 15 seconds. These changes are to remove unnecessary Datastore reads.

## Issues

Fixes https://github.com/flutter/flutter/issues/45735

## Preview

https://testchillers-dot-flutter-dashboard.appspot.com/#/build

### DIFF
--- a/app_dart/lib/src/service/build_status_provider.dart
+++ b/app_dart/lib/src/service/build_status_provider.dart
@@ -56,10 +56,8 @@ class BuildStatusProvider {
           /// commit may be done that we can base off of.
           final bool checked = checkedTasks[task.name] ?? false;
 
-          if (isRelevantToLatestStatus && task.isFlaky) {
-            checkedTasks[task.name] = true;
-          } else if (isRelevantToLatestStatus && !checked) {
-            if (_isSuccessful(task)) {
+          if (isRelevantToLatestStatus && !checked) {
+            if (task.isFlaky || _isSuccessful(task)) {
               checkedTasks[task.name] = true;
             } else if (_isFailed(task) || _isRerunning(task)) {
               return BuildStatus.failed;
--- a/app_dart/lib/src/service/build_status_provider.dart
+++ b/app_dart/lib/src/service/build_status_provider.dart
@@ -23,7 +23,8 @@ class BuildStatusProvider {
   final DatastoreServiceProvider datastoreProvider;
 
   @visibleForTesting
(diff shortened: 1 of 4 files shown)


app_dart/lib/src/service/build_status_provider.dart
```diff
@@ -22,36 +22,50 @@ class BuildStatusProvider {
...
   /// Calculates and returns the "overall" status of the Flutter build.
   ///
   /// This calculation operates by looking for the most recent success or
   /// failure for every (non-flaky) task in the manifest.
   Future<BuildStatus> calculateCumulativeStatus() async {
     final Map<String, bool> checkedTasks = <String, bool>{};
-    bool isLatestBuild = true;
+    bool isLatestCommitStatus = true;
 
     await for (CommitStatus status in retrieveCommitStatus()) {
```
digiter: I have to wrap around my head on this section. Is there a way to simply it?

CaseyHillers: I think inherently this is hard algorithm to follow. The hardest and most time consuming part of this PR was just understanding the PR. I have two ideas for making it a bit more self explanatory.

1. Renaming:
  a. `!checked` => `taskIsInProgress`
  b. `checkedTask` => `tasksInProgress`
2. ASCII diagrams of a build dashboard. I think this is the most beneficial as it's what got me to understand the algorithm.

What do you think?

CaseyHillers:
@digiter I implemented my approaches to making this algorithm more readable. I think the next steps for making the algorithm more readable would be flattening all the nesting that `CommitStatus`, `Stage`, and `Task` loops creates. Do you think that's necessary for this PR?

app_dart/lib/src/service/build_status_provider.dart
digiter: Those are helpful! In addition, we could reduce the nesting ifs with "early returns", see https://testing.googleblog.com/2017/06/code-health-reduce-nesting-reduce.html

CaseyHillers: Thanks for the resource! I was able to extract the latest commit status logic to its own block which seems to have cleaned up most of the complexity.

PR #619, opened 2020-01-10
Remove unused Datastore composite indexes

CaseyHillers:
According to [Datastore best practices](https://cloud.google.com/datastore/docs/best-practices#indexes)

> Avoid having too many composite indexes. Excessive use of composite indexes could result in increased latency and increased storage costs of index entries. If you need to execute ad hoc queries on large datasets without previously defined indexes, use BigQuery.

This will help reduce the latency on Datastore operations for `Task`. It seems like a few of these were added in https://github.com/flutter/cocoon/pull/321 for future features, but the indexes were never used. We have a few cases where we stack simple indexes that could take advantage of these composite index.

### DIFF
--- a/app_dart/index.yaml
+++ b/app_dart/index.yaml
@@ -2,8 +2,13 @@
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE file.
 
-# This file must be manually deployed to updates the Datastore index with:
+# This file must be manually deployed to update Datastore
+# Ensure gcloud sdk is configured to the Cocoon GCP project:
+# $ gcloud config list
+# Then deploy to Datastore for the project:
 # $ gcloud datastore indexes create index.yaml
+# If indexes are removed:
+# $ gcloud datastore indexes cleanup index.yaml
 indexes:
 - kind: Checklist
   ancestor: yes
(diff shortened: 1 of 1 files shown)


CaseyHillers:
@keyonghan do you know if any of these indexes are being used in the Dart backend?

keyonghan:
We are using all of them. Refer to:
https://github.com/flutter/cocoon/blob/master/app_dart/lib/src/service/datastore.dart#L99
https://github.com/flutter/cocoon/blob/master/app_dart/lib/src/service/datastore.dart#L102
https://github.com/flutter/cocoon/blob/master/app_dart/lib/src/request_handlers/debug/reset_pending_tasks.dart#L39

CaseyHillers:
@keyonghan sorry I misworded it. This removes composite indexes. So something like `filter.And(name filter, agentId filter)` would use these composite indexes. To my knowledge those filters linked are using the simple indexes.

CaseyHillers:
Closing this as these indexes are used in Cocoon. Opened https://github.com/flutter/flutter/issues/48914 to track removing these composite indexes to improve the latency of Cocoon's Datastore.

PR #620, opened 2020-01-13
Prompt sign in process when making authenticated requests if necessary

CaseyHillers:
If a user is not signed in, and goes to make an authenticated request in Cocoon they will see an error snackbar prompting them to sign in. This updates it so whenever we attempt to make an authenticated request and the user is not signed in, we start the sign in process.

## Issues

Fixes https://github.com/flutter/flutter/issues/44854

## Tested

Since this creates special behavior for id token, I removed some of the basic checks for id token.

### DIFF
--- a/app_flutter/lib/service/google_authentication.dart
+++ b/app_flutter/lib/service/google_authentication.dart
@@ -57,8 +57,11 @@ class GoogleSignInService {
       await signIn();
     }
 
-    return user?.authentication
+    final String idToken = await user?.authentication
         ?.then((GoogleSignInAuthentication key) => key.idToken);
+    assert(idToken != null && idToken.isNotEmpty);
+
+    return idToken;
   }
 
   /// Initiate the Google Sign In process.
--- a/app_flutter/lib/service/google_authentication.dart
+++ b/app_flutter/lib/service/google_authentication.dart
@@ -49,8 +49,17 @@ class GoogleSignInService {
   GoogleSignInAccount user;
 
   /// Authentication token to be sent to Cocoon Backend to verify API calls.
-  Future<String> get idToken => user?.authentication
-      ?.then((GoogleSignInAuthentication key) => key.idToken);
+  ///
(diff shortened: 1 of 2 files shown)


app_flutter/lib/service/google_authentication.dart
```diff
@@ -49,8 +49,17 @@ class GoogleSignInService {
...
-  Future<String> get idToken => user?.authentication
-      ?.then((GoogleSignInAuthentication key) => key.idToken);
+  ///
+  /// If there is no currently signed in user, it will prompt the sign in
+  /// process before attempting to return an id token.
+  Future<String> get idToken async {
+    if (!await isAuthenticated) {
+      await signIn();
+    }
+
```
digiter: Nit: assert idToken isn't null to make sure `signIn` function succeeded.

app_flutter/test/service/google_authentication_test.dart
```diff
@@ -21,7 +21,8 @@ void main() {
           .thenAnswer((_) => const Stream<GoogleSignInAccount>.empty());
       when(mockSignIn.isSignedIn())
           .thenAnswer((_) => Future<bool>.value(false));
-      authService = GoogleSignInService(googleSignIn: mockSignIn);
+      authService = GoogleSignInService(googleSignIn: mockSignIn)
+        ..notifyListeners = () => null;
```
digiter: I don't understand why notifyListeners needs to be set.

app_flutter/lib/service/google_authentication.dart
CaseyHillers: Added

app_flutter/test/service/google_authentication_test.dart
CaseyHillers: Good point! This needs to be set for the id token test where we check this new functionality. I moved this line out of the general set up and to the specific tests that need it. It's use is for updating the dashboard with the latest state. In unit tests, it just needs to be stubbed out.

PR #622, opened 2020-01-14
Enable flexibility to reduce dashboard commit number according to input parameter

keyonghan:
This PR solves the following two issues: reducing dashboard commit number from 100 to 50.

https://github.com/flutter/flutter/issues/43098
https://github.com/flutter/flutter/issues/44478

### DIFF
--- a/app_dart/lib/src/service/build_status_provider.dart
+++ b/app_dart/lib/src/service/build_status_provider.dart
@@ -68,7 +68,8 @@ class BuildStatusProvider {
   Stream<CommitStatus> retrieveCommitStatus() async* {
     const int maxRecords = 50;
     final DatastoreService datastore = datastoreProvider();
-    await for (Commit commit in datastore.queryRecentCommits(limit: maxRecords)) {
+    await for (Commit commit
+        in datastore.queryRecentCommits(limit: maxRecords)) {
       final List<Stage> stages =
           await datastore.queryTasksGroupedByStage(commit);
       yield CommitStatus(commit, stages);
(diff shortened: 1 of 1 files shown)


app_dart/lib/src/service/build_status_provider.dart
```diff
@@ -66,8 +66,9 @@ class BuildStatusProvider {
   /// The returned stream will be ordered by most recent commit first, then
   /// the next newest, and so on.
   Stream<CommitStatus> retrieveCommitStatus() async* {
+    const int maxRecords = 50;
```
CaseyHillers: This should be a parameter passed in to the function. That way we can update `/api/public/get-build-status` to only look at the 25 most recent commits.

keyonghan: Oh, so we want the api to have the controllability over the commit number then.

CaseyHillers: I was approaching this problem in another PR and I implemented it like https://github.com/flutter/cocoon/pull/617/files#diff-82ed9182ac5fc63b1aa6aa311754159dR83 to work around `/api/public/get-status` and `/api/public/get-build-status` using the same retrieval code.

keyonghan: I saw you removed yourself from issue: https://github.com/flutter/flutter/issues/44478, and thought you may not work on this issue.

Since you have started working on this issue. I will close this PR, and issue: https://github.com/flutter/flutter/issues/44478. 

And leave https://github.com/flutter/flutter/issues/43098 to you. 

Please let me know if you have a different opinion.

CaseyHillers: That works. Would you be able to pick up the backend functionality for https://github.com/flutter/flutter/issues/43098? I was thinking of creating something similar to `/api/public/get-commit-statuses?lastKey=[LastKnownTaskKey]` for the backend api call. I can implement the frontend functionality, and then we can get rid of the current `/api/public/get-status`.

PR #624, opened 2020-01-15
Reduce commit rows in dashboard

keyonghan:
This PR reduces the default commit records shown on build dashboard from 100 to 25. 

This is the backend change, and enables `/api/public/get-status` to support input variable: ownerKey (commit). When the input variable is specified, this API will return the next 25 records whose timestamp is smaller than that of the input commit.

This will make it possible to append more records to build page when requested, and effectively reduce datastore queries.

### DIFF
--- a/app_dart/lib/src/datastore/cocoon_config.dart
+++ b/app_dart/lib/src/datastore/cocoon_config.dart
@@ -114,7 +114,7 @@ class Config {
   int get maxTaskRetries => 2;
 
   /// The default number of commit shown in flutter build dashboard.
-  int get commitNumber => 25;
+  int get commitNumber => 1;
 
   // TODO(keyonghan): update all existing APIs to use this reference, https://github.com/flutter/flutter/issues/48987.
   KeyHelper get keyHelper =>
--- a/app_dart/lib/src/datastore/cocoon_config.dart
+++ b/app_dart/lib/src/datastore/cocoon_config.dart
@@ -112,6 +112,11 @@ class Config {
 
   int get maxTaskRetries => 2;
 
+  /// The default number of commit shown in flutter build dashboard.
+  int get commitNumber => 25;
+
+  AppEngineContext get applicationContext => context.applicationContext;
+
   String get cqLabelName => 'CQ+1';
 
   String get waitingForTreeToGoGreenLabelName => 'waiting for tree to go green';
--- a/app_dart/lib/src/datastore/cocoon_config.dart
+++ b/app_dart/lib/src/datastore/cocoon_config.dart
@@ -16,6 +16,7 @@ import 'package:googleapis/bigquery/v2.dart' as bigquery;
 import 'package:meta/meta.dart';
 
 import '../../cocoon_service.dart';
+import '../model/appengine/key_helper.dart';
 import '../model/appengine/service_account_info.dart';
 import '../service/access_client_provider.dart';
 import '../service/bigquery.dart';
@@ -115,7 +116,7 @@ class Config {
   /// The default number of commit shown in flutter build dashboard.
   int get commitNumber => 25;
 
-  AppEngineContext get applicationContext => context.applicationContext;
+  KeyHelper get keyHelper => KeyHelper(applicationContext: context.applicationContext);
 
   String get cqLabelName => 'CQ+1';
 
--- a/app_dart/lib/src/datastore/cocoon_config.dart
+++ b/app_dart/lib/src/datastore/cocoon_config.dart
@@ -116,6 +116,7 @@ class Config {
   /// The default number of commit shown in flutter build dashboard.
   int get commitNumber => 25;
 
+  // TODO(keyonghan): update all existing APIs to use this reference, https://github.com/flutter/flutter/issues/48987.
   KeyHelper get keyHelper =>
       KeyHelper(applicationContext: context.applicationContext);
 
--- a/app_dart/lib/src/datastore/cocoon_config.dart
+++ b/app_dart/lib/src/datastore/cocoon_config.dart
@@ -114,7 +114,7 @@ class Config {
   int get maxTaskRetries => 2;
 
   /// The default number of commit shown in flutter build dashboard.
-  int get commitNumber => 1;
+  int get commitNumber => 30;
 
   // TODO(keyonghan): update all existing APIs to use this reference, https://github.com/flutter/flutter/issues/48987.
   KeyHelper get keyHelper =>
--- a/app_dart/lib/src/datastore/cocoon_config.dart
+++ b/app_dart/lib/src/datastore/cocoon_config.dart
@@ -116,7 +116,8 @@ class Config {
   /// The default number of commit shown in flutter build dashboard.
   int get commitNumber => 25;
 
-  KeyHelper get keyHelper => KeyHelper(applicationContext: context.applicationContext);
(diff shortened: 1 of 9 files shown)


keyonghan:
Related issues:
https://github.com/flutter/flutter/issues/43098
https://github.com/flutter/flutter/issues/44478

app_dart/lib/src/datastore/cocoon_config.dart
```diff
@@ -112,6 +113,12 @@ class Config {
 
   int get maxTaskRetries => 2;
 
+  /// The default number of commit shown in flutter build dashboard.
+  int get commitNumber => 25;
+
+  KeyHelper get keyHelper =>
```
CaseyHillers: Since key helper is used throughout the code base, can you update all the constructions to just use this reference instead? This can be done in a separate PR.

app_dart/lib/src/request_handlers/get_status.dart
```diff
@@ -30,15 +31,36 @@ class GetStatus extends RequestHandler<Body> {
   final DatastoreServiceProvider datastoreProvider;
   final BuildStatusProvider buildStatusProvider;
 
+  static const String ownerKeyParam = 'ownerKey';
```
CaseyHillers: I am unclear of what this does. It could be more descriptive if called `commitKey`, but I still don't know what it's for.

CaseyHillers: Maybe `lastCommitKey`?

app_dart/lib/src/request_handlers/get_status.dart
```diff
@@ -30,15 +31,36 @@ class GetStatus extends RequestHandler<Body> {
...
+    /// list whose timestamp is smaller than [timestamp].
+    int timestamp = DateTime.now().millisecondsSinceEpoch;
+
+    if (encodedOwnerKey != null) {
+      final Key ownerKey = keyHelper.decode(encodedOwnerKey);
+      final Commit commit =
+          await datastore.db.lookupValue<Commit>(ownerKey, orElse: () => null);
+
+      timestamp = commit?.timestamp;
+    }
```
CaseyHillers: I think this logic is enough to just be moved its own function

app_dart/lib/src/datastore/cocoon_config.dart
keyonghan: Yeah, I plan to do it in another PR.

app_dart/lib/src/request_handlers/get_status.dart
keyonghan: `lastCommitKey` works!

dnfield:
Don't let this derail you but, it seems like it might be good to have a view _only_ of tests that are either running, red, or passed on re-run.

app_dart/lib/src/request_handlers/get_status.dart
keyonghan: Done.

app_dart/lib/src/datastore/cocoon_config.dart
CaseyHillers: Can we add a TODO?

keyonghan:
@dnfield I believe current dashboard lists red or yellow tests on the left to make it easy to see. As for running ones, I am not sure about the logic. @CaseyHillers may know.

CaseyHillers:
Another note, if we move this to a separate handler it's easier to configure longer ttls on the `CacheRequestHandler`. I don't think these tasks will get changed very often, and we could probably store these pages for a day to a week compared to this handler only storing for the past minute.

However, I don't think there will be many of these requests where we would need to prioritize this caching (and there's a 1 in 25 chance in this current PR the cache could get hit).

CaseyHillers:
@keyonghan I believe Dan is talking about optimizing how much data we send down. Over 50% of the tasks are green, and their information isn't too relevant. We could optimize the size of the network request by excluding these tasks, and have the build dashboard fill in unknown tasks as green.

The edge case being tasks that are skipped will need to be populated by the backend.

dnfield:
We could both optimize the network request _and_ optimize the amount of data we actually show. As we add more tests, it gets harder to parse which test is which in the grid - and it takes longer to render.

keyonghan:
Yeah, I can imagine how hard to parse tests when we have more. Optimizing the network request should be a good solution to this pain point.

app_dart/lib/src/datastore/cocoon_config.dart
keyonghan: Added.

app_dart/lib/src/request_handlers/get_status.dart
```diff
@@ -30,15 +31,30 @@ class GetStatus extends RequestHandler<Body> {
...
+    final String encodedLastCommitKey =
+        request.uri.queryParameters[lastCommitKeyParam];
+    final DatastoreService datastore = datastoreProvider();
+    final KeyHelper keyHelper = config.keyHelper;
+    final int commitNumber = config.commitNumber;
+
+    /// [timestamp] is initially set as the current time, which allows query return
+    /// latest commit list. If [owerKeyParam] is not empty, [timestamp] will be set
+    /// as the timestamp of that [commit], and the query will return lastest commit
+    /// list whose timestamp is smaller than [timestamp].
```
CaseyHillers: This documentation should be moved to document `_obtainTimestamp`.

app_dart/lib/src/request_handlers/get_status.dart
```diff
@@ -30,15 +31,30 @@ class GetStatus extends RequestHandler<Body> {
...
+    final int timestamp =
```
CaseyHillers: Maybe `lastCommitTimestamp` is clearer?

app_dart/test/request_handlers/get_status_test.dart
```diff
@@ -74,5 +83,65 @@ void main() {
...
+
+      const String expectedLastCommitKeyEncoded =
+          'ahNzfmZsdXR0ZXItZGFzaGJvYXJkckcLEglDaGVja2xpc3QiOGZsdXR0ZXIvZmx1dHRlci9lYTI4YTljMzRkYzcwMWRlODkxZWFmNzQ1MDNjYTQ3MTcwMTlmODI5DA';
+
+      tester.request = FakeHttpRequest(queryParametersValue: <String, String>{
+        GetStatus.lastCommitKeyParam: expectedLastCommitKeyEncoded,
+      });
+      final Map<String, dynamic> result = await decodeHandlerBody();
+
+      expect(result['Statuses'].length, 1);
```
CaseyHillers: Let's verify that `result['Statuses'].first` is the status for `commit2`.

app_dart/lib/src/request_handlers/get_status.dart
keyonghan: moved.

app_dart/lib/src/request_handlers/get_status.dart
keyonghan: updated.

app_dart/test/request_handlers/get_status_test.dart
keyonghan: Added.

CaseyHillers:
I landed https://github.com/flutter/cocoon/pull/625, so you'll need to update

PR #625, opened 2020-01-16
Fix /api/public/get-status only returning 20 commits

CaseyHillers:
Introduced in https://github.com/flutter/cocoon/pull/617 where `retrieveCommitStatus` had a hard coded value for the number of commits to store. Instead it needed it to be passed in as a variable.

The reason for 20 commits is that's how many the build status algorithm looks at to validate the current tree state. Before it used to be 100.

# Preview

https://v350-dot-flutter-dashboard.appspot.com/#/build


CaseyHillers:
If https://github.com/flutter/cocoon/pull/624 lands before this, we can close this.

keyonghan:
Let's land this PR first. I see there are also changes regarding tree status.

PR #627, opened 2020-01-17
improve review handling logic

dnfield:
The old logic allows for non-members/owners to approve

It also doesn't handle when someone goes from changes requested to approved.

This does that and adds tests.


dnfield:
I found this while looking into fixing https://github.com/flutter/flutter/issues/49061

I will address https://github.com/flutter/flutter/issues/49061 once this lands.

PR #630, opened 2020-01-17
Unlabel PRs with failing checks

dnfield:
This change makes the bot unlabel PRs that have statsuses or checks that are failing _other than_ luci-engine or flutter-build.

The idea is that if any other check/status is failing, the author needs to take action to actually fix those (as opposed to luci-engine and flutter-build, which get resolved without additional action on the PR).

This is important for autorollers, as they will otherwise not know when to close their PR.

@rmistry

fixes https://github.com/flutter/flutter/issues/49061


dnfield:
Fixes https://github.com/flutter/flutter/issues/49061

dnfield:
I'm fine with an¥ refactor to make these things clearer.

I'm not sure if it's a bug or not.  There are definitely a few tracking things where const lints aren't applied in a sensible manner though.

PR #631, opened 2020-01-21
Remove cirrus support for cocoon project.

godofredoc:
We will be using this repo to improve the APIs communicating with
github, luci and dashboard.

https://github.com/flutter/flutter/issues/43348


godofredoc:
Yes, we will need to that too.

PR #632, opened 2020-01-22
Write X-Flutter-IdToken cookie on sign in changes

CaseyHillers:
The cookie `X-Flutter-IdToken` is used by the Angular Dart dashboards for authenticating with Cocoon. The new Flutter app sends it as a header instead so that authentication in the app works on both mobile and web. When making direct requests to the api and using only the Flutter app, the request will fail when it previously worked on the old dashboards. This is because this cookie isn't being set.

This updates the code to write the auth cookie in the same fashion as the old dashboard so API requests can be made from the browser. That is, when the signed in user changes, write their id token as a cookie.

This removes the arbitrary wait on the cookie being written for downloading logs.

# Tested

With the main functionality being the html cookie, I was only able to test that the cookie service is called correctly when the signed in user changes.

### DIFF
--- a/app_flutter/lib/service/cookie_web.dart
+++ b/app_flutter/lib/service/cookie_web.dart
@@ -9,9 +9,9 @@ import 'cookie_interface.dart' as i;
 /// Utility service for managing HTML cookies.
 class Cookie implements i.Cookie {
   @override
-  Future<void> set(String name, String value, {String options}) async {
+  Future<void> set(String name, String value, {String options = ''}) async {
     // This line is dangerous as it fails silently. Be careful.
-    html.document.cookie = '$name=$value;path=/';
+    html.document.cookie = '$name=$value;$options';
 
     // This wait is a work around as the above line is not synchronous.
     // The cookie needs to be set for the request to be authenticated.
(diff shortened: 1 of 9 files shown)


CaseyHillers:
\cc @dnfield this fixes the issue you ran into last week where you had to sign in to the performance dashboard before making API requests to Cocoon via the browser.

app_flutter/lib/service/cookie_web.dart
```diff
@@ -0,0 +1,23 @@
...
+  Future<void> set(String name, String value, {String options = ''}) async {
+    // This line is dangerous as it fails silently. Be careful.
+    html.document.cookie = '$name=$value;$options';
+
+    // This wait is a work around as the above line is not synchronous.
+    // The cookie needs to be set for the request to be authenticated.
+    //
+    // dart:html will say the cookie has been written, but the browser
+    // is still writing it.
+    await Future<void>.delayed(const Duration(milliseconds: 500));
```
digiter: Is there a way to query the cookie to make sure it's written?

CaseyHillers: Follow up from https://github.com/flutter/cocoon/pull/531#discussion_r348190293, I made a demo script to try and achieve a way of polling for it, but was unsuccessful.

digiter: Ah, right

PR #633, opened 2020-01-22
Remove dead code from debug handlers

CaseyHillers:
I searched the logs on all AppEngine versions from the past 31 days, and found no use of either of these end points. Neither are on cron jobs.

I believe `reset-pending-tasks` was renamed `vacuum-clean`, but it was never removed (both have similar functionality). The other being a test handler when Cocoon was starting out on the Dart custom runtime.

I plan to have a follow up PR that modifies vacuum clean to not use `status` as an index so we can remove that index from Datastore, and improve the overall latency of task queries.


digiter:
Removed technical debt, yay!

PR #635, opened 2020-01-23
Remove agent id composite index on task

CaseyHillers:
I scanned all of `app_dart` and could not find any `order('reservedForAgentID')` or `filter('reservedForArgentID')` calls. I believe this is safe to remove.

# Issues

https://github.com/flutter/flutter/issues/48914

### DIFF
--- a/app_dart/index.yaml
+++ b/app_dart/index.yaml
@@ -19,11 +19,6 @@ indexes:
   - name: CreateTimestamp
     direction: desc
 - kind: Task
-  properties:
-  - name: ReservedForAgentID
-  - name: CreateTimestamp
-    direction: desc
-- kind: Task
   ancestor: yes
   properties:
   - name: Name


CaseyHillers:
@keyonghan this is follow up to https://github.com/flutter/cocoon/pull/619. You didn't list any sources for the index for reservedForAgentId, and I was unable to find any. Do you think this is okay to remove?

keyonghan:
This is the one as mentioned in #619 https://github.com/flutter/cocoon/blob/master/app_dart/lib/src/request_handlers/debug/reset_pending_tasks.dart#L39
But that file seems to have been removed. Any chance you know who removed it?

CaseyHillers:
@keyonghan that is correct. See https://github.com/flutter/cocoon/pull/633 since I removed it as the first step to make this PR possible.

CaseyHillers:
Since this index is the least likely to break anything, I am going to start with it.

PR #639, opened 2020-01-27
Throw more details when GithubBuildStatusUpdate has more than one entries

digiter:


app_dart/lib/src/service/datastore.dart
```diff
@@ -151,7 +151,8 @@ class DatastoreService {
         updates: 0,
       );
     } else {
-      assert(previousStatusUpdates.length == 1);
+      assert(previousStatusUpdates.length == 1,
```
dnfield: We're not running with asserts on GCP.

CaseyHillers:
```suggestion
      assert(previousStatusUpdates.length == 1);

      if (previousStatusUpdates.length != 1) {
        print('GithubBuildStatusUpdate should have no more than one entries on repository $slug.fullName, pr $pr.number, head $pr.head.sha');
      }

      return previousStatusUpdates[0];
```
If this is an urgent issue, we could just start by printing to stdout. I believe GCP is capturing stdout of the server file and logging it. Otherwise, we should probably inject Logger to this class and start logging this specific case. Would be good to find out what caused this to create duplicate keys.

digiter: Changed to throwing an exception instead. How about this way?

app_dart/lib/src/service/datastore.dart
```diff
@@ -151,6 +151,9 @@ class DatastoreService {
         updates: 0,
       );
     } else {
+      if (previousStatusUpdates.length > 1) {
+        throw 'GithubBuildStatusUpdate should have no more than one entries on repository ${slug.fullName}, pr ${pr.number}, head ${pr.head.sha}';
```
dnfield: We should throw an actual exception here - maybe a `StateError`?

dnfield: And with this, you can get rid of the assert below.

digiter: Done.

app_dart/lib/src/service/datastore.dart
CaseyHillers: SGTM. I didn't know if we wanted it to make a best effort approach.

dnfield: Having multiple entries in here with the same hash indicates something has gone wrong or our understanding of the way all this is working is wrong.

Although maybe something got mixed up where multiple PRs had the same hashes in them for some reason.  Perhaps because of something like https://github.com/flutter/flutter/pull/48985

digiter:
A duplicate found by this change.
WN6LA8vFYuO

PR #640, opened 2020-01-28
Do 2 PRs per cycle, oldest first (by create date)

dnfield:
Right now, the bot is not specifying a sort order when querying for PRs, which can starve older PRs.

Also bumps the number of PRs we merge every 5 minutes to 2 instead of 1.

### DIFF
--- a/app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
+++ b/app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
@@ -46,7 +46,7 @@ class CheckForWaitingPullRequests extends ApiRequestHandler<Body> {
     Logging log,
     GraphQLClient client,
   ) async {
-    bool hasMerged = false;
+    int mergeCount = 0;
     final Map<String, dynamic> data = await _queryGraphQL(
       owner,
       name,
@@ -54,13 +54,16 @@ class CheckForWaitingPullRequests extends ApiRequestHandler<Body> {
       client,
     );
     for (_AutoMergeQueryResult queryResult in _parseQueryData(data)) {
-      if (!hasMerged && queryResult.shouldMerge) {
-        hasMerged = await _mergePullRequest(
+      if (mergeCount < 2 && queryResult.shouldMerge) {
+        final bool merged = await _mergePullRequest(
           queryResult.graphQLId,
           queryResult.sha,
           log,
           client,
         );
+        if (merged) {
+          mergeCount++;
+        }
       } else if (queryResult.shouldRemoveLabel) {
         await _removeLabel(
           queryResult.graphQLId,
--- a/app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
+++ b/app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
@@ -18,6 +18,11 @@ import '../request_handling/body.dart';
 
 import 'check_for_waiting_pull_requests_queries.dart';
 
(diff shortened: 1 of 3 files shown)


app_dart/lib/src/request_handlers/check_for_waiting_pull_requests_queries.dart
```diff
@@ -8,7 +8,7 @@ query LabeledPullRequcodeestsWithReviews($sOwner: String!, $sName: String!, $sLa
     labels(first: 1, query: $sLabelName) {
       nodes {
         id
-        pullRequests(first: 100, states: OPEN) {
+        pullRequests(first: 100, states: OPEN, orderBy: {direction: ASC, field: CREATED_AT}) {
```
dnfield: Not sure how to test this other than asserting the query contains it, which seems excessive.

app_dart/test/request_handlers/check_for_waiting_pull_requests_test.dart
```diff
@@ -206,7 +206,8 @@ void main() {
       );
     });
 
-    test('Merges first PR in list, all successful', () async {
+    test('Merges first 2 PRs in list, all successful', () async {
```
CaseyHillers: Can we have a test for 3 PRs, where either the first or second are not merged, but it still process all 3 PRs?

goderbauer:
LGTM

app_dart/test/request_handlers/check_for_waiting_pull_requests_test.dart
CaseyHillers: Yes. For example, the oldest PR has failing checks so we don't merge it. However, the next oldest can be merged. I want to make sure that we don't break this functionality in the future.

dnfield: Done.

app_dart/test/request_handlers/check_for_waiting_pull_requests_test.dart
```diff
@@ -220,7 +221,66 @@ void main() {
...
+            variables: <String, dynamic>{
+              'id': engineRepoPRs.first.id,
+              'oid': oid,
+            },
+          ),
+        ],
+      );
+    });
+
+    test('Merges first 2 PRs in list, one not successful', () async {
```
digiter: Nit: Merges first 2 PRs in list, but the 3rd one failed

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
```diff
@@ -46,21 +46,24 @@ class CheckForWaitingPullRequests extends ApiRequestHandler<Body> {
...
     final Map<String, dynamic> data = await _queryGraphQL(
       owner,
       name,
       log,
       client,
     );
     for (_AutoMergeQueryResult queryResult in _parseQueryData(data)) {
-      if (!hasMerged && queryResult.shouldMerge) {
-        hasMerged = await _mergePullRequest(
+      if (mergeCount < 2 && queryResult.shouldMerge) {
```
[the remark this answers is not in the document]
CaseyHillers: This could be moved to the config.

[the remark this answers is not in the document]
dnfield: Done.  If we find we're tweaking this much we should move it to config, but I don't expect us to be moving it around much.

dnfield:
Deploying as v357

PR #643, opened 2020-01-30
Fix status checks for waiting label

dnfield:
/cc @rmistry @fkorotkov

More context in: https://github.com/flutter/engine/pull/16220

The top level conclusion field for a check run is not reliable.  This uses the conclusion field from all sub-jobs of the check run instead.  This also should have the benefit of saying which specific job failed, rather than just the top level app name.


dnfield:
Resolves part of https://github.com/flutter/flutter/issues/49817

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
```diff
@@ -257,12 +257,14 @@ bool _checkStatuses(
...
-      if (check['conclusion'] == 'FAILURE' &&
-          !notInAuthorsControl.contains(name)) {
-        failures.add(name);
+    for (Map<String, dynamic> checkRun in check['checkRuns']['nodes']) {
+      if (checkRun['conclusion'] != 'SUCCESS') {
+        final String name = checkRun['name'];
+        allSuccess = false;
+        if (checkRun['conclusion'] == 'FAILURE' &&
+            !notInAuthorsControl.contains(name)) {
+          failures.add(name);
```
fkorotkov: BTW how do you check that CI is in progress? Something like `!allSuccess && failures.isEmpty`?

dnfield: We delay actually checking the PR for an hour after the last push event. 

The idea is that should be more than enough time to get CI started, and avoids needing to hard code anything about how many jobs we expect to run.

PR #644, opened 2020-01-31
reduce nodes touched

dnfield:
GitHub is telling us we're touching too many nodes with this query as it stands - it hits 100 PRs x 100 check suites x 100 jobs per suite, which is 1m nodes.

We only have two check suites.  I'm reducing it to 10, which keeps us under the limits.  A bigger solution would be to actually do paging.


dnfield:
An example error log:

```
By the time this query traverses to the checkRuns connection, it is requesting up to 1,000,000 possible nodes which exceeds the maximum limit of 500,000.: Undefined location
```

dnfield:
To unblock rollers, I'm publishing this as v359.  If we want to make other changes to it we can do that and republish.

PR #646, opened 2020-02-04
Disable attempts increase when resetting devicelab task

keyonghan:
This PR fixes issue: https://github.com/flutter/flutter/issues/44780

### DIFF
--- a/app_dart/lib/src/request_handlers/reset_devicelab_task.dart
+++ b/app_dart/lib/src/request_handlers/reset_devicelab_task.dart
@@ -35,8 +35,9 @@ class ResetDevicelabTask extends ApiRequestHandler<Body> {
   Future<Body> post() async {
     checkRequiredParameters(<String>[keyParam]);
     final String encodedKey = requestData[keyParam];
+    final ClientContext clientContext = authContext.clientContext;
     final KeyHelper keyHelper =
-        KeyHelper(applicationContext: context.applicationContext);
+        KeyHelper(applicationContext: clientContext.applicationContext);
     final Key key = keyHelper.decode(encodedKey);
 
     final DatastoreService datastore = datastoreProvider();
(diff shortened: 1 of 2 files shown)


app_dart/lib/src/request_handlers/reset_devicelab_task.dart
```diff
@@ -35,8 +35,9 @@ class ResetDevicelabTask extends ApiRequestHandler<Body> {
   Future<Body> post() async {
     checkRequiredParameters(<String>[keyParam]);
     final String encodedKey = requestData[keyParam];
+    final ClientContext clientContext = authContext.clientContext;
     final KeyHelper keyHelper =
-        KeyHelper(applicationContext: context.applicationContext);
+        KeyHelper(applicationContext: clientContext.applicationContext);
```
CaseyHillers: Why was this changed?

app_dart/test/request_handlers/reset_devicelab_task_test.dart
```diff
@@ -0,0 +1,54 @@
...
+          key: commit.key.append(Task, id: 4590522719010816),
+          commitKey: commit.key,
+          attempts: 0);
+      config.db.values[task.key] = task;
+
+      expect(task.attempts, 0);
+
+      await tester.post(handler);
+      // Reset devicelab task will not increase [attempts].
+      expect(task.attempts, 0);
```
CaseyHillers: Can we add an expectation that the task is marked new?

keyonghan: Added.

app_dart/lib/src/request_handlers/reset_devicelab_task.dart
keyonghan: `context` tries to find `_APPENGINE_CONTEXT`, but test fails in local development.

CaseyHillers: That's interesting. Is this how the other Authenticated RequestHandlers use keyhelper then?

keyonghan: I believe so. If not, then we need to update when creating corresponding tests.

PR #648, opened 2020-02-04
Reschedule infra failures on try jobs

dnfield:
Fixes https://github.com/flutter/flutter/issues/50118

Rather than failing immediately on any failure, reschedule infra failures up to 2 times.


rmistry:
> /cc @rmistry

👍 👍 👍

app_dart/lib/src/datastore/cocoon_config.dart
```diff
@@ -113,6 +113,9 @@ class Config {
 
   int get maxTaskRetries => 2;
 
+  /// The number of times to retry a LUCI job on infra failures.
+  int get luciTryJobRetries => 2;
```
CaseyHillers: nit: the variable name should mention it's for infra failure attempts, but I don't have a better name other than `maxLuciAttemptsOnInfraFailure`

app_dart/lib/src/request_handlers/luci_status.dart
```diff
@@ -81,6 +88,78 @@ class LuciStatusHandler extends RequestHandler<Body> {
     return Body.empty;
   }
 
+  Future<void> _processCompleted({
```
CaseyHillers: Does `_taskCompleted(...)` make more sense in this situation?

app_dart/lib/src/request_handlers/luci_status.dart
```diff
@@ -81,6 +88,78 @@ class LuciStatusHandler extends RequestHandler<Body> {
...
+    }
+    await _setCompletedStatus(
+      ref: sha,
+      builderName: builderName,
+      buildUrl: build.url,
+      result: build.result,
+    );
+  }
+
+  Future<void> _rescheduleBuild({
```
CaseyHillers: method documentation? (also for _processComplete)

app_dart/lib/src/request_handlers/luci_status.dart
```diff
@@ -12,10 +12,12 @@ import 'package:meta/meta.dart';
 
 import '../datastore/cocoon_config.dart';
 import '../model/appengine/service_account_info.dart';
+import '../model/luci/buildbucket.dart' as bb;
```
CaseyHillers: Not a comment, but I like this prefix :)

app_dart/test/request_handlers/luci_status_test.dart
```diff
@@ -234,6 +237,55 @@ void main() {
     );
   });
 
+  test('Reschedules an infra failure', () async {
```
CaseyHillers: Does this also check that we do not get into a reschedule loop?

app_dart/lib/src/datastore/cocoon_config.dart
dnfield: Going with luciTryInfraFailureRetries

app_dart/lib/src/request_handlers/luci_status.dart
dnfield: Ahh this is ambiguous.  The ideas is this method should process completed jobs, it's a bad name.  I'll update and document.

app_dart/test/request_handlers/luci_status_test.dart
dnfield: Added another test to verify the negative side of what we do when we have exceeded the retries.

This test does verify that we increment the retries counter on the user_data.  The new test verifies that if that value >= the config value, we set the status to failure.

dnfield:
Deploying as v364

PR #649, opened 2020-02-07
Switch cirrus status calculation from Github check-runs to Cirrus API

keyonghan:
This PR solved issue https://github.com/flutter/flutter/issues/50293

### DIFF
--- a/app_dart/lib/src/datastore/cocoon_config.dart
+++ b/app_dart/lib/src/datastore/cocoon_config.dart
@@ -295,6 +295,17 @@ class Config {
     );
   }
 
+  Future<GraphQLClient> createCirrusGraphQLClient() async {
+    final HttpLink httpLink = HttpLink(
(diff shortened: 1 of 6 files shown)


keyonghan:
> Lgtm. Should we do this in the waiting for tree to go green logic as well?

@dnfield The logic in /api/check-waiting-pull-requests seems to be enough? It checks all PRs with `waiting for tree to go green`, and it checks all tests status (including the cirrus result) when merging.

Here this API only updates the cirrus status. Doesn't seem necessary to consider the go green logic.

Please correct me if I understand incorrectly.

keyonghan:
> Lgtm. Should we do this in the waiting for tree to go green logic as well?

@dnfield You are right. I see that logic is still using checkruns. Will merge this PR now so the build will perform well. And create a separate issue to track: https://github.com/flutter/flutter/issues/50350

dnfield:
Great, thanks!

PR #650, opened 2020-02-11
Use cirrus api in waiting for tree to go green logic

keyonghan:
Thus is a follow up PR solving issue: https://github.com/flutter/flutter/issues/50293

### DIFF
--- a/app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
+++ b/app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
@@ -17,7 +17,7 @@ import '../request_handling/authentication.dart';
 import '../request_handling/body.dart';
 
 import 'check_for_waiting_pull_requests_queries.dart';
-import 'refresh_cirrus_status_queries.dart';
+import 'refresh_cirrus_status.dart';
 
 /// Maximum number of pull requests to merge on each check.
 /// This should be kept reasonably low to avoid flooding infra when the tree
@@ -109,41 +109,6 @@ class CheckForWaitingPullRequests extends ApiRequestHandler<Body> {
     }
 
     return result.data;
-  }
-
-  Future<List<dynamic>> _queryCirrusGraphQL(
-    String sha,
-    GraphQLClient client,
-  ) async {
-    assert(client != null);
-    const String owner = 'flutter';
-    const String name = 'flutter';
-    final QueryResult result = await client.query(
-      QueryOptions(
-        document: cirusStatusQuery,
-        fetchPolicy: FetchPolicy.noCache,
-        variables: <String, dynamic>{
-          'owner': owner,
-          'name': name,
-          'SHA': sha,
-        },
-      ),
-    );
-
-    if (result.hasErrors) {
-      for (GraphQLError error in result.errors) {
-        log.error(error.toString());
-      }
-      throw const BadRequestException('GraphQL query failed');
-    }
-
-    final List<dynamic> tasks = <dynamic>[];
-    if (result.data == null) {
-      return tasks;
-    }
-    final Map<String, dynamic> searchBuilds = result.data['searchBuilds'].first;
-    tasks.addAll(searchBuilds['latestGroupTasks']);
-    return tasks;
   }
 
   Future<bool> _removeLabel(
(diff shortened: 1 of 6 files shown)


app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
```diff
@@ -110,6 +110,49 @@ class CheckForWaitingPullRequests extends ApiRequestHandler<Body> {
...
+    assert(client != null);
+    const String owner = 'flutter';
+    const String name = 'flutter';
+    const String cirusStatusQuery = r'''
+    query BuildBySHAQuery($owner: String!, $name: String!, $SHA: String) { 
+      searchBuilds(repositoryOwner: $owner, repositoryName: $name, SHA: $SHA) { 
+        id latestGroupTasks { 
+          id name status 
+        } 
+      }
```
dnfield: Should this now be moved to a common query file to share between both places?

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
```diff
@@ -223,52 +263,36 @@ class CheckForWaitingPullRequests extends ApiRequestHandler<Body> {
...
-    final String name = status['context'];
-    if (status['state'] != 'SUCCESS') {
-      allSuccess = false;
-      if (status['state'] == 'FAILURE' && !notInAuthorsControl.contains(name)) {
-        failures.add(name);
-      }
+  /// Returns whether all statuses are successful.
+  ///
+  /// Also fills [failures] with the names of any status/check that has failed.
+
```
dnfield: nit: remove blank line

also nit: why remove the documentation?

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
keyonghan: Created a query file: refresh_cirrus_status_queries

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
keyonghan: Since here we avoid the checks/statuses but use the cirrus API, I just keep those directly related to the cirrus API results. Do you see any part we should keep? I will add them back then.

keyonghan:
removed in both query script and the test.

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests_queries.dart
```diff
@@ -23,27 +23,6 @@ query LabeledPullRequcodeestsWithReviews($sOwner: String!, $sName: String!, $sLa
                   oid
                   committedDate
                   pushedDate
-                  status {
-                    contexts {
-                      context
-                      state
-                    }
-                  }
```
dnfield: Don't we still need this?

keyonghan: I doubt. Both the removed checkSuites and statuses below were used for cirrus status check if I understand correctly. After using cirrus API to get and calculate the status, we do not seem to need any of those two.
```final List<Map<String, dynamic>> checkSuites =
          commit['checkSuites']['nodes'].cast<Map<String, dynamic>>();
      final List<Map<String, dynamic>> statuses =
          commit['status']['contexts'].cast<Map<String, dynamic>>();```

Comment (dnfield) [reply app_dart/lib/src/request_handlers/check_for_waiting_pull_requests_queries.dart]: The status check is needed to see if the tree is greeen for luci-engine and flutter-build

Comment (keyonghan) [reply app_dart/lib/src/request_handlers/check_for_waiting_pull_requests_queries.dart]: Sorry I missed that part. Now added statuses back.

File: app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart [hunk 4]
```diff
@@ -110,6 +111,41 @@ class CheckForWaitingPullRequests extends ApiRequestHandler<Body> {
     return result.data;
   }
 
+  Future<List<dynamic>> _queryCirrusGraphQL(
```
dnfield: This seems like it has potential to live in a shared file

keyonghan: Kept that part in refresh_cirrus_status.dart only, and referenced it from check_for_waiting_pull_requests.dart.

keyonghan:
@dnfield One question: `flutter` and `engine` share `_checkPRs`, `_parseQueryData`, `_checkStatuses`, and then `queryCirrusGraphQL`. However in `queryCirrusGraphQL`, I used `ower='flutter'`, and `name='flutter'`.

Does `engine` have cirrus tests run?

dnfield:
> @dnfield One question: `flutter` and `engine` share `_checkPRs`, `_parseQueryData`, `_checkStatuses`, and then `queryCirrusGraphQL`. However in `queryCirrusGraphQL`, I used `ower='flutter'`, and `name='flutter'`.
> 
> Does `engine` have cirrus tests run?

The engine does have cirrus tests run. We should pass the name parameter into the queryCirrusGraphQL.  I missed that.

keyonghan:
Added repo name as parameter.

PR #653, opened 2020-02-14
Slow vacuum clean cron to run hourly

CaseyHillers:
The vacuum clean handler looks for tasks that are running for more than an hour.

Currently it checks every 10 minutes, which seems excessive with the current usage.

### DIFF
--- a/app_dart/cron.yaml
+++ b/app_dart/cron.yaml
@@ -9,7 +9,7 @@ cron:
   schedule: every 3 minutes
 - description: clean up stale datastore records
   url: /api/vacuum-clean
-  schedule: every 10 minutes
+  schedule: every 1 hours
 - description: sends build status to GitHub to annotate PRs and commits
   url: /api/push-build-status-to-github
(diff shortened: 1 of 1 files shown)


keyonghan:
Do you think it is good to change to half an hour? For the worst case, if the cronjob is called when a test runs for 59 mins, then it will wait another hour to clean it up, which adds the total cleaning time to be two hours.

CaseyHillers:
@keyonghan the vacuum clean works by removing tasks that have been running for over an hour. Thus, we would see instead worst case being a task having run for 2 hours, instead of currently 1 hour and 10 minutes.

I don't have the data (if it's in bigquery feel free to share the data), but I could add logging to the vacuum clean to see how much of an effect it is having. I think the infra is less flaky than when this was implemented, so I haven't seen tasks running for a long time.

keyonghan:
I think it should be fine to change it to one hour for now. We can always monitor and change if needed.

PR #655, opened 2020-02-20
Make health checks resilient to flutter failures.

godofredoc:
One health check is creating a flutter app and compiling it. If there is
and error with flutter then the agent will quarantine until we manually
checkout a new version of flutter.

https://github.com/flutter/flutter/issues/51085

### DIFF
--- a/agent/lib/src/health.dart
+++ b/agent/lib/src/health.dart
@@ -66,6 +66,7 @@ Future<AgentHealth> performHealthChecks(Agent agent) async {
             } else {
               results['able-to-build-and-sign'] = HealthCheckResult.failure(
                   'Failed to build and sign iOS app.');
+              await getFlutterAt('master');
             }
           });
         } finally {


agent/lib/src/health.dart
```diff
@@ -66,6 +66,7 @@ Future<AgentHealth> performHealthChecks(Agent agent) async {
             } else {
               results['able-to-build-and-sign'] = HealthCheckResult.failure(
                   'Failed to build and sign iOS app.');
+              await getFlutterAt('master');
```
digiter: Sometimes we need to run agent against a specific commit, I'm afraid this line would defect that usage.

godofredoc: This should not affect when running the test using run and using the revision parameter. Also this will run only if the health check fails.

The only way this could affect is if we want to run a test with whatever is already available on $HOME/.cocoon.

digiter: LGTM for this change. I do feel the health check is pretty heavy though.

godofredoc: Let's evaluate if it is still needed and if not we can remove it.

PR #658, opened 2020-02-25
Backend support branching for flutter repo

keyonghan:
This PR supports Flutter to be a branch based build, test and release process.

Right now, `master`, `beta`, and `stable` are supported.


app_dart/lib/src/request_handlers/refresh_github_commits.dart
```diff
@@ -62,47 +63,43 @@ class RefreshGithubCommits extends ApiRequestHandler<Body> {
...
-    const String projectId = 'flutter-dashboard';
-    const String dataset = 'cocoon';
-    const String table = 'Checklist';
-
-    final GitHub github = await config.createGitHubClient();
+    final GithubService githubService = await config.createGithubService();
     const RepositorySlug slug = RepositorySlug('flutter', 'flutter');
-    final Stream<RepositoryCommit> commits =
-        github.repositories.listCommits(slug);
+    const List<String> branches = <String>['master', 'beta', 'stable'];
```
tvolkert: We're going to be using a release branch scheme, where each dev (`x.y.z`) release gets its own branch, and testing is done on those branches.  Then dev, beta, and stable will just point to hashes that have already been tested as part of the release branch.

So rather than enumerating the branches here, you'll want to query GitHub for the list of branches, then filter that list to include only branches that match a specified list of patterns (e.g. ['master', r'v[0-9]+\.[0-9]+\.[0-9]+']).

keyonghan: I see there are existing branches: `vx.x.x-hotfixes` in flutter repo. Shall we exclude those, namely only strictly matching `vx.x.x` pattern?

tvolkert: You can exclude them.  We'll want a way to update the pattern of the branches we match though, since the proposal for those branch names is still in flux.  Maybe a config file that we query so we can update the pattern without needing to update Cocoon?

godofredoc: Should we add a top level folder called release to both flutter/flutter and flutter/engine with a file called branch_regs.txt with one regex per line?

tvolkert: I'd put it somewhere in the `dev/` folder hierarchy.

app_dart/lib/src/request_handlers/refresh_github_commits.dart
```diff
@@ -62,47 +63,47 @@ class RefreshGithubCommits extends ApiRequestHandler<Body> {
...
-    final Stream<RepositoryCommit> commits =
-        github.repositories.listCommits(slug);
+    final Stream<Branch> branches = github.repositories.listBranches(slug);
     final DatastoreService datastore = datastoreProvider();
-    final TabledataResourceApi tabledataResourceApi =
-        await config.createTabledataResourceApi();
-    final List<Commit> newCommits = <Commit>[];
-    final List<Map<String, Object>> tableDataInsertAllRequestRows =
-        <Map<String, Object>>[];
+    final RegExp exp = RegExp(r'v[0-9]+.[0-9]+.[0-9]+');
```
tvolkert: No need to explicitly exclude the existing hotfix branches - just change the regexp to be stricter - r`^rv[0-9]+\.[0-9]+\.[0-9]+$`

app_dart/lib/src/service/github_service.dart
```diff
@@ -27,4 +25,37 @@ class GithubService {
...
+
+  /// Lists the commits of the provided repository [slug] and [branch].
+  Future<List<RepositoryCommit>> listCommits(
+      RepositorySlug slug, String branch) async {
+    const Duration time = Duration(hours: 1);
+    ArgumentError.checkNotNull(slug);
+    final PaginationHelper paginationHelper = PaginationHelper(github);
+
+    final List<dynamic> commits = <dynamic>[];
+    await for (Response response in paginationHelper.fetchStreamed(
```
tvolkert: May be worth sending a PR to `package:github` to add the functionality to query commits on a branch.

app_dart/lib/src/request_handlers/refresh_github_commits.dart
keyonghan: Thanks for the suggestion.

app_dart/lib/src/service/github_service.dart
keyonghan: Will consider sending that soon.

app_dart/lib/src/request_handlers/refresh_github_commits.dart
keyonghan: added.

keyonghan:
> Approach LGTM.
> 
> This should probably also contain a change to the endpoint that drives the dashboard commit status table, so as to only list commits on a specific branch (defaulting to master).

@tvolkert added issue to track: https://github.com/flutter/flutter/issues/51807

app_dart/dev/branch_regexps.txt
```diff
@@ -0,0 +1,2 @@
+^v[0-9]+.[0-9]+.[0-9]+$
```
godofredoc: This file has to go into the different supported repos, e.g flutter/flutter and flutter/engine

tvolkert: That was my first reaction as well, but I couldn't think of a reason why it needs to...  @godofredoc is there a reason why?

app_dart/lib/src/model/appengine/commit.dart
```diff
@@ -50,6 +51,10 @@ class Commit extends Model {
   @StringProperty(propertyName: 'FlutterRepositoryPath', required: true)
   String repository;
 
+  /// The branch of the commit.
+  @StringProperty(propertyName: 'Branch', required: true)
```
tvolkert: How does making this required play with the existing entries that don't have it?

app_dart/lib/src/request_handlers/refresh_github_commits.dart
```diff
@@ -62,47 +63,49 @@ class RefreshGithubCommits extends ApiRequestHandler<Body> {
...
-    final Stream<RepositoryCommit> commits =
-        github.repositories.listCommits(slug);
+    final Stream<Branch> branches = github.repositories.listBranches(slug);
     final DatastoreService datastore = datastoreProvider();
-    final TabledataResourceApi tabledataResourceApi =
-        await config.createTabledataResourceApi();
-    final List<Commit> newCommits = <Commit>[];
-    final List<Map<String, Object>> tableDataInsertAllRequestRows =
-        <Map<String, Object>>[];
+    final File file = File('dev/branch_regexps.txt');
```
tvolkert: This will require a new Cocoon deployment to pick up the new file.  I think we'll have to request it from GitHub over http.

app_dart/dev/branch_regexps.txt
tvolkert: dots should be preceded by backslashes, or the regexp is saying "any character"

keyonghan: right!

app_dart/lib/src/model/appengine/commit.dart
keyonghan: For the safety purpose, removing `required`. Thanks.

app_dart/lib/src/request_handlers/refresh_github_commits.dart
keyonghan: > nit: against.

fixed. Thanks.

keyonghan:
> nit: against.

@godofredoc Fixed. Thanks.

app_dart/lib/src/request_handlers/refresh_github_commits.dart
keyonghan: Considering moving the file to `flutter` or `engine` repo, will change to http request.

app_dart/dev/branch_regexps.txt
godofredoc: One possibility is to support different branches per repo(not sure this is really necessary) the other one will be give more control to the people working on the different repos on which branches will be tested.

The main reason I added this comment is to be able to load it without a new deployment but it will be ok to keep it here and loading it over http.

tvolkert: Yeah living here and requesting over http sgtm

app_dart/lib/src/request_handlers/refresh_github_commits.dart
keyonghan: Done.

app_dart/dev/branch_regexps.txt
keyonghan: Done.

app_dart/lib/src/request_handlers/refresh_github_commits.dart
```diff
@@ -244,4 +275,47 @@ class RefreshGithubCommits extends ApiRequestHandler<Body> {
...
+                'Attempt to download branch_regexps.txt failed (HTTP $status)');
+            return <String>['master'];
+          }
+        } catch (error, stackTrace) {
+          log.error(
+              'Attempt to download branch_regexps.txt failed:\n$error\n$stackTrace');
+        }
+
+        await Future<void>.delayed(gitHubBackoffCalculator(attempt));
+        return <String>['master'];
```
tvolkert: This belongs outside the for loop

app_dart/lib/src/request_handlers/refresh_github_commits.dart
```diff
@@ -244,4 +275,47 @@ class RefreshGithubCommits extends ApiRequestHandler<Body> {
...
+      }
+    } finally {
+      client.close(force: true);
+    }
+
+    log.error('GitHub not responding; giving up');
```
tvolkert: This is unreachable

app_dart/lib/src/request_handlers/refresh_github_commits.dart
keyonghan: Moved outside. Thank you for pointing this out.

app_dart/lib/src/request_handlers/refresh_github_commits.dart
keyonghan: Removed.

PR #659, opened 2020-02-27
Drain subprocess stdout and stderr before throwing an exception

zanderso:
If a subprocess fails quickly, an exception may be thrown before its output can be echoed to the logger. This PR drains the subprocess output before checking the exit code.


zanderso:
@keyonghan @godofredoc How does this get deployed? Is landing the change sufficient?

godofredoc:
This needs manually pulling the change and restarting the agents. Please ping keyonghan@ to update and restart the agents, he is the on call this week.

keyonghan:
Will do

PR #660, opened 2020-02-27
Consider SKIPPED status for cirrus graphql API

keyonghan:
Existing cirrus status logic misses `SKIPPED` status, which causes hang when one task is `SKIPPED`. This will cause PRs not merging forever.

This PR puts `SKIPPED` as a success flag for tasks in cirrus.


keyonghan:
Some context: https://github.com/flutter/flutter/pull/51523

PR #662, opened 2020-03-03
Make task name selectable

cbracken:
Currently the git commit SHA is a SelectableText, but task name is not.
That said, when tasks fail, developers frequently want to search the
source for the task in question in order to debug.


cbracken:
See https://github.com/flutter/cocoon/blob/900c9eb99db0422436bf2d2b1d3f5320875fb7a5/app_flutter/lib/commit_box.dart#L147 for the commit SHA equivalent.

cbracken:
Thanks! I'm okay with selectability for now and once the web engine folks get the clipboard working, we're set to go. In the meantime selecting it will make me feel slightly better even if I can't actually copy it, like one of those elevator close buttons that's not actually wired up :)

PR #663, opened 2020-03-03
Make author email and cocoon agent selectable

cbracken:
In case of task failure, these fields are sometimes useful for pasting
to chat or grepping in source.


cbracken:
Related: #662

cbracken:
I'm okay with us just getting selection enabled now and then when the web engine folks fix the clipboard we'll just get it for free -- unless we think we'll need more work to enable copy/paste other than the engine fix?

PR #666, opened 2020-03-04
Logging for Cirrus queries

Piinks:
There appears to be an issue with the Cirrus GraphQL queries. This adds more logging to help diagnose which query is causing the issue.

### DIFF
--- a/app_dart/lib/src/request_handlers/refresh_cirrus_status.dart
+++ b/app_dart/lib/src/request_handlers/refresh_cirrus_status.dart
@@ -124,7 +124,7 @@ Future<List<dynamic>> queryCirrusGraphQL(
   try {
     final Map<String, dynamic> searchBuilds = result.data['searchBuilds'].first;
     tasks.addAll(searchBuilds['latestGroupTasks']);
-  } catch(_) {
+  } catch (_) {
     log.debug('Did not receive expected result from Cirrus:');
     log.debug(result.data);
   }
--- a/app_dart/lib/src/request_handlers/refresh_cirrus_status.dart
+++ b/app_dart/lib/src/request_handlers/refresh_cirrus_status.dart
@@ -121,7 +121,12 @@ Future<List<dynamic>> queryCirrusGraphQL(
   if (result.data == null) {
     return tasks;
   }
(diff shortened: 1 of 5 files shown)


Piinks:
Thank you! I have formatted and added a little bit more so other checks can be  completed rather than the whole cron job failing. PTAL. :)

app_dart/lib/src/request_handlers/refresh_cirrus_status.dart
```diff
@@ -121,7 +121,12 @@ Future<List<dynamic>> queryCirrusGraphQL(
...
   }
-  final Map<String, dynamic> searchBuilds = result.data['searchBuilds'].first;
-  tasks.addAll(searchBuilds['latestGroupTasks']);
+  try {
+    final Map<String, dynamic> searchBuilds = result.data['searchBuilds'].first;
+    tasks.addAll(searchBuilds['latestGroupTasks']);
+  } catch (_) {
+    log.debug('Did not receive expected result from Cirrus:');
+    log.debug(result.data);
+  }
```
keyonghan: Thank you for adding this catch.

app_dart/lib/src/request_handlers/refresh_cirrus_status.dart
```diff
@@ -121,7 +121,12 @@ Future<List<dynamic>> queryCirrusGraphQL(
   if (result.data == null) {
     return tasks;
```
fkorotkov: Maybe also log if data is missing. 🤔

Piinks: Hey @fkorotkov!
Can do! 
After merging this and checking the logs it looks like there wasn't an issue with the query.
The change that caused the issue was https://github.com/flutter/flutter/pull/51919, which didn't have any cirrus tasks scheduled to run, which is actually working as intended. No cirrus tasks were run for that change since we run tests conditionally, and it didn't change any relevant files to do so. Do you think the query `result.data` should have been null if there wasn't a build?

fkorotkov: `data` can be null if there is `error` so if everything works as expected you should be good.

Glad queries worked as expected!

PR #668, opened 2020-03-05
Support branch parameter for API get-status

keyonghan:
This PR prepares for frontend support to list commits w.r.t. different branches. Default lists are for `master`, otherwise it shows corresponding lists when `branch` parameter is specified.

Related issue: https://github.com/flutter/flutter/issues/51807

### DIFF
--- a/app_dart/index.yaml
+++ b/app_dart/index.yaml
@@ -8,6 +8,11 @@ indexes:
   properties:
   - name: CreateTimestamp
     direction: desc
+- kind: Checklist
+  properties:
+  - name: Branch
+  - name: CreateTimestamp
+    direction: desc
 - kind: Task
   ancestor: yes
   properties:
--- a/app_dart/lib/src/service/datastore.dart
+++ b/app_dart/lib/src/service/datastore.dart
@@ -43,10 +43,13 @@ class DatastoreService {
   /// The [limit] argument specifies the maximum number of commits to retrieve.
   ///
   /// The returned commits will be ordered by most recent [Commit.timestamp].
-  Stream<Commit> queryRecentCommits({int limit = 100, int timestamp}) {
+  Stream<Commit> queryRecentCommits(
+      {int limit = 100, int timestamp, String branch}) {
     timestamp ??= DateTime.now().millisecondsSinceEpoch;
+    branch ??= 'master';
     final Query<Commit> query = db.query<Commit>()
       ..limit(limit)
+      ..filter('branch =', branch)
       ..order('-timestamp')
       ..filter('timestamp <', timestamp);
     return query.run();
--- a/app_dart/test/request_handlers/get_status_test.dart
+++ b/app_dart/test/request_handlers/get_status_test.dart
@@ -158,5 +158,58 @@ void main() {
         'Stages': <String>[]
       });
     });
+
+
+    test('reports statuses with input branch', () async {
+      final Commit commit1 = Commit(
+          key: config.db.emptyKey.append(Commit,
+              id: 'flutter/flutter/ea28a9c34dc701de891eaf74503ca4717019f829'),
+          timestamp: 3,
+          branch: 'master');
+      final Commit commit2 = Commit(
+          key: config.db.emptyKey.append(Commit,
+              id: 'flutter/flutter/d5b0b3c8d1c5fd89302089077ccabbcfaae045e4'),
+          timestamp: 1,
+          branch: 'v0.0.0');
+      config.db.values[commit1.key] = commit1;
+      config.db.values[commit2.key] = commit2;
+      buildStatusProvider = FakeBuildStatusProvider(
+          commitStatuses: <CommitStatus>[
+            CommitStatus(commit1, const <Stage>[]),
+            CommitStatus(commit2, const <Stage>[])
+          ]);
+      handler = GetStatus(
+        config,
+        datastoreProvider: () => DatastoreService(db: config.db),
+        buildStatusProvider: buildStatusProvider,
+      );
+
+      const String expectedLastCommitKeyEncoded =
+          'ahNzfmZsdXR0ZXItZGFzaGJvYXJkckcLEglDaGVja2xpc3QiOGZsdXR0ZXIvZmx1dHRlci9lYTI4YTljMzRkYzcwMWRlODkxZWFmNzQ1MDNjYTQ3MTcwMTlmODI5DA';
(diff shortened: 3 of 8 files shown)


app_dart/lib/src/service/datastore.dart
```diff
@@ -43,10 +43,13 @@ class DatastoreService {
   /// The [limit] argument specifies the maximum number of commits to retrieve.
   ///
   /// The returned commits will be ordered by most recent [Commit.timestamp].
-  Stream<Commit> queryRecentCommits({int limit = 100, int timestamp}) {
+  Stream<Commit> queryRecentCommits(
```
CaseyHillers: Tests for this?

app_dart/test/request_handlers/get_status_test.dart
```diff
@@ -115,11 +117,13 @@ void main() {
...
               id: 'flutter/flutter/ea28a9c34dc701de891eaf74503ca4717019f829'),
-          timestamp: 3);
+          timestamp: 3,
+          branch: 'master');
       final Commit commit2 = Commit(
           key: config.db.emptyKey.append(Commit,
               id: 'flutter/flutter/d5b0b3c8d1c5fd89302089077ccabbcfaae045e4'),
-          timestamp: 1);
+          timestamp: 1,
+          branch: 'master');
```
CaseyHillers: What about a test for when branch isn't given?

app_dart/test/src/request_handling/fake_http.dart
```diff
@@ -563,7 +563,7 @@ class FakeHttpHeaders implements HttpHeaders {
   List<String> operator [](String name) => _values[name];
 
   @override
-  void add(String name, Object value) {
+  void add(String name, Object value, {bool preserveHeaderCase = false}) {
```
CaseyHillers: Was there an update to the http library?

app_dart/test/request_handlers/get_status_test.dart
```diff
@@ -152,7 +156,59 @@ void main() {
...
+        datastoreProvider: () => DatastoreService(db: config.db),
+        buildStatusProvider: buildStatusProvider,
+      );
+
+      const String branch = 'v0.0.0';
+
+      expect(config.db.values.length, 2);
+
+      tester.request = FakeHttpRequest(queryParametersValue: <String, String>{
+        //GetStatus.lastCommitKeyParam: expectedLastCommitKeyEncoded,
```
CaseyHillers: No longer needed?

CaseyHillers:
Rebase with master, https://github.com/flutter/cocoon/pull/670 should have fixed the issues LUCI is showing.

app_dart/index.yaml
```diff
@@ -8,6 +8,11 @@ indexes:
   properties:
   - name: CreateTimestamp
     direction: desc
+- kind: Checklist
```
CaseyHillers: This will not create indexes for previous data. Is there a plan to create these indexes?

Explanation: https://stackoverflow.com/questions/35742786/how-to-add-a-composite-index-in-google-datastore

app_dart/lib/src/service/datastore.dart
keyonghan: Please refer to the last test in `get_status_test.dart`

app_dart/test/request_handlers/get_status_test.dart
keyonghan: Updated the `commit` model to have branch with default value `master.

app_dart/test/src/request_handling/fake_http.dart
keyonghan: Yes.

app_dart/test/request_handlers/get_status_test.dart
keyonghan: removed.

keyonghan:
> Rebase with master, #670 should have fixed the issues LUCI is showing.

Thanks for the fix!

app_dart/index.yaml
tvolkert: In the past we've just had the person who updated the yaml file run the gcloud command to create the index.

app_dart/lib/src/service/datastore.dart
```diff
@@ -43,10 +43,13 @@ class DatastoreService {
...
+      {int limit = 100, int timestamp, String branch}) {
     timestamp ??= DateTime.now().millisecondsSinceEpoch;
+    branch ??= 'master';
     final Query<Commit> query = db.query<Commit>()
       ..limit(limit)
+      ..filter('branch =', branch)
```
tvolkert: Did we back-fill existing data to add 'master' as the branch?

keyonghan:
> Should we add support for the tree status to be based on branch too?

Yes, we should. I think we also need to update the cirrus/luci/devicelab test status logic to be based on branch as well. I updated https://github.com/flutter/flutter/issues/51807 to track
/cc @godofredoc @tvolkert 

This should be fixing https://github.com/flutter/flutter/issues/51884
/cc @christopherfujino

app_dart/lib/src/service/datastore.dart
CaseyHillers: Doesn't have to be this PR due to the priority, but we should start moving some of the datastore logic we do to be tested in datastore_test.dart. datastore_test.dart doesn't even exist :)

app_dart/test/request_handlers/get_status_test.dart
CaseyHillers: Does the datastore data have all tasks currently set to master if they're null? Otherwise wouldn't the model get overwritten from that null in datastore.

keyonghan: I guess you meant `commit`, rather than `task`. Yes, if null, `master` is the default value now. (just pushed the changes)

app_dart/lib/src/service/datastore.dart
tvolkert: I had written a debug handler for updating data at one point for this purpose, but never checked it in for security reasons...

As you say, it's lower priority if the affected rows are old enough to not show on the dashboard by default.

app_dart/index.yaml
keyonghan: As talked about updating old data with 'master' below with Todd, this is of lower priority as long as those data is old enough, which actually is. I am gonna to merge this PR for now.

Created https://github.com/flutter/flutter/issues/52066 to track.

PR #670, opened 2020-03-05
Upgrade Cocoon dependencies

CaseyHillers:
Cocoon LUCI task is throwing issues building `app_dart`. This upgrades the dependencies and any reliances on those dependencies to get it running again.


CaseyHillers:
There's two issues on Flutter dev that Cocoon will have to address
 (1) Find a new way of testing that doesn't rely on colors of widgets
 (2) NetworkImage changes

To make sure breaking changes like this are not introduced, I'll follow up and see if we can get Cocoon added to flutter/tests so that changes like this have to be sent out as breaking changes.

For now, I am setting Flutter to beta. It was set to dev as web support had not made it to beta.

PR #651, opened 2020-02-12
Adding gold status check to framework PRs

Piinks:
This change will add a Gold status check to all framework PRs that reflects the state of the current tryjobs run through Flutter Gold for the pr and commit.

This was originally attempted in https://github.com/flutter/flutter/pull/49370 as a Cirrus check. That implementation was subject to timeouts from Cirrus, so it will be implemented here instead.

Adding this check will do a couple of things for improved workflow and productivity.
- This will streamline framework changes.
  - Currently, framework tests fail when a golden file change is detected. After triaging on the Flutter Gold dashboard, the affected test shards need to be re-executed, which in some cases could take over an hour. A follow-up to this change in the framework would remove the failure case on individual test shards, making this check the gatekeeper. It will notify when a golden file change is detected, and once triaged, will update in 3 minutes or less.
- This will help the engine roll go more smoothly when golden files are affected.
  - Currently, the engine auto-roller will repeatedly close attempts to roll into the framework since golden file shards fail. Without a notification or delay for images to be triaged and shards to be re-triggered, this often results in the need for a manual engine roll to check in golden file changes. This will stop the auto-roller from closing the pull request, leaving this check in a running state and notifying the engine sheriff to triage images on the dashboard. Once the images have been triaged, the auto-roller will proceed. Related issue: https://github.com/flutter/flutter/issues/48744

PR #672, opened 2020-03-05
Performance Improvement for Gold Check

Piinks:
- Noticed in the logs that there was some unnecessary work going on (no need to query cirrus if the last known gold status was completed), rearranged the logic to make it leaner. 
- Also updated the messaging, https://github.com/flutter/flutter/pull/51968 is my validator and the comment wasn't as clear as it could be.
- Moves cron job to every 5 minutes instead of 3
- Re-introduces considering what tests are running to reduce gold checks
  - In #651 this was considered liable to break, but I have implemented it differently now. Instead of only checking framework tests, it is just using the presence of a framework test to indicate that golden file tests are executing for this PR
- Reduced comment noise after initial feedback

### DIFF
--- a/app_dart/lib/src/datastore/cocoon_config.dart
+++ b/app_dart/lib/src/datastore/cocoon_config.dart
@@ -90,7 +90,7 @@ class Config {
       'and make sure this patch meets those guidelines before LGTMing.';
 
   String get goldenBreakingChangeMessage =>
-      'Changes to golden files are considered breaking, so consult '
+      'Changes to golden files are considered breaking changes, so consult '
       '[Handling Breaking Changes](https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes) '
       'to proceed. While there are exceptions to this rule, if this patch modifies '
       'an existing golden file, it is probably not an exception. Only new golden '
--- a/app_dart/lib/src/datastore/cocoon_config.dart
+++ b/app_dart/lib/src/datastore/cocoon_config.dart
@@ -89,13 +89,15 @@ class Config {
       '(https://github.com/flutter/flutter/wiki/Tree-hygiene#how-to-review-code) '
       'and make sure this patch meets those guidelines before LGTMing.';
 
-  String get goldenBreakingChangeMessage => 'Check '
-      '[Handling Breaking Changes](https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes). '
-      'While there are exceptions to this rule, if this patch modifies an existing '
-      'golden file, it is probably not an exception. Only new golden files are not '
-      'considered breaking changes.\n\n'
-      '[Writing a golden file test for `package:flutter`](https://github.com/flutter/flutter/wiki/Writing-a-golden-file-test-for-package:flutter) '
-      'may also provide guidance for this change.\n\n'
+  String get goldenBreakingChangeMessage =>
+      'Changes to golden files are considered breaking, so consult '
+      '[Handling Breaking Changes](https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes) '
+      'to proceed. While there are exceptions to this rule, if this patch modifies '
+      'an existing golden file, it is probably not an exception. Only new golden '
+      'file tests, or downstream changes like those from skia updates are '
+      'considered non-breaking.\n\n'
+      'For more guidance, visit '
+      '[Writing a golden file test for `package:flutter`](https://github.com/flutter/flutter/wiki/Writing-a-golden-file-test-for-package:flutter).\n\n'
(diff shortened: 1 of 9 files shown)


app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
```diff
@@ -59,7 +59,22 @@ class PushGoldStatusToGithub extends ApiRequestHandler<Body> {
...
+      CreateStatus statusRequest;
+
+      log.debug('Last known Gold status for #${pr.number} was with sha: '
+        '${lastUpdate.head}, status: ${lastUpdate.status}');
+
+      if (lastUpdate.status == GithubGoldStatusUpdate.statusCompleted &&
+        lastUpdate.head == pr.head.sha) {
+        log.debug('Completed status already reported for this commit.');
+        // We have already seen this commit and it is completed.
+        continue;
```
Piinks: This is the case for an overwhelming number of PRs.

app_dart/lib/src/datastore/cocoon_config.dart
```diff
@@ -89,13 +89,15 @@ class Config {
...
 
-  String get goldenBreakingChangeMessage => 'Check '
-      '[Handling Breaking Changes](https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes). '
-      'While there are exceptions to this rule, if this patch modifies an existing '
-      'golden file, it is probably not an exception. Only new golden files are not '
-      'considered breaking changes.\n\n'
-      '[Writing a golden file test for `package:flutter`](https://github.com/flutter/flutter/wiki/Writing-a-golden-file-test-for-package:flutter) '
-      'may also provide guidance for this change.\n\n'
+  String get goldenBreakingChangeMessage =>
+      'Changes to golden files are considered breaking, so consult '
```
CaseyHillers: nit: are considered breaking changes,

Piinks:
```suggestion
      'Changes to golden files are considered breaking changes, so consult '
```

Piinks:
I have added a few more changes after touching base with @kjlubick about the amount of traffic on the gold endpoint
@dnfield had some concerns earlier (#651) about checking for framework tests, I've reintroduced this in a different way that is more reliable.
I've also moved the cron job to 5 minutes from 3.

PTAL. :)

Piinks:
I have to fix the formatting, but oddly dartfmt is different on my laptop. 😖

app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
```diff
@@ -90,53 +74,80 @@ class PushGoldStatusToGithub extends ApiRequestHandler<Body> {
...
-          await _commentAndApplyGoldLabel(gitHubClient, pr, slug);
+      log.debug('Querying Cirrus for pull request #${pr.number}...');
+      cirrusCheckStatuses.clear();
+      bool runsGoldenFileTests = false;
+      // Query current checks for this pr.
+      final List<dynamic> cirrusChecks =
+          await queryCirrusGraphQL(pr.head.sha, cirrusClient, log, 'flutter');
+      for (dynamic check in cirrusChecks) {
+        final String status = check['status'];
+        final String taskName = check['name'];
```
dnfield: This is reminding me that we should turn off `avoid_as` and set `implicit-casts: false`.

I was going to suggest some ways to avoid dynamic typing here but we can save that for a PR that just changes those lints around.

app_dart/test/request_handlers/push_gold_status_to_github_test.dart
```diff
@@ -552,63 +645,6 @@ void main() {
           ));
         });
       });
-
-      test(
-          'Completed pull request does not skip follow-up prs with early return',
```
CaseyHillers: To verify, this was removed because this was the performance issue?

Piinks: That was an error. 😅  It's back.

app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
```diff
@@ -131,6 +131,7 @@ class PushGoldStatusToGithub extends ApiRequestHandler<Body> {
                 .createStatus(slug, pr.head.sha, statusRequest);
             lastUpdate.status = statusRequest.state;
             lastUpdate.head = pr.head.sha;
+            print('About to update: ${lastUpdate.pr}, ${lastUpdate.updates}');
```
CaseyHillers: What's the reasoning for switching from using log statements to prints? IIRC AppEngine will still log stdout, but it doesn't assign a log level.

Piinks: Not intentional, just local testing, it ended up getting pushed. I'm considering this WIP until I can find out how I broke a test. :confused:

Piinks:
This is all fixed now. Sorry for the noise. :)

PR #674, opened 2020-03-06
Turn off implicit casts and "avoid as" to prep for NNBD migration

dnfield:
This should result in no semantic/runtime differences,and is required for eventual null-safety migration.


digiter:
Did you use some tool in generating these changes?

dnfield:
Unfortunately it is not at all automated :)

dnfield:
I changed analysis_options.yaml and then fixed all the call sites, sometimes using regexes to find and replace common things.

app_dart/lib/src/model/appengine/key_helper.dart
```diff
@@ -157,9 +157,9 @@ class KeyHelper {
         }
         if (key.id != null) {
           if (key.id is String) {
-            element.name = key.id;
+            element.name = key.id as String;
           } else if (key.id is int) {
-            element.id = Int64(key.id);
+            element.id = Int64(key.id as int);
           }
         }
```
a14n: You can avoid the 2 casts by extracting a `id` local variable.

app_dart/lib/src/model/appengine/key_wrapper.dart
```diff
@@ -43,9 +43,9 @@ class KeyWrapper {
       }
       if (slice.id != null) {
         if (slice.id is String) {
-          current.name = slice.id;
+          current.name = slice.id as String;
         } else if (slice.id is int) {
-          current.uid = Int64(slice.id);
+          current.uid = Int64(slice.id as int);
```
a14n: ditto

app_dart/lib/src/model/appengine/key_helper.dart
dnfield: Done and done, thanks

a14n: Actually you can also remove the surrounding if-not-null

dnfield: That thought occurred to me and then for some reason I didn't do it.  Thanks!

PR #683, opened 2020-03-11
Add transaction retries.

godofredoc:
Most of the 500 errors in the backend are related to failing
transactions. This change adds retries with quadratic backoffs to remove
the flakiness.

Bugs:

https://github.com/flutter/flutter/issues/42524
https://github.com/flutter/flutter/issues/43112
https://github.com/flutter/flutter/issues/49673
https://github.com/flutter/flutter/issues/49672

### DIFF
--- a/app_dart/lib/src/foundation/utils.dart
+++ b/app_dart/lib/src/foundation/utils.dart
@@ -4,15 +4,20 @@
 
 import 'package:gcloud/datastore.dart';
 import 'package:retry/retry.dart';
+import 'package:grpc/grpc.dart';
 
 typedef RetryHandler = Function();
 
 // Runs a db transaction with retries.
 //
 // It uses quadratic backoff starting with 50ms and 3 max attempts.
-Future<void> runTransactionWithRetries(RetryHandler retryHandler) {
-  const RetryOptions r =
-      RetryOptions(delayFactor: Duration(milliseconds: 50), maxAttempts: 3);
-  return r.retry(retryHandler,
-      retryIf: (Exception e) => e is TransactionAbortedError);
+Future<void> runTransactionWithRetries(RetryHandler retryHandler,
+    {int delayMilliseconds = 50, int maxAttempts = 3}) {
+  final RetryOptions r = RetryOptions(
+      delayFactor: Duration(milliseconds: delayMilliseconds),
+      maxAttempts: maxAttempts);
+  return r.retry(
+    retryHandler,
+    retryIf: (Exception e) => e is TransactionAbortedError || e is GrpcError,
+  );
 }
--- a/app_dart/lib/src/foundation/utils.dart
+++ b/app_dart/lib/src/foundation/utils.dart
@@ -1,0 +1,18 @@
+// Copyright 2020 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:gcloud/datastore.dart';
+import 'package:retry/retry.dart';
+
+typedef RetryHandler = Function();
+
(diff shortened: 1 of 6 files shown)


app_dart/lib/src/foundation/utils.dart
```diff
@@ -0,0 +1,18 @@
...
+import 'package:retry/retry.dart';
+
+typedef RetryHandler = Function();
+
+// Runs a db transaction with retries.
+//
+// It uses quadratic backoff starting with 50ms and 3 max attempts.
+Future<void> runTransactionWithRetries(RetryHandler retryHandler) {
+  const RetryOptions r =
+      RetryOptions(delayFactor: Duration(milliseconds: 50), maxAttempts: 3);
```
CaseyHillers: I think we should add these as optional parameters to the function with these as the default values. It could be useful to modify these if there are some operations we want to modify this on.

godofredoc: Done, added them as optional parameters.

godofredoc:
@digiter would you mind expanding on reducing the nesting scope? is it within the new function or nesting in general where the function is being used?

CaseyHillers:
@godofredoc one way I imagine reducing the nesting scope it to have `datastore.dart` implement it's own `withTransaction` function and we switch the Cocoon usage to use that function.

For example this PR has:
```
await runTransactionWithRetries(() async {
          await datastore.db
              .withTransaction<void>((Transaction transaction) async {
            final TimeSeries series =
                await _getOrCreateTimeSeries(transaction, task, scoreKey);
            final num value = resultData[scoreKey] as num;

            final TimeSeriesValue seriesValue = TimeSeriesValue(
              key: series.key.append(TimeSeriesValue),
              createTimestamp: DateTime.now().millisecondsSinceEpoch,
              revision: commit.sha,
              taskKey: task.key,
              value: value.toDouble(),
            );

            transaction.queueMutations(inserts: <TimeSeriesValue>[seriesValue]);
            await transaction.commit();
          });
        });
```

instead we could do:

```
final TimeSeries series = await _getOrCreateTimeSeries(transaction, task, scoreKey);
final num value = resultData[scoreKey] as num;
final TimeSeriesValue seriesValue = TimeSeriesValue(
    key: series.key.append(TimeSeriesValue),
    createTimestamp: DateTime.now().millisecondsSinceEpoch,
    revision: commit.sha,
    taskKey: task.key,
    value: value.toDouble(),
);
await datastore.withTransaction<void>(insert: <TimeSeriesValue>[seriesValue]); // this being the end goal of removing the extra nesting
```

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
```diff
@@ -49,25 +50,26 @@ class RefreshChromebotStatus extends ApiRequestHandler<Body> {
...
+            rethrow;
           }
-          await transaction.commit();
-        } catch (error) {
-          await transaction.rollback();
-          rethrow;
-        }
+        });
       });
     }
```
digiter: @godofredoc I'm talking about reducing nesting in general, for example, this change has 6 nested scopes, which made it a little difficult to read. Also see go/tott/455.

godofredoc: I see what you mean, thanks for the explanation. Those several layers of nested scopes are  everywhere where the backend is using transactions.

Added a new bug for the cleanup as it will require to update most of the handlers:
 
https://github.com/flutter/flutter/issues/52436

godofredoc:
@CaseyHillers thanks for the explanation added flutter/flutter#52436 for a follow up cl on cleaning up the transactional code in the backend.

PR #684, opened 2020-03-11
Add branch to commit key

keyonghan:
This solves https://github.com/flutter/flutter/issues/52443

### DIFF
--- a/app_dart/test/request_handlers/refresh_github_commits_test.dart
+++ b/app_dart/test/request_handlers/refresh_github_commits_test.dart
@@ -80,9 +80,9 @@ void main() {
       return commits;
     }
 
-    Commit shaToCommit(String sha) {
+    Commit shaToCommit(String sha, String branch) {
       return Commit(
-          key: db.emptyKey.append(Commit, id: 'flutter/flutter/$sha'),
+          key: db.emptyKey.append(Commit, id: 'flutter/flutter/$branch/$sha'),
           sha: sha);
     }
 
@@ -145,9 +145,8 @@ void main() {
       httpClient.request.response.body = singleTaskManifestYaml;
       branchHttpClient.request.response.body = branchRegExp;
       await tester.get<Body>(handler);
-      //expect(body.branches, null);
-      final Commit commit = db.values.values.whereType<Commit>().last;
-      //expect(body, null);
+      final Commit commit = db.values.values.whereType<Commit>().first;
+      expect(db.values.values.whereType<Commit>().length, 2);
       expect(commit.branch, 'flutter-1.1-candidate.1');
     });
 
@@ -157,7 +156,7 @@ void main() {
       githubBranches = <String>['master'];
       const List<String> dbCommits = <String>['3', '4', '5', '6'];
       for (String sha in dbCommits) {
-        final Commit commit = shaToCommit(sha);
+        final Commit commit = shaToCommit(sha, 'master');
         db.values[commit.key] = commit;
       }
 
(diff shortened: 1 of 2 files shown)


app_dart/test/request_handlers/refresh_github_commits_test.dart
```diff
@@ -80,9 +80,9 @@ void main() {
       return commits;
     }
 
-    Commit shaToCommit(String sha) {
+    Commit shaToCommit(String sha, String branch) {
       return Commit(
-          key: db.emptyKey.append(Commit, id: 'flutter/flutter/$sha'),
+          key: db.emptyKey.append(Commit, id: 'flutter/flutter/$branch/$sha'),
```
tvolkert: This needs to be rolled out in a way that handles all existing entries nicely.  Otherwise the dashboard will experience pandemonium 🙂.

I'm guessing you'll need to transiently check both the old style and the new style keys, then once it's rolled out, you can update all the old style keys to be new style. Then you can update this here to just handle new style keys.

keyonghan: It does not seem to have any effect to me. Please correct me if I am not correct.

Dashboard gets latest commits status by `time` and `branch` right now, both of which are not related to commit Key (which behaves like a unique Key). Tasks are then queried by checking their parent key, which is equal to commit Key.

This logic is same for both old style and new style. So there doesn't seem to be any conflict between them.

We are changing the commit Key, so the task Key will also be changed accordingly.
https://github.com/flutter/cocoon/blob/master/app_dart/lib/src/request_handlers/refresh_github_commits.dart#L154
https://github.com/flutter/cocoon/blob/master/app_dart/lib/src/request_handlers/refresh_github_commits.dart#L188
Old style's tasks are paired with old commits, whereas new style's tasks will be paired with new commits. From the dashboard's side, it simply lists all commits out, regardless of new or old style.

tvolkert: I was afraid that it would create a bunch of "new" entries for commits that had already been tested, thus flooding the device lab with jobs that had already run -- and in turn making the dashboard look like all tasks were in progress.

If that's not a risk, then carry on 🙂

keyonghan: Understand your concern. I will do some test before moving forward :-)

keyonghan: Tested, and it seems to work fine.
@CaseyHillers Do you think there is any changes needed from frontend if we update the commit key?

CaseyHillers: The frontend doesn't store any data, so this should be fine. Caches may create some inconsistency for the 1-15 minute interval that this rolls out. I don't know if any of our cached data interacts with each other, so if it does you may want to manually clear memorystore when rolling this change out.

keyonghan: @tvolkert Existing logic checks new github commits within past one hour to refresh. If we don't check old Key, then it may insert repeated commits fell in the past one hour.

To avoid that, I just added the old Key check to make sure no single repeated commit inserted. And added a TODO to remove that check after a while.

keyonghan: It has been verified to work as expected: new commits are inserted, dashboard is showing both new and old commits, new commits are with new key format, etc.

tvolkert:
Out of an abundance of caution, you may also want to back up the datastore before rolling this out.

keyonghan:
will do a backup later today.

keyonghan:
A quick update, I have backed up all datastore entities over the weekend. Am gonna to merge and redeploy. Please let me know if any other concerns or comments.

keyonghan:
@tvolkert seems your approval is necessary. Could you please help review/comment/approve? Thank you.

PR #685, opened 2020-03-11
Upgrade to graphql 3.0 for type safety and autocompletion

liyuqian:


liyuqian:
@dnfield @godofredoc : I wonder if you'd like to upgrade graphql to 3.0.0 for a type-safe GraphQL? See, for example, https://github.com/flutter/cocoon/pull/685/commits/ed92c05d056db06a02340a022ace236a51a643ad#diff-f957dc75adaf3b64a6d32b77e02f96e3L196

keyonghan:
I think this would make codes readable friendly with latest fancy features. How are below scripts generated? don't seem to be created manually..
```labeled_pull_requests_with_reviews.ast.gql.dart
labeled_pull_requests_with_reviews.data.gql.dart
labeled_pull_requests_with_reviews.op.gql.dart
labeled_pull_requests_with_reviews.var.gql.dart 
schema.public.ast.gql.dart
schema.public.graphql
schema.public.schema.gql.dart```

**Comment by liyuqian** [comment]
@keyonghan : they're auto generated by `pub run build_runner build`.

File: app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart [hunk 1]
```diff
@@ -6,6 +6,10 @@ import 'dart:async';
 
 import 'package:appengine/appengine.dart';
 import 'package:cocoon_service/src/request_handling/exceptions.dart';
+import 'package:cocoon_service/src/service/github/labeled_pull_requests_with_reviews.data.gql.dart';
+import 'package:cocoon_service/src/service/github/labeled_pull_requests_with_reviews.op.gql.dart';
+import 'package:cocoon_service/src/service/github/labeled_pull_requests_with_reviews.var.gql.dart';
+import 'package:cocoon_service/src/service/github/schema.public.schema.gql.dart' show StatusState, CommentAuthorAssociation, PullRequestReviewState;
```
keyonghan, line 12: How about these import, are they auto added to the file? If not, do we need to run the build first, and then import then here?

dnfield:
Ah sorry this fell off my radar. This looks great.

I think we have to add a step to the CI/build scripts to do the generation so that we're sure we're testing the right code.  

The checked in code also needs a format.

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
```diff
@@ -162,53 +159,41 @@ class CheckForWaitingPullRequests extends ApiRequestHandler<Body> {
...
+            pullRequest.reviews.nodes,
             changeRequestAuthors,
           );
 
-      final String sha = commit['oid'] as String;
-      final List<Map<String, dynamic>> statuses =
-          (commit['status']['contexts'] as List<dynamic>)
-              .cast<Map<String, dynamic>>();
+      final String sha = commit.oid.value;
+      final List<$LabeledPullRequcodeestsWithReviews$repository$labels$nodes$pullRequests$nodes$commits$nodes$commit$status$contexts> statuses = commit.status.contexts;
```
keyonghan, line 196: As for such long type references, not sure if is there any way to have short replacement names? like defining type mapping in a central location, and then refer the short name here.

dnfield:
I think I was reviewing this and something either had a merge conflict or a CI issue - happy to review it once it's ready again.

liyuqian:
@keyonghan @dnfield : I finally fixed all the tests!

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
```diff
@@ -162,50 +162,51 @@ class CheckForWaitingPullRequests extends ApiRequestHandler<Body> {
...
-          (commit['status']['contexts'] as List<dynamic>).cast<Map<String, dynamic>>();
-      List<Map<String, dynamic>> checkRuns;
-      if (commit['checkSuites']['nodes'] != null && (commit['checkSuites']['nodes'] as List<dynamic>).isNotEmpty) {
-        checkRuns =
-            (commit['checkSuites']['nodes']?.first['checkRuns']['nodes'] as List<dynamic>).cast<Map<String, dynamic>>();
-      }
-      checkRuns = checkRuns ?? <Map<String, dynamic>>[];
+      final String sha = commit.oid.value;
+      final List<
+              $LabeledPullRequestsWithReviews$repository$labels$nodes$pullRequests$nodes$commits$nodes$commit$status$contexts>
```
dnfield, line 200: These seem unfortunate. I wonder if there's any setting to have it generate better names than this?

dnfield (commented):
Is there some way that we can verify that our copy of the graphql schema from GitHub is valid?

Should we be pulling it down at CI time and making sure it generates code that still works for us?

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
liyuqian: I asked in the gql discord chat room and the answer seems to be know. For this issue, our best hope is for https://github.com/dart-lang/language/issues/65 to be fixed.

For now, I can also use `// ignore: always_specify_types` to save as many type names as possible. Please see my latest patch.

liyuqian:
> Is there some way that we can verify that our copy of the graphql schema from GitHub is valid?
> 
> Should we be pulling it down at CI time and making sure it generates code that still works for us?

That's a good question. How does the existing graphql query deal with [the breaking schema change][1]? Such change seems to happen once every quarter.

One easy way of doing this might be to just add a CI test that verifies that https://docs.github.com/public/schema.docs.graphql and our schema matches. If not, fail the test and suggest the fix.

[1]: https://developer.github.com/v4/breaking_changes/

dnfield:
Need to make dartfmt ignore the generated files (or format them with 120 line length)

liyuqian:
Formated.

PR #686, opened 2020-03-13
Cocoon getFlutterBranches()

CaseyHillers:
This adds the necessary backend in the Flutter app to get the list of branches from Cocoon. It defaults to showing just the `master` branch until data is loaded. The branch list is not refreshed after the initial load of the app.

# Tests

Tested all possible points of error like the other cocoon service functions are.

Tested that on start of the Flutter build state this `getFlutterBranches()` is called.

# Issues 

https://github.com/flutter/flutter/issues/51807 - Flutter branching for releases

# Future Work

Enable UI in the Flutter app to switch the current branch

### DIFF
--- a/app_flutter/lib/service/appengine_cocoon.dart
+++ b/app_flutter/lib/service/appengine_cocoon.dart
@@ -116,6 +116,28 @@ class AppEngineCocoonService implements CocoonService {
   }
 
   @override
+  Future<CocoonResponse<List<String>>> fetchFlutterBranches() async {
+    final String getBranchesUrl = _apiEndpoint('/api/public/get-branches');
+
+    /// This endpoint returns JSON [List<Agent>, List<CommitStatus>]
+    final http.Response response = await _client.get(getBranchesUrl);
+
+    if (response.statusCode != HttpStatus.ok) {
+      print(response.body);
+      return CocoonResponse<List<String>>()
+        ..error = '/api/public/get-branches returned ${response.statusCode}';
+    }
+
+    try {
+      final List<String> jsonResponse = jsonDecode(response.body);
+      return CocoonResponse<List<String>>()
+        ..data = jsonResponse;
+    } catch (error) {
+      return CocoonResponse<List<String>>()..error = error.toString();
+    }
+  }
+
+  @override
   Future<bool> rerunTask(Task task, String idToken) async {
     assert(idToken != null);
     final String postResetTaskUrl = _apiEndpoint('/api/reset-devicelab-task');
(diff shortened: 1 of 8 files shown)


app_flutter/lib/state/flutter_build.dart
```diff
@@ -58,6 +62,10 @@ class FlutterBuildState extends ChangeNotifier {
   static const String errorMessageFetchingTreeStatus =
       'An error occured fetching tree status from Cocoon';
 
+  @visibleForTesting
+  static const String errorMessageFetchingBranches =
+      'An error occured fetching branches from flutter/flutter on Cocoon.';
+
   /// Start a fixed interval loop that fetches build state updates based on [refreshRate].
   Future<void> startFetchingBuildStateUpdates() async {
```
keyonghan: Nit: good to have a general function name.

app_flutter/lib/state/flutter_build.dart
```diff
@@ -68,6 +76,9 @@ class FlutterBuildState extends ChangeNotifier {
     /// [Timer.periodic] does not necessarily run at the start of the timer.
     _fetchBuildStatusUpdate();
 
+    _fetchFlutterBranches()
+        .then((List<String> branchResponse) => _branches = branchResponse);
```
keyonghan: Is it possible to use different `refreshRate` for these two functions? `branch` shall not be changed frequently.

CaseyHillers: This doesn't use a timer or refresh for `_fetchFlutterBranches`. `_fetchBuildStatusUpdate` has a `Timer.periodic` does whereas the current implementation of `_fetchFlutterBranches` just requests the branches once at start of `FlutterBuildState`.

app_flutter/lib/state/flutter_build.dart
CaseyHillers: Does `startFetchingUpdates()` SG?

keyonghan: SGTM.

app_flutter/lib/service/appengine_cocoon.dart
```diff
@@ -115,6 +115,28 @@ class AppEngineCocoonService implements CocoonService {
...
 
+  @override
+  Future<CocoonResponse<List<String>>> fetchFlutterBranches() async {
+    final String getBranchesUrl = _apiEndpoint('/api/public/get-branches');
+
+    /// This endpoint returns JSON {"Branches": List<String>}
+    final http.Response response = await _client.get(getBranchesUrl);
+
+    if (response.statusCode != HttpStatus.ok) {
+      print(response.body);
```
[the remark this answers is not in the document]
CaseyHillers: IIRC in the past we never integrated a logger service because it was (1) small detail during the EngRes project and (2) Flutter for web didn't necessarily have the plugin support for it. I believe both of these issues have been wiped away so I created https://github.com/flutter/flutter/issues/52697 to track this.

PR #687, opened 2020-03-13
Add skia perf links to Cocoon home page and sidebar

CaseyHillers:
# Issues

Fixes https://github.com/flutter/flutter/issues/52557

### DIFF
--- a/app_flutter/lib/index_page.dart
+++ b/app_flutter/lib/index_page.dart
@@ -117,6 +117,17 @@ class Index extends StatelessWidget {
               SizedBox(
                 width: 300,
                 child: RaisedButton(
+                  child: const Text('Skia Perf'),
+                  onPressed: () => launch('https://flutter-perf.skia.org/'),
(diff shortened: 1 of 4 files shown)


app_flutter/lib/index_page.dart
```diff
@@ -114,6 +114,17 @@ class Index extends StatelessWidget {
                 ),
               ),
               Container(height: 25),
+              SizedBox(
+                width: 300,
+                child: RaisedButton(
+                  child: const Text('Skia Perf'),
```
liyuqian: Nit: "Skia perf" is a little confusing. Maybe call it "Flutter on skia perf"? Once we also imported data from cocoon benchmarks to Flutter's skia perf instance, we may also just call it "Flutter perf" and put our current "Benchmarks" in a less promising position.

CaseyHillers: Updated to "Benchmarks on Skia Perf". Let me know when we can make it "Benchmarks" :)

PR #688, opened 2020-03-16
Support branching for cirrus status

keyonghan:
This is part of https://github.com/flutter/flutter/issues/51807

Instead of assuming only one branch exists for each commit in cirrus, we here consider all branches. When updating cirrus status in cocoon, we consider the branch property as well.

### DIFF
--- a/app_dart/lib/src/foundation/utils.dart
+++ b/app_dart/lib/src/foundation/utils.dart
@@ -39,7 +39,7 @@ Future<List<String>> loadBranchRegExps(
 
   final HttpClient client = branchHttpClientProvider();
   try {
-    // TODO(keyonghan): Apply retry logic here to simply, https://github.com/flutter/flutter/issues/52427
+    // TODO(keyonghan): apply retry logic here to simply, https://github.com/flutter/flutter/issues/52427
     for (int attempt = 0; attempt < 3; attempt++) {
       final HttpClientRequest clientRequest = await client.getUrl(url);
 
--- a/app_dart/lib/src/foundation/utils.dart
+++ b/app_dart/lib/src/foundation/utils.dart
@@ -2,11 +2,74 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'dart:async';
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:appengine/appengine.dart';
 import 'package:gcloud/datastore.dart';
-import 'package:retry/retry.dart';
+import 'package:github/server.dart';
 import 'package:grpc/grpc.dart';
+import 'package:retry/retry.dart';
+
+import '../datastore/cocoon_config.dart';
+import '../foundation/typedefs.dart';
 
 typedef RetryHandler = Function();
+
+/// Signature for a function that calculates the backoff duration to wait in
+/// between requests when GitHub responds with an error.
+///
+/// The [attempt] argument is zero-based, so if the first attempt to request
+/// from GitHub fails, and we're backing off before making the second attempt,
+/// the [attempt] argument will be zero.
+typedef GitHubBackoffCalculator = Duration Function(int attempt);
+
+/// Default backoff calculator.
+Duration twoSecondLinearBackoff(int attempt) {
+  return const Duration(seconds: 2) * (attempt + 1);
+}
+
+Future<List<String>> loadBranchRegExps(
+    HttpClientProvider branchHttpClientProvider,
+    Logging log,
+    GitHubBackoffCalculator gitHubBackoffCalculator) async {
+  const String path = '/flutter/cocoon/master/app_dart/dev/branch_regexps.txt';
+  final Uri url = Uri.https('raw.githubusercontent.com', path);
+
+  final HttpClient client = branchHttpClientProvider();
+  try {
+    for (int attempt = 0; attempt < 3; attempt++) {
+      final HttpClientRequest clientRequest = await client.getUrl(url);
+
+      try {
+        final HttpClientResponse clientResponse = await clientRequest.close();
+        final int status = clientResponse.statusCode;
+
+        if (status == HttpStatus.ok) {
+          final String content = await utf8.decoder.bind(clientResponse).join();
+          final List<String> branches = content
+              .split('\n')
+              .map((String branch) => branch.trim())
+              .toList();
+          branches.removeWhere((String branch) => branch.isEmpty);
+          return branches;
+        } else {
+          log.warning(
+              'Attempt to download branch_regexps.txt failed (HTTP $status)');
+        }
+      } catch (error, stackTrace) {
+        log.error(
+            'Attempt to download branch_regexps.txt failed:\n$error\n$stackTrace');
+      }
+      await Future<void>.delayed(gitHubBackoffCalculator(attempt));
+    }
+  } finally {
+    client.close(force: true);
+  }
+  log.error('GitHub not responding; giving up');
+  return <String>['master'];
+}
 
 // Runs a db transaction with retries.
 //
@@ -20,4 +83,25 @@ Future<void> runTransactionWithRetries(RetryHandler retryHandler,
     retryHandler,
     retryIf: (Exception e) => e is TransactionAbortedError || e is GrpcError,
   );
+}
+
+Future<List<Branch>> getBranches(
+    Config config,
+    HttpClientProvider branchHttpClientProvider,
+    Logging log,
+    GitHubBackoffCalculator gitHubBackoffCalculator) async {
+  final GitHub github = await config.createGitHubClient();
+  const RepositorySlug slug = RepositorySlug('flutter', 'flutter');
+  final Stream<Branch> branchList = github.repositories.listBranches(slug);
+  final List<String> regExps = await loadBranchRegExps(
+      branchHttpClientProvider, log, gitHubBackoffCalculator);
+  final List<Branch> branches = <Branch>[];
+
+  await for (Branch branch in branchList) {
+    if (!regExps.any((String regExp) => RegExp(regExp).hasMatch(branch.name))) {
+      continue;
+    }
+    branches.add(branch);
+  }
+  return branches;
 }
--- a/app_dart/lib/src/foundation/utils.dart
+++ b/app_dart/lib/src/foundation/utils.dart
@@ -39,6 +39,7 @@ Future<List<String>> loadBranchRegExps(
 
   final HttpClient client = branchHttpClientProvider();
   try {
+    // TODO(keyonghan): Apply retry logic here to simply, https://github.com/flutter/flutter/issues/52427
     for (int attempt = 0; attempt < 3; attempt++) {
       final HttpClientRequest clientRequest = await client.getUrl(url);
 
--- a/app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
+++ b/app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
@@ -264,8 +264,16 @@ class CheckForWaitingPullRequests extends ApiRequestHandler<Body> {
     const List<String> _failedStates = <String>['FAILED', 'ABORTED'];
     const List<String> _succeededStates = <String>['COMPLETED', 'SKIPPED'];
     final GraphQLClient cirrusClient = await config.createCirrusGraphQLClient();
-    final List<Map<String, dynamic>> cirrusStatuses =
+    final List<CirrusResult> cirrusResults =
         await queryCirrusGraphQL(sha, cirrusClient, log, name);
+    if (!cirrusResults
+        .any((CirrusResult cirrusResult) => cirrusResult.branch == 'master')) {
+      return allSuccess;
+    }
+    final List<Map<String, dynamic>> cirrusStatuses = cirrusResults
+        .singleWhere(
+            (CirrusResult cirrusResult) => cirrusResult.branch == 'master')
+        .tasks;
     if (cirrusStatuses == null) {
       return allSuccess;
     }
--- a/app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
+++ b/app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
@@ -244,6 +244,9 @@ class CheckForWaitingPullRequests extends ApiRequestHandler<Body> {
     assert(failures != null && failures.isEmpty);
     bool allSuccess = true;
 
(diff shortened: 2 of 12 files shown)


app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
```diff
@@ -78,8 +78,16 @@ class PushGoldStatusToGithub extends ApiRequestHandler<Body> {
       cirrusCheckStatuses.clear();
       bool runsGoldenFileTests = false;
       // Query current checks for this pr.
-      final List<dynamic> cirrusChecks =
+      final List<CirrusResult> cirrusResults =
           await queryCirrusGraphQL(pr.head.sha, cirrusClient, log, 'flutter');
+      if (!cirrusResults.any(
```
Piinks: So this will only apply the `flutter-gold` status to pre-submit checks that will be landing on master?

When Gold is able to support branching in the future, I think we will want the `flutter-gold` check to work on these branched pre-submit checks. Right now individual pre-submit tests with golden files will not fail since they rely on this `flutter-gold` check. 

Can you add a TODO here to explain that?

keyonghan: Yes. It will check only `master` branch based on the above logic.
I see your comment: https://github.com/flutter/flutter/issues/51807#issuecomment-599790674
Do you mind create corresponding issue, and then I will add a TODO pointing to that issue?

Piinks: https://github.com/flutter/flutter/issues/52700 :) Thanks!

keyonghan: TODO added.

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
```diff
@@ -264,8 +264,16 @@ class CheckForWaitingPullRequests extends ApiRequestHandler<Body> {
     const List<String> _failedStates = <String>['FAILED', 'ABORTED'];
     const List<String> _succeededStates = <String>['COMPLETED', 'SKIPPED'];
     final GraphQLClient cirrusClient = await config.createCirrusGraphQLClient();
-    final List<Map<String, dynamic>> cirrusStatuses =
+    final List<CirrusResult> cirrusResults =
         await queryCirrusGraphQL(sha, cirrusClient, log, name);
+    if (!cirrusResults
+        .any((CirrusResult cirrusResult) => cirrusResult.branch == 'master')) {
```
godofredoc: Should we use a branch passed as a parameter instead of fixing it to master?

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
```diff
@@ -264,8 +264,16 @@ class CheckForWaitingPullRequests extends ApiRequestHandler<Body> {
...
+      return allSuccess;
+    }
+    final List<Map<String, dynamic>> cirrusStatuses = cirrusResults
+        .singleWhere(
+            (CirrusResult cirrusResult) => cirrusResult.branch == 'master')
```
godofredoc: Same question as above.

app_dart/lib/src/request_handlers/refresh_cirrus_status.dart
```diff
@@ -37,60 +42,91 @@ class RefreshCirrusStatus extends ApiRequestHandler<Body> {
...
+  final GitHubBackoffCalculator gitHubBackoffCalculator;
 
   @override
   Future<Body> get() async {
     final DatastoreService datastore = datastoreProvider();
     final GraphQLClient client = await config.createCirrusGraphQLClient();
+    final GitHub github = await config.createGitHubClient();
+    const RepositorySlug slug = RepositorySlug('flutter', 'flutter');
+    final Stream<Branch> branches = github.repositories.listBranches(slug);
+    final List<String> regExps = await loadBranchRegExps(
```
godofredoc: I wonder if this logic can be abstracted to a utility function that can be used everywhere(I've seen this logic replicated in several places).

Ideally something like:

List<Branch> branches = await getSupportedBranches();
for (Branch branch in branches) { ...

app_dart/lib/src/request_handlers/refresh_cirrus_status.dart
```diff
@@ -37,60 +42,91 @@ class RefreshCirrusStatus extends ApiRequestHandler<Body> {
...
-        final String taskName = runStatus['name'] as String;
-        log.debug('Found Cirrus build status for $sha: $taskName ($status)');
-        statuses.add(status);
+    await for (Branch branch in branches) {
+      if (!regExps
+          .any((String regExp) => RegExp(regExp).hasMatch(branch.name))) {
+        continue;
       }
+      await for (FullTask task in datastore.queryRecentTasks(
+          taskName: 'cirrus', commitLimit: 15, branch: branch.name)) {
```
godofredoc: Should we add a constant for this limit? it can make it easier to change in the future.

app_dart/lib/src/request_handlers/refresh_cirrus_status.dart
```diff
@@ -37,60 +42,91 @@ class RefreshCirrusStatus extends ApiRequestHandler<Body> {
...
-      } else if (statuses.any(kCirrusInProgressStates.contains)) {
-        newTaskStatus = Task.statusInProgress;
-      } else {
-        newTaskStatus = Task.statusSucceeded;
-        task.task.endTimestamp = DateTime.now().millisecondsSinceEpoch;
-      }
+        log.debug(
+            'Found Cirrus task for branch ${branch.name}, commit $sha, with existing status $existingTaskStatus');
+
+        final List<String> statuses = <String>[];
```
godofredoc: Seems like we have too much logic in a single method, can we split the functionality into multiple methods? This will also simplify tests for specific functionality.

app_dart/lib/src/request_handlers/refresh_cirrus_status.dart
```diff
@@ -37,60 +42,91 @@ class RefreshCirrusStatus extends ApiRequestHandler<Body> {
...
 
-      if (newTaskStatus != existingTaskStatus) {
-        task.task.status = newTaskStatus;
-        await config.db.withTransaction<void>((Transaction transaction) async {
-          transaction.queueMutations(inserts: <Task>[task.task]);
-          await transaction.commit();
-        });
+        if (newTaskStatus != existingTaskStatus) {
+          task.task.status = newTaskStatus;
+          await config.db
```
godofredoc: Maybe add the retries wrapper? everything related with datastore has a high failure rate.

app_dart/test/request_handlers/check_for_waiting_pull_requests_test.dart
```diff
@@ -758,6 +758,7 @@ QueryResult createCirrusQueryResult(List<dynamic> statuses) {
     'searchBuilds': <dynamic>[
       <String, dynamic>{
         'id': '1',
+        'branch': 'master',
```
godofredoc: Can we also tests that used a branch different than master? My assumption is that master is hardcoded everywhere and we may miss some places that need changes if we use master.

app_dart/test/request_handlers/refresh_cirrus_status_test.dart
```diff
@@ -104,20 +150,47 @@ void main() {
       config.db.values[task.key] = task;
 
       expect(task.status, 'New');
+      branchHttpClient.request.response.body = branchRegExp;
       await tester.get(handler);
       expect(task.status, 'In Progress');
     });
+
+    test('skip updating cirrus status when there is no matching branch',
```
godofredoc: What about adding a test for a matching branch different than master?

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
keyonghan: Done.

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
keyonghan: Done.

app_dart/lib/src/request_handlers/refresh_cirrus_status.dart
keyonghan: Rewrite this part as suggested. Now multiple APIs call this shared logic defined in utils.dart. Additionally merged the /request_handler/utils.dart with /foundation/utils.dart.

app_dart/lib/src/request_handlers/refresh_cirrus_status.dart
keyonghan: Added.

app_dart/lib/src/request_handlers/refresh_cirrus_status.dart
keyonghan: Added two sub functions: _getStatuses and _getNewTaskStatus to simply.

app_dart/lib/src/request_handlers/refresh_cirrus_status.dart
keyonghan: Added.

app_dart/test/request_handlers/check_for_waiting_pull_requests_test.dart
keyonghan: The reason we fix `master` here is that `check for waiting pull requests` applied to `master` only. As suggested, I have added one more test to consider the case with a different branch name, for which we will ignore cirrus statues.

app_dart/test/request_handlers/refresh_cirrus_status_test.dart
keyonghan: Added.

app_dart/lib/src/foundation/utils.dart
```diff
@@ -2,12 +2,75 @@
...
+Future<List<String>> loadBranchRegExps(
+    HttpClientProvider branchHttpClientProvider,
+    Logging log,
+    GitHubBackoffCalculator gitHubBackoffCalculator) async {
+  const String path = '/flutter/cocoon/master/app_dart/dev/branch_regexps.txt';
+  final Uri url = Uri.https('raw.githubusercontent.com', path);
+
+  final HttpClient client = branchHttpClientProvider();
+  try {
+    for (int attempt = 0; attempt < 3; attempt++) {
```
godofredoc: We can simplify this function removing the for loop and the try..catch by using "retry" with retryIf.

app_dart/lib/src/foundation/utils.dart
```diff
@@ -2,12 +2,75 @@
...
+              .split('\n')
+              .map((String branch) => branch.trim())
+              .toList();
+          branches.removeWhere((String branch) => branch.isEmpty);
+          return branches;
+        } else {
+          log.warning(
+              'Attempt to download branch_regexps.txt failed (HTTP $status)');
+        }
+      } catch (error, stackTrace) {
```
godofredoc: Is is possible to use specific errors rather than catching all?

app_dart/lib/src/foundation/utils.dart
```diff
@@ -2,12 +2,75 @@
...
+        log.error(
+            'Attempt to download branch_regexps.txt failed:\n$error\n$stackTrace');
+      }
+      await Future<void>.delayed(gitHubBackoffCalculator(attempt));
+    }
+  } finally {
+    client.close(force: true);
+  }
+  log.error('GitHub not responding; giving up');
+  return <String>['master'];
```
godofredoc: This can potentially hide errors, I'd suggest throwing an error rather than logging and then defaulting to master.

app_dart/lib/src/request_handlers/refresh_cirrus_status.dart
```diff
@@ -37,60 +42,99 @@ class RefreshCirrusStatus extends ApiRequestHandler<Body> {
...
+            transaction.queueMutations(inserts: <Task>[task.task]);
+            await transaction.commit();
+          });
         });
       }
     }
     return Body.empty;
   }
+
+  String _getNewTaskStatus(List<String> statuses, FullTask task) {
```
godofredoc: Some docs will be great.

app_dart/lib/src/request_handlers/refresh_cirrus_status.dart
```diff
@@ -37,60 +42,99 @@ class RefreshCirrusStatus extends ApiRequestHandler<Body> {
...
+    } else if (statuses.any(kCirrusInProgressStates.contains)) {
+      newTaskStatus = Task.statusInProgress;
+    } else {
+      newTaskStatus = Task.statusSucceeded;
+      task.task.endTimestamp = DateTime.now().millisecondsSinceEpoch;
+    }
+    return newTaskStatus;
+  }
+
+  List<String> _getStatuses(
```
godofredoc: Some docs will be great.

app_dart/lib/src/foundation/utils.dart
keyonghan: This is a TODO as mentioned in the above issue: https://github.com/flutter/flutter/issues/52427. Since this PR is big already, do you think it would be good to finish this in a separate PR later?

app_dart/lib/src/request_handlers/refresh_cirrus_status.dart
keyonghan: added.

app_dart/lib/src/request_handlers/refresh_cirrus_status.dart
keyonghan: added.

app_dart/lib/src/foundation/utils.dart
godofredoc: Can we add the TODO here to ensure we don't forget about it?

app_dart/lib/src/foundation/utils.dart
keyonghan: ditto.

app_dart/lib/src/foundation/utils.dart
keyonghan: ditto.

app_dart/lib/src/foundation/utils.dart
keyonghan: Added, and pointed to the above issue.

keyonghan:
> LGTM after adding TODO to simplify loadBranchRegExps

Thank you for all the detailed comments!

PR #689, opened 2020-03-16
Workaround fix for performance benchmarks  dashboard with limited data

keyonghan:
This fixes: https://github.com/flutter/flutter/issues/52692

### DIFF
--- a/app_dart/lib/src/service/datastore.dart
+++ b/app_dart/lib/src/service/datastore.dart
@@ -54,6 +54,17 @@ class DatastoreService {
       ..filter('timestamp <', timestamp);
     return query.run();
   }
+  // Queries for recent commits without considering branches.
+  // TODO(keyonghan): combine this function with the above `queryRecentCommits`, 
+  // this needs to fix https://github.com/flutter/flutter/issues/52694.
+  Stream<Commit> queryRecentCommitsNoBranch({int limit = 100, int timestamp}) {
+    timestamp ??= DateTime.now().millisecondsSinceEpoch;
+    final Query<Commit> query = db.query<Commit>()
+      ..limit(limit)
+      ..order('-timestamp')
+      ..filter('timestamp <', timestamp);
+    return query.run();
+  }
 
   /// queryRecentTimeSerialsValues fetches the latest benchmark results starting from
   /// [startFrom] and up to a given [limit].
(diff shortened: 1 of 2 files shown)


app_dart/lib/src/service/datastore.dart
```diff
@@ -55,6 +55,18 @@ class DatastoreService {
     return query.run();
   }
 
+  // Queries for recent commits without considering branches.
+  // TODO(keyonghan): combine this function with the above `queryRecentCommits`,
+  // this needs to fix https://github.com/flutter/flutter/issues/52694.
+  Stream<Commit> queryRecentCommitsNoBranch({int limit = 100, int timestamp}) {
```
liyuqian: Are 100 commits enough? I remember that Casey once extended the memcache from 3 months of commits to 6 months of commits to properly triage some performance issues. If there are 10 commits a day, that's a limit of 1800 commits.

keyonghan: That's the default number for this function. The actual limit for history API is 6000.
https://github.com/flutter/cocoon/blob/c2bf4d635f65dc9203e621a23504ec9b35eeeebf/app_dart/lib/src/request_handlers/get_timeseries_history.dart#L42

liyuqian: I see. LGTM!

CaseyHillers: To clarify, the issue @liyuqian : There was an update to the Datastore records to use a new index for the branching release strategy in Flutter. This an unintended side effect as Cocoon never saved previous records with a branch until the past few weeks. This adds a new query that doesn't use the branch field (at the moment there are very few timeseries datapoints that will impact the perf graphs). 

The long term solution is going to be to update old timeseries data to have a master branch (as Cocoon only processed data on master before).

liyuqian: Thanks @CaseyHillers for the explanation. And thank you @keyonghan for a quick fix!

liyuqian:
Shall I merge it now? Is there any extra step to get the fix deployed?

CaseyHillers:
@liyuqian there's some issues in Flutter dev that created issues for the Cocoon project where it can't be deployed. Keyong is working on resolving them first.

keyonghan:
@liyuqian Just deployed. You should be able to see the data now.

PR #690, opened 2020-03-17
Add force-deploy and help flags to deploy script

CaseyHillers:
There was an issue this past week in deploying where the latest Flutter dev channel broke Cocoon. To prevent these breakages is WIP and being tracked in https://github.com/flutter/flutter/issues/52147

This allows us to pass --force-deploy to ignore Flutter not being a recent version from the past 3 weeks.

Added help dialog to have the deploy script document itself.

PR #692, opened 2020-03-17
Fix agent lint issue

CaseyHillers:
https://github.com/flutter/cocoon/pull/691 and https://github.com/flutter/cocoon/pull/690 (and any new PRs to Cocoon) are blocked due to some lint issue with the agent:

```
  error - The type 'Iterable<dynamic>' used in the 'for' loop must implement Iterable with a type argument that can be assigned to 'String' at lib/src/utils.dart:636:31 - (for_in_of_invalid_element_type)
```

I believe this is due to the Dart tests using the latest dev channel, and this issue is now rolled into Cocoon.


CaseyHillers:
Good catch, trying that. Followed up with https://github.com/flutter/flutter/issues/52771 to update the agent lint to be current with flutter/flutter.

PR #693, opened 2020-03-18
Explicitly use the master branch in build dashboard

CaseyHillers:
Create `currentBranch` in `FlutterBuildState` and plumbed it through everywhere that uses it.

Refactored some of the test code to remove some of the duplication spread throughout it and fixed some spacing issues.

# Tests

Added tests for `apiEndpoint` refactor to handle query parameters.

Updated tests to check for defaulting to the `master` branch.

# Issues 

https://github.com/flutter/flutter/issues/51807 - Flutter branching for releases

# Future work

Create an `updateBranch` method in `FlutterBuildState` to change `currentBranch` to update the UI.

### DIFF
--- a/app_flutter/lib/state/flutter_build.dart
+++ b/app_flutter/lib/state/flutter_build.dart
@@ -51,7 +51,7 @@ class FlutterBuildState extends ChangeNotifier {
   List<String> _branches = <String>['master'];
   List<String> get branches => _branches;
 
-  String _currentBranch = 'master';
+  final String _currentBranch = 'master';
   String get currentBranch => _currentBranch;
 
   /// A [ChangeNotifer] for knowing when errors occur that relate to this [FlutterBuildState].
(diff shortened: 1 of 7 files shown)


app_flutter/lib/state/flutter_build.dart
```diff
@@ -51,6 +51,9 @@ class FlutterBuildState extends ChangeNotifier {
   List<String> _branches = <String>['master'];
   List<String> get branches => _branches;
 
+  final String _currentBranch = 'master';
+  String get currentBranch => _currentBranch;
```
keyonghan: Is this used somewhere? I see `_currentBranch` is being used all the time.

CaseyHillers: The followup PR to this will use this variable directly. I am going to leave this in to maintain consistency of the build state variables.

keyonghan: SGTM.

CaseyHillers:
Added the test for specific branches. Good catch! Fixed some groups that were nested in another group.

PR #696, opened 2020-03-18
Update branch name for gold status

keyonghan:
This PR changes the default branch name from `master` to `pull/${pr.name}`.

### DIFF
--- a/app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
+++ b/app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
@@ -82,13 +82,13 @@ class PushGoldStatusToGithub extends ApiRequestHandler<Body> {
           await queryCirrusGraphQL(pr.head.sha, cirrusClient, log, 'flutter');
 
       // TODO(katelovett): flutter gold will need to support branching, https://github.com/flutter/flutter/issues/52700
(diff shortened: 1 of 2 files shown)


app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
```diff
@@ -82,13 +82,13 @@ class PushGoldStatusToGithub extends ApiRequestHandler<Body> {
           await queryCirrusGraphQL(pr.head.sha, cirrusClient, log, 'flutter');
 
       // TODO(katelovett): flutter gold will need to support branching, https://github.com/flutter/flutter/issues/52700
-      if (!cirrusResults.any(
-          (CirrusResult cirrusResult) => cirrusResult.branch == 'master')) {
+      if (!cirrusResults.any((CirrusResult cirrusResult) =>
+          cirrusResult.branch == 'pull/${pr.number}')) {
         continue;
```
Piinks: Can you add a `log.debug(...);` here specifying that the PR is being skipped since no valid CirrusResult was found? That should make it easier to diagnose if there is ever an issue like this again.

keyonghan: Added.

PR #697, opened 2020-03-19
Update analysis_options

Hixie:


app_flutter/lib/state/agent.dart
```diff
@@ -81,7 +81,7 @@ class AgentState extends ChangeNotifier {
           .fetchAgentStatuses()
           .then((CocoonResponse<List<Agent>> response) {
         if (response.error != null) {
-          print(response.error);
+          print(response.error); // TODO(ianh): report this rather than printing it
```
CaseyHillers: I believe these TODOs should also link to https://github.com/flutter/flutter/issues/52697 (I don't know if flutter/cocoon has that enabled for the linter)

Hixie: Oh interesting I was imagining they'd just get logged to a snackbar or some such. Logging seems good too though.

Hixie: updated the TODO, will commit on green

Hixie:
Analysis failed because dartfmt tried to reformat the lines I made more readable; I have a PR to switch us to 120 character lines (https://github.com/flutter/cocoon/pull/699) which should be a reasonable compromise between the readability we can get without a formatter and the consistency we can get with a formatter.

PR #698, opened 2020-03-19
Update CommitStatus to have branch field

CaseyHillers:
This lets the frontend track data based on branch.

# Issues

flutter/flutter#51807 - Flutter branching for releases

# Tests

app_dart tests pass

app_flutter appengine json updated to match latest which includes the branch field

# Future work

Use this to invalidate previous branch data to show the currently selected branch.

### DIFF
--- a/app_flutter/test/service/appengine_cocoon_test.dart
+++ b/app_flutter/test/service/appengine_cocoon_test.dart
@@ -26,6 +26,7 @@ const String jsonGetStatsResponse = '''
           "Checklist": {
             "Key": "iamatestkey", 
             "Checklist": {
+              "Branch": "master",
               "FlutterRepositoryPath": "flutter/cocoon", 
               "CreateTimestamp": 123456789, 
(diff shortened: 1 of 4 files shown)


app_flutter/test/service/appengine_cocoon_test.dart
```diff
@@ -26,6 +26,7 @@ const String jsonGetStatsResponse = '''
           "Checklist": {
             "Key": "iamatestkey", 
             "Checklist": {
+              "Branch": "master",
               "FlutterRepositoryPath": "flutter/cocoon", 
               "CreateTimestamp": 123456789, 
               "Commit": {
```
keyonghan: Nit: shall we use single quote for all of them?

CaseyHillers: JSON doesn't send single quotes, only double quotes. This is supposed to be raw JSON.

PR #699, opened 2020-03-19
Move to a 120 character line length

Hixie:
`dartfmt` output is not great at 120, but it's better than at 80 at least.


CaseyHillers:
\cc @keyonghan we may want to consider updating app_dart and agent to use 120 character line limits instead of 80 too

Hixie:
Thanks!

PR #700, opened 2020-03-20
Branch dropdown menu on build dashboard

CaseyHillers:
This adds a dropdown button to the build dashboard appbar that allows for switching the current branch. It adds updates `FlutterBuildState` to be able to change the branch that data is shown from.

It currently erases previous branch data when switching to a new branch.

# Open Questions

I feel `moreStatusesExist` can have a better name. I am looking for a variable name that says the build state knows that more commit statuses exist from the backend, and they have to be loaded over the network. In certain cases (such as branches not master), there is only a few commits and Cocoon should not keep trying the infinite load approach

# Issues

flutter/flutter#51807 - Flutter branching for releases

# Tests

Tests for update branch and new behavior.

Widget tests to check the dropdown exists

# Preview

https://testchillers-dot-flutter-dashboard.appspot.com/#/build


CaseyHillers:
/cc @Hixie if you have time, can you take a look at this? It's for adding a dropdown to the build dashboard to change branches.

I'm running into an issue where I can't get the text color to be white when the dropdown is idle, and have it be black when the dropdown menu is open.

app_flutter/lib/build_dashboard_page.dart
```diff
@@ -119,6 +119,29 @@ class BuildDashboard extends StatelessWidget {
           title: statusTable[buildState.isTreeBuilding],
           backgroundColor: colorTable[buildState.isTreeBuilding],
           actions: <Widget>[
+            DropdownButton<String>(
+                value: buildState.currentBranch,
+                icon: Icon(
+                  Icons.arrow_downward,
+                  color: Colors.white,
```
Hixie: in theory the color should be picked up automatically by the ambient theme, something like how i made it happen for the SIGN IN button. cc @HansMuller for advice.

app_flutter/lib/build_dashboard_page.dart
```diff
@@ -119,6 +119,29 @@ class BuildDashboard extends StatelessWidget {
...
+                onChanged: (String branch) {
+                  buildState.updateCurrentBranch(branch);
+                },
+                items: buildState.branches.map<DropdownMenuItem<String>>((String value) {
+                  return DropdownMenuItem<String>(
+                    value: value,
+                    child: Text(value, style: TextStyle(color: Colors.black)),
+                  );
+                }).toList()),
+            Container(width: 10), // Padding between branches and sign in
```
Hixie: `SizedBox(width: 10)` is more efficient

Hixie:
I'm still learning the codebase, but LGTM.

app_flutter/lib/build_dashboard_page.dart
CaseyHillers: Thanks! Added.

app_flutter/lib/build_dashboard_page.dart
CaseyHillers: Thanks! I am going to submit on green to enable testing of the branching release strategy. Will follow up with smoothing this out next week.

PR #703, opened 2020-03-20
Do not remove .pub-cache as it is required by the agent.

godofredoc:
If we delete .pub-cache the agent execution is affected even if we
immediately run pub get. Pub-cache will need to be cleaned out as part
of the agent update.

https://github.com/flutter/flutter/issues/52957


godofredoc:
I tried removing and then running pub get right after but it fails. I'm working on a different PR that will allow the agent to auto-update and reboot daily. I'll plug the logic to remove .pub-cache there.

PR #704, opened 2020-03-20
Bring the index page more in line with the Material spec.

Hixie:
...and lots of random cleanup, including:
 * Have the Drawer automatically figure out what page is up
 * Test that the dashboard and the index page buttons match
 * Prettier logo in drawer
 * Show about/licenses

### DIFF
--- a/app_flutter/lib/index_page.dart
+++ b/app_flutter/lib/index_page.dart
@@ -83,76 +83,65 @@ class Index extends StatelessWidget {
 
   final GlobalKey<ScaffoldState> scaffoldKey;
 
+  static const Widget separator = SizedBox(height: 24.0);
+
   @override
   Widget build(BuildContext context) {
+    final ThemeData theme = Theme.of(context);
     return Consumer<IndexState>(
-      builder: (_, IndexState indexState, Widget child) => Scaffold(
+      builder: (BuildContext context, IndexState indexState, Widget child) => Scaffold(
         key: scaffoldKey,
         appBar: AppBar(
           title: const Text('Cocoon'),
           actions: <Widget>[
-            SignInButton(authService: indexState.authService),
+            SignInButton(
+              authService: indexState.authService,
+              colorBrightness: theme.appBarTheme.brightness ?? theme.primaryColorBrightness,
+            ),
           ],
         ),
         body: Center(
-          child: Column(
-            children: <Widget>[
-              Container(height: 50),
-              SizedBox(
-                width: 300,
-                child: RaisedButton(
-                  child: const Text('Build'),
+          child: IntrinsicWidth(
+            stepWidth: 80.0,
+            child: Column(
+              crossAxisAlignment: CrossAxisAlignment.stretch,
+              children: <Widget>[
+                separator,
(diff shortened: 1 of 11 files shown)


app_flutter/test/navigation_drawer_test.dart
```diff
@@ -114,7 +119,7 @@ void main() {
         ),
       );
 
-      const String skiaPerfText = 'Benchmarks on Skia Perf';
+      const String skiaPerfText = 'Skia Perf';
```
CaseyHillers: +@liyuqian mentioned Skia perf was unclear in https://github.com/flutter/cocoon/pull/687

app_flutter/lib/index_page.dart
```diff
@@ -83,76 +83,65 @@ class Index extends StatelessWidget {
...
   final GlobalKey<ScaffoldState> scaffoldKey;
 
+  static const Widget separator = SizedBox(height: 24.0);
+
   @override
   Widget build(BuildContext context) {
+    final ThemeData theme = Theme.of(context);
     return Consumer<IndexState>(
-      builder: (_, IndexState indexState, Widget child) => Scaffold(
+      builder: (BuildContext context, IndexState indexState, Widget child) => Scaffold(
```
CaseyHillers: Q: Is it better practice to leave the variable in instead of ignoring them?

app_flutter/lib/main.dart
```diff
@@ -9,14 +9,18 @@ import 'build_dashboard_page.dart';
...
 import 'service/google_authentication.dart';
 
-void main() => runApp(MyApp());
+void main() => runApp(const MyApp());
 
-final ThemeData theme = ThemeData(
-  appBarTheme: const AppBarTheme(color: Colors.green),
-  primarySwatch: Colors.blue,
+final ThemeData theme = ThemeData.from(
+  colorScheme: ColorScheme.light(
```
CaseyHillers: Dark mode dashboard? 👀

app_flutter/test/navigation_drawer_test.dart
Hixie: I'm happy to change it back. I was just trying to avoid the word Benchmarks twice but I don't feel strongly about it.

app_flutter/lib/index_page.dart
Hixie: In general I would encourage it just because it makes it easier to read the code (you don't have to wonder what the argument is), but in this case I added it back just because I needed it. :-)

See also: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#avoid-anonymous-parameter-names

app_flutter/lib/main.dart
Hixie: We can add support for dark mode if you want, it's pretty easy these days. :-)

Hixie:
Updated PR with "Benchmarks for Skia Perf" and dark mode support.

app_flutter/lib/index_page.dart
Hixie: (to clarify: i needed it, then later took out the code that needed it)

PR #711, opened 2020-03-23
Factor out the AppBar widget

Hixie:
That way we don't need to keep setting the color on each page.

### DIFF
--- a/app_flutter/lib/app_bar.dart
+++ b/app_flutter/lib/app_bar.dart
@@ -1,0 +1,39 @@
+// Copyright (c) 2019 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter/material.dart';
+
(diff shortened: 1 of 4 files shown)


app_flutter/lib/app_bar.dart
```diff
@@ -0,0 +1,39 @@
...
+  Size get preferredSize => Size.fromHeight(kToolbarHeight);
+
+  @override
+  Widget build(BuildContext context) {
+    final ThemeData theme = Theme.of(context);
+    return AppBar(
+      title: title,
+      backgroundColor: backgroundColor,
+      actions: <Widget>[
+        ...?actions,
```
CaseyHillers: TIL!

CaseyHillers: (about the ...? operator)

Hixie: not my favourite part of the language (i argued that it should be `actions?...`) but yup.

PR #714, opened 2020-03-23
Split the AgentHealthDetailsBar out into its own file

Hixie:
Eventually I would like to move all the widgets into their own directory
and all the data objects into their own directory. This is one of the few
files that has both in one file.


CaseyHillers:
nit: There's a test in agent_health_details that checks this widget, so that should be moved to an agent_health_details_bar_test file too

Hixie:
done. will land on green unless you have more comments!

PR #715, opened 2020-03-24
Remove dependencies on DateTime.now

Hixie:
Having DateTime.now dependencies in logic has two problems:

- It causes the view to become stale over time (because nothing
  changes when the time changes, so e.g. "10 minutes ago" stops being
  accurate).

- It increases the chance of flakes in tests (because any time you run
  the tests, you're running them with new previously untested data,
  and because if there's multiple places in the code checking the
  current time, they race each other).

Instead this provides the time as an inherited widget.

### DIFF
--- a/app_flutter/lib/agent_health_details.dart
+++ b/app_flutter/lib/agent_health_details.dart
@@ -24,39 +24,32 @@ import 'package:cocoon_service/protos.dart' show Agent;
 /// ```
 class AgentHealthDetails {
   factory AgentHealthDetails(Agent agent) {
-    final DateTime currentTime = DateTime.now();
-    final DateTime agentTime = DateTime.fromMillisecondsSinceEpoch(agent.healthCheckTimestamp.toInt());
-    final Duration agentLastUpdateDuration = currentTime.difference(agentTime);
-
     final String healthDetails = agent.healthDetails;
     return AgentHealthDetails._(
       agent,
+      DateTime.fromMillisecondsSinceEpoch(agent.healthCheckTimestamp.toInt()),
       healthDetails.contains(_hasHealthyDevices),
       healthDetails.contains(_cocoonAuthentication),
       healthDetails.contains(_cocoonConnection),
       healthDetails.contains(_ableToPerformhealthCheck),
       healthDetails.contains(_hasSshConnectivity),
-      agentLastUpdateDuration.inMinutes < minutesUntilAgentIsUnresponsive,
     );
   }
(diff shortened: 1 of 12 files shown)


Hixie:
https://github.com/flutter/cocoon/pull/714 should land first; this will be easier to review once https://github.com/flutter/cocoon/pull/714 has landed.

Hixie:
@CaseyHillers this is ready to review now

app_flutter/lib/agent_health_details.dart
```diff
@@ -70,6 +63,17 @@ class AgentHealthDetails {
   /// The agent to show health details from.
   final Agent agent;
 
+  /// The date and time of the last time this agent pinged Cocoon to show it is
+  /// still responsive and able to take tasks.
+  ///
+  /// Agents that become unresponsive tend to be unresponsive because a process
```
CaseyHillers: nitty nit: there's a lot of "unresponsive" in this sentence

app_flutter/lib/now.dart
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  void _tick() {
+    value = DateTime.now();
+    _scheduleTick();
+    notifyListeners();
+  }
+
+  void _scheduleTick() {
+    _timer = Timer(Duration(milliseconds: 1000 - (value.millisecondsSinceEpoch % 1000)), _tick);
```
CaseyHillers: nit: This took some time to understand. Can you document it?

app_flutter/test/utils/wrapper.dart
```diff
@@ -2,9 +2,11 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'package:flutter/material.dart';
```
CaseyHillers: Thank you for bringing order to this file! :)

app_flutter/lib/agent_health_details.dart
Hixie: i mean, you're not wrong...

fixed.

app_flutter/test/utils/wrapper.dart
Hixie: well now that i know what convention we're following... :-)

app_flutter/lib/now.dart
Hixie: done

Hixie:
thanks for the review, will land on green

PR #716, opened 2020-03-24
Fix a regression introduced by https://github.com/flutter/cocoon/pull/707

Hixie:


Hixie:
FWIW, you can tell github to give you a whitespace-only diff (see the cog at the top of the Files tab).

Hixie:
(i am working on the error in the tests, not sure what's going on yet...)

Hixie:
This PR now has two commits, the first is a straight restoration of #707, the second is the patch to make it work. I somewhat accidentally also addressed another issue in that patch (the Column/Expanded thing), which ended up being the bulk of the actual diff. The only part of the patch that's actually fixing the regression #707 introduced is switching from Consumer to AnimatedBuilder, because #707 changed how the state is introduced (it used to be a ChangeNotifier-based provider, now it's just a regular value-based provider, so that you can fetch it without getting notified of every change).

Hixie:
Landing with the earlier approval, let me know if you want additional changes.

PR #719, opened 2020-03-24
[cocoon] Adds datastore wrapper class.

godofredoc:
This class will be used to simplify datastore transactional code.

Bugs:

https://github.com/flutter/flutter/issues/53148
https://github.com/flutter/flutter/issues/49671
https://github.com/flutter/flutter/issues/52436
https://github.com/flutter/flutter/issues/52427


app_dart/lib/src/datastore/datastore_wrapper.dart
```diff
@@ -0,0 +1,77 @@
...
+
+/// Wrapper class for datastore operations.
+///
+/// Datastore operations generate high rates of errors related to
+/// Transaction and/or GRPC problems. Fixing this required to add retries
+/// to all the database operations causing 4 or 5 levels of nesting. This
+/// class was created to simplify all the database related operations to a
+/// single nesting level in most cases or two at most while still providing
+/// retry logic.
+class DatastoreWrapper {
```
CaseyHillers: I thought `datastore.dart` is already the wrapper class for Datastore code. What's the reasoning for having these in a separate file?

godofredoc: datastore.dart is the dart implementation for AppEngine datastore. This wrapper class is to provide a consistent retry on failure behavior to the cocoon backend.

This class will also be used to simplify the transactional code everywhere in the cocoon backend removing the several nesting level required if you are using datastore.dart directly.

This is part 1 of many other PRs that will simplify the code in the handlers.

app_dart/lib/src/datastore/datastore_wrapper.dart
```diff
@@ -0,0 +1,77 @@
...
+
+  /// Runs a function inside a transaction providing a [Transaction] parameter.
+  Future<T> withTransaction<T>(Future<T> Function(Transaction) handler) async {
+    T result;
+    await runTransactionWithRetries(() async {
+      await _db.withTransaction<void>((Transaction transaction) async {
+        result = await handler(transaction);
+      });
+    });
+    return result;
```
keyonghan: seems all transaction errors happen on 'insert' operation. will this lookup wrapper add extra cost when querying? This comment is based on lookup doesn't hit the transaction error.

app_dart/lib/src/datastore/datastore_wrapper.dart
CaseyHillers: `import 'package:gcloud/db.dart';` is the Dart implementation of Datastore. `datastore.dart` is Cocoon's wrapper for that library. Granted, there is datastore code sprinkled throughout the codebase which should be moved to `datastore.dart`.

It seems more idiomatic to call `datastore.insert(...)` where `datastore` already exists in the codebase versus adding a new variable and calling it like `datastoreWrapper.insert(...)`.

godofredoc: @CaseyHillers Not sure which datastore are you referring to, the one that is everywhere in the backend code comes from import 'package:gcloud/datastore.dart'; and all the handlers are using Config.db rather than datastore directly.

@keyonghan why we'll need to change the queries, sharding the inserts was already happening in most places where more than 1 model was being inserted, would you mind expanding on that?

app_dart/lib/src/datastore/datastore_wrapper.dart
godofredoc: There won't be overhead as retries only happen if there is a transaction or grpc error.

app_dart/lib/src/datastore/datastore_wrapper.dart
keyonghan: Oh, what I meant is changing the reference point from datastore to the wrapper for all APIs.

godofredoc: yes, I made all those changes locally but will send in separate reviews to simplify the reviews and minimize the impact during the releases.

CaseyHillers: FYI `package:gcloud/db.dart` is just another layer on top of `package:gcloud/datastore.dart`, so they're just the same code :)
https://github.com/dart-lang/gcloud/blob/master/lib/db.dart

CaseyHillers: To clarify my earlier point about this change being more idiomatic in the DatastoreService, see https://github.com/flutter/cocoon/blob/master/app_dart/lib/src/request_handlers/update_benchmark_targets.dart

```
    /// ...
    await datastore.db.commit(inserts: <TimeSeries>[timeSeries]); // could be changed to just datastore.commit(...);
```

godofredoc: To clarify what you are suggesting is to move the content of datastore_wrapper to service/datastore.dart and remove the new variable and db from Config. Is that correct?

CaseyHillers: Yes

godofredoc: Code moved to service/datastore.

app_dart/lib/src/datastore/datastore_wrapper.dart
godofredoc: @keyonghan added tests for these functions. There will be a follow up PR to migrate the code to remove several layers of nested code.

app_dart/lib/src/foundation/utils.dart
```diff
@@ -73,20 +68,6 @@ Future<List<String>> loadBranchRegExps(
   return <String>['master'];
 }
 
-// Runs a db transaction with retries.
-//
-// It uses quadratic backoff starting with 50ms and 3 max attempts.
-Future<void> runTransactionWithRetries(RetryHandler retryHandler,
```
CaseyHillers, line 79: nit: Why was this moved out of utils.dart? I thought this was a great addition, and was hoping to import it into the CacheRequestHandler to reuse the logic there.

godofredoc: I moved the function to ./services/datastore.dart without changes. I think that's a better place for it because it is checking for errors in the datastore transactions.

I have a follow up PR to move all the existing code to use the wrappers to keep this PR as short as possible.

app_dart/lib/src/datastore/datastore_wrapper.dart
keyonghan: Thanks for adding tests for datastore!

PR #722, opened 2020-03-25
Make AgentState automatically start fetching data when listened.

Hixie:
This means we don't have to explicitly tell it to start fetching.


Hixie:
If you like this I can do the same for build state.

Hixie:
cc @CaseyHillers

PR #723, opened 2020-03-26
Make FlutterBuildStatus autotrigger on listen

Hixie:
...rather than having an explicit init step

### DIFF
--- a/app_flutter/lib/state/flutter_build.dart
+++ b/app_flutter/lib/state/flutter_build.dart
@@ -27,30 +27,21 @@ class FlutterBuildState extends ChangeNotifier {
   /// Authentication service for managing Google Sign In.
   final GoogleSignInService authService;
 
-  /// How often to query the Cocoon backend for the current build state.
-  @visibleForTesting
-  final Duration refreshRate = const Duration(seconds: 10);
-
-  /// Timer that calls [_fetchBuildStatusUpdate] on a set interval.
-  @visibleForTesting
-  @protected
-  Timer refreshTimer;
-
-  /// The current status of the commits loaded.
-  List<CommitStatus> _statuses = <CommitStatus>[];
-  List<CommitStatus> get statuses => _statuses;
-
-  /// Whether or not flutter/flutter currently passes tests.
-  bool _isTreeBuilding;
-  bool get isTreeBuilding => _isTreeBuilding;
-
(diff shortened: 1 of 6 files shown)


Hixie:
cc @CaseyHillers same as the previous PR but for build state.

app_flutter/lib/state/flutter_build.dart
```diff
@@ -73,69 +64,101 @@ class FlutterBuildState extends ChangeNotifier {
...
-    _fetchBuildStatusUpdate();
+  /// Timer that calls [_fetchStatusUpdates] on a set interval.
+  @visibleForTesting
+  @protected
+  Timer refreshTimer;
 
-    _fetchFlutterBranches().then((List<String> branchResponse) => _branches = branchResponse);
+  // There's no way to cancel futures in the standard library so instead we just track
+  // if we've been disposed, and if so, we drop everything on the floor.
+  bool _active = true;
```
CaseyHillers: Obviously we implement this with Streams then :P

app_flutter/lib/state/flutter_build.dart
```diff
@@ -73,69 +64,101 @@ class FlutterBuildState extends ChangeNotifier {
...
-      cocoonService.fetchCommitStatuses(branch: _currentBranch).then((CocoonResponse<List<CommitStatus>> response) {
+  Future<void> _fetchStatusUpdates([Timer timer]) async {
+    await Future.wait<void>(<Future<void>>[
+      () async {
+        final CocoonResponse<List<String>> response = await cocoonService.fetchFlutterBranches();
+        if (!_active) {
+          return null;
+        }
+        if (response.error != null) {
+          _errors.send('$errorMessageFetchingBranches: ${response.error}');
```
CaseyHillers: the error brook handles notifying it's listeners, right? (I saw the notifyListener was brought into the else logic block)

app_flutter/lib/state/flutter_build.dart
Hixie: Or some sort of cancelable API, yeah. We could. Not sure it's really worth it though. At the end of the day this boils down to a network request and it's not clear how much we win from canceling that.

app_flutter/lib/state/flutter_build.dart
Hixie: Yeah, the error brook has a different set of listeners than the state.

PR #724, opened 2020-03-26
Clean up some widgets to use Align instead of Column

Hixie:


Hixie:
cc @CaseyHillers A simple one for once. :-)

PR #725, opened 2020-03-26
Remove redundant navigator observer

Hixie:


Hixie:
cc @CaseyHillers as discussed earlier

PR #726, opened 2020-03-26
Tests should not have randomness.

Hixie:
It introduces the risk of flakes.

### DIFF
--- a/app_flutter/test/status_grid_test.dart
+++ b/app_flutter/test/status_grid_test.dart
@@ -2,16 +2,17 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'dart:math' as math;
+
+import 'package:cocoon_service/protos.dart' show Commit, CommitStatus, Stage, Task;
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_test/flutter_test.dart';
 
-import 'package:cocoon_service/protos.dart' show Commit, CommitStatus, Stage, Task;
-
+import 'package:app_flutter/commit_box.dart';
 import 'package:app_flutter/service/cocoon.dart';
 import 'package:app_flutter/service/dev_cocoon.dart';
 import 'package:app_flutter/state/flutter_build.dart';
-import 'package:app_flutter/commit_box.dart';
 import 'package:app_flutter/state_provider.dart';
 import 'package:app_flutter/status_grid.dart';
 import 'package:app_flutter/task_box.dart';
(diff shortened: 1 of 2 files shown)


Hixie:
cc @CaseyHillers another simple one

app_flutter/test/status_grid_test.dart
```diff
@@ -2,21 +2,22 @@
...
 import 'package:app_flutter/service/cocoon.dart';
 import 'package:app_flutter/service/dev_cocoon.dart';
 import 'package:app_flutter/state/flutter_build.dart';
-import 'package:app_flutter/commit_box.dart';
 import 'package:app_flutter/state_provider.dart';
 import 'package:app_flutter/status_grid.dart';
 import 'package:app_flutter/task_box.dart';
 import 'package:app_flutter/task_icon.dart';
 import 'package:app_flutter/task_matrix.dart' show TaskMatrix;
+import 'package:cocoon_service/protos.dart' show Commit, CommitStatus, Stage, Task;
```
CaseyHillers: nit: Why was this moved?

Hixie: Isn't cocoon_service from the same repo as app_flutter? Maybe I misunderstood the convention. I thought it was dart: packages, then foreign packages, then packages from your own project, then local imports. I'm happy to use whatever order you want. :-)

CaseyHillers: They're both in the same repo. I always had a hard time telling if these imports should be grouped together or not (if only there was a linter that would tell me 😃). I default to leaving the app_dart imports above app_flutter since it's outside of the flutter codebase.

CaseyHillers: This nit was for me to probe and see if there is a set way of grouping these imports :)

Hixie: so the order you want is:

1. dart: imports
1. package: imports except package:app_flutter
1. package:app_flutter (only in tests)
1. local imports

Is that right?

Hixie: I've changed it to that, let me know if you want something else.

CaseyHillers: The complex ordering style I tried to follow in app_flutter was:

1. dart imports
**2. package: imports excluding cocoon_service and app_flutter**
**3. package: imports of cocoon_service**
4. package imports of app_flutter (only in tests)
5. local imports

Your current way SGTM. I wanted to give insight to why it might look a certain way in the codebase.

Hixie: i'm happy to use whatever pattern you want :-)

I'll try to do the five block style you describe here. I'll document it in the README.

PR #730, opened 2020-03-31
Add v1.12.13-hotfixes to cocoon branch list

jmagman:


app_dart/dev/branch_regexps.txt
```diff
@@ -1,2 +1,3 @@
 ^flutter-[0-9]+\.[0-9]+-candidate\.[0-9]+$
+v1.12.13-hotfixes
```
keyonghan, line 2: `\` may be needed before `.`

jmagman: https://regex101.com/r/IdQ3GF/1

PR #732, opened 2020-04-03
Refactor reserve_task handler to use datastore service.

godofredoc:
Bugs:

flutter/flutter#42524
flutter/flutter#43112
flutter/flutter#49673
flutter/flutter#49672

### DIFF
--- a/app_dart/lib/src/service/reservation_provider.dart
+++ b/app_dart/lib/src/service/reservation_provider.dart
@@ -1,0 +1,50 @@
+// Copyright 2020 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:cocoon_service/src/model/appengine/task.dart';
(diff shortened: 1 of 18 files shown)


app_dart/lib/src/service/reservation_provider.dart
```diff
@@ -0,0 +1,50 @@
...
+// found in the LICENSE file.
+
+import 'package:cocoon_service/src/model/appengine/task.dart';
+import 'package:cocoon_service/src/service/datastore.dart';
+
+/// Function signature for a [ReservationService] provider.
+typedef ReservationServiceProvider = ReservationService Function(
+    DatastoreService datastoreService);
+
+class ReservationService {
```
keyonghan, line 12: Shall we combine this `reservation_provider` with the `task_provider`?

godofredoc: That's the way it used to be I split them to simplify the refactoring and to rewrite them as service + provider to make them consistent with the other code.

PR #734, opened 2020-04-03
Refresh github commits of release branches

keyonghan:
This fixes: https://github.com/flutter/flutter/issues/53711

### DIFF
--- a/app_dart/lib/src/datastore/cocoon_config.dart
+++ b/app_dart/lib/src/datastore/cocoon_config.dart
@@ -128,6 +128,12 @@ class Config {
   /// The default number of commit shown in flutter build dashboard.
   int get commitNumber => 30;
 
+  /// The default hours to look back for github commits refresh - existing branch.
+  int get existingBranchHours => 1;
+
+  /// The default hours to look back for github commits refresh - new branch.
+  int get newBranchHours => 168;
+
   // TODO(keyonghan): update all existing APIs to use this reference, https://github.com/flutter/flutter/issues/48987.
   KeyHelper get keyHelper =>
       KeyHelper(applicationContext: context.applicationContext);
--- a/app_dart/lib/src/request_handlers/refresh_github_commits.dart
+++ b/app_dart/lib/src/request_handlers/refresh_github_commits.dart
@@ -61,6 +61,7 @@ class RefreshGithubCommits extends ApiRequestHandler<Body> {
         config, branchHttpClientProvider, log, gitHubBackoffCalculator);
 
     for (Branch branch in branches) {
+      // TODO(keyonghan): use time of latest commit for [hours] in github API query, https://github.com/flutter/flutter/issues/54284
       int hours = config.existingBranchHours;
 
       if (branch.name != 'master') {
--- a/app_dart/lib/src/request_handlers/refresh_github_commits.dart
+++ b/app_dart/lib/src/request_handlers/refresh_github_commits.dart
@@ -61,8 +61,21 @@ class RefreshGithubCommits extends ApiRequestHandler<Body> {
         config, branchHttpClientProvider, log, gitHubBackoffCalculator);
 
     for (Branch branch in branches) {
+      int hours = config.existingBranchHours;
+
+      if (branch.name != 'master') {
+        final Set<Commit> commits = await datastore
+            .queryRecentCommits(limit: 1, branch: branch.name)
+            .toSet();
+
+        /// Set [hours] to be longer if a new branch is found.
+        if (commits.isEmpty) {
+          hours = config.newBranchHours;
+        }
+      }
+
       final List<RepositoryCommit> commits =
-          await githubService.listCommits(slug, branch.name);
+          await githubService.listCommits(slug, branch.name, hours, config);
       final List<Commit> newCommits =
           await _getNewCommits(commits, datastore, branch.name);
 
--- a/app_dart/test/request_handlers/refresh_github_commits_test.dart
+++ b/app_dart/test/request_handlers/refresh_github_commits_test.dart
@@ -50,6 +50,7 @@ void main() {
     List<String> githubCommits;
     List<String> githubBranches;
     int yieldedCommitCount;
+    int newBranchHours;
 
     Stream<Branch> branchStream() async* {
       for (String branchName in githubBranches) {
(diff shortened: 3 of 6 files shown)


app_dart/test/request_handlers/refresh_github_commits_test.dart
```diff
@@ -173,6 +178,18 @@ void main() {
...
+      githubBranches = <String>['flutter-0.0-candidate.0'];
+      config.newBranchHoursValue = 168;
+
+      expect(db.values.values.whereType<Commit>().length, 0);
+      httpClient.request.response.body = singleTaskManifestYaml;
+      branchHttpClient.request.response.body = branchRegExp;
+      await tester.get<Body>(handler);
+      expect(db.values.values.whereType<Commit>().length, 1);
+    });
+
```
CaseyHillers, line 192: What happens in the case that a hotfix branch is created and it has several commits cherry picked to it?

keyonghan: My understanding is that the release branch is created from master first (this case it will check the latest commit only), and then cherry picks happen later (this case we will check all commits picked up).

Does the case Casey mentioned above happen @tvolkert ?

keyonghan: cc @godofredoc

app_dart/lib/src/request_handlers/refresh_github_commits.dart
```diff
@@ -61,8 +61,21 @@ class RefreshGithubCommits extends ApiRequestHandler<Body> {
         config, branchHttpClientProvider, log, gitHubBackoffCalculator);
 
     for (Branch branch in branches) {
+      int hours = config.existingBranchHours;
+
+      if (branch.name != 'master') {
+        final Set<Commit> commits = await datastore
+            .queryRecentCommits(limit: 1, branch: branch.name)
```
godofredoc, line 68: Any reason to limit it to 1? this can potentially limit processing one commit per cron job iteration.

app_dart/lib/src/request_handlers/refresh_github_commits.dart
```diff
@@ -61,8 +61,21 @@ class RefreshGithubCommits extends ApiRequestHandler<Body> {
...
+            .toSet();
+
+        /// Set [hours] to be longer if a new branch is found.
```
godofredoc, line 71: Is this to trigger tests when the branch is created? if that is the case is there a way to get the commit the branch was created on and using it as the start point rather than going back several days in the query.

app_dart/lib/src/request_handlers/refresh_github_commits.dart
```diff
@@ -61,8 +61,21 @@ class RefreshGithubCommits extends ApiRequestHandler<Body> {
...
+        if (commits.isEmpty) {
+          hours = config.newBranchHours;
+        }
+      }
+
       final List<RepositoryCommit> commits =
-          await githubService.listCommits(slug, branch.name);
+          await githubService.listCommits(slug, branch.name, hours, config);
```
godofredoc, line 78: Using hours may lead to a race condition e.g the cocoon backend goes down for 1:15 mins then when it gets active it will just query 1 hour window not processing the commits that happened between current_time - 1 hour 15 mins to current_time - 1 hour.

app_dart/test/request_handlers/refresh_github_commits_test.dart
godofredoc: Added some comments along the different files

app_dart/lib/src/datastore/cocoon_config.dart
```diff
@@ -128,6 +128,12 @@ class Config {
   /// The default number of commit shown in flutter build dashboard.
   int get commitNumber => 30;
 
+  /// The default hours to look back for github commits refresh - existing branch.
+  int get existingBranchHours => 1;
```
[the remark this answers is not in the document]
keyonghan: every 1 min.

app_dart/lib/src/datastore/cocoon_config.dart
```diff
@@ -128,6 +128,12 @@ class Config {
...
+
+  /// The default hours to look back for github commits refresh - new branch.
+  int get newBranchHours => 168;
```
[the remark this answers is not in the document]
keyonghan: This is used to distinguish whether a new branch is released. Related issue: https://github.com/flutter/flutter/issues/53711

When a new branch is released from `master`, it should pick up the latest one without pulling in all (those older ones should have been tested in `master` branch).

If branches already exist in datastore, all detected commits are subsequent new ones on those released branches. For this case, we need to test all of them.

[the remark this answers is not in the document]
keyonghan: btw: could you comment on https://github.com/flutter/cocoon/pull/734#discussion_r403382313? Todd seems to be offline these days. Based on what @jmagman commented in issue: https://github.com/flutter/flutter/issues/53711, the latest single commit is enough.

app_dart/lib/src/request_handlers/refresh_github_commits.dart
keyonghan: Please refer to above comment. The case when limiting 1 happens only when a new release branch is detected for the first time.

app_dart/lib/src/request_handlers/refresh_github_commits.dart
keyonghan: Thanks for pointing this out. Days may sound more reasonable.

app_dart/lib/src/request_handlers/refresh_github_commits.dart
keyonghan: From github API https://developer.github.com/v3/repos/commits/#list-commits-on-a-repository, `since` seems the only parameter we can use.

app_dart/lib/src/datastore/cocoon_config.dart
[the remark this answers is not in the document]
jmagman: I think when the branch is first detected only the most recent commit would be needed.

app_dart/lib/src/request_handlers/refresh_github_commits.dart
godofredoc: Not a blocker for this PR but we can optimize this logic by selecting the last processed commit from the datastore, getting its datetime and grabbing only the commits after that date time. Can you please open a bug for that and add a TODO here?

keyonghan: Sounds good! Created: https://github.com/flutter/flutter/issues/54284

PR #735, opened 2020-04-03
Reorganize the files from the main lib/ directory

Hixie:
I've moved the non-top-level files from lib/:

* Widgets in widgets/

* Logic in logic/

In addition, I renamed "FlutterBuildState" to "BuildState" since that's what we called it in a bunch of places anyway and it was confusing.

Also I made all the imports follow the scheme now described in the README.


app_flutter/README.md
```diff
@@ -6,4 +6,51 @@
...
+
+The `lib` directory is organized in subdirectories as follows:
+
+* `lib/`: main.dart and top-level widgets (e.g. the build dashboard
+  widget). Can import anything.
+
+* `lib/widgets/`: Custom widgets used in this project. The files in
+  this directory should import either the Flutter widgets package or
+  the Flutter material package, and may import any local files except
+  those in the top-level `lib/` directory.
```
CaseyHillers, line 24: To clarify, is this saying it can only imports files from `lib/widgets` or is saying do not import files from only the top level of  `lib/`?

app_flutter/README.md
```diff
@@ -6,4 +6,51 @@
...
+  services. Files in this directory should not import any Flutter
+  packages other than foundation, and should not import local files
+  that relate to UI (e.g. those in `widgets/`).
+  
+* `lib/service/`: Services, objects that interface with the network.
+  Files in this directory should not import any Flutter packages other
+  than foundation, and should not import any local files from other
+  directories.
+  
+* `lib/logic/`: Other code, mainly non-UI utility methods and data
```
CaseyHillers, line 36: TIL utils is an unclear name :)

Hixie: well we already have a utils in the tests directory, so... :-)

app_flutter/README.md
Hixie: tried to clarify

Hixie:
Thanks for the review, will land on green.

PR #736, opened 2020-04-06
Refactor handlers to use service/datastore.

godofredoc:
* Simplifies transactional code.
* Pass datastore parameters explicitly.
* Adds retries mechanism to all the handlers.

Bugs:

  flutter/flutter#42524
  flutter/flutter#43112
  flutter/flutter#49673
  flutter/flutter#49672

### DIFF
--- a/app_dart/lib/src/request_handlers/append_log.dart
+++ b/app_dart/lib/src/request_handlers/append_log.dart
@@ -17,6 +17,7 @@ import '../request_handling/api_request_handler.dart';
 import '../request_handling/authentication.dart';
 import '../request_handling/body.dart';
 import '../request_handling/exceptions.dart';
+import '../service/datastore.dart';
 import '../service/stackdriver_logger.dart';
 
 @immutable
@@ -24,6 +25,8 @@ class AppendLog extends ApiRequestHandler<Body> {
   AppendLog(
     Config config,
     AuthenticationProvider authenticationProvider, {
+    @visibleForTesting
+        this.datastoreProvider = DatastoreService.defaultProvider,
     StackdriverLoggerService stackdriverLogger,
     @visibleForTesting Uint8List requestBodyValue,
   })  : stackdriverLogger =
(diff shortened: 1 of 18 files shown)


app_dart/lib/src/request_handlers/reset_devicelab_task.dart
```diff
@@ -34,31 +34,26 @@ class ResetDevicelabTask extends ApiRequestHandler<Body> {
...
+    final Task task = await datastore.lookupByValue(key);
+    if (task == null) {
+      throw const BadRequestException('Invalid key. Entity does not exist.');
+    }
+
+    if (task.status == Task.statusInProgress) {
+      throw const BadRequestException(
+          'Not allowed to restart task in progress.');
+    }
+    task.reason = '';
```
CaseyHillers, line 53: nit: Dart offers cascading
```
task..reason = ''
    ..status = Task.statusNew
    ..reservedForAgentId = '';
```

app_dart/lib/src/request_handlers/update_task_status.dart
```diff
@@ -95,44 +96,30 @@ class UpdateTaskStatus extends ApiRequestHandler<UpdateTaskStatusResponse> {
...
-              value: value.toDouble(),
-            );
-
-            transaction.queueMutations(inserts: <TimeSeriesValue>[seriesValue]);
-            await transaction.commit();
-          });
-        });
-      }
+    if (newStatus != Task.statusSucceeded || scoreKeys.isEmpty) {
+      return UpdateTaskStatusResponse(task);
```
CaseyHillers, line 106: When does this occur? At a glance it's hard to tell why we exit early here (other than to maintain the existing logic), and it would be good to document it.

app_dart/lib/src/request_handlers/append_log.dart
```diff
@@ -34,11 +37,14 @@ class AppendLog extends ApiRequestHandler<Body> {
...
 
   final StackdriverLoggerService stackdriverLogger;
+  final DatastoreServiceProvider datastoreProvider;
 
   static const String ownerKeyParam = 'ownerKey';
 
   @override
   Future<Body> post() async {
+    final DatastoreService datastore = datastoreProvider(
+        db: config.db, maxEntityGroups: config.maxEntityGroups);
```
CaseyHillers, line 47: With the refactor, I don't think `maxEntityGroups` needs to live in the `CocoonConfig`. It could be moved into `DatastoreService` and the `maxEntityGroups` parameter could be removed.

godofredoc: I still need to update a couple of handlers. I'll move the maxEntityGroups to datastore once all the handlers are using it and I'll remove it from CocoonConfig.

app_dart/lib/src/request_handlers/update_task_status.dart
godofredoc: Actually I just rewrote the logic using the datastore object but looking this part closely seems like the intent was an optimization but in reality it is duplicating code. If scoreKeys is empty the code inside the loop does not execute and goes directly to UpdateTaskStatusResponse.

Removed those lines.

PR #738, opened 2020-04-08
Branch fix - Github API rate limit exceeded

keyonghan:
This PR fixes https://github.com/flutter/flutter/issues/54293

This is a temporary fix. Will create a follow up PR to save branch list info in datastore, and retrieve it from cocoon config. This should get rid of github API rate limit.

### DIFF
--- a/app_dart/lib/src/foundation/utils.dart
+++ b/app_dart/lib/src/foundation/utils.dart
@@ -75,7 +75,7 @@ Future<List<String>> getBranchList(
     GitHubBackoffCalculator gitHubBackoffCalculator) async {
   final List<String> regExps = await loadBranchRegExps(
       branchHttpClientProvider, log, gitHubBackoffCalculator);
-  // TODO(keyonghan): save this list in cocoon config datastore, https://github.com/flutter/flutter/issues/54297
+  // TODO(keyonghan): save this list in the cocoon config datastore, https://github.com/flutter/flutter/issues/54297
   const List<String> branchList = <String>[
     'master',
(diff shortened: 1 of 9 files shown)


app_dart/lib/src/foundation/utils.dart
```diff
@@ -68,6 +68,39 @@ Future<List<String>> loadBranchRegExps(
   return <String>['master'];
 }
 
+Future<List<String>> getBranchList(
+    Config config,
+    HttpClientProvider branchHttpClientProvider,
+    Logging log,
+    GitHubBackoffCalculator gitHubBackoffCalculator) async {
+  final List<String> regExps = await loadBranchRegExps(
+      branchHttpClientProvider, log, gitHubBackoffCalculator);
```
godofredoc, line 77: Is this a temporary fix? we probably should avoid hardcoding branch names.

godofredoc: If it is a temporary fix please add a todo with the bug number to remove it.

keyonghan: Added that in this PR description. I will have a follow up PR soon.

godofredoc: I think is better to add todos and bug number to the code. Is difficult to keep track of them from PR descriptions.

keyonghan: Yeah. Just added: https://github.com/flutter/flutter/issues/54297

PR #739, opened 2020-04-08
Save flutter branch list to cocoon config in datastore to reduce github query

keyonghan:
This PR fixes https://github.com/flutter/flutter/issues/54297.

### DIFF
--- a/app_dart/lib/src/datastore/cocoon_config.dart
+++ b/app_dart/lib/src/datastore/cocoon_config.dart
@@ -60,6 +60,9 @@ class Config {
 
   DatastoreDB get db => _db;
 
+  // TODO(keyonghan): [FlutterBranches] should be refreshed periodically, https://github.com/flutter/flutter/issues/54315
+  Future<String> get flutterBranches => _getSingleValue('FlutterBranches');
+
   Future<String> get oauthClientId => _getSingleValue('OAuthClientId');
 
   Future<String> get githubOAuthToken => _getSingleValue('GitHubPRToken');
--- a/app_dart/test/request_handlers/get_branches_test.dart
+++ b/app_dart/test/request_handlers/get_branches_test.dart
@@ -60,7 +60,7 @@ void main() {
     });
 
     test('returns branches matching regExps', () async {
-      githubBranches = <String>['flutter-1.1-candidate.1', 'master', 'test'];
+      config.flutterBranchesValue = 'flutter-1.1-candidate.1,master,test';
 
       branchHttpClient.request.response.body = branchRegExp;
 
--- a/app/lib/benchmark/benchmark_grid.dart
+++ b/app/lib/benchmark/benchmark_grid.dart
@@ -7,7 +7,6 @@ import 'dart:convert' show json;
 import 'dart:html';
 
 import 'package:angular/angular.dart';
-import 'package:angular_components/angular_components.dart';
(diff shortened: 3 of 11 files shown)


app_dart/lib/src/datastore/cocoon_config.dart
```diff
@@ -60,6 +60,9 @@ class Config {
 
   DatastoreDB get db => _db;
 
+  // TODO(keyonghan): [FlutterBranches] should be refreshed periodically, https://github.com/flutter/flutter/issues/54315
+  Future<String> get flutterBranches => _getSingleValue('FlutterBranches');
+
```
godofredoc, line 65: Seems like branching management is getting more and more complicated have you considered adding a branching service + provider class?

app_dart/test/request_handlers/get_branches_test.dart
```diff
@@ -60,7 +60,7 @@ void main() {
     });
 
     test('returns branches matching regExps', () async {
-      githubBranches = <String>['flutter-1.1-candidate.1', 'master', 'test'];
+      config.flutterBranchesValue = 'flutter-1.1-candidate.1,master,test';
```
godofredoc, line 63: Any reasons for not keeping the list of strings? I think branches are more readable as a list rather than a comma separated list.

app_dart/lib/src/datastore/cocoon_config.dart
keyonghan: Actually it is not. 
Multiple APIs (around 5) are reading Github branch API via cronjobs every one to five minutes. That would be more than one thousand reads per day.
Now, with this info. saved in datastore, we only need to refresh that property via a cronjob. This cronjob can be called every hour, or even less frequently, since release branches are generated every couple of days.

app_dart/test/request_handlers/get_branches_test.dart
keyonghan: Let me try saving that info as an array.

keyonghan: Oh, it supports only `String` type: https://github.com/keyonghan/cocoon/blob/master/app_dart/lib/src/datastore/cocoon_config.dart#L40

godofredoc: Would it be possible to use memcache instead of datastore? it will be faster and we don't need to hit datastore for branching data.

keyonghan: Actually here redisCache is used to cache this single value.  @CaseyHillers may have a better idea to choose redishCache over memcache.

CaseyHillers: IIRC memcache was deprecated in the Dart AppEngine libraries when I implemented the redis caching (my initial approach was to implement memcache).

CaseyHillers: Redis does only store strings, so `flutterBranches` getter would need a wrapper to convert the string to `List<String>`. There's some examples in the code base for how to serialize a value to bytes and serialize it back to cache it in redis.

CaseyHillers: For example, https://github.com/flutter/cocoon/blob/master/app_dart/lib/src/datastore/cocoon_config.dart#L153 deserializes the service account which is a JSON object.

PR #740, opened 2020-04-09
Fix one off index calculation for datastore.

godofredoc:
https://github.com/flutter/flutter/issues/54298


godofredoc:
Cocoon task completed but status hasn't updated because github API quota.

PR #742, opened 2020-04-09
Tweak the names of the fake agents to avoid duplicates

Hixie:
I got the maths wrong when writing this.

### DIFF
--- a/app_flutter/test/service/dev_cocoon_test.dart
+++ b/app_flutter/test/service/dev_cocoon_test.dart
@@ -1,0 +1,22 @@
+// Copyright (c) 2019 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter_test/flutter_test.dart';
+
+import 'package:cocoon_service/protos.dart' show Agent;
+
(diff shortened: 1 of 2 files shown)


app_flutter/test/service/dev_cocoon_test.dart
```diff
@@ -0,0 +1,22 @@
...
+import 'package:cocoon_service/protos.dart' show Agent;
+
+import 'package:app_flutter/service/cocoon.dart';
+import 'package:app_flutter/service/dev_cocoon.dart';
+
+void main() {
+  testWidgets('DevelopmentCocoonService agents don\'t duplicate', (WidgetTester tester) async {
+    final DevelopmentCocoonService cocoon = DevelopmentCocoonService(DateTime(0));
+    CocoonResponse<List<Agent>> agents;
+    await tester.runAsync<void>(() async {
```
CaseyHillers, line 16: Is this async zone necessary? Seems weird to have an async zone nested in a widget test.

Hixie: Yeah, it's to break out of the FakeAsync jail. The precaching code does real async work, so it can't be awaited inside FakeAsync (it would hang).

PR #744, opened 2020-04-09
Cache github branches rather than save in datastore and update logic to fetch branches

keyonghan:
This PR fixes 
https://github.com/flutter/flutter/issues/54315
https://github.com/flutter/flutter/issues/54293

Cocoon has been fetching github branch list via github API (together with existing backend logic to query github other APIs, such as commits, pulls, etc) so frequently that rate limit is hit easily.

With this PR, `branches` is updated in datastore less frequently, maybe hourly. This largely reduces github API calls, from thousands to tens a day.


app_dart/lib/src/request_handlers/get_branches.dart
```diff
@@ -7,34 +7,20 @@ import 'dart:async';
...
-import '../foundation/providers.dart';
-import '../foundation/typedefs.dart';
-import '../foundation/utils.dart';
 import '../request_handling/body.dart';
 import '../request_handling/request_handler.dart';
 
-/// Queries GitHub for the list of all available branches on
-/// [config.flutterSlug] repo, and returns list of branches
-/// that match pre-defined branch regular expressions.
 @immutable
```
CaseyHillers, line 13: dart doc? This still gets branches from the framework repo :)

app_dart/lib/src/request_handlers/get_branches.dart
```diff
@@ -7,34 +7,20 @@ import 'dart:async';
...
 
   @override
   Future<Body> get() async {
-    final List<String> branchList = await getBranchList(
-        config, branchHttpClientProvider, log, gitHubBackoffCalculator);
+    final String branches = await config.flutterBranches;
 
-    return Body.forJson(<String, List<String>>{'Branches': branchList});
+    return Body.forJson(
+        <String, List<String>>{'Branches': branches.split(',')});
```
CaseyHillers, line 24: The config should be returning `List<String>` if that's how it is used everywhere else in the code. These split calls are littered through out the code when we could just do it once in the getter for this variable.

app_dart/lib/src/request_handlers/refresh_github_branches.dart
```diff
@@ -0,0 +1,63 @@
...
+import '../datastore/cocoon_config.dart';
+import '../foundation/providers.dart';
+import '../foundation/typedefs.dart';
+import '../foundation/utils.dart';
+import '../request_handling/api_request_handler.dart';
+import '../request_handling/authentication.dart';
+import '../request_handling/body.dart';
+import '../service/datastore.dart';
+
+/// Queries GitHub for the list of available branches, fliters ones according
```
CaseyHillers, line 19: fliters -> filters

app_dart/lib/src/request_handlers/get_branches.dart
keyonghan: Added wrapper.
//cc @godofredoc related to your comment: https://github.com/flutter/cocoon/pull/739/files#r405849586

app_dart/lib/src/request_handlers/refresh_github_branches.dart
keyonghan: Updated.

app_dart/lib/src/request_handlers/get_branches.dart
keyonghan: Have removed.

keyonghan:
> Is the cronjob missing in this PR? I don't see an update to the yaml to include this endpoint.
> 
> Does this need to be done as a request handler? It seems like this logic could be put behind a service, and if the config cannot get it from cache, it can delegate to the service to get the branches. I think this would be fine to be cached with a ttl of a day, and people in a rush can clear the cache for the key in GCP.

app_dart/lib/src/request_handlers/get_branches.dart
CaseyHillers: Sorry I could have been clearer. Can you add a dart doc explaining this class? There should be class level documentation for every class.

CaseyHillers:
To clarify, I think this should implement something like

```
Future<String> get flutterBranches => _getFlutterBranches();

Future<List<String>> _getFlutterBranches() async {
    final Uint8List cacheValue = await _cache.getOrCreate(
      configCacheName,
      'flutterBranches',
      createFn: () => await getBranches(config, branchHttpClientProvider, log, gitHubBackoffCalculator), // TODO: Update to return List<String>
      ttl: configCacheTtl,
    );

    return String.fromCharCodes(cacheValue);
}
```

This lets us avoid storing it in Datastore and reduces the logic by an extra layer.

app_dart/lib/src/request_handlers/get_branches.dart
keyonghan: Added.

keyonghan:
@CaseyHillers Thanks for the suggestion! Have used cache to save `branches` rather than using datastore. With this, no cronjob is needed, together with `refresh-github-branch` API.

app_dart/bin/server.dart
```diff
@@ -40,6 +40,8 @@ Future<void> main() async {
       '/api/refresh-chromebot-status':
           RefreshChromebotStatus(config, authProvider),
       '/api/refresh-github-commits': RefreshGithubCommits(config, authProvider),
+      //'/api/refresh-github-branches':
```
CaseyHillers, line 43: nit: If this isn't needed anymore, we should remove it

keyonghan: definitely.

PR #746, opened 2020-04-10
Query commit list using branch filter for benchmark history

keyonghan:
This PR fixes: https://github.com/flutter/flutter/issues/52694

### DIFF
--- a/app_dart/lib/src/service/datastore.dart
+++ b/app_dart/lib/src/service/datastore.dart
@@ -89,8 +89,6 @@ class DatastoreService {
   }
 
   // Queries for recent commits without considering branches.
-  // TODO(keyonghan): combine this function with the above `queryRecentCommits`,
-  // this needs to fix https://github.com/flutter/flutter/issues/52694.
   Stream<Commit> queryRecentCommitsNoBranch({int limit = 100, int timestamp}) {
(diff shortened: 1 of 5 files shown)


app_dart/lib/src/service/datastore.dart
```diff
@@ -89,8 +89,6 @@ class DatastoreService {
   }
 
   // Queries for recent commits without considering branches.
-  // TODO(keyonghan): combine this function with the above `queryRecentCommits`,
-  // this needs to fix https://github.com/flutter/flutter/issues/52694.
   Stream<Commit> queryRecentCommitsNoBranch({int limit = 100, int timestamp}) {
```
digiter, line 92: Is it ready to remove this function everywhere?

keyonghan: Not yet, task reservation is still using it: https://github.com/flutter/cocoon/blob/master/app_dart/lib/src/service/task_provider.dart#L30

digiter: Can we remove that one altogether?

keyonghan: Actually it is not a simple removal thing. As for task reservation, we need to treat all tasks equally no matter which branch it is, and reserve based on only time. So, we need to keep the NoBranch function at this moment.

PR #748, opened 2020-04-12
Update backend retry logic.

godofredoc:
It will modify to a quadratic backoff starting at 200ms and will retry
up to 8 times.

https://github.com/flutter/flutter/issues/54615

### DIFF
--- a/app_dart/lib/src/service/datastore.dart
+++ b/app_dart/lib/src/service/datastore.dart
@@ -33,11 +33,15 @@ typedef RetryHandler = Function();
 
 /// Runs a db transaction with retries.
 ///
-/// It uses quadratic backoff starting with 50ms and 3 max attempts.
+/// It uses quadratic backoff starting with 200ms and 3 max attempts.
+/// for context please read https://github.com/flutter/flutter/issues/54615.
 Future<void> runTransactionWithRetries(RetryHandler retryHandler,
     {RetryOptions retryOptions}) {
-  final RetryOptions r =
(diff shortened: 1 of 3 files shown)


godofredoc:
Those errors are happening at a lower level than the dart library and as far as I know a common problem with datastore. This is coming from the grpc error:

```
gRPC Error (14, The service is temporarily unavailable. Please retry with exponential backoff.)
```

godofredoc:
Ping! this changes reduce the backend errors to 1/3. Looking at the last 6 hours we got 45 errors compared to 15 with a version including this change.

Also that 1/3 of errors are actually real errors.

app_dart/lib/src/service/datastore.dart
```diff
@@ -35,10 +35,9 @@ typedef RetryHandler = Function();
 ///
 /// It uses quadratic backoff starting with 50ms and 3 max attempts.
 Future<void> runTransactionWithRetries(RetryHandler retryHandler,
-    {int delayMilliseconds = 50, int maxAttempts = 3}) {
-  final RetryOptions r = RetryOptions(
-      delayFactor: Duration(milliseconds: delayMilliseconds),
-      maxAttempts: maxAttempts);
+    {RetryOptions retryOptions}) {
+  final RetryOptions r =
```
CaseyHillers, line 39: I would add a link to a GitHub issue that explains the rationale for this.

godofredoc: Added the bug for context and added the rationale for the retries. Also kept the 3 retries only updates the backoff.

PR #749, opened 2020-04-13
Add branching support for performance dashboard

keyonghan:
This PR is part of https://github.com/flutter/flutter/issues/51807

It adds branch select on performance dashboard;
It adds branch parameter injection in get-benchmark API
It adds branch property in TimeseriesValue
It saves branch property when inserting records to TimeseriesValue


app/lib/benchmark/benchmark_grid.html
```diff
@@ -26,4 +31,4 @@
         [data]="benchmark"
         (onZoomIn)="zoomInto = benchmark">
     </benchmark-card>
-</div>
\ No newline at end of file
+</div>
```
CaseyHillers, line 34: nit: the project leaves a new line at EOF

app_dart/lib/src/model/appengine/time_series.dart
```diff
@@ -111,6 +111,18 @@ class SerializableTimeSeries {
...
 
+  Map<String, dynamic> get facade {
+    return <String, dynamic>{
+      'Archived': series.archived,
+      'Baseline': series.baseline,
+      'Goal': series.goal,
+      'ID': series.timeSeriesId,
+      'Label': series.label,
+      'TaskName': series.taskName,
+      'Unit': series.unit,
```
CaseyHillers, line 122: nitty nit: I would add branch here and update the `TimeSeriesValue`.toJson() to not serialize it. Every single time series value shared will be repeating the same bytes. (I think it's fine to just add a TODO)

app_dart/lib/src/model/appengine/time_series_value.dart
```diff
@@ -64,6 +65,10 @@ class TimeSeriesValue extends Model {
   @JsonKey(name: 'Revision')
   String revision;
 
+  /// The branch of [commit].
+  @StringProperty(propertyName: 'Branch')
+  String branch;
```
CaseyHillers, line 70: Is there a plan in place to update past time series data include the branch field? Or will this only work on new timeseries data?

app/lib/benchmark/benchmark_grid.dart
```diff
@@ -39,13 +45,22 @@ class BenchmarkGrid implements OnInit, OnDestroy {
   Timer _reloadTimer;
   bool _isShowArchived = false;
   bool _userIsAuthenticated = false;
+  List<String> _values = ['master'];
+  String _selectedValue = 'master';
```
CaseyHillers, line 49: I think `Uri.base.queryParameters['branch'] ?? 'master'` would allow users to share the performance page of the branch they are viewing. `set selectedValue(...)` will have to be updated to set the query parameter.

CaseyHillers: (This was a request on the build dashboard branch pages)

app/lib/benchmark/benchmark_grid.html
keyonghan: Added.

app_dart/lib/src/model/appengine/time_series.dart
```diff
@@ -111,6 +111,18 @@ class SerializableTimeSeries {
   @KeyConverter()
   Key get key => series.key;
 
+  Map<String, dynamic> get facade {
```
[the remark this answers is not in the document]
keyonghan: Good catch!

app_dart/lib/src/model/appengine/time_series.dart
keyonghan: I was thinking about that too. But it didn't work out based on current logic in https://github.com/flutter/cocoon/blob/master/app_dart/lib/src/request_handlers/get_benchmarks.dart
It queries recent 50 commits, and recent 50 TimeseriesValue, then if the `sha` matches between those two entities, a point is recorded, otherwise it will be noted as missing value.
When we consider branching, different branches may share same `sha`. Results will be messed up.
Another reason why I add the branch property to TimeseriesValue: existing logic queries data based on timestamp, which is the latest 50 records. If without branch property, there is no way to query the exact records of one specific branch. For example: in the latest 50 commits, if 48 is with `master`, 2 with `release branch`, the dashboard will show only the 2 records for that release branch, even though we may have older records beyond the 50 limit.

app_dart/lib/src/model/appengine/time_series_value.dart
keyonghan: I have updated all historical data to have the branch property over the weekend.

app_dart/lib/src/model/appengine/time_series.dart
CaseyHillers: I agree with the model needing to have the branch field for TimeSeriesValue. Only the backend needs to know about what branch a TimeSeriesValue is on. The frontend shouldn't need to further parse the data.

This was moreso on just hiding the TimeSeriesValue.branch field when returning the JSON.

app/lib/benchmark/benchmark_grid.dart
keyonghan: Add it in issue: https://github.com/flutter/flutter/issues/53714

app_dart/lib/src/model/appengine/time_series.dart
keyonghan: `branch` in TimeseriesValue is only as a Property, rather than a JsonKey: https://github.com/flutter/cocoon/blob/9f90c73ab083b49c65c43cba871bd6f6652e74ad/app_dart/lib/src/model/appengine/time_series_value.dart#L69.
So it should not be parsed in the frontend side.

PR #752, opened 2020-04-14
Add branch filter when querying TimeseriesValue for benchmark history

keyonghan:
Before this PR, it queries recent history data regardless of branches. 
With this PR, it uses default `master`.


digiter:
Could you explain why the change in addition to what?

keyonghan:
This is a follow up PR for https://github.com/flutter/cocoon/pull/746, which deals with `commits`. This PR focuses on `TimeseriesValue`. 
API `/api/public/get-timeseries-history` queries both `commits` and `TimeseriesValue`, and they both need to consider `branch`.

PR #753, opened 2020-04-15
Append old master data to release branch for /api/public/get-benchmarks

keyonghan:
Benchmark dashboards shows only commits of specific release branches right now. It would be easier to view performance trending/regression if we can have historical `master` data prior to this release branch.

This pr adds the above feature.

### DIFF
--- a/app_dart/lib/src/datastore/cocoon_config.dart
+++ b/app_dart/lib/src/datastore/cocoon_config.dart
@@ -148,6 +148,8 @@ class Config {
 
   String get cqLabelName => 'CQ+1';
 
+  String get defaultBranch => 'master';
+
   // Default number of commits to return for benchmark dashboard.
   int get maxRecords => 50;
 
--- a/app_dart/lib/src/datastore/cocoon_config.dart
+++ b/app_dart/lib/src/datastore/cocoon_config.dart
@@ -148,6 +148,9 @@ class Config {
 
   String get cqLabelName => 'CQ+1';
 
+  // Default number of commits to return for benchmark dashboard.
+  int get maxRecords => 50;
+
   // Repository status context for github status.
   String get flutterBuild => 'flutter-build';
 
--- a/app_dart/lib/src/request_handlers/get_benchmarks.dart
+++ b/app_dart/lib/src/request_handlers/get_benchmarks.dart
@@ -35,41 +35,48 @@ class GetBenchmarks extends RequestHandler<Body> {
   Future<Body> get() async {
     const String master = 'master';
 
-    final int maxRecords = config.maxRecords;
     final String branch = request.uri.queryParameters[branchParam] ?? master;
     final DatastoreService datastore = datastoreProvider(config.db);
     final List<Map<String, dynamic>> benchmarks = <Map<String, dynamic>>[];
 
     Map<String, Result> releaseBranchMap = <String, Result>{};
     Map<String, Result> masterMap = <String, Result>{};
-    int masterLimit = maxRecords;
-    int timestamp = DateTime.now().millisecondsSinceEpoch;
+    int numberOfMasterCommits = config.maxRecords;
+    int lastTimestampOfReleaseBranchCommits =
+        DateTime.now().millisecondsSinceEpoch;
 
     /// Query all commits of the release branch first. Then calcalute the
     /// number of commits to retrieve from master branch, and obtain
     /// the starting [timestamp] to filter master commits.
     if (branch != master) {
       final List<Commit> releaseBranchCommits = await datastore
-          .queryRecentCommits(limit: maxRecords, branch: branch)
+          .queryRecentCommits(limit: config.maxRecords, branch: branch)
           .toList();
-      releaseBranchMap = await _getBenchmarks(releaseBranchCommits.length,
-          branch, datastore, releaseBranchCommits, benchmarks, timestamp);
-      masterLimit = maxRecords > releaseBranchCommits.length
-          ? maxRecords - releaseBranchCommits.length
+      releaseBranchMap = await _getBenchmarks(
+          releaseBranchCommits.length,
+          branch,
+          datastore,
+          releaseBranchCommits,
+          benchmarks,
+          lastTimestampOfReleaseBranchCommits);
+      numberOfMasterCommits = config.maxRecords > releaseBranchCommits.length
+          ? config.maxRecords - releaseBranchCommits.length
           : 0;
-      timestamp = releaseBranchCommits.last.timestamp;
+      lastTimestampOfReleaseBranchCommits = releaseBranchCommits.last.timestamp;
     }
 
     /// Query all remaining commits from master.
-    if (branch == master || masterLimit > 0) {
+    if (branch == master || numberOfMasterCommits > 0) {
       /// `+1` is to guarantee picking up the master commit, from which
       /// the release branch is derived.
       final List<Commit> masterCommits = await datastore
-          .queryRecentCommits(timestamp: timestamp + 1, limit: masterLimit)
+          .queryRecentCommits(
+              timestamp: lastTimestampOfReleaseBranchCommits + 1,
+              limit: numberOfMasterCommits)
           .toList();
 
-      masterMap = await _getBenchmarks(
-          masterLimit, master, datastore, masterCommits, benchmarks, timestamp);
+      masterMap = await _getBenchmarks(numberOfMasterCommits, master, datastore,
+          masterCommits, benchmarks, lastTimestampOfReleaseBranchCommits);
     }
 
     _combineValues(releaseBranchMap, masterMap, benchmarks);
@@ -79,7 +86,9 @@ class GetBenchmarks extends RequestHandler<Body> {
     });
   }
 
-  /// Combine results for both release and master branches.
+  /// Combine results for both release and master branches. [releaseBranchMap] contains
+  /// data from `release branch`, whereas [masterMap] contains data from `master`. Combined
+  /// results will be saved/returned via [benchmarks].
   void _combineValues(Map<String, Result> releaseBranchMap,
       Map<String, Result> masterMap, List<Map<String, dynamic>> benchmarks) {
     final KeyHelper keyHelper = config.keyHelper;
--- a/app_dart/lib/src/request_handlers/get_benchmarks.dart
+++ b/app_dart/lib/src/request_handlers/get_benchmarks.dart
@@ -43,8 +43,7 @@ class GetBenchmarks extends RequestHandler<Body> {
     Map<String, Result> releaseBranchMap = <String, Result>{};
     Map<String, Result> masterMap = <String, Result>{};
     int numberOfMasterCommits = config.maxRecords;
-    int lastTimestampOfReleaseBranchCommits =
-        DateTime.now().millisecondsSinceEpoch;
+    int lastMsOfReleaseBranchCommit = DateTime.now().millisecondsSinceEpoch;
 
     /// Query all commits of the release branch first. Then calcalute the
     /// number of commits to retrieve from [defaultBranch] branch, and obtain
@@ -59,11 +58,11 @@ class GetBenchmarks extends RequestHandler<Body> {
           datastore,
           releaseBranchCommits,
           benchmarks,
-          lastTimestampOfReleaseBranchCommits);
+          lastMsOfReleaseBranchCommit);
       numberOfMasterCommits = config.maxRecords > releaseBranchCommits.length
           ? config.maxRecords - releaseBranchCommits.length
           : 0;
-      lastTimestampOfReleaseBranchCommits = releaseBranchCommits.last.timestamp;
+      lastMsOfReleaseBranchCommit = releaseBranchCommits.last.timestamp;
     }
 
     /// Query all remaining commits from [defaultBranch].
@@ -72,17 +71,12 @@ class GetBenchmarks extends RequestHandler<Body> {
       /// the release branch is derived.
       final List<Commit> masterCommits = await datastore
           .queryRecentCommits(
-              timestamp: lastTimestampOfReleaseBranchCommits + 1,
+              timestamp: lastMsOfReleaseBranchCommit + 1,
               limit: numberOfMasterCommits)
           .toList();
 
-      masterMap = await _getBenchmarks(
-          numberOfMasterCommits,
-          defaultBranch,
-          datastore,
-          masterCommits,
-          benchmarks,
-          lastTimestampOfReleaseBranchCommits);
+      masterMap = await _getBenchmarks(numberOfMasterCommits, defaultBranch,
+          datastore, masterCommits, benchmarks, lastMsOfReleaseBranchCommit);
     }
 
     _combineValues(releaseBranchMap, masterMap, benchmarks);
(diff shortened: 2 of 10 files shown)


app_dart/lib/src/datastore/cocoon_config.dart
```diff
@@ -150,6 +150,9 @@ class Config {
 
   String get cqLabelName => 'CQ+1';
 
+  // Default number of commits to return for benchmark dashboard.
+  int get maxRecords => 50;
```
godofredoc, line 154: How did we get this number?

app_dart/lib/src/request_handlers/get_benchmarks.dart
```diff
@@ -30,26 +33,102 @@ class GetBenchmarks extends RequestHandler<Body> {
 
   @override
   Future<Body> get() async {
-    final String branch = request.uri.queryParameters[branchParam] ?? 'master';
-    const int maxRecords = 50;
+    const String master = 'master';
```
godofredoc, line 36: Should this be a module or class level constant?

app_dart/lib/src/request_handlers/get_benchmarks.dart
```diff
@@ -30,26 +33,102 @@ class GetBenchmarks extends RequestHandler<Body> {
...
+
+    final int maxRecords = config.maxRecords;
```
godofredoc, line 38: Use config.maxRecords everywhere, you have three variables pointing to the same value.

app_dart/lib/src/request_handlers/get_benchmarks.dart
```diff
@@ -30,26 +33,102 @@ class GetBenchmarks extends RequestHandler<Body> {
...
+    final String branch = request.uri.queryParameters[branchParam] ?? master;
     final DatastoreService datastore = datastoreProvider(config.db);
-    final DatastoreDB db = datastore.db;
+    final List<Map<String, dynamic>> benchmarks = <Map<String, dynamic>>[];
+
+    Map<String, Result> releaseBranchMap = <String, Result>{};
+    Map<String, Result> masterMap = <String, Result>{};
+    int masterLimit = maxRecords;
```
godofredoc, line 45: rm and use the config.maxRecords.

app_dart/lib/src/request_handlers/get_benchmarks.dart
```diff
@@ -30,26 +33,102 @@ class GetBenchmarks extends RequestHandler<Body> {
...
+      releaseBranchMap = await _getBenchmarks(releaseBranchCommits.length,
+          branch, datastore, releaseBranchCommits, benchmarks, timestamp);
+      masterLimit = maxRecords > releaseBranchCommits.length
+          ? maxRecords - releaseBranchCommits.length
+          : 0;
+      timestamp = releaseBranchCommits.last.timestamp;
+    }
+
+    /// Query all remaining commits from master.
+    if (branch == master || masterLimit > 0) {
```
godofredoc, line 64: This variable name is really confusing I'd recommend using a different name for it. In this case is not clear if you want 'master' instead.

app_dart/lib/src/request_handlers/get_benchmarks.dart
```diff
@@ -30,26 +33,102 @@ class GetBenchmarks extends RequestHandler<Body> {
...
+      /// `+1` is to guarantee picking up the master commit, from which
+      /// the release branch is derived.
+      final List<Commit> masterCommits = await datastore
+          .queryRecentCommits(timestamp: timestamp + 1, limit: masterLimit)
+          .toList();
```
godofredoc, line 69: What +1 means in this context? 1 ms, 1m? are we introducing a race condition?

app_dart/lib/src/request_handlers/get_benchmarks.dart
```diff
@@ -30,26 +33,102 @@ class GetBenchmarks extends RequestHandler<Body> {
...
+
+    _combineValues(releaseBranchMap, masterMap, benchmarks);
+
+    return Body.forJson(<String, dynamic>{
+      'Benchmarks': benchmarks,
+    });
+  }
+
+  /// Combine results for both release and master branches.
+  void _combineValues(Map<String, Result> releaseBranchMap,
```
godofredoc, line 83: please add docs for the parameters.

app_dart/lib/src/request_handlers/get_benchmarks.dart
```diff
@@ -30,26 +33,102 @@ class GetBenchmarks extends RequestHandler<Body> {
...
+      List<Map<String, dynamic>> benchmarks,
+      int timestamp) async {
+    final Map<String, Result> map = <String, Result>{};
+
+    await for (TimeSeries series in datastore.db.query<TimeSeries>().run()) {
       final Map<String, TimeSeriesValue> valuesByCommit =
           <String, TimeSeriesValue>{};
-      await for (TimeSeriesValue value in query.run()) {
+
+      /// Adding `1` hour to [timestamp] to guarantee all [values] belonging to
```
godofredoc, line 121: 1 hour is too much, how did we get to this value?

app_dart/lib/src/request_handlers/get_benchmarks.dart
```diff
@@ -81,17 +160,17 @@ class GetBenchmarks extends RequestHandler<Body> {
...
     }
-
-    return Body.forJson(<String, dynamic>{
-      'Benchmarks': benchmarks,
-    });
+    return map;
   }
 }
+
+class Result {
```
godofredoc, line 169: Should we place this class in models?

app_dart/lib/src/service/datastore.dart
```diff
@@ -114,10 +114,15 @@ class DatastoreService {
   /// If startFrom is nil, starts from the latest available record.
   /// [startFrom] to be implemented...
   Stream<TimeSeriesValue> queryRecentTimeSeriesValues(TimeSeries timeSeries,
-      {int limit = 1500, String startFrom, String branch = 'master'}) {
+      {int limit = 1500,
```
godofredoc, line 117: Please add a constant for this value it will make easier to change in a single place.

app_dart/test/request_handlers/get_benchmarks_test.dart
```diff
@@ -92,6 +96,7 @@ void main() {
     });
 
     test('reports statuses with input branch', () async {
+      config.maxRecordsValue = 2;
```
godofredoc, line 99: Do we need to add tests that are adding records from master and one release branch?

app_dart/lib/src/datastore/cocoon_config.dart
keyonghan: This number inherits from existing logic (from GO backend more earlier). Up to change if necessary.

app_dart/lib/src/request_handlers/get_benchmarks.dart
keyonghan: It is only used in function level.

app_dart/lib/src/request_handlers/get_benchmarks.dart
keyonghan: Updated.

app_dart/lib/src/request_handlers/get_benchmarks.dart
keyonghan: Replace maxRecords with config.maxRecords. But we need to keep variable masterLimit, which is used to track the number of `master` commits.

app_dart/lib/src/request_handlers/get_benchmarks.dart
keyonghan: changed to `numberOfMasterCommits` to make it clear.

app_dart/lib/src/request_handlers/get_benchmarks.dart
keyonghan: Added.

app_dart/lib/src/request_handlers/get_benchmarks.dart
keyonghan: Like already mentioned in the comment doc: `values` are inserted into datastore later than `commit`. The workflow inserts `commit` first into datastore, suppose with `timestamp1`, and then inserts `values` of different tests results later once tests finish, suppose with `timestampX`. `timestampX` varies for different tests/commits, depending on test execution time and environment. Based on our infra metrics collected/analyzed, executing time of a commit row is less than 1 hour, that's why I put the 1 hour timeframe.  If too small, we may miss some `values`.

app_dart/lib/src/request_handlers/get_benchmarks.dart
keyonghan: This class is used to hold temporary results, without a general-wide usage purpose across the whole app. We may just keep it here loally.

app_dart/lib/src/service/datastore.dart
keyonghan: Added.

app_dart/test/request_handlers/get_benchmarks_test.dart
keyonghan: `reports statuses with input branch` is testing this case.

app_dart/lib/src/request_handlers/get_benchmarks.dart
keyonghan: timestamp is in milliseconds, which should allow only the commit with same timestamp as the release branch.

app_dart/lib/src/request_handlers/get_benchmarks.dart
godofredoc: This looks like a constant, how is this different from the one you defined in cocoon config?

Having the same value defined in several places makes an update of constant error prone.

app_dart/lib/src/request_handlers/get_benchmarks.dart
godofredoc: We are mixing timestamp and milliseconds since epoch. We should use one or the other and name properties parameters accordingly.

app_dart/lib/src/request_handlers/get_benchmarks.dart
godofredoc: can we use a better name? from the name we don't know what it refers to.

app_dart/lib/src/request_handlers/get_benchmarks.dart
godofredoc: Adding, subtracting time can cause race conditions(we don't control the time) should we fix the query to return only commits that have all the data instead?

app_dart/lib/src/request_handlers/get_benchmarks.dart
godofredoc: In that case please document its purpose.

app_dart/test/request_handlers/get_benchmarks_test.dart
godofredoc: In that case can we refactor the tests to cover more atomic cases? Having tests that are testing too much logic require to understand the entire workflow to fix them when they  break because we don't know exactly what is failing.

app_dart/lib/src/request_handlers/get_benchmarks.dart
keyonghan: Replace with `defaultBranch`, and move it to cocoon config.

app_dart/lib/src/request_handlers/get_benchmarks.dart
keyonghan: It has been designed as to use `timestamp` variable name in cocoon to hold values of milliseconds since epoch for all APIs, and in datastore as well. If we want to distinguish them, we may need a separate PR to change from the base.

app_dart/lib/src/request_handlers/get_benchmarks.dart
keyonghan: I guess the current logic is designed like this to reduce datastore traffic. If we query timeseriesValue based on commit, then we need to make `config.maxRecords` requests each time the web refreshes.

With current logic, it needs to query datastore only once. 

Also, I don't think race condition is an issue here. This API reads data all the time, rather than writes anything. If data is not there yet, it simply show empty value on the web.

app_dart/lib/src/request_handlers/get_benchmarks.dart
keyonghan: Added.

app_dart/test/request_handlers/get_benchmarks_test.dart
keyonghan: could you please elaborate on what you mean by covering atomic cases. Do you mean test their sub functions individually? In that case, we may need to move those private functions public.

BTW: Each API in cocoon performs a specific purpose. It seems that all APIs' tests test their whole logic no matter the API is as complex as https://github.com/flutter/cocoon/blob/master/app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart, or as simply as https://github.com/flutter/cocoon/blob/master/app_dart/lib/src/request_handlers/get_branches.dart.

To make the test more easily readable, I have added some documents.

godofredoc: Sent the link to tests best practices by email, please read it and refactor accordingly.

app_dart/lib/src/request_handlers/get_benchmarks.dart
godofredoc: My comment was not about changing the current types but to make the variable names consistent with the content of them. In this case name your variable something like currentMsSinceEpoch instead of timestamp to make it consistent with the content.

app_dart/lib/src/request_handlers/get_benchmarks.dart
godofredoc: My suggestion was to update the query to return only commits with complete results up to the passed timestamp. That should not affect the current performance and will eliminate the need to guess if a commit is complete based on estimates of the execution time.

keyonghan: Unluckily that is not possible due to datastore query limitation.
To achieve what you suggested, we need to add a `commit` list which contains all commits to filter out our interested records of timeseriesValue. However, dart db supports only simple filter options like `=` (https://github.com/dart-lang/gcloud/blob/master/lib/src/db/db.dart#L128).

app_dart/test/request_handlers/get_benchmarks_test.dart
keyonghan: Thanks for the link! Have added more atomic tests.

app_dart/lib/src/request_handlers/get_benchmarks.dart
keyonghan: Added a TODO: https://github.com/flutter/flutter/issues/56731

app_dart/test/request_handlers/get_benchmarks_test.dart
```diff
@@ -39,106 +39,210 @@ void main() {
...
       final List<dynamic> benchmarks = result['Benchmarks'] as List<dynamic>;
       final Map<String, dynamic> benchmark =
           benchmarks.first as Map<String, dynamic>;
+
       expect(benchmark['Values'].length, 2);
     });
 
-    test('reports statuses with input branch', () async {
+    test(
+        'returns only release branch commits when there are more release branch commits than the maxRecordsValue - with input branch',
```
godofredoc, line 134: Here and below, let's try to keep test names short. If we need to add full descriptions let's add it as docs.

keyonghan: Use shorter test names, and put extra words to doc.

PR #754, opened 2020-04-16
[app_dart] Upgrade the github package to 6.1.3

digiter:
This PR upgrades the github package to prepare the integration with https://github.com/liyuqian/metrics_center.

The github_webhook.dart is the most affected file in by this change, as the github package no longer exports PullRequestEvent. See https://github.com/SpinlockLabs/github.dart/issues/220

Bug: https://github.com/flutter/flutter/issues/51607

### DIFF
--- a/app_dart/lib/src/request_handlers/github_webhook.dart
+++ b/app_dart/lib/src/request_handlers/github_webhook.dart
@@ -8,7 +8,6 @@ import 'dart:io';
 
 import 'package:crypto/crypto.dart';
 import 'package:github/github.dart';
-// import 'package:github/src/server/hooks.dart';
 import 'package:meta/meta.dart';
 
 import '../datastore/cocoon_config.dart';
@@ -73,8 +72,6 @@ class GithubWebhook extends RequestHandler<Body> {
     if (event == null) {
       throw const BadRequestException('Expected pull request event.');
     }
-    final List<IssueLabel> existingLabels = event.labels;
-    final bool isDraft = event.draft;
     final String eventAction = _getEventProperty(rawRequest, 'action');
 
     // See the API reference:
@@ -87,7 +84,7 @@ class GithubWebhook extends RequestHandler<Body> {
         // We'll leave unfinished jobs if it was merged since we care about those
         // results.
         if (event.merged) {
-          await _checkForGoldenTriage(eventAction, event, existingLabels);
+          await _checkForGoldenTriage(eventAction, event, event.labels);
         } else {
           await _cancelLuci(event.head.repo.name, event.number, event.head.sha,
               'Pull request closed');
@@ -96,26 +93,26 @@ class GithubWebhook extends RequestHandler<Body> {
       case 'edited':
         // Editing a PR should not trigger new jobs, but may update whether
         // it has tests.
-        await _checkForLabelsAndTests(eventAction, event, isDraft);
+        await _checkForLabelsAndTests(eventAction, event);
         break;
       case 'opened':
       case 'ready_for_review':
       case 'reopened':
         // These cases should trigger LUCI jobs.
-        await _checkForLabelsAndTests(eventAction, event, isDraft);
-        await _scheduleIfMergeable(event, labels: existingLabels);
+        await _checkForLabelsAndTests(eventAction, event);
+        await _scheduleIfMergeable(event);
         break;
       case 'labeled':
         // This should only trigger a LUCI job for flutter/flutter right now,
         // since it is in the needsCQLabelList.
         if (needsCQLabelList.contains(event.head.repo.fullName.toLowerCase())) {
-          await _scheduleIfMergeable(event, labels: existingLabels);
+          await _scheduleIfMergeable(event);
         }
         break;
       case 'synchronize':
         // This indicates the PR has new commits. We need to cancel old jobs
         // and schedule new ones.
-        await _scheduleIfMergeable(event, labels: existingLabels);
+        await _scheduleIfMergeable(event);
         break;
       case 'unlabeled':
         // Cancel the jobs if someone removed the label on a repo that needs
@@ -124,7 +121,7 @@ class GithubWebhook extends RequestHandler<Body> {
             .contains(event.head.repo.fullName.toLowerCase())) {
           break;
         }
-        if (!await _checkForCqLabel(existingLabels)) {
+        if (!await _checkForCqLabel(event.labels)) {
           await _cancelLuci(event.head.repo.name, event.number, event.head.sha,
               'Tryjobs canceled (label removed)');
         }
@@ -142,16 +139,13 @@ class GithubWebhook extends RequestHandler<Body> {
 
   /// This method assumes that jobs should be cancelled if they are already
   /// runnning.
-  Future<void> _scheduleIfMergeable(
-    PullRequest event, {
-    @required List<IssueLabel> labels,
-  }) async {
+  Future<void> _scheduleIfMergeable(PullRequest event) async {
     // The mergeable flag may be null. False indicates there's a merge conflict,
     // null indicates unknown. Err on the side of allowing the job to run.
 
     // For flutter/flutter tests need to be optimized before enforcing CQ.
     if (needsCQLabelList.contains(event.head.repo.fullName.toLowerCase())) {
-      if (!await _checkForCqLabel(labels)) {
+      if (!await _checkForCqLabel(event.labels)) {
         return;
       }
     }
@@ -304,7 +298,7 @@ class GithubWebhook extends RequestHandler<Body> {
   }
 
   // Eliminate when github.dart exports PullRequestEvent again.
-  // TODO(wutong): ...
+  // TODO(wutong): https://github.com/SpinlockLabs/github.dart/issues/220
   T _getEventProperty<T>(String pullRequestJson, String property) {
     final Map<String, dynamic> event =
         json.decode(pullRequestJson) as Map<String, dynamic>;
@@ -390,8 +384,7 @@ class GithubWebhook extends RequestHandler<Body> {
     }
   }
 
-  Future<void> _pingForTriage(
-      GitHub gitHubClient, PullRequest event) async {
+  Future<void> _pingForTriage(GitHub gitHubClient, PullRequest event) async {
     final String body = config.goldenTriageMessage;
     final RepositorySlug slug = event.head.repo.slug();
     await gitHubClient.issues.createComment(slug, event.number, body);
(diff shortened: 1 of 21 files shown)


app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -68,87 +68,62 @@ class GithubWebhook extends RequestHandler<Body> {
   }
 
   Future<void> _handlePullRequest(String rawRequest) async {
-    final PullRequestEvent event = await _getPullRequest(rawRequest);
-    final List<IssueLabel> existingLabels =
```
godofredoc, line 72: Why are we removing the labels?

app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -68,87 +68,62 @@ class GithubWebhook extends RequestHandler<Body> {
...
-        _getProperty<List<dynamic>>(rawRequest, 'labels')
-            .cast<Map<String, dynamic>>()
-            .map<IssueLabel>(IssueLabel.fromJSON)
-            .toList();
-    final bool isDraft = _getProperty(rawRequest, 'draft');
```
godofredoc, line 77: Aren't we using this anymore?

app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -68,87 +68,62 @@ class GithubWebhook extends RequestHandler<Body> {
...
+    final PullRequest event = await _getPullRequest(rawRequest);
```
godofredoc, line 71: Is a pull request the only event we can get here? if not how do we optimize to only process pull requests?

app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -68,87 +68,62 @@ class GithubWebhook extends RequestHandler<Body> {
...
+          await _checkForGoldenTriage(eventAction, event, event.labels);
         } else {
-          await _cancelLuci(
-            event.repository.name,
-            event.number,
-            event.pullRequest.head.sha,
-            'Pull request closed',
-          );
+          await _cancelLuci(event.head.repo.name, event.number, event.head.sha,
+              'Pull request closed');
```
godofredoc, line 90: add a ',' at the end for better formatting.

app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -68,87 +68,62 @@ class GithubWebhook extends RequestHandler<Body> {
...
-          labels: existingLabels,
-        );
+        await _scheduleIfMergeable(event);
         break;
       case 'unlabeled':
         // Cancel the jobs if someone removed the label on a repo that needs
         // them.
         if (!needsCQLabelList
-            .contains(event.repository.fullName.toLowerCase())) {
+            .contains(event.head.repo.fullName.toLowerCase())) {
```
godofredoc, line 121: Why forcing this to only head?

app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -68,87 +68,62 @@ class GithubWebhook extends RequestHandler<Body> {
...
       case 'labeled':
         // This should only trigger a LUCI job for flutter/flutter right now,
         // since it is in the needsCQLabelList.
-        if (needsCQLabelList
-            .contains(event.repository.fullName.toLowerCase())) {
-          await _scheduleIfMergeable(
-            event,
-            labels: existingLabels,
-          );
+        if (needsCQLabelList.contains(event.head.repo.fullName.toLowerCase())) {
```
godofredoc, line 108: Please fix formatting.

app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -68,87 +68,62 @@ class GithubWebhook extends RequestHandler<Body> {
...
-        if (!await _checkForCqLabel(existingLabels)) {
-          await _cancelLuci(
-            event.repository.name,
-            event.number,
-            event.pullRequest.head.sha,
-            'Tryjobs canceled (label removed)',
-          );
+        if (!await _checkForCqLabel(event.labels)) {
+          await _cancelLuci(event.head.repo.name, event.number, event.head.sha,
+              'Tryjobs canceled (label removed)');
```
godofredoc, line 126: add a ',' at the end for better formatting.

app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -164,31 +139,28 @@ class GithubWebhook extends RequestHandler<Body> {
...
+    if (needsCQLabelList.contains(event.head.repo.fullName.toLowerCase())) {
+      if (!await _checkForCqLabel(event.labels)) {
         return;
       }
     }
 
     // Always cancel running builds so we don't ever schedule duplicates.
     await _cancelLuci(
-      event.repository.name,
+      event.head.repo.name,
```
godofredoc, line 155: Same question for this property in  all the occurrences in this file, why does head means in this context?

app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -370,7 +339,7 @@ class GithubWebhook extends RequestHandler<Body> {
     await buildBucketClient.batch(BatchRequest(requests: requests));
   }
 
-  Future<bool> _isIgnoredForGold(PullRequestEvent event) async {
+  Future<bool> _isIgnoredForGold(String eventAction, PullRequest event) async {
     bool ignored = false;
     String rawResponse;
     try {
```
godofredoc, line 345: Is this something we can get directly from event?

app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -400,11 +369,12 @@ class GithubWebhook extends RequestHandler<Body> {
...
 
   Future<void> _checkForGoldenTriage(
-    PullRequestEvent event,
+    String eventAction,
+    PullRequest event,
     List<IssueLabel> labels,
   ) async {
-    if (event.repository.fullName.toLowerCase() == 'flutter/flutter' &&
-        await _isIgnoredForGold(event)) {
+    if (needsCQLabelList.contains(event.head.repo.fullName.toLowerCase()) &&
```
godofredoc, line 376: This logic is different from the previous one, actually in a few days we will remove the requirement for explicitly adding the CQ+1 label on sdk. I'd recommend created a top level constant for this use case.

app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -414,37 +384,33 @@ class GithubWebhook extends RequestHandler<Body> {
...
   Future<void> _checkForLabelsAndTests(
-    PullRequestEvent event,
-    bool isDraft,
+    String eventAction,
+    PullRequest event,
   ) async {
-    if (event.repository.fullName.toLowerCase() == 'flutter/flutter') {
+    if (needsCQLabelList.contains(event.head.repo.fullName.toLowerCase())) {
       final GitHub gitHubClient = await config.createGitHubClient();
       try {
```
godofredoc, line 399: needsCQLabelList should only be used to require CQ+1 for presubmits and not as a synonym of 'flutter/flutter'.

app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -7,7 +7,7 @@ import 'dart:convert';
 import 'dart:io';
 
 import 'package:crypto/crypto.dart';
-import 'package:github/server.dart';
+import 'package:github/github.dart';
```
godofredoc, line 10: I'm surprised that so many things changed on this file and nothing changed in tests, please add tests for the use cases that changed the number of parameters and or types.

liyuqian:
@digiter : thanks for the change! BTW, if upgrading github to 6.0.0^ is taking a lot of time, I can just downgrade the github dependency in `metrics_center` to be 5.0.0^. My package has only one usage of github package so it should be fairly trivial to change.

app_dart/lib/src/request_handlers/github_webhook.dart
digiter: This is expected as this PR is a pure refactor, and the tests are using public APIs while the changed methods (`_scheduleIfMergeable`, `_isIgnoredForGold`, `_checkForGoldenTriage`, `_checkForLabelsAndTests`, `_applyLabels`) are implementation details. See http://go/unit-test-practices?polyglot=dart#public-apis

app_dart/lib/src/request_handlers/github_webhook.dart
godofredoc: In that case I'd recommend renaming the variable to something like pullRequest. It will be easier to correlate to the object.

app_dart/lib/src/request_handlers/github_webhook.dart
godofredoc: the if "{" moving at the beginning of the line

app_dart/lib/src/request_handlers/github_webhook.dart
godofredoc: It was eventAction at the call to _isIgnoredForGold

app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -19,10 +19,16 @@ import '../request_handling/request_handler.dart';
 import '../service/buildbucket.dart';
 
 /// List of Github supported repos.
-const Set<String> supportedRepos = <String>{'engine', 'flutter', 'cocoon'};
+const Set<String> kSupportedRepos = <String>{'engine', 'flutter', 'cocoon'};
```
godofredoc, line 22: Why the k?

app_dart/lib/src/request_handlers/github_webhook.dart
digiter: The left curly bracket has been on the same line with the if.

app_dart/lib/src/request_handlers/github_webhook.dart
digiter: No, eventAction is not a field.

app_dart/lib/src/request_handlers/github_webhook.dart
digiter: Renamed it to "pr".

app_dart/lib/src/request_handlers/github_webhook.dart
digiter: Just following the guideline at https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#begin-global-constant-names-with-prefix-k. I know it's not required for this cocoon repo, but it sounds like a good idea.

PR #758, opened 2020-04-26
Update author headers to Flutter Authors.

godofredoc:


Hixie:
Can you add a test to make sure people always use the same header? There's a similar test in Flutter/flutter if that helps. (and one in engine too I think)

godofredoc:
> Can you add a test to make sure people always use the same header? There's a similar test in Flutter/flutter if that helps. (and one in engine too I think)

Added the test to verify the license headers.

licenses/check_licenses.dart
```diff
@@ -0,0 +1,142 @@
...
+
+import 'dart:async';
+import 'dart:convert';
+import 'dart:core' hide print;
+import 'dart:io' hide exit;
+import 'package:meta/meta.dart';
+import 'package:path/path.dart' as path;
+
+import 'run_command.dart';
+import 'utils.dart';
```
Hixie, line 13: i recommend inlining these files into this one, and removing as much of the code as possible, e.g. it seems runCommand isn't needed at least

Hixie:
Is the dart analyzer running against this new code? (as in `flutter analyze`)

licenses/check_licenses.dart
godofredoc: Done, moved all the functions to check_licenses.dat and removed the utility files.

godofredoc:
> Is the dart analyzer running against this new code? (as in `flutter analyze`)

It was not running but added license folder to the tests.yaml to run the analyze checks.

godofredoc:
> sis is actually running -- I would t

You were 

> > Is the dart analyzer running against this new code? (as in `flutter analyze`)
> 
> It was not running but added license folder to the tests.yaml to run the analyze checks.

Actually you were correct I forgot to upload the yaml file with the change for analyze and test license folder.

PR #760, opened 2020-04-29
[WIP] allow re-running LUCI tasks from cocoon

jonahwilliams:


app_dart/lib/src/request_handlers/reset_luci_task.dart
```diff
@@ -0,0 +1,62 @@
...
+    final KeyHelper keyHelper =
+        KeyHelper(applicationContext: clientContext.applicationContext);
+    final Key key = keyHelper.decode(encodedKey);
+
+    final LuciBuildService luciBuildService = LuciBuildService(
+      config: config,
+      buildBucketClient: buildBucketClient,
+    );
+    // TODO
+    await luciBuildService.rescheduleBuild(
```
jonahwilliams, line 52: I'm looking for the best way to get the build info to schedule a rerun, without routing it through the front end.

@keyonghan does the backend store the luci build info anywhere today?

keyonghan: AFAIK, backend only stores luci build info when refreshing new github commits: https://github.com/flutter/cocoon/blob/master/app_dart/lib/src/request_handlers/refresh_github_commits.dart#L218

keyonghan: Then refresh status via: https://github.com/flutter/cocoon/blob/master/app_dart/lib/src/request_handlers/refresh_chromebot_status.dart

jonahwilliams: Does the LuciTask object have this info? I could lookup the recent ones via https://github.com/flutter/cocoon/blob/master/app_dart/lib/src/request_handlers/refresh_chromebot_status.dart#L53 somehow?

Alternatively, can this info be stored on an entity linked to the task id?

keyonghan: Right now, we do not save LuciTask in datastore, but definitely we can create a new entity to save this info. if this can make it easier.

jonahwilliams: If the API to look them up from LUCI is fast enough, it seems like it would be better to not cache them to avoid filling up datastore. WDUT?

keyonghan: I would use the current API call first before saving them to datastore. On the other hand, reset-luci doesn't seem to be called that frequently if implemented.

app_dart/lib/src/request_handlers/reset_luci_task.dart
```diff
@@ -0,0 +1,64 @@
...
+  static const String luciBuilderNameParam = 'LuciBuilderName';
+  static const String commitShaParam = 'CommitSha';
+
+  @override
+  Future<Body> post() async {
+    checkRequiredParameters(<String>[luciBuilderNameParam, commitShaParam]);
+    final String luciBuilderName = requestData[luciBuilderNameParam] as String;
+    final String commitSha = requestData[commitShaParam] as String;
+    await buildBucketClient.scheduleBuild(bb.ScheduleBuildRequest(
+      builderId: bb.BuilderId(
```
jonahwilliams, line 42: Not sure which values are required here. It looks like buildset is usually something like:

`commit/git/fdc6f38a440472a696c5347515d89fe4dd9d699f`

app_dart/lib/src/request_handlers/reset_luci_task.dart
jonahwilliams: Updated this - based on my understanding of the API, I think I just need to schedule a new build? I think luci_status would pick it up again If I understand how it works

keyonghan: I guess so. but one thing to make sure is whether we just need the latest status of one SHA (based on the assumption that the returned status is the latest), or we want to maintain all statuses. If the latter case, the frontend may consider to present them all, similar to the devicelab task.

app_dart/lib/src/request_handlers/reset_luci_task.dart
keyonghan: Didn't look into details before, but find some related codes:
https://github.com/flutter/cocoon/blob/17dda0d7bfa89ae83c2d1f15c64ce90272dc7876/app_dart/lib/src/model/luci/buildbucket.dart#L401
https://github.com/flutter/cocoon/blob/dbd113890d14171bfda5b55bdc7b3e104ce7e7cf/app_dart/lib/src/request_handlers/github_webhook.dart#L204
https://github.com/flutter/cocoon/blob/dbd113890d14171bfda5b55bdc7b3e104ce7e7cf/app_dart/lib/src/request_handlers/luci_status.dart#L67

jonahwilliams:
I guess the next question is "how do I test this?" Deploy it and see if it works :)

keyonghan:
> I guess the next question is "how do I test this?" Deploy it and see if it works :)

We can run the server locally and test the api by calling it locally as well.
Please refer `Running a local dev server` in https://github.com/flutter/cocoon

jonahwilliams:
My b on the dartfmt, I'll revert that before this is ready for review

app_dart/lib/src/request_handlers/reset_luci_task.dart
```diff
@@ -0,0 +1,64 @@
...
+        project: 'flutter',
+        bucket: 'try',
```
godofredoc, line 44: The bucket for postsubmit is called 'prod'

app_dart/lib/src/request_handlers/reset_luci_task.dart
```diff
@@ -0,0 +1,64 @@
...
+        builder: luciBuilderName,
+      ),
+      gitilesCommit: GitilesCommit(
+        host: 'chromium.googlesource.com',
+        project: 'external/github.com/flutter/flutter',
+        ref: 'refs/heads/master',
+        hash: commitSha,
```
godofredoc, line 51: For presubmit these values need to point to github. Eg. 

```
host: 'github.com',
project: 'flutter/flutter',
ref: 'refs/pull/16950/head'
```

keyonghan:
Hi Jonah, just want to follow up to see if you need anything from my side, or you have any blockers. Please let me know if any:-)

jonahwilliams:
Hi, I worked on this a bit more, but I'm a bit stuck because I'm not really comfortable with the testing requirements of requiring access to the production instances to run it. Is there a way I can run this against simulated/fake data without affecting the actual LUCI instances?

keyonghan:
> Hi, I worked on this a bit more, but I'm a bit stuck because I'm not really comfortable with the testing requirements of requiring access to the production instances to run it. Is there a way I can run this against simulated/fake data without affecting the actual LUCI instances?

@godofredoc Do you know any way around to test luci? I only read luci, rather than write so far.

keyonghan:
Hi Jonah, this is what godofredo suggested for luci test: https://github.com/flutter/cocoon/pull/787#discussion_r433621098
> unit testing with mocked data is the best option for testing buildbucket related changes

PR #761, opened 2020-04-30
Remove unnecessary Gold requests

Piinks:
This trims the number of requests we are making to Gold for the flutter-gold pull request status. After @kjlubick reported increased traffic, I noticed in the logs that we could eliminated hitting the endpoint if the PR is a 'draft' PR, as well as when other checks are failing since some other action will already need to be taken to resolve those tests before proceeding.
In these two cases, draft and failing PRs, a pending state will be pushed and maintained without checking Gold.

### DIFF
--- a/app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
+++ b/app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
@@ -255,7 +255,7 @@ class PushGoldStatusToGithub extends ApiRequestHandler<Body> {
           'Click [here to view](${_getTriageUrl(pr.number)}).\n\n';
     }
     body += 'If you are still iterating on this change and are not ready to '
-        'resolve the images on the Flutter Gold dashboard, consider making this PR '
+        'resolve the images on the Flutter Gold dashboard, consider marking this PR '
         'as a draft pull request above. You will still be able to view image results '
         'on the dashboard, and the check will not try to resolve itself until '
         'marked ready for review.\n\n'
--- a/app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
+++ b/app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
@@ -109,8 +109,12 @@ class PushGoldStatusToGithub extends ApiRequestHandler<Body> {
       }
 
       if (runsGoldenFileTests) {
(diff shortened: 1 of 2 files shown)


app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
```diff
@@ -250,7 +254,11 @@ class PushGoldStatusToGithub extends ApiRequestHandler<Body> {
       body = 'Golden file changes are available for triage from new commit, '
           'Click [here to view](${_getTriageUrl(pr.number)}).\n\n';
     }
-    body +=
+    body += 'If you are still iterating on this change and are not ready to '
+        'resolve the images on the Flutter Gold dashboard, consider making this PR '
```
Piinks, line 258:
```suggestion
        'resolve the images on the Flutter Gold dashboard, consider marking this PR '
```

app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
```diff
@@ -109,8 +109,12 @@ class PushGoldStatusToGithub extends ApiRequestHandler<Body> {
       }
 
       if (runsGoldenFileTests) {
-        if (cirrusCheckStatuses.any(kCirrusInProgressStates.contains)) {
-          // Checks are still running, we have to wait.
+        if (cirrusCheckStatuses.any(kCirrusInProgressStates.contains) ||
+            cirrusCheckStatuses.any(kCirrusFailedStates.contains) ||
```
keyonghan, line 113: Shall we push gold status for finished runs, no matter the checkStatus is `Failed` or `succeeded`?

Piinks: Yes, there will always be a status reported to Github, either pending or completed.

I noticed this AM in the logs that we were checking gold status when there were failing tests on other shards, and that seemed like an unnecessary request in that case, so it should just report pending (without checking Gold) until all other failures are resolved.

Piinks:
> LGTM.
> 
> One question: this should reduce the traffic for sure. But not sure if any other root cause for the traffic increase happened in the morning?

The increase this morning was deemed working as intended. There was a draft PR that was hanging out unresolved just hitting the endpoint every 5 minutes, plus two other PRs in similar gold-limbo. The Gold team is also working on mitigating the cost of these requests with a new feature that will index CLs better.

PR #763, opened 2020-05-04
Dark mode improvements

dnfield:
- Changes the background color in Dark mode to not be flat black
- Avoids black text in dark mode for drop down menu for branches
- Lightens the circle avatars in dark mode
- Inverts the color of the Cirrus and Apple images in dark mode (note: this only works in CanvasKit, which is fine since that's what we use).

Trying to make sure I updated goldens correctly, then I'll add some tests for dark mode.

Before:

image

After:

image


app_flutter/lib/build_dashboard_page.dart
```diff
@@ -68,7 +68,7 @@ class BuildDashboardPage extends StatelessWidget {
                   (String value) {
                     return DropdownMenuItem<String>(
                       value: value,
-                      child: Text(value, style: const TextStyle(color: Colors.black)),
+                      child: Text(value),
```
CaseyHillers, line 71: Does this work in light theme? When I originally implemented this, the text was white so it blended into the dropdown.

app_flutter/lib/widgets/task_icon.dart
```diff
@@ -25,22 +25,46 @@ class TaskIcon extends StatelessWidget {
...
-      icon = stageIcons[qualifiedTask.stage];
+      return Image.asset(
+        'assets/fuchsia.png',
+      );
     }
+    switch (qualifiedTask.stage) {
+      case StageName.cirrus:
+        return Image.asset(
+          'assets/cirrus.png',
+          color: blendFilter,
```
CaseyHillers, line 40: Do you know if the DOM implementation is going to get this support? I lean towards duplicating the assets if this only works in CanvasKit since the DOM was the quicker version.

app_flutter/lib/build_dashboard_page.dart
dnfield: Light theme:
image

image

app_flutter/lib/widgets/task_icon.dart
dnfield: I'm not sure. Whatever is currently up on flutter-dashboard.appspot.com is still using CanvasKit for sure though.

dnfield:
If we need the images to work in DOM mode, we could also just add light mode images - or swap them for vector assets that we could more easily control the colors of.

app_flutter/lib/build_dashboard_page.dart
CaseyHillers: How about the items in the dropdown menu?

CaseyHillers:
I deployed http://testchillers.flutter-dashboard.appspot.com/#/build to test out the DOM implementation. It seems much slower on my machine, so I wouldn't worry about it.

The color blend is fine, just leave a comment stating it won't work on the DOM implementation.

dnfield:
Added a comment.

app_flutter/lib/build_dashboard_page.dart
dnfield: Ahhh I see, I'll fix that.

dnfield: Unfortunately, the best I can do right now is to use white only for dark mode and white for light mode. I'm not thrilled with that, filed https://github.com/flutter/flutter/issues/56399

app_flutter/lib/widgets/task_icon.dart
yjbanov: This was implemented (at least partially) in the HTML backend last week: https://github.com/flutter/engine/commit/a0983d36f7bf2b4f45a5762ce40bb81d8a3f8f9f.

dnfield:
Friendly ping - I've received offline feedback that this is better, but it still needs an approving review or we need to figure out what else has to change :)

CaseyHillers:
Is there a golden image test for dark theme on the task grid? The `TaskGridContainer with DevelopmentCocoonService` test can be reused with dark theme as that encapsulates what the application looks like when deployed.

dnfield:
Hah, I got so caught up in fixing image loading stuff I forgot to add the updated test.  Added it.

PR #766, opened 2020-05-12
Mark linux_bot as flaky

jmagman:
Linux bots are experiencing infra failures downloading and installing tiles_ctl dependencies for the fuchsia test.

Temporarily mark as flaky.

### DIFF
--- a/app_dart/lib/src/request_handlers/refresh_github_commits.dart
+++ b/app_dart/lib/src/request_handlers/refresh_github_commits.dart
@@ -218,7 +218,7 @@ class RefreshGithubCommits extends ApiRequestHandler<Body> {
       newTask(
           'mac_bot', 'chromebot', <String>['can-update-chromebots'], false, 0),
       newTask('linux_bot', 'chromebot', <String>['can-update-chromebots'],
-          false, 0),
+          true, 0),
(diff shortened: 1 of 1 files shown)


app_dart/lib/src/request_handlers/refresh_github_commits.dart
```diff
@@ -218,7 +218,7 @@ class RefreshGithubCommits extends ApiRequestHandler<Body> {
       newTask(
           'mac_bot', 'chromebot', <String>['can-update-chromebots'], false, 0),
       newTask('linux_bot', 'chromebot', <String>['can-update-chromebots'],
-          false, 0),
+          true, 0),
```
jmagman, line 221: I'm not really sure if this will work, let's see!

jmagman:
Created https://github.com/flutter/flutter/issues/57042 to track reverting this once infra issues are resolved.

app_dart/lib/src/request_handlers/refresh_github_commits.dart
godofredoc: luci bots follow a different workflow but let's try it. If this does not work I'll disable the test.

PR #773, opened 2020-05-18
[app_dart] Adds a trigger for fuchsia_ctl builder

digiter:
Bug: https://github.com/flutter/flutter/issues/56138

### DIFF
--- a/app_dart/lib/src/request_handlers/github_webhook.dart
+++ b/app_dart/lib/src/request_handlers/github_webhook.dart
@@ -21,9 +21,9 @@ import '../service/buildbucket.dart';
 
 /// List of Github supported repos.
 const Set<String> kSupportedRepos = <String>{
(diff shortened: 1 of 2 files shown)


app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -19,7 +19,12 @@ import '../request_handling/request_handler.dart';
 import '../service/buildbucket.dart';
 
 /// List of Github supported repos.
-const Set<String> kSupportedRepos = <String>{'engine', 'flutter', 'cocoon'};
+const Set<String> kSupportedRepos = <String>{
+  'engine',
```
godofredoc, line 23: Alphabetical order?

digiter: Done.

PR #774, opened 2020-05-18
Fix test failures caused by new version of flutter in beta channel.

godofredoc:
Bug:
  https://github.com/flutter/flutter/issues/57256

### DIFF
--- a/app_dart/lib/src/request_handlers/github_webhook.dart
+++ b/app_dart/lib/src/request_handlers/github_webhook.dart
@@ -9,6 +9,7 @@ import 'dart:io';
 import 'package:crypto/crypto.dart';
 import 'package:github/github.dart';
 import 'package:meta/meta.dart';
+import 'package:github/hooks.dart';
 
 import '../datastore/cocoon_config.dart';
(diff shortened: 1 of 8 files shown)


app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -579,7 +580,8 @@ class GithubWebhook extends RequestHandler<Body> {
     }
     try {
       // ignore: undefined_identifier, return_of_invalid_type
```
digiter, line 582: The ignores and the TODO in the comment can be removed.

app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -9,6 +9,7 @@ import 'dart:io';
 import 'package:crypto/crypto.dart';
 import 'package:github/github.dart';
 import 'package:meta/meta.dart';
+import 'package:github/hooks.dart';
```
keyonghan, line 12: You may want to move this up.

godofredoc:
Missed the comment, will send a follow up to fix them!

PR #775, opened 2020-05-18
Remove todos and ignores from github webhook.

godofredoc:
Bug:
  https://github.com/flutter/flutter/issues/57256


godofredoc:
> There are other ignores can be removed in the file.

Removed the other ignores.

PR #776, opened 2020-05-22
Update task ID in bigquery to be consistent with commit ID

keyonghan:
Old `task` ID follows format: `/flutter/flutter/<SHA>`, whereas `commit` ID follows: `/flutter/flutter/<branch>/<SHA>`. This make it difficult to join these tables in bigquery.

This PR changes `task` ID to have the same format as `commit`.

### DIFF
--- a/app_dart/lib/src/request_handlers/update_task_status.dart
+++ b/app_dart/lib/src/request_handlers/update_task_status.dart
@@ -126,7 +126,7 @@ class UpdateTaskStatus extends ApiRequestHandler<UpdateTaskStatusResponse> {
 
     requestRows.add(<String, Object>{
       'json': <String, Object>{
-        'ID': 'flutter/flutter/${commit.sha}',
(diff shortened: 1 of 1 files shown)


app_dart/lib/src/request_handlers/update_task_status.dart
```diff
@@ -126,7 +126,7 @@ class UpdateTaskStatus extends ApiRequestHandler<UpdateTaskStatusResponse> {
 
     requestRows.add(<String, Object>{
       'json': <String, Object>{
-        'ID': 'flutter/flutter/${commit.sha}',
+        'ID': 'flutter/flutter/${commit.branch}/${commit.sha}',
```
godofredoc, line 129: Is this compatible with all the data that shows up in the current view of the dashboard?

keyonghan: This does not affect the dashboard. This is for query purpose in BigQuery only. And dashboard is already using `branch/commit`.

PR #777, opened 2020-05-26
Fix cocoon deployments.

godofredoc:
The latest version of build compilers is failing on the cocoon app
folder. This change adds an upper limit for the dependency.


digiter:
Does this change solve the malformed "<!DOCTYPE HTML>" issue on benchmark page?

godofredoc:
> Does this change solve the malformed "" issue on benchmark page?

Would you mind sharing the failure to take a look?

keyonghan:
I have a PR to fix that on the way.

PR #778, opened 2020-05-26
Fix benchmark malformed issue

keyonghan:
This PR solves the malformed <!DOCTYPE HTML> issue on flutter benchmark page.

(The above issue happens in both benchmark_grid and benchmark_card).

### DIFF
--- a/app/lib/benchmark/benchmark_card.html
+++ b/app/lib/benchmark/benchmark_card.html
@@ -1,4 +1,3 @@
-<!DOCTYPE HTML>
 <!-- Copyright 2018 The Flutter Authors. All rights reserved.
 Use of this source code is governed by a BSD-style license that can be
 found in the LICENSE file. -->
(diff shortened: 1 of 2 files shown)


keyonghan:
@digiter @godofredoc Removing `<!DOCTYPE HTML>` solves the malformed issue. However it violates the license rule defined:
https://github.com/flutter/cocoon/blob/master/licenses/check_licenses.dart#L72

Is it safe to remove the license check? Or any idea for an alternative fix?

keyonghan:
This is the issue on benchmark page:
Screen Shot 2020-05-26 at 2 39 31 PM

godofredoc:
> <!DOCTYPE HTML>

What happens if you try the following?

<!DOCTYPE HTML/>

keyonghan:
> What happens if you try the following?

Do you mean try `disable the license check for html` or something else?

godofredoc:
> > What happens if you try the following?
> 
> Do you mean try `disable the license check for html` or something else?

upps I think I pushed the button too early. We should not disable the license check. You can remove the doctype line and then remove the prefix for html files in the license checker. I'd recommend doing that in a separate pr as you'll need to modify the headers for all the existing html files.

The alternative is adding a "\" at the end <DOCTYPE HTML\>

keyonghan:
Thanks. Merging this PR and will create a follow up one to modify all html headers.

PR #785, opened 2020-05-29
Add luci build number support

keyonghan:
Existing cocoon keeps track of latest status for luci build, and log url is linked to the general builder page: such as [mac](https://ci.chromium.org/p/flutter/builders/prod/Mac).

It would be handy to have separate links to difference builds, similar to devicelab tasks.

This PR enables build ID support, which makes the above requirement possible.

(part of https://github.com/flutter/flutter/issues/56048)


app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
```diff
@@ -79,19 +80,79 @@ class RefreshChromebotStatus extends ApiRequestHandler<Body> {
...
+      for (int i = tasks.length - 1; i >= 0; i--) {
+        final FullTask task = tasks[i];
+        for (int j = allLuciTasks.length - 1; j >= 0; j--) {
+          final LuciTask luciTask = allLuciTasks[j];
           if (luciTask.commitSha == task.commit.sha &&
-              luciTask.ref == 'refs/heads/$branch') {
+              luciTask.ref == 'refs/heads/$branch' &&
+              (task.task.buildId == null ||
+                  task.task.buildId == luciTask.buildId) &&
+              !updatedLuciTasks.contains(luciTask)) {
```
godofredoc, line 100: Looks like a very complex if to compare two objects. Do we need to override operator == in either luci or task to avoid this?

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
```diff
@@ -79,19 +80,79 @@ class RefreshChromebotStatus extends ApiRequestHandler<Body> {
...
+          .toList();
+      final List<LuciTask> allLuciTasks = luciTasks[builder];
+      final Set<LuciTask> updatedLuciTasks = <LuciTask>{};
+
+      /// Scan both [tasks] and [luciTasks] from old to new, since [luciTasks] may
+      /// contain new re-run builds which are not in datastore yet. If matched, then
+      /// update accordingly.
+      for (int i = tasks.length - 1; i >= 0; i--) {
+        final FullTask task = tasks[i];
+        for (int j = allLuciTasks.length - 1; j >= 0; j--) {
```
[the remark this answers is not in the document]
keyonghan: I guess these three loops are necessary. `branches` need to be iterated for sure. Then we need to find matched pairs of `luciTasks` and `tasks`. These two lists are ordered according to time by themselves, but no direct mapping between each other.

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
keyonghan: I have put these conditions to a sub-function to make it look simpler.

keyonghan:
> I don't really have enough context into what the LUCI service should do to review this change beyond a surface level inspection

No problem at all. Thanks for letting me know. @jonahwilliams

app_dart/lib/src/model/appengine/task.dart
```diff
@@ -32,6 +32,7 @@ class Task extends Model {
     this.requiredCapabilities,
     this.reservedForAgentId = '',
     this.stageName,
+    this.buildId,
```
digiter, line 35: Since a task can represent one of Devicelab/LUCI/Cirrus task. How do we know which fields need to be filled when constructing a Devicelab task (or LUCI/Cirrus)?

app_dart/lib/src/model/appengine/task.dart
```diff
@@ -132,6 +133,11 @@ class Task extends Model {
   @JsonKey(name: 'Reason')
   String reason;
 
+  /// The build ID of luci build.
```
digiter, line 136: Is there an protobuf definition or documentation that explains this field?

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
```diff
@@ -79,19 +80,82 @@ class RefreshChromebotStatus extends ApiRequestHandler<Body> {
...
+  ///
+  /// The best time to add re-run luci builds to datastore is when
+  /// re-running them (push rather than pull).
+  // TODO(keyonghan): add new tasks to datastore when re-running luci builds,
+  // https://github.com/flutter/flutter/issues/58268
+  List<Task> _getNewTasks(Set<LuciTask> updatedLuciTasks, LuciBuilder builder,
+      List<LuciTask> allLuciTasks, DatastoreService datastore) {
+    final List<Task> tasks = <Task>[];
+    for (LuciTask luciTask in allLuciTasks) {
+      if (!updatedLuciTasks.contains(luciTask) &&
```
godofredoc, line 136: Instead of filtering here why not just get the list of task we are interested in? both buildBucket and datastore supports retrieving only what is necessary.

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
```diff
@@ -79,19 +80,82 @@ class RefreshChromebotStatus extends ApiRequestHandler<Body> {
...
             await datastore.insert(<Task>[update]);
-            // Stop updating task whenever we find the latest status of the same commit.
+            updatedLuciTasks.add(luciTask);
             break;
           }
         }
       }
+
+      final List<Task> newTasks =
+          _getNewTasks(updatedLuciTasks, builder, allLuciTasks, datastore);
```
godofredoc, line 109: Why we need to filter again with a new loop? can we filter instead in previous iteration?

app_dart/lib/src/model/appengine/task.dart
keyonghan: `stageName` tells the difference: https://github.com/flutter/cocoon/blob/master/app_dart/lib/src/model/appengine/task.dart#L34

app_dart/lib/src/model/appengine/task.dart
keyonghan: https://chromium.googlesource.com/infra/luci/luci-go/+/master/buildbucket/proto/build.proto#146
Added to doc.

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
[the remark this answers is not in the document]
keyonghan: Thanks for the `Map` suggestion, but it works when we have a strict mapping between a `datastoreTask` and a `luciTask`. 

However for a specific triple `{branch, commitSha, builder}`, there may be multiple runs (this is exactly why we create this PR to deal with the re-run cases). For such case, we have to iterate from the OLD to NEW to pair `datastoreTask` and `luciTask` one by one.

If we keep to use `Map`, it should be `{'branch': {'commitSha': {'builderName', ['buildList']}}}`, but it doesn't help reduce the loop cycles as I see it.

Due to the above mentioned `from OLD to NEW` iteration requirement, we have to iterate the ordered `datastoreTasks` list queried according to a `branch`-`builder` pair, rather than randomly iterate through. Based on the returned `datastoreTasks` list, we then try finding the paired `luciTask`.

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
```diff
@@ -79,19 +80,82 @@ class RefreshChromebotStatus extends ApiRequestHandler<Body> {
...
+      final List<FullTask> tasks = await datastore
+          .queryRecentTasks(taskName: builder.taskName, branch: branch)
+          .toList();
+      final List<LuciTask> allLuciTasks = luciTasks[builder];
+      final Set<LuciTask> updatedLuciTasks = <LuciTask>{};
+
+      /// Scan both [tasks] and [luciTasks] from old to new, since [luciTasks] may
+      /// contain new re-run builds which are not in datastore yet. If matched, then
+      /// update accordingly.
+      for (int i = tasks.length - 1; i >= 0; i--) {
```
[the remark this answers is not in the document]
keyonghan: Replaced.

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
```diff
@@ -79,19 +80,82 @@ class RefreshChromebotStatus extends ApiRequestHandler<Body> {
       DatastoreService datastore,
       Map<LuciBuilder, List<LuciTask>> luciTasks) async {
     for (String branch in branches) {
-      await for (FullTask task in datastore.queryRecentTasks(
-          taskName: builder.taskName, branch: branch)) {
-        for (LuciTask luciTask in luciTasks[builder]) {
-          if (luciTask.commitSha == task.commit.sha &&
-              luciTask.ref == 'refs/heads/$branch') {
+      final List<FullTask> tasks = await datastore
```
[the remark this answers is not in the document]
keyonghan: Use `datastoreTask` instead.

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
```diff
@@ -79,19 +80,82 @@ class RefreshChromebotStatus extends ApiRequestHandler<Body> {
...
         }
       }
+
+      final List<Task> newTasks =
+          _getNewTasks(updatedLuciTasks, builder, allLuciTasks, datastore);
+      await datastore.insert(newTasks);
+    }
+  }
+
+  bool _taskMatched(FullTask task, LuciTask luciTask, String branch) {
```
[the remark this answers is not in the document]
keyonghan: This is not a function belonging to a specific class, but across both `Task` and `Luci` models.

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
```diff
@@ -79,19 +80,82 @@ class RefreshChromebotStatus extends ApiRequestHandler<Body> {
...
+    return luciTask.commitSha == task.commit.sha &&
+        luciTask.ref == 'refs/heads/$branch' &&
+        (task.task.buildId == null || task.task.buildId == luciTask.buildId);
```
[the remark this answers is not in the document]
keyonghan: No. This updates both existing build and retry. A retry will have `null` `builderID` initially.

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
godofredoc: Moving it to a to a function is just moving the complexity to a different place. I'd suggest trying to remove the complexity instead.

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
```diff
@@ -79,19 +80,82 @@ class RefreshChromebotStatus extends ApiRequestHandler<Body> {
...
+  /// corresponding tasks in datastore to be updated.
+  ///
+  /// When there are re-run luci builds, `_getNewTasks` returns them to be
+  /// inserted into datastore.
+  ///
+  /// The best time to add re-run luci builds to datastore is when
+  /// re-running them (push rather than pull).
+  // TODO(keyonghan): add new tasks to datastore when re-running luci builds,
+  // https://github.com/flutter/flutter/issues/58268
+  List<Task> _getNewTasks(Set<LuciTask> updatedLuciTasks, LuciBuilder builder,
```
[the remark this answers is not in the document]
keyonghan: Added.

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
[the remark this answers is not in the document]
godofredoc: Using a map was a suggestion not the final solution. 4 nested loops strongly suggests problems in the way we are getting the data or the way we are processing the data I'd suggest to model classes, queries to buildbucket or queries to datastore to simplify this.

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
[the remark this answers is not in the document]
keyonghan: oh, I can add the logic to either model.

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
```diff
@@ -79,19 +80,82 @@ class RefreshChromebotStatus extends ApiRequestHandler<Body> {
...
+      List<LuciTask> allLuciTasks, DatastoreService datastore) {
+    final List<Task> tasks = <Task>[];
+    for (LuciTask luciTask in allLuciTasks) {
+      if (!updatedLuciTasks.contains(luciTask) &&
+          updatedLuciTasks.any((LuciTask e) =>
+              e.commitSha == luciTask.commitSha && e.ref == luciTask.ref)) {
```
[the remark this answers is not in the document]
keyonghan: Moded to `LuciTask` Model.

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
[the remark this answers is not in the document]
keyonghan: Added to `LuciTask` model.

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
keyonghan: To avoid confusion, I decide to skip the `newTasks` insert in this PR. Instead, will use https://github.com/flutter/flutter/issues/58268 to take care of new re-run builds.

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
keyonghan: Have removed the section of `new builds insert` from this PR.

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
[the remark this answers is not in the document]
keyonghan: Managed to reduce one loop level by changing the returned `luciTasks` to consider `branch`. With this, we need to iterate only `branch` list occurred in these `luciTasks`, rather than iterating all `branch` options. Thank you for the suggestion.

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
keyonghan: Inspired one you your comments above about using `Map`, I have including `sha` in the returned `luciTasks`. This naturally removes two conditions, together with the `branch`.

keyonghan:
> Only adding build id isn't enough. Currently the "build url" is synthesized on the frontend:
> https://github.com/flutter/cocoon/blob/a6db0e2603dc9e101dd3e54f3e73a8e0c79533b7/app_flutter/lib/logic/qualified_task.dart#L62-L72
> 
> Which is vulnerable to future LUCI builder renames. I would save the pool name, builder name in addition, or even the "build url" as a whole.

@digiter Those info. is better to be saved in config rather than in each task record. I have created issue: https://github.com/flutter/flutter/issues/58450 to track.

On the other hand, we already have builder info. in the task entity.

app_dart/lib/src/model/appengine/task.dart
```diff
@@ -132,6 +133,11 @@ class Task extends Model {
   @JsonKey(name: 'Reason')
   String reason;
 
+  /// The build ID of luci build: https://chromium.googlesource.com/infra/luci/luci-go/+/master/buildbucket/proto/build.proto#146
+  @IntProperty(propertyName: 'BuildID')
+  @JsonKey(name: 'BuildID')
+  int buildId;
```
digiter, line 139: > > Only adding build id isn't enough. Currently the "build url" is synthesized on the frontend:
> > https://github.com/flutter/cocoon/blob/a6db0e2603dc9e101dd3e54f3e73a8e0c79533b7/app_flutter/lib/logic/qualified_task.dart#L62-L72
> > 
> > Which is vulnerable to future LUCI builder renames. I would save the pool name, builder name in addition, or even the "build url" as a whole.
> 
> @digiter Those info. is better to be saved in config rather than in each task record. I have created issue: [flutter/flutter#58450](https://github.com/flutter/flutter/issues/58450) to track.
> 
> On the other hand, we already have builder info. in the task entity.

Moving the discussion here. Saving the info in `config` would cause troubles when we rename the LUCI pools or LUCI builders. As new task records and old ones will have different URL prefixes, a single `config` wouldn't handle both cases.

keyonghan: What is the chance to have such change? The reason I ask is that `task` entity is kind of complex already. Additionally, if we add these info., all records of these property will be same until any change happens. If there is no chance those values change, then it may be good enough to have them in config.

I am okay to save those info., but just want to make sure it makes sense.
/cc @godofredoc

digiter: I have been thinking about renaming the builders to be more machine friendly and we will probably add devicelab builders in the future.

godofredoc: In the future we'll run tasks from different places, e.g. chromium, chrome

keyonghan: @digiter Added both `builderName` and `poolName`.

keyonghan:
@digiter a kind reminder.

app_dart/lib/src/model/appengine/task.dart
digiter: But we still don't know what fields need to be filled. Please add constructors for each variant.

keyonghan: Added.

app_dart/lib/src/model/appengine/task.dart
```diff
@@ -236,3 +255,96 @@ class FullTask {
...
+    Key key,
+    Key commitKey,
+    int createTimestamp,
+    int startTimestamp = 0,
+    int endTimestamp = 0,
+    String name,
+    int attempts = 0,
+    bool isFlaky = false,
+    int timeoutInMinutes = 0,
+    List<String> requiredCapabilities = const <String>['can-update-chromebots'],
```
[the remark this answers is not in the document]
keyonghan: Removed `requiredCapabilities`, `isFalky`, and `timeoutInMinutes` for both luci and cirrus.

app_dart/lib/src/service/luci.dart
```diff
@@ -58,6 +94,31 @@ class LuciService {
...
+      final String ref = build.input?.gitilesCommit?.ref ?? 'unknown';
+      final LuciBuilder builder = builders.singleWhere((LuciBuilder builder) {
+        return builder.name == build.builderId.builder;
+      });
+      results[builder] ??= <LuciTask>[];
+      results[builder].add(LuciTask(
+        commitSha: commit,
+        ref: ref,
+        status: _luciStatusToTaskStatus[build.status],
+        buildId: build.number,
```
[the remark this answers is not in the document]
keyonghan: Changed `buildId` to `buildNumber`.

app_dart/lib/src/service/luci.dart
```diff
@@ -58,6 +94,31 @@ class LuciService {
...
+      ));
+    }
+    return results;
+  }
+
+  Future<Iterable<Build>> getBuilds(
```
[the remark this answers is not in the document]
keyonghan: Added.

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
```diff
@@ -73,25 +71,38 @@ class RefreshChromebotStatus extends ApiRequestHandler<Body> {
...
+    for (FullTask datastoreTask in datastoreTasks.reversed) {
+      if (!luciTasks.containsKey(datastoreTask.commit.sha)) {
+        continue;
+      }
+      for (LuciTask luciTask in luciTasks[datastoreTask.commit.sha].reversed) {
+        if (!updatedLuciTasks.contains(luciTask) &&
+            _buildIdMatched(datastoreTask, luciTask)) {
+          final Task update = datastoreTask.task;
+          update.status = luciTask.status;
+          update.buildId = luciTask.buildNumber;
```
digiter, line 93: Change to `update.buildNumber`.

keyonghan: Done

app_dart/lib/src/model/appengine/task.dart
```diff
@@ -132,6 +135,21 @@ class Task extends Model {
   @JsonKey(name: 'Reason')
   String reason;
 
+  /// The build ID of luci build: https://chromium.googlesource.com/infra/luci/luci-go/+/master/buildbucket/proto/build.proto#146
+  @IntProperty(propertyName: 'BuildID')
+  @JsonKey(name: 'BuildID')
```
digiter, line 140: BuildNumber?

keyonghan: Ha, thanks for catching this! just updated.

app_dart/lib/src/datastore/cocoon_config.dart
```diff
@@ -192,6 +192,9 @@ class Config {
         'engine-flutter-autoroll',
       };
 
+  /// Luci flutter prod pool.
+  String get luciFlutterProdPool => 'luci.flutter.prod';
```
digiter, line 196: Nit: just inline this constant.

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
```diff
@@ -73,25 +71,38 @@ class RefreshChromebotStatus extends ApiRequestHandler<Body> {
...
+          update.luciPoolName = config.luciFlutterProdPool;
+          await datastore.insert(<Task>[update]);
+          updatedLuciTasks.add(luciTask);
+          break;
         }
       }
     }
   }
+
+  bool _buildIdMatched(FullTask task, LuciTask luciTask) {
```
digiter, line 104: `buildNumberMatched`

app_dart/lib/src/service/luci.dart
```diff
@@ -49,6 +49,42 @@ class LuciService {
...
 
+  /// Gets the list of recent LUCI tasks, broken out by the [BranchLuciBuilder]
+  ///  that owns them.
+  ///
+  /// The list of known LUCI builders is specified in [LuciBuilder.all].
+  Future<Map<BranchLuciBuilder, Map<String, List<LuciTask>>>>
+      getBranchRecentTasks({
+    String repo,
+    bool requireTaskName = false,
+  }) async {
```
digiter, line 60: Nit: the return value of this function is too complicated. We can add `LuciBuilder` to `LuciTask` since every LUCI task has to run on a builder. Then this return value can be simplified to a list of LUCI tasks without two levels of nested mapping.

app_dart/lib/src/service/luci.dart
```diff
@@ -168,4 +229,7 @@ class LuciTask {
 
   /// The status of this task. See the [Task] class for supported values.
   final String status;
+
+  /// The build number of this task.
+  final int buildNumber;
```
digiter, line 234: It's more readable to add a `LuciBuilder` field since every LuciTask must have been executed by a builder.

And add a method to return the branch calculated from `ref`.

app_dart/lib/src/model/appengine/task.dart
```diff
@@ -132,6 +135,21 @@ class Task extends Model {
   @JsonKey(name: 'Reason')
   String reason;
 
+  /// The build ID of luci build: https://chromium.googlesource.com/infra/luci/luci-go/+/master/buildbucket/proto/build.proto#146
```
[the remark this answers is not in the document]
keyonghan: Done.

app_dart/lib/src/datastore/cocoon_config.dart
keyonghan: Done.

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
keyonghan: Done.

app_dart/lib/src/service/luci.dart
keyonghan: I agree that will make the returning list simple, but it will make the follow-up status update challenging.

If you look into the update process, you will find that `luciBuilder` is the key base for looping around. Iterating through each `luciBuilder`, we then query DB for that builder. At the same time, `luciTask` needs to be ordered chronologically according to `luciBuilder`, and even a specific `sha`.

This is designed to reduce the time complexity.

app_dart/lib/src/service/luci.dart
keyonghan: Please see the comment above.

keyonghan:
> How many will be returned?

IIRC, around 30 builds will be returned.

app_dart/lib/src/service/luci.dart
digiter: Exactly, the double nested mapping only makes sense in the high level refresh_chromebot_status, while this low level `getBranchRecentTasks` function is overfitting that. They are tightly coupled.

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
```diff
@@ -73,25 +71,38 @@ class RefreshChromebotStatus extends ApiRequestHandler<Body> {
...
+    /// Scan both [datastoreTasks] and [luciTasks] from old to new, since [luciTasks] may
+    /// contain new re-run builds which are not in datastore yet. If matched, then
+    /// update accordingly.
+    for (FullTask datastoreTask in datastoreTasks.reversed) {
+      if (!luciTasks.containsKey(datastoreTask.commit.sha)) {
+        continue;
+      }
+      for (LuciTask luciTask in luciTasks[datastoreTask.commit.sha].reversed) {
+        if (!updatedLuciTasks.contains(luciTask) &&
+            _buildNumberMatched(datastoreTask, luciTask)) {
```
digiter, line 90: Nit: I think the `!updatedLuciTasks.contains(luciTask)` check is unnecessary, since the inner loop breaks right after it finds the first match.

app_dart/test/request_handlers/refresh_chromebot_status_test.dart
```diff
@@ -48,89 +48,284 @@ void main() {
...
+                Map<BranchLuciBuilder, Map<String, List<LuciTask>>>>.value(
+            luciTasks);
+      });
+
+      expect(task.status, Task.statusNew);
+      await tester.get(handler);
+      expect(task.status, Task.statusNew);
+    });
+
+    test('update task status and buildId when buildId is null', () async {
```
digiter, line 123: s/buildId/buildNumber

keyonghan: Done.

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
keyonghan: It is needed for cases where there are reruns (this is exactly why this PR) for a commit.

app_dart/lib/src/model/appengine/task.g.dart
```diff
@@ -18,6 +18,9 @@ Map<String, dynamic> _$TaskToJson(Task instance) => <String, dynamic>{
       'Flaky': instance.isFlaky,
       'TimeoutInMinutes': instance.timeoutInMinutes,
       'Reason': instance.reason,
+      'BuildID': instance.buildNumber,
+      'BuilderName': instance.builderName,
+      'LuciPoolName': instance.luciPoolName,
```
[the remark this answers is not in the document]
keyonghan: Done.

app_dart/lib/src/model/appengine/task.dart
```diff
@@ -195,6 +213,7 @@ class Task extends Model {
       ..write(', reservedForAgentId: $reservedForAgentId')
       ..write(', stageName: $stageName')
       ..write(', status: $status')
+      ..write(', buildNumber: $buildNumber')
```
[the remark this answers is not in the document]
keyonghan: Added.

app_dart/lib/src/service/luci.dart
keyonghan: Understand. Low level functions are not supposed to be used by a single target, the more general the better.

Per that, in the beginning, I tried using existing function `getRecentTasks` to implement the API, however, as @godofredoc pointed out, that resulted in too many loop cycles and complicated condition logics. It definitely will cause potential issues.

To solve that, here comes the specific function for our purpose here. Though tightly coupled with the high level purpose, it eases the logic greatly: reducing both loop cycles and logic conditions.

digiter: There's no extra loop cycles, it's just a choice of having them in the low level API or high level functions. The current approach is less readable as one has to read both functions to understand the structure of the return value, and the reason why it's structured that way.

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
digiter: A `datastoreTask` is expected to be updated by no more than one `luciTask`. 
- When `datastoreTask.task.buildNumber == null`, it expects the newest `luciTask`
- When not null, it expects the newest `luciTask` with matching `buildNumber`.

Therefore we can loop the luciTask from new to old and break when a first match is found.

digiter: A `datastoreTask` is expected to be updated by no more than one `luciTask`. 
- When `datastoreTask.task.buildNumber == null`, it expects the newest `luciTask`
- When not null, it expects the newest `luciTask` with matching `buildNumber`.

Therefore we can loop the luciTask from new to old and break when a first match is found.

app_dart/lib/src/service/luci.dart
digiter: There's no extra loop cycles, it's just a choice of having them in the low level API or high level functions. The current approach is less readable as one has to read both functions to understand the structure of the return value, and the reason why it's structured that way.

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
keyonghan: When new builds or re-runs happen, they will be inserted to datastore immediately, but may not scheduled yet in luci. For such case, there will not be one-to-one mapping between `datastoreTask` and `luciTasks` from new to old. That's why I changed the looping order from old to new. 

The old strategy only needs to keep the latest status of a commit, so it works as expected when looping from new to old.

digiter: Very nit: OK then this apparently can be simplified to linear time similar to merging two sorted arrays. And the comments need more clarification for future readers.

keyonghan: The logic is already "almost" in linear time complexity, thanks to the "complex" map we created. When iterating through `datastoreTask`, we compare directly with `luciTasks[datastoreTask.commit.sha]`. For commits without reruns, this is just one time comparison. 

I have updated the doc. to make it clearer.

PR #788, opened 2020-06-01
[app_dart] Add a handler that pushes benchmark results to Metrics Center

digiter:
The handler reads benchmark results after [sinceMillis] parameter, transforms them into MetricPoints format, then writes to Metrics Center. The Metrics Center will transfer the data to Skia Perf later and let us evaluate the effectiveness of using Skia Perf.

Follow up PRs will add a cron job that runs this handler periodically.

Bug: https://github.com/flutter/flutter/issues/58442

Currently the data store read operations of `flutter-dashboard` is about 906 millions per day. 
This handler reads about 20 thousands data store records when transferring benchmark data for one day, so it's a tiny fraction of the overall usage.


app_dart/lib/src/datastore/cocoon_config.dart
```diff
@@ -396,6 +396,12 @@ class Config {
   bool githubPresubmitSupportedRepo(String repositoryName) {
     return supportedRepos.contains(repositoryName);
   }
+
+  Future<Map<String, dynamic>> get metricsCenterServiceAccountJson async {
+    final String rawValue =
+        await _getSingleValue('MetricsCenterServiceAccount');
```
godofredoc, line 402: Is this a new service account or an existing one? Let's create a doc to document the origin, purpose and location. Going forward we'll need to track very closely all the service accounts used in flutter infrastructure.

app_dart/lib/src/request_handlers/push_benchmark_to_center.dart
```diff
@@ -0,0 +1,124 @@
+// Copyright 2020 The Chromium Authors. All rights reserved.
```
godofredoc, line 1: Flutter? this should have failed the licenses check.

app_dart/lib/src/request_handlers/push_benchmark_to_center.dart
```diff
@@ -0,0 +1,124 @@
...
+import 'package:metrics_center/flutter.dart' as mc;
+
+import '../datastore/cocoon_config.dart';
+import '../foundation/providers.dart';
+import '../model/appengine/time_series.dart';
+import '../model/appengine/time_series_value.dart';
+import '../request_handling/body.dart';
+import '../request_handling/request_handler.dart';
+import '../service/datastore.dart';
+
```
godofredoc, line 18: Don't we have a handler to push metrics? can we reuse the existing one instead of creating a new ad hoc handler?

app_dart/lib/src/request_handlers/push_benchmark_to_center.dart
```diff
@@ -0,0 +1,124 @@
...
+        continue;
+      }
+      map[ts.id as String] = ts;
+    }
+    return map;
+  }
+
+  /// Reads benchmark records with [createTimestamp] later than [sinceMillis].
+  /// And limits the result size to no larger than [batchSize].
+  Future<List<TimeSeriesValue>> readTimeSeriesValue(int sinceMillis,
```
godofredoc, line 79: I'm a bit confused, why we need to read the benchmark from datastore instead of directly processing it from the origin. e.g the request payload.

Inserting to datastore and processing later is adding unnecessary jumps and complexity IMHO.

app_dart/lib/src/request_handlers/push_benchmark_to_center.dart
```diff
@@ -0,0 +1,124 @@
...
+  int parseSinceMillis(Map<String, String> queryParameters) {
+    final String s = queryParameters['sinceMillis'] ?? '';
+    return int.tryParse(s);
+  }
+
+  /// Returns a map containing all time series keyed by its id.
+  Future<Map<String, TimeSeries>> readTimeSeriesMap() async {
+    final Map<String, TimeSeries> map = <String, TimeSeries>{};
+    await for (TimeSeries ts in datastoreService.db.query<TimeSeries>().run()) {
+      if (!(ts.id is String)) {
```
keyonghan, line 68: this condition seems not necessary: https://github.com/flutter/cocoon/blob/master/app_dart/lib/src/model/appengine/time_series.dart#L57

app_dart/lib/src/request_handlers/push_benchmark_to_center.dart
```diff
@@ -0,0 +1,124 @@
...
+      {int batchSize = 1000}) async {
+    final Query<TimeSeriesValue> query =
+        datastoreService.db.query<TimeSeriesValue>()
+          ..filter('createTimestamp >', sinceMillis)
+          ..limit(batchSize)
+          ..order('createTimestamp');
+    return query.run().toList();
+  }
+
+  Future<List<mc.MetricPoint>> transform(Map<String, TimeSeries> timeSeriesMap,
```
godofredoc, line 89: You can potentially move most of the content of this class to a service class and use it within the existing metrics handler to push the data. That way you won't need to synchronize timestamps and/or create an additional cronjob

app_dart/test/request_handlers/push_benchmark_to_center_test.dart
```diff
@@ -0,0 +1,97 @@
+// Copyright 2020 The Chromium Authors. All rights reserved.
```
godofredoc, line 1: Flutter?

app_dart/test/request_handlers/push_benchmark_to_center_test.dart
```diff
@@ -0,0 +1,97 @@
...
+
+  setUp(() {
+    config = FakeConfig();
+    logger = FakeLogging();
+    handler = PushBenchmarkToCenter(config, logger: logger);
+  });
+
+  group('parseSinceMillis', () {
+    test('with missing parameter', () {
+      final int sinceMillis = handler.parseSinceMillis(<String, String>{});
```
godofredoc, line 26: Would parseSinceMillis make more sense in the utilities class? it would be easier to reuse and I don't think it depends on the handler.

app_dart/lib/src/request_handlers/push_benchmark_to_center.dart
```diff
@@ -0,0 +1,124 @@
...
+      List<TimeSeriesValue> timeSeriesValues) async {
+    final List<mc.MetricPoint> points = <mc.MetricPoint>[];
+    for (TimeSeriesValue tsv in timeSeriesValues) {
+      if (!(tsv.parentKey.id is String)) {
```
keyonghan, line 93: ditto.

app_dart/lib/src/request_handlers/push_benchmark_to_center.dart
keyonghan, line 79: move these functions private if not intended to be called by others?

app_dart/test/src/datastore/fake_datastore.dart
```diff
@@ -59,6 +59,11 @@ class FakeDatastoreDB implements DatastoreDB {
     onQuery[T] = untypedCallback;
   }
 
+  /// Adds a [value] to the [values] map.
+  void addValue<T extends Model>(T value) {
```
keyonghan, line 63: No need to add this function. You can use `values[key]=value` to add value.

app_dart/test/request_handlers/push_benchmark_to_center_test.dart
```diff
@@ -0,0 +1,97 @@
...
+          handler.parseSinceMillis(<String, String>{'sinceMillis': '123'});
+      expect(sinceMillis, 123);
+    });
+  });
+
+  group('readTimeSeriesMap', () {
+    test('returns a map', () async {
+      final TimeSeries ts =
+          TimeSeries(key: config.db.emptyKey.append(TimeSeries, id: 'tsId'));
+      config.db.addValue(ts);
```
keyonghan, line 41: see comment below.

app_dart/test/request_handlers/push_benchmark_to_center_test.dart
```diff
@@ -0,0 +1,97 @@
...
+            'originId': 'devicelab',
+            'taskName': null,
+            'unit': null,
+            'name': null,
+            'archived': 'false',
+          },
+          'devicelab',
+        )
+      ]);
+    });
```
keyonghan, line 95: Test for `destination` result is missing.

app_dart/lib/src/request_handlers/push_benchmark_to_center.dart
```diff
@@ -0,0 +1,124 @@
...
+
+import '../datastore/cocoon_config.dart';
+import '../foundation/providers.dart';
+import '../model/appengine/time_series.dart';
+import '../model/appengine/time_series_value.dart';
+import '../request_handling/body.dart';
+import '../request_handling/request_handler.dart';
+import '../service/datastore.dart';
+
+/// Pushes devicelab benchmark results to Metrics Center storage.
```
keyonghan, line 19: It would be nice to add more doc. about `Metrics center storage`.

app_dart/pubspec.yaml
```diff
@@ -24,6 +24,10 @@ dependencies:
   http: ^0.12.0
   json_annotation: ^3.0.0
   meta: ^1.1.7
+  metrics_center:
+    # TODO(liyuqian): once metrics_center is properly reviewed, add it to
```
keyonghan, line 28: a tracking issue?

app_dart/lib/src/request_handlers/push_benchmark_to_center.dart
godofredoc: > The existing metrics handler only handles new data, @liyuqian and I would like to also transfer existing data to skia perf. Although we could export the existing data, however there's no easy way to process the exported data (leveldb log format).
> 
> Besides adding more burden to existing handler will increase its latency, complexity and decrease stability. For example, a failed pushing to metrics center could be caused by network interruptions temporarily, after the incident is over, there's no way to retry the pushing.

If that is the case then I'd recommend to split the process in two parts one for the data migration and the other one for just sending the data as it arrives.

Data migration will require different processes, maybe running for days, data validation on the target, running after work hours to minimize disruption, etc. 

For the second part the agent or the bots can call the two handlers with the same collected json.

app_dart/test/request_handlers/push_benchmark_to_center_test.dart
digiter: Done.

app_dart/lib/src/request_handlers/push_benchmark_to_center.dart
digiter: This PR is partially implementing the data migration process.

For the second part, do you mean letting cocoon agent calls `/api/push_benchmark_to_center` in addition to `/api/update-task-status`?

app_dart/lib/src/request_handlers/push_benchmark_to_center.dart
```diff
@@ -0,0 +1,144 @@
...
+  Future<Map<String, TimeSeries>> readTimeSeriesMap() async {
+    final Map<String, TimeSeries> map = <String, TimeSeries>{};
+    await for (TimeSeries ts in datastoreService.db.query<TimeSeries>().run()) {
+      // The id is string typed, see the defination of [TimeSeries].
+      map[ts.id as String] = ts;
+    }
+    return map;
+  }
+
+  /// Reads benchmark records with [createTimestamp] later than [sinceMillis].
```
liyuqian, line 81: Maybe change "later than" to "strictly later than" to clarify that we're doing ">" instead of ">=" comparison.

app_dart/lib/src/request_handlers/push_benchmark_to_center.dart
liyuqian: We can directly translate the payload in the future. I think this is mainly to migrate old metrics from cocoon to Skia perf for historical checks.

app_dart/lib/src/request_handlers/push_benchmark_to_center.dart
```diff
@@ -0,0 +1,144 @@
...
+  /// And limits the result size to no larger than [batchSize].
+  Future<List<TimeSeriesValue>> readTimeSeriesValue(int sinceMillis,
+      {int batchSize = 1000}) async {
+    final Query<TimeSeriesValue> query =
+        datastoreService.db.query<TimeSeriesValue>()
+          ..filter('createTimestamp >', sinceMillis)
+          ..limit(batchSize)
```
liyuqian, line 88: Do you think we'll need a more sophisticated batch size limit like https://github.com/liyuqian/metrics_center/blob/master/lib/src/flutter/center.dart#L121 ? Notice that we're returning points that are strictly later than a timestamp, and we would miss some points if many points have the same timestamp, and we cut the batch in the middle.

app_dart/lib/src/request_handlers/push_benchmark_to_center.dart
liyuqian: BTW, @kf6gpe has a plan to have the new metrics co-exist in both SkiaPerf/MetricsCenter and Cocoon. That way, he can compare between two dashboards' regression alerts, and make sure that SkiaPerf's alerts are properly set before sunsetting the Cocoon dashboard.

app_dart/lib/src/request_handlers/push_benchmark_to_center.dart
godofredoc: Actually reusing this handler for data migration is what I think we should avoid. Data migration should be addressed separately with a plan that includes: validation of the data migration in the target, how much resources is going to use, estimated duration of the migration, rollback plan, etc.

The dashboard is a production app that the team relies on and we need to avoid unnecessary risks as much as possible. For migration I'd suggest you create a tool that runs separately from the dashboard and we have more control over the processing rate it will be using.

app_dart/lib/src/request_handlers/push_benchmark_to_center.dart
godofredoc: > This PR is partially implementing the data migration process.
> 
> For the second part, do you mean letting cocoon agent calls `/api/push_benchmark_to_center` in addition to `/api/update-task-status`?

Yes, that is correct.

godofredoc:
Closing as outdated.

PR #789, opened 2020-06-02
[cocoon] Add timestamp for entity GithubBuildStatusUpdate

keyonghan:
This is to fix issue: https://github.com/flutter/flutter/issues/55377

### DIFF
--- a/app_dart/lib/src/model/appengine/github_build_status_update.dart
+++ b/app_dart/lib/src/model/appengine/github_build_status_update.dart
@@ -15,6 +15,7 @@ class GithubBuildStatusUpdate extends Model {
     this.head,
     this.status,
     this.updates,
+    this.updateTimestamp,
   }) {
     parentKey = key?.parent;
     id = key?.id;
@@ -39,6 +40,9 @@ class GithubBuildStatusUpdate extends Model {
   @IntProperty(propertyName: 'Updates', required: true)
   int updates;
 
+  @IntProperty(propertyName: 'UpdateTimestamp')
+  int updateTimestamp;
+
   @override
   String toString() {
     final StringBuffer buf = StringBuffer()
@@ -51,6 +55,7 @@ class GithubBuildStatusUpdate extends Model {
       ..write(', head: $head')
       ..write(', lastStatus: $status')
       ..write(', updates: $updates')
+      ..write(', updateTimestamp: $updateTimestamp')
       ..write(')');
     return buf.toString();
   }
--- a/app_dart/lib/src/model/appengine/github_build_status_update.dart
+++ b/app_dart/lib/src/model/appengine/github_build_status_update.dart
@@ -15,7 +15,7 @@ class GithubBuildStatusUpdate extends Model {
     this.head,
     this.status,
     this.updates,
-    this.updateTimestamp,
+    this.updateTimeMillis,
   }) {
     parentKey = key?.parent;
     id = key?.id;
@@ -40,8 +40,9 @@ class GithubBuildStatusUpdate extends Model {
   @IntProperty(propertyName: 'Updates', required: true)
   int updates;
 
-  @IntProperty(propertyName: 'UpdateTimestamp')
-  int updateTimestamp;
+  /// The last time when the status is updated for the PR.
+  @IntProperty(propertyName: 'UpdateTimeMillis')
+  int updateTimeMillis;
 
   @override
   String toString() {
@@ -55,7 +56,7 @@ class GithubBuildStatusUpdate extends Model {
       ..write(', head: $head')
       ..write(', lastStatus: $status')
       ..write(', updates: $updates')
-      ..write(', updateTimestamp: $updateTimestamp')
+      ..write(', updateTimeMillis: $updateTimeMillis')
       ..write(')');
     return buf.toString();
   }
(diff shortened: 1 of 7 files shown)


app_dart/lib/src/model/appengine/github_build_status_update.dart
```diff
@@ -39,6 +40,9 @@ class GithubBuildStatusUpdate extends Model {
   @IntProperty(propertyName: 'Updates', required: true)
   int updates;
 
+  @IntProperty(propertyName: 'UpdateTimestamp')
+  int updateTimestamp;
```
digiter, line 44: It's better to explain the purpose: "The last time when the status is applied to the PR".

Nit: rename it `appliedTimeMillis` to indicate the unit is millisecond.

app_dart/lib/src/request_handlers/push_build_status_to_github.dart
```diff
@@ -89,6 +89,7 @@ class PushBuildStatusToGithub extends ApiRequestHandler<Body> {
             await github.repositories.createStatus(slug, pr.head.sha, request);
             update.status = buildStatus.githubStatus;
             update.updates += 1;
+            update.updateTimestamp = DateTime.now().millisecondsSinceEpoch;
```
digiter, line 92: Add a test?

app_dart/lib/src/service/datastore.dart
```diff
@@ -208,12 +208,13 @@ class DatastoreService {
...
+        updateTimestamp: DateTime.now().millisecondsSinceEpoch,
       );
     } else {
       if (previousStatusUpdates.length > 1) {
-        throw StateError(
-            'GithubBuildStatusUpdate should have no more than one entries on '
-            'repository ${slug.fullName}, pr ${pr.number}, head ${pr.head.sha}');
+        return previousStatusUpdates.reduce((GithubBuildStatusUpdate current,
+                GithubBuildStatusUpdate next) =>
+            current.updateTimestamp < next.updateTimestamp ? next : current);
```
digiter, line 217: It would be less confusing to directly query the latest record, e.g. adding `.order('-timestamp').limit(1)` to the query.

app_dart/lib/src/model/appengine/github_build_status_update.dart
```diff
@@ -39,6 +40,9 @@ class GithubBuildStatusUpdate extends Model {
   @IntProperty(propertyName: 'Updates', required: true)
   int updates;
```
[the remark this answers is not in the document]
keyonghan: Currently we do not seem to use it, but it tells how many times the value has been updated. It shows the tree healthiness in some sense.

I do not have much context why this property is here.

app_dart/lib/src/model/appengine/github_build_status_update.dart
keyonghan: Added doc., and change name to `updateTimeMillis` (seems update is more appropriate?).

app_dart/lib/src/request_handlers/push_build_status_to_github.dart
keyonghan: Added.

app_dart/lib/src/service/datastore.dart
keyonghan: It is a good point, and I have thought about this as well. 

To enable `order().limt()`, a complex composite index is needed. From the query, there are already three filter options there. Building that index is heavy, and it doesn't cove old data. Additionally it will be rarely used if we add that index (duplicate cases happen rarely).

On the other hand, we have been trying to get rid of composite indexes as much as possible to lower datastore traffic.

So I have decided to use `reduce` instead, and it is only called when the rare duplicate cases occure.

Added some doc to make it clear.

app_dart/lib/src/request_handlers/push_engine_status_to_github.dart
```diff
@@ -82,6 +82,7 @@ class PushEngineStatusToGithub extends ApiRequestHandler<Body> {
           await github.repositories.createStatus(slug, pr.head.sha, request);
           update.status = latestStatus;
           update.updates += 1;
+          update.updateTimeMillis = DateTime.now().millisecondsSinceEpoch;
```
digiter, line 85: A test for this line?

keyonghan: Have added corresponding test for `/push_build_status_to_github` for the same functionality.

Since we are missing the test for this API, I have created an issue: https://github.com/flutter/flutter/issues/58613 to track all missing tests.

So will skip the test here.

digiter: Nit: instead of adding all the tests, we could just copy a small portion of `push_build_status_to_github_test.dart` to `push_engine_status_to_github_test.dart` to test this line.

app_dart/test/src/datastore/fake_datastore.dart
```diff
@@ -173,7 +173,8 @@ class FakeQuery<T extends Model> implements Query<T> {
     for (FakeFilterSpec filter in filters) {
       final String filterString = filter.filterString;
       final Object value = filter.comparisonObject;
-      if (filterString.contains('branch =') || filterString.contains('pr =')) {
+      if (filterString.contains('branch =') ||
+          filterString.contains('head =')) {
```
digiter, line 177: Why changing this line? Is it relevant to this PR?

keyonghan: Yes. This is for test `update if statuses have changed since last update - multiple branches`. 

Fake_datastore doesn't deal with `filter` for queries by default. `head` is added here since we need to test results filtered by it. (somehow the old `pr` doesn't work out)

app_dart/lib/src/request_handlers/push_engine_status_to_github.dart
keyonghan: Added.

PR #790, opened 2020-06-02
Capture stderr stream in the deploy script

iskakaushik:


iskakaushik:
It definitely worked for me locally. I'm not sure why its failing the presubmits though. Though, i've only redirected the streams for `pub` failures, not the `gcloud` failures.

It's still better than before but it doesn't cover all the commands.

iskakaushik:
The luci error seems unrelated. I'm landing this.

PR #791, opened 2020-06-02
Use timestmap of last commit rather than a fixed time period to query github commits API

keyonghan:
This PR fixes: https://github.com/flutter/flutter/issues/54284

It uses `timestamp` of last commit in datastore as the criteria to query Github api. This makes the query more accurate than the case using a fixed time period.

### DIFF
--- a/app_dart/lib/src/request_handlers/refresh_github_commits.dart
+++ b/app_dart/lib/src/request_handlers/refresh_github_commits.dart
@@ -61,13 +61,13 @@ class RefreshGithubCommits extends ApiRequestHandler<Body> {
     for (String branch in await config.flutterBranches) {
       final List<Commit> lastCommit =
           await datastore.queryRecentCommits(limit: 1, branch: branch).toList();
-      int lastCommitTimestamp = 0;
+      int lastCommitTimestampMills = 0;
       if (lastCommit.isNotEmpty) {
-        lastCommitTimestamp = lastCommit[0].timestamp;
+        lastCommitTimestampMills = lastCommit[0].timestamp;
       }
 
       final List<RepositoryCommit> commits = await githubService.listCommits(
-          slug, branch, lastCommitTimestamp, config);
+          slug, branch, lastCommitTimestampMills, config);
 
       final List<Commit> newCommits =
           await _getNewCommits(commits, datastore, branch);
--- a/app_dart/lib/src/request_handlers/refresh_github_commits.dart
+++ b/app_dart/lib/src/request_handlers/refresh_github_commits.dart
@@ -59,22 +59,16 @@ class RefreshGithubCommits extends ApiRequestHandler<Body> {
     final DatastoreService datastore = datastoreProvider(config.db);
 
     for (String branch in await config.flutterBranches) {
-      // TODO(keyonghan): use time of latest commit for [hours] in github API query, https://github.com/flutter/flutter/issues/54284
-      int hours = config.existingBranchHours;
-
-      if (branch != 'master') {
-        final Set<Commit> commits = await datastore
-            .queryRecentCommits(limit: 1, branch: branch)
-            .toSet();
-
-        /// Set [hours] to be longer if a new branch is found.
-        if (commits.isEmpty) {
-          hours = config.newBranchHours;
-        }
+      final List<Commit> lastCommit =
+          await datastore.queryRecentCommits(limit: 1, branch: branch).toList();
+      int lastCommitTimestamp = 0;
+      if (lastCommit.isNotEmpty) {
+        lastCommitTimestamp = lastCommit[0].timestamp;
       }
 
-      final List<RepositoryCommit> commits =
-          await githubService.listCommits(slug, branch, hours, config);
+      final List<RepositoryCommit> commits = await githubService.listCommits(
+          slug, branch, lastCommitTimestamp, config);
+
       final List<Commit> newCommits =
           await _getNewCommits(commits, datastore, branch);
 
@@ -88,7 +82,6 @@ class RefreshGithubCommits extends ApiRequestHandler<Body> {
       //Save [Commit] to BigQuery and create [Task] in Datastore.
       await _saveData(newCommits, datastore);
     }
-
     return Body.empty;
   }
 
--- a/app_dart/lib/src/request_handlers/refresh_github_commits.dart
+++ b/app_dart/lib/src/request_handlers/refresh_github_commits.dart
@@ -59,15 +59,15 @@ class RefreshGithubCommits extends ApiRequestHandler<Body> {
     final DatastoreService datastore = datastoreProvider(config.db);
 
     for (String branch in await config.flutterBranches) {
-      final List<Commit> lastCommit =
+      final List<Commit> lastProcessedCommit =
           await datastore.queryRecentCommits(limit: 1, branch: branch).toList();
       int lastCommitTimestampMills = 0;
-      if (lastCommit.isNotEmpty) {
-        lastCommitTimestampMills = lastCommit[0].timestamp;
+      if (lastProcessedCommit.isNotEmpty) {
+        lastCommitTimestampMills = lastProcessedCommit[0].timestamp;
       }
 
       final List<RepositoryCommit> commits = await githubService.listCommits(
-          slug, branch, lastCommitTimestampMills, config);
+          slug, branch, lastCommitTimestampMills);
 
       final List<Commit> newCommits =
           await _getNewCommits(commits, datastore, branch);
--- a/app_dart/lib/src/request_handlers/refresh_github_commits.dart
+++ b/app_dart/lib/src/request_handlers/refresh_github_commits.dart
@@ -61,6 +61,8 @@ class RefreshGithubCommits extends ApiRequestHandler<Body> {
     for (String branch in await config.flutterBranches) {
       final List<Commit> lastProcessedCommit =
           await datastore.queryRecentCommits(limit: 1, branch: branch).toList();
+
+      /// That [lastCommitTimestampMills] equals 0 means a new release branch is detected.
       int lastCommitTimestampMills = 0;
       if (lastProcessedCommit.isNotEmpty) {
         lastCommitTimestampMills = lastProcessedCommit[0].timestamp;
--- a/app_dart/lib/src/service/github_service.dart
+++ b/app_dart/lib/src/service/github_service.dart
@@ -14,8 +14,9 @@ class GithubService {
 
   /// Lists commits of the provided repository [slug] and [branch]. When
   /// [lastCommitTimestampMills] equals 0, it means a new release branch is
-  /// found and only the branched commit will be returned. Otherwise, it returns
-  /// all newer commits since [lastCommitTimestampMills].
+  /// found and only the branched commit will be returned for now, though the
+  /// rare case that multiple commits exist. For other cases, it returns all
+  /// newer commits since [lastCommitTimestampMills].
   Future<List<RepositoryCommit>> listCommits(
       RepositorySlug slug, String branch, int lastCommitTimestampMills) async {
     ArgumentError.checkNotNull(slug);
--- a/app_dart/lib/src/service/github_service.dart
+++ b/app_dart/lib/src/service/github_service.dart
@@ -51,6 +51,10 @@ class GithubService {
     }
 
     /// When a release branch is first detected only the most recent commit would be needed.
+    ///
+    /// If for the worst case, a new release branch consists of a useful cherry pick commit
+    /// which should be considered as well, here is the todo.
+    // TODO(keyonghan): https://github.com/flutter/flutter/issues/59275
     if (lastCommitTimestampMills == 0) {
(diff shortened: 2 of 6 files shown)


app_dart/lib/src/request_handlers/refresh_github_commits.dart
```diff
@@ -59,22 +59,16 @@ class RefreshGithubCommits extends ApiRequestHandler<Body> {
...
-        final Set<Commit> commits = await datastore
-            .queryRecentCommits(limit: 1, branch: branch)
-            .toSet();
-
-        /// Set [hours] to be longer if a new branch is found.
-        if (commits.isEmpty) {
-          hours = config.newBranchHours;
-        }
+      final List<Commit> lastCommit =
+          await datastore.queryRecentCommits(limit: 1, branch: branch).toList();
```
godofredoc, line 63: queryRecentCommits does not provide all the context of what the method is doing, should we rename it to queryNotProcessedCommits?

app_dart/lib/src/request_handlers/refresh_github_commits.dart
```diff
@@ -59,22 +59,16 @@ class RefreshGithubCommits extends ApiRequestHandler<Body> {
...
+      int lastCommitTimestampMills = 0;
+      if (lastCommit.isNotEmpty) {
+        lastCommitTimestampMills = lastCommit[0].timestamp;
       }
 
-      final List<RepositoryCommit> commits =
-          await githubService.listCommits(slug, branch, hours, config);
+      final List<RepositoryCommit> commits = await githubService.listCommits(
```
godofredoc, line 69: What happens if last commit is empty? should we just return here, most likely and empty commit means nothing to process.

app_dart/lib/src/service/github_service.dart
```diff
@@ -15,27 +15,40 @@ class GithubService {
...
   /// Lists the commits of the provided repository [slug] and [branch].
-  Future<List<RepositoryCommit>> listCommits(
-      RepositorySlug slug, String branch, int hours, Config config) async {
-    final Duration time = Duration(hours: hours);
+  Future<List<RepositoryCommit>> listCommits(RepositorySlug slug, String branch,
+      int lastCommitTimestampMills, Config config) async {
     ArgumentError.checkNotNull(slug);
     final PaginationHelper paginationHelper = PaginationHelper(github);
 
+    /// Return only one page when this is a new branch. Otherwise it will
```
godofredoc, line 23: maybe expand a little bit on what a page means in this context?

app_dart/lib/src/service/github_service.dart
```diff
@@ -15,27 +15,40 @@ class GithubService {
...
+    /// return all commits prior to this release branch commit, leading to
+    /// heavy workload.
+    int pages;
+    if (lastCommitTimestampMills == 0) {
```
godofredoc, line 27: What happens if this check fails? should we add a default for pages?

app_dart/lib/src/service/github_service.dart
```diff
@@ -15,27 +15,40 @@ class GithubService {
...
       },
+      pages: pages,
     )) {
       commits.addAll((json.decode(response.body) as List<dynamic>)
           .cast<Map<String, dynamic>>());
     }
 
     /// Take the latest single commit for a new release branch.
-    if (hours == config.newBranchHours) {
+    if (lastCommitTimestampMills == 0) {
```
godofredoc, line 51: Under which conditions are we expecting this to be true?

app_dart/lib/src/request_handlers/refresh_github_commits.dart
```diff
@@ -59,22 +59,16 @@ class RefreshGithubCommits extends ApiRequestHandler<Body> {
...
-      if (branch != 'master') {
-        final Set<Commit> commits = await datastore
-            .queryRecentCommits(limit: 1, branch: branch)
-            .toSet();
-
-        /// Set [hours] to be longer if a new branch is found.
-        if (commits.isEmpty) {
-          hours = config.newBranchHours;
-        }
+      final List<Commit> lastCommit =
```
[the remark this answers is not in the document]
keyonghan: Updated.

app_dart/lib/src/request_handlers/refresh_github_commits.dart
keyonghan: The returned results are commits that have been processed. Other APIs are calling `queryRecentCommits` as well.

To me, it does what it suggests: query the recent commits in datastore filtered by the `limit` number and the `branch`.

Here it returns the most recent single (processed) commit in datastore, which is expected.

Please let me know if I understand incorrectly.

app_dart/lib/src/request_handlers/refresh_github_commits.dart
keyonghan: If `lastProcessedCommit ` is empty, it is possible that a new release branch is available. We still need to query github to insert the corresponding data.

app_dart/lib/src/service/github_service.dart
keyonghan: Added doc: `The [pages] defines the number of pages of returned http request results.`

app_dart/lib/src/service/github_service.dart
keyonghan: It will return all results. No need for a default value: https://github.com/SpinlockLabs/github.dart/blob/master/lib/src/common/util/pagination.dart#L54

The `null` has been taken care of.

app_dart/lib/src/service/github_service.dart
keyonghan: As mentioned in the doc., this happens when a new release branch is found.

app_dart/lib/src/service/github_service.dart
```diff
@@ -15,27 +15,40 @@ class GithubService {
   final GitHub github;
 
   /// Lists the commits of the provided repository [slug] and [branch].
-  Future<List<RepositoryCommit>> listCommits(
-      RepositorySlug slug, String branch, int hours, Config config) async {
-    final Duration time = Duration(hours: hours);
+  Future<List<RepositoryCommit>> listCommits(RepositorySlug slug, String branch,
+      int lastCommitTimestampMills, Config config) async {
```
digiter, line 19: The config is no longer used.

app_dart/lib/src/service/github_service.dart
```diff
@@ -15,27 +15,40 @@ class GithubService {
   final GitHub github;
 
   /// Lists the commits of the provided repository [slug] and [branch].
```
digiter, line 17: Missing documentation about how this function behaves against the `lastCommitTimestampMills` parameter.

app_dart/lib/src/service/github_service.dart
```diff
@@ -15,27 +15,40 @@ class GithubService {
...
+    /// heavy workload.
+    int pages;
+    if (lastCommitTimestampMills == 0) {
+      pages = 1;
+    }
+
     List<Map<String, dynamic>> commits = <Map<String, dynamic>>[];
+
+    /// [lastCommitTimestamp+1] excludes last commit itself.
     await for (Response response in paginationHelper.fetchStreamed(
```
digiter, line 34: Nit: comment the url to the github api: https://developer.github.com/v3/repos/commits/#list-commits

app_dart/lib/src/service/github_service.dart
```diff
@@ -15,27 +15,40 @@ class GithubService {
...
+      pages: pages,
     )) {
       commits.addAll((json.decode(response.body) as List<dynamic>)
           .cast<Map<String, dynamic>>());
     }
 
     /// Take the latest single commit for a new release branch.
-    if (hours == config.newBranchHours) {
+    if (lastCommitTimestampMills == 0) {
       commits = commits.take(1).toList();
```
digiter, line 52: There could be more than one commits in a new release branch.

app_dart/lib/src/service/github_service.dart
keyonghan: Removed.

app_dart/lib/src/service/github_service.dart
keyonghan: Added.

app_dart/lib/src/service/github_service.dart
keyonghan: added.

app_dart/lib/src/service/github_service.dart
keyonghan: Only the most recent commit is needed: https://github.com/flutter/cocoon/pull/734#discussion_r405195385

app_dart/lib/src/request_handlers/refresh_github_commits.dart
godofredoc: queryRecentCommits is just grabbing all the commits on datastore and returning the one on the top when ordering by timestamp. That means that we are potentially processing the same commit several times. I'd suggest to improve the query to return only commits that haven't been processed yet.

app_dart/lib/src/request_handlers/refresh_github_commits.dart
godofredoc: If this is the only case I'd recommend adding docs to make it clear lastProcessedCommit==0 means new branch.

app_dart/lib/src/service/github_service.dart
digiter: That's a heuristic and I understand that works based on the assumption that Cocoon processes every new commit. However if Cocoon is down for a while (the reason of this PR), there can be multiple unprocessed commits in the new branch.

Also it'd better to explain every heuristics in comment, so that future readers won't have trouble in understanding the assumption and reasoning.

app_dart/lib/src/request_handlers/refresh_github_commits.dart
keyonghan: Yes. That is the only case. I have added the doc.

app_dart/lib/src/request_handlers/refresh_github_commits.dart
keyonghan: I think there is some misunderstanding here.

We are trying to get the latest commit that has been processed in datastore, rather than get the commits that haven't been processed. All commits in datastore are already processed.

Then we use this commit's timestamp as the benchmark to call Github API to return newer commits.

app_dart/lib/src/service/github_service.dart
keyonghan: Ah, you meant new commits submitted to the branch, that has branched out from main. This case does exist, but it usually occurs after couple of days since branching out.

It is a rare case that our cocoon service keeps down for days, but this corner case still may happen. Considering this, I would prefer to have this as a TODO, since this is not an easy implementation and we are not in need of this urgently.

At the same time, have added doc to explain.

keyonghan: TODO added.

app_dart/lib/src/service/github_service.dart
```diff
@@ -7,35 +7,55 @@ import 'dart:convert' show json;
...
   const GithubService(this.github);
 
   final GitHub github;
 
-  /// Lists the commits of the provided repository [slug] and [branch].
+  /// Lists commits of the provided repository [slug] and [branch]. When
+  /// [lastCommitTimestampMills] equals 0, it means a new release branch is
+  /// found and only the branched commit will be returned. Otherwise, it returns
+  /// all newer commits since [lastCommitTimestampMills].
   Future<List<RepositoryCommit>> listCommits(
```
digiter, line 19: The description doesn't mention the fact that when a new release branch is found, there can be more than one branched commits in rare cases.

app_dart/test/request_handlers/refresh_github_commits_test.dart
```diff
@@ -172,6 +171,15 @@ void main() {
...
       config.flutterBranchesValue = <String>['master'];
+
+      /// Pre-insert one commit first, otherwise it will insert only one
+      /// commit for a new branch.
+      const List<String> dbCommits = <String>['4'];
+      for (String sha in dbCommits) {
+        final Commit commit = shaToCommit(sha, 'master');
+        db.values[commit.key] = commit;
+      }
+
```
digiter, line 182: 1. The comment doesn't address why a pre-insert is needed. The reason is that this test is simulating an existing branch, which must already have at least one processed commit in the datastore.
2. The existing processed commit should have smaller timestamp than `githubCommits`.
3. Since there's only one element, there's no need for a loop.

app_dart/lib/src/service/github_service.dart
keyonghan: Updated.

app_dart/test/request_handlers/refresh_github_commits_test.dart
keyonghan: 1. updated.
2. updated existing commit to have smaller timestamp.
3. removed loop.

keyonghan:
@digiter can you please unblock the PR by marking the requested change completed?

PR #792, opened 2020-06-03
Add nag for engine tests

xster:


app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -26,7 +26,9 @@ const Set<String> kNeedsCQLabelList = <String>{'flutter/flutter'};
 const Set<String> kNeedsCheckGoldenTriage = <String>{'flutter/flutter'};
 
 /// List of repos that require check for labels and tests.
-const Set<String> kNeedsCheckLabelsAndTests = <String>{'flutter/flutter'};
+const Set<String> kNeedsCheckLabelsAndTests = <String>{'flutter/flutter', 'flutter/engine'};
+
+final RegExp kEngineTestRegExp = RegExp(r'tests?\.(dart|java|mm|m|cc)$');
```
dnfield, line 31: this is not the only place tests live - for example, C++ tests live along side their counterparts, conventionally named `*_unittests.cc`.

xster: this covers that case no?

xster: https://github.com/flutter/cocoon/pull/792/files#diff-304fad7784d8986a3af98d8752b20bc4R502 is what you meant right? (added a specific test for that just to make sure I'm not going crazy)

dnfield: Ahh yeah, ok. I misread the regexp.

xster:
@godofredoc do we need to deploy the cocoon code somehow or is it built per commit already?

godofredoc:
It requires a manual deployment.

xster:
Are there instructions I can follow? Or you guys will just do them on a regular cadence?

godofredoc:
> Are there instructions I can follow? Or you guys will just do them on a regular cadence?

We have been doing it per request, let me push a new version

xster:
Thanks!

godofredoc:
Released in version 403

xster:
Seems to work. Thanks @godofredoc

PR #798, opened 2020-06-04
Create common test utils

keyonghan:
This PR simplifies tests by putting common codes/classes in a util file.

Related issue: https://github.com/flutter/flutter/issues/53280.


app_dart/test/src/utils.dart
```diff
@@ -0,0 +1,75 @@
+// Copyright 2020 The Flutter Authors. All rights reserved.
```
godofredoc, line 1: Should we call this file mocks.dart inside a utilities folder?

keyonghan: Done.

PR #799, opened 2020-06-05
Avoid unwanted branches scan

keyonghan:
This PR skips unwanted branches scan by changing the general branch regexp to more detailed format.

Related issue:
https://github.com/flutter/flutter/issues/58627
https://github.com/flutter/flutter/issues/55579

### DIFF
--- a/app_dart/dev/branch_regexps.txt
+++ b/app_dart/dev/branch_regexps.txt
@@ -1,2 +1,4 @@
-^flutter-[0-9]+\.[0-9]+-candidate\.[0-9]+$
+^flutter-1+\.17+-candidate\.3+$
+^flutter-1+\.18+-candidate\.11+$
+^flutter-1+\.19+-candidate\.4+$
 master
--- a/app_dart/README.md
+++ b/app_dart/README.md
@@ -114,4 +114,4 @@ $ dart dev/deploy.dart --help
 
 ### Branching support for flutter repo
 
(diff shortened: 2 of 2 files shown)


app_dart/dev/branch_regexps.txt
```diff
@@ -1,2 +1,4 @@
-^flutter-[0-9]+\.[0-9]+-candidate\.[0-9]+$
+^flutter-1+\.17+-candidate\.3+$
+^flutter-1+\.18+-candidate\.11+$
+^flutter-1+\.19+-candidate\.4+$
```
godofredoc, line 3: This means we'll still call the github api instead of using the names here. I'd suggest to use directly:

flutter-1.17-candidate.3
flutter-1.18-candidate.11
flutter-1.19-candidate.4

And use the content of the file directly.

keyonghan: Yes. You are right. It will still call the github api.

The main github traffic should fall on follow-up github prs/commits calls when iterating branch list.

The change here should reduce the github traffic a lot, but I will open another PR to remove the github call for the branch list.

christopherfujino: @keyonghan we don't need the `+`'s in these regexes, right?

christopherfujino: nvm, I see with your followup PR you're gonna remove this anyway.

PR #800, opened 2020-06-05
Removing Gold checks on non-master branches

Piinks:
Related issues: https://github.com/flutter/flutter/issues/58627 https://github.com/flutter/flutter/issues/58814

This removes the gold check from branches that are not staged to land on master (i.e. release branches) We've been hitting Github API rate limits, so this should help cut back on those requests.

PR #801, opened 2020-06-05
Updating log info for Gold

Piinks:
I merged #800 before seeing @keyonghan's comment to fix the log. 😅  Adds some extra logging as we have changed the skip case to reduce github requests.

### DIFF
--- a/app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
+++ b/app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
@@ -67,7 +67,7 @@ class PushGoldStatusToGithub extends ApiRequestHandler<Body> {
           '${lastUpdate.head}, status: ${lastUpdate.status}, description: ${lastUpdate.description}');
 
       if (lastUpdate.status == GithubGoldStatusUpdate.statusCompleted &&
-              lastUpdate.head == pr.head.sha) {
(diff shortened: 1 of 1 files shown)


Piinks:
It looks like there is an unrelated golden file failure, is master broken?

keyonghan:
> It looks like there is an unrelated golden file failure, is master broken?

That failed test seems to be flaky. I have seen it now and then. Will file a bug.

keyonghan:
Here it is
https://github.com/flutter/flutter/issues/58833

app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
```diff
@@ -69,7 +69,9 @@ class PushGoldStatusToGithub extends ApiRequestHandler<Body> {
       if ((lastUpdate.status == GithubGoldStatusUpdate.statusCompleted &&
               lastUpdate.head == pr.head.sha) ||
           (pr.base.ref != 'master')) {
-        log.debug('Completed status already reported for this commit.');
+        log.debug(pr.base.ref == 'master'
```
christopherfujino, line 72: drive by nit: this whole code would be more readable by changing line 69 to an `if`...`else if`, with each having an explicit log message.

PR #802, opened 2020-06-05
[cocoon] Avoid GitHub API call after changing branch regexps to branches

keyonghan:
This PR addressed comment: https://github.com/flutter/cocoon/pull/799#discussion_r436174916

Cocoon is using fixed branch name list to iterate wanted branches after PR: https://github.com/flutter/cocoon/pull/799. However it still keeps querying GitHub API to unnecessarily screen through all available branches. This helps nothing but wasting API calls.

This PR removes the above GitHub api call, and uses branches directly.

### DIFF
--- a/app_dart/lib/src/foundation/utils.dart
+++ b/app_dart/lib/src/foundation/utils.dart
@@ -67,10 +67,8 @@ Future<List<String>> loadBranches(HttpClientProvider branchHttpClientProvider,
   return <String>['master'];
 }
 
-Future<Uint8List> getBranches(
-    HttpClientProvider branchHttpClientProvider,
-    Logging log,
-    GitHubBackoffCalculator gitHubBackoffCalculator) async {
+Future<Uint8List> getBranches(HttpClientProvider branchHttpClientProvider,
+    Logging log, GitHubBackoffCalculator gitHubBackoffCalculator) async {
   final List<String> branches = await loadBranches(
       branchHttpClientProvider, log, gitHubBackoffCalculator);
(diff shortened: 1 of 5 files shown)


app_dart/lib/src/foundation/utils.dart
```diff
@@ -69,28 +64,10 @@ Future<List<String>> loadBranchRegExps(
...
-    HttpClientProvider branchHttpClientProvider,
-    Logging log,
-    GitHubBackoffCalculator gitHubBackoffCalculator) async {
-  final GithubService githubService = await config.createGithubService();
-  final GitHub github = githubService.github;
-  final RepositorySlug slug = RepositorySlug('flutter', 'flutter');
-  final Stream<Branch> branchList = github.repositories.listBranches(slug);
-  final List<String> regExps = await loadBranchRegExps(
+Future<Uint8List> getBranches(HttpClientProvider branchHttpClientProvider,
+    Logging log, GitHubBackoffCalculator gitHubBackoffCalculator) async {
```
godofredoc, line 68: I think we can remove gitHubBackoffCalculator

keyonghan: `gitHubBackoffCalculator` is for http request. We are still reading branch contents of `branches.txt` via http client.

This can be removed if we consider applying the retry logic.

godofredoc: ok, seems we may want to call it http backoff calculator in the future.

PR #807, opened 2020-06-11
Discard state pubsub message for builders that were deleted.

godofredoc:
When a builder is deleted there may be still pending or running tasks
that will try to update state. GCP pubsub will try to deliver the
message several times potentially generating several 500 http status.

Bug:
  https://github.com/flutter/flutter/issues/59270

### DIFF
--- a/app_dart/lib/src/datastore/cocoon_config.dart
+++ b/app_dart/lib/src/datastore/cocoon_config.dart
@@ -394,6 +394,9 @@ class Config {
       orElse: () => <String, dynamic>{'repo': ''},
     );
     final String repoName = builderConfig['repo'] as String;
+    // If there is no builder config for the builderName then we
+    // return null. This is to allow the code calling this method
+    // to skip changes that depend on a builder configuration.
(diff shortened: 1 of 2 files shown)


app_dart/lib/src/datastore/cocoon_config.dart
```diff
@@ -389,9 +389,14 @@ class Config {
 
   Future<RepositorySlug> repoNameForBuilder(String builderName) async {
     final List<Map<String, dynamic>> builders = luciTryBuilders;
-    final String repoName = builders.firstWhere(
-        (Map<String, dynamic> builder) =>
-            builder['name'] == builderName)['repo'] as String;
+    final Map<String, dynamic> builderConfig = builders.firstWhere(
```
keyonghan, line 392: Nit: some doc would be helpful for the empty case.

godofredoc: > Nit: some doc would be helpful for the empty case.

Added docs.

PR #809, opened 2020-06-12
Fix return type for getRepoForBuilder and add tests.

godofredoc:
Bug:
  https://github.com/flutter/flutter/issues/59270

### DIFF
--- a/app_dart/test/datastore/cocoon_config_test.dart
+++ b/app_dart/test/datastore/cocoon_config_test.dart
@@ -1,0 +1,26 @@
+import 'package:cocoon_service/cocoon_service.dart';
+import 'package:github/github.dart';
+import 'package:test/test.dart';
+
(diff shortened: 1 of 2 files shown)


app_dart/test/datastore/cocoon_config_test.dart
```diff
@@ -0,0 +1,30 @@
...
+    });
+    test('Builder config does not exist', () async {
+      final RepositorySlug result =
+          await config.repoNameForBuilder('DoesNotExist');
+      expect(result, isNull);
+    });
+
+    test('Builder exists', () async {
+      final RepositorySlug result = await config.repoNameForBuilder('Cocoon');
+      expect(result, isNotNull);
```
keyonghan, line 27: Using the real value?

godofredoc: Added validation for repo name given that this is an object.

PR #814, opened 2020-06-23
Apply vacuum clean to branch tasks

keyonghan:
Existing `vacuum-clean` api runs on `master` tasks only. This causes unhealthy status for tasks of release branches.

This PR updates `vacuum-clean` api to consider release branches as well.

Related issue: https://github.com/flutter/flutter/issues/60056

### DIFF
--- a/app_dart/lib/src/service/datastore.dart
+++ b/app_dart/lib/src/service/datastore.dart
@@ -166,6 +166,20 @@ class DatastoreService {
     }
   }
 
+  // Queries for recent tasks without considering branches.
+  Stream<FullTask> queryRecentTasksNoBranch(
+      {int commitLimit = 20, int taskLimit = 20}) async* {
+    assert(commitLimit != null);
+    assert(taskLimit != null);
+    await for (Commit commit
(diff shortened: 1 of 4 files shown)


app_dart/lib/src/service/datastore.dart
```diff
@@ -166,6 +166,20 @@ class DatastoreService {
     }
   }
 
+  // Queries for recent tasks without considering branches.
+  Stream<FullTask> queryRecentTasksNoBranch(
```
godofredoc, line 170: A new test in datastore_test for this new method?

app_dart/test/request_handlers/vacuum_clean_test.dart
```diff
@@ -0,0 +1,113 @@
+// Copyright 2019 The Flutter Authors. All rights reserved.
```
godofredoc, line 1: 2020?

keyonghan: Updated.

app_dart/lib/src/service/datastore.dart
keyonghan: Added.

PR #818, opened 2020-06-29
Update luci build number list

keyonghan:
This PR simplifies luci number support for luci reruns.

Existing strategy keeps track of different luci rerun builds by maintaining multiple task records in datastore.

With this PR, only one task record is needed: buildNumberList keeps track of all different builds. This will also make frontend change easiler.

### DIFF
--- a/app_dart/lib/src/model/appengine/task.dart
+++ b/app_dart/lib/src/model/appengine/task.dart
@@ -141,7 +141,12 @@ class Task extends Model {
   @JsonKey(name: 'BuildNumber')
   int buildNumber;
 
-  /// The build number list (reruns) of luci build.
+  /// The build number list of luci builds: comma joined string of
+  /// different build numbers. 
+  /// 
+  /// For the case with single run 123, [buildNumberList] = '123';
+  /// For the case with multiple reruns 123, 456, 789, 
+  /// [buildNumberList] = '123,456,789'.
   @StringProperty(propertyName: 'BuildNumberList')
   @JsonKey(name: 'BuildNumberList')
   String buildNumberList;
--- a/app_dart/lib/src/model/appengine/task.dart
+++ b/app_dart/lib/src/model/appengine/task.dart
@@ -33,6 +33,7 @@ class Task extends Model {
     this.reservedForAgentId = '',
     this.stageName,
     this.buildNumber,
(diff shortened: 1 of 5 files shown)


app_dart/lib/src/model/appengine/task.dart
```diff
@@ -140,6 +141,11 @@ class Task extends Model {
   @JsonKey(name: 'BuildNumber')
   int buildNumber;
 
+  /// The build number list (reruns) of luci build.
```
godofredoc, line 144: Can we expand this doc a little bit more to explain the structure and one example?

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
```diff
@@ -71,38 +71,36 @@ class RefreshChromebotStatus extends ApiRequestHandler<Body> {
...
-    /// and [luciTasks] from old to new. If matched, then update accordingly.
-    for (FullTask datastoreTask in datastoreTasks.reversed) {
-      if (!luciTasks.containsKey(datastoreTask.commit.sha)) {
-        continue;
-      }
-      for (LuciTask luciTask in luciTasks[datastoreTask.commit.sha].reversed) {
-        if (!updatedLuciTasks.contains(luciTask) &&
-            _buildNumberMatched(datastoreTask, luciTask)) {
+    /// Update [devicelabTask] when first [luciTask] run finishes. There may be
+    /// reruns for the same commit and same builder. Update [devicelabTask]
```
godofredoc, line 84: What happens if there is a rerun? are they ignored?

app_dart/lib/src/model/appengine/task.dart
keyonghan: Done.

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
keyonghan: Suppose the first run with build number 123, then `builderNumberList='123'`.
When rerun happens with build number 456, the `builderNumberList` will be updated to `123,456`.

The status will be updated to the status of the new run of build 456 if changed.

With newly added doc. in `Task`, this should be easy to follow now.

PR #819, opened 2020-06-30
Add packages repository.

godofredoc:
LUCI support for packages repository was removed in one of the many
refactorings. This PR is adding it back.


godofredoc:
\cc @nturgut this is the first step to add LUCI support for a repo.

PR #826, opened 2020-07-02
Add new properties to Task in app flutter appengine

keyonghan:
This is a follow-up pr of https://github.com/flutter/cocoon/pull/820

In addition to adding properties to Task in app_dart, we also need to add those in app_flutter for frontend usage.

### DIFF
--- a/app_flutter/lib/service/appengine_cocoon.dart
+++ b/app_flutter/lib/service/appengine_cocoon.dart
@@ -421,8 +421,8 @@ class AppEngineCocoonService implements CocoonService {
       ..reservedForAgentId = taskData['ReservedForAgentID']
       ..stageName = taskData['StageName']
       ..status = taskData['Status']
-      ..buildNumberList = taskData['BuildNumberList']??''
-      ..builderName = taskData['BuilderName']??''
-      ..luciPoolName = taskData['LuciPoolName']??'';
+      ..buildNumberList = taskData['BuildNumberList'] ?? ''
+      ..builderName = taskData['BuilderName'] ?? ''
+      ..luciPoolName = taskData['LuciPoolName'] ?? '';
   }
 }
--- a/app_flutter/lib/service/appengine_cocoon.dart
+++ b/app_flutter/lib/service/appengine_cocoon.dart
@@ -423,6 +423,6 @@ class AppEngineCocoonService implements CocoonService {
       ..status = taskData['Status']
       ..buildNumberList = taskData['BuildNumberList'] ?? ''
       ..builderName = taskData['BuilderName'] ?? ''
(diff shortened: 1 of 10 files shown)


app_flutter/lib/service/appengine_cocoon.dart
```diff
@@ -420,6 +420,9 @@ class AppEngineCocoonService implements CocoonService {
       ..requiredCapabilities.add(objectRequiredCapabilities.toString())
       ..reservedForAgentId = taskData['ReservedForAgentID']
       ..stageName = taskData['StageName']
-      ..status = taskData['Status'];
+      ..status = taskData['Status']
+      ..buildNumberList = taskData['BuildNumberList'] ?? ''
+      ..builderName = taskData['BuilderName'] ?? ''
+      ..luciPoolName = taskData['LuciPoolName'] ?? '';
```
godofredoc, line 426: This should be bucket instead of pool name.

app_flutter/lib/service/dev_cocoon.dart
```diff
@@ -246,6 +246,9 @@ class DevelopmentCocoonService implements CocoonService {
       ..requiredCapabilities.add('[linux/android]')
       ..reservedForAgentId = 'linux1'
       ..stageName = stageName
-      ..status = status;
+      ..status = status
+      ..buildNumberList = '$index'
+      ..builderName = 'Linux'
+      ..luciPoolName = 'luci.flutter.prod';
```
godofredoc, line 252: Same here, bucket.

app_flutter/lib/service/dev_cocoon.dart
```diff
@@ -246,6 +246,9 @@ class DevelopmentCocoonService implements CocoonService {
       ..requiredCapabilities.add('[linux/android]')
       ..reservedForAgentId = 'linux1'
       ..stageName = stageName
-      ..status = status;
+      ..status = status
+      ..buildNumberList = '$index'
```
CaseyHillers, line 250: This will update every task to have these values. Check if the the task will be a luci task before assigning these.

app_flutter/lib/service/appengine_cocoon.dart
keyonghan: Done.

app_flutter/lib/service/dev_cocoon.dart
keyonghan: Good point. Added luci check before assigning.

keyonghan:
> Don't forget to add tests! The appengine_cocoon has a pretty robust test since this adapter code does some tricky casts to make the proto

Fixed the test as well.

PR #828, opened 2020-07-06
Schedule luci builds using CheckSuite and CheckRun events.

godofredoc:
This is required to add functionality to re-trigger tasks on failure.

Bug:

### DIFF
--- a/app_dart/lib/src/request_handlers/luci_status.dart
+++ b/app_dart/lib/src/request_handlers/luci_status.dart
@@ -72,11 +72,9 @@ class LuciStatusHandler extends RequestHandler<Body> {
     final PushMessageEnvelope envelope = PushMessageEnvelope.fromJson(
       json.decode(requestString) as Map<String, dynamic>,
     );
-    final BuildPushMessage buildMessage = BuildPushMessage.fromJson(
+    final BuildPushMessage buildPushMessage = BuildPushMessage.fromJson(
         json.decode(envelope.message.data) as Map<String, dynamic>);
-    final Build build = buildMessage.build;
-    final Map<String, dynamic> userData =
-        jsonDecode(buildMessage.userData) as Map<String, dynamic>;
+    final Build build = buildPushMessage.build;
     final String builderName = build.tagsByName('builder').single;
     final RepositorySlug slug = await config.repoNameForBuilder(builderName);
 
@@ -91,15 +89,13 @@ class LuciStatusHandler extends RequestHandler<Body> {
         .tagsByName('buildset')
         .firstWhere((String tag) => tag.startsWith(shaPrefix))
         .substring(shaPrefix.length);
-    log.debug('Setting status: ${buildMessage.toJson()} for $builderName');
-    switch (buildMessage.build.status) {
+    log.debug('Setting status: ${buildPushMessage.toJson()} for $builderName');
+    switch (buildPushMessage.build.status) {
       case Status.completed:
-        await _rescheduleOrMarkCompleted(
+        await _markCompleted(
           sha: sha,
           builderName: builderName,
           build: build,
-          retries: userData['retries'] as int,
-          luciBuildService: luciBuildService,
           githubStatusService: githubStatusService,
           slug: slug,
         );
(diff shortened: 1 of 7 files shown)


app_dart/lib/src/request_handlers/luci_status.dart
```diff
@@ -123,47 +119,16 @@ class LuciStatusHandler extends RequestHandler<Body> {
   /// Reschedules jobs that failed for infra reasons up to
   /// [CocoonConfig.luciTryInfraFailureRetries] times, and updates statuses on
   /// GitHub for all other cases.
-  Future<void> _rescheduleOrMarkCompleted({
+  Future<void> _markCompleted({
```
digiter, line 122: The function doc needs update.

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -4,7 +4,12 @@
 
 import 'dart:convert';
 
+import 'package:appengine/appengine.dart';
+import 'package:cocoon_service/src/foundation/github_checks_util.dart';
+import 'package:cocoon_service/src/model/github/checks.dart';
 import 'package:cocoon_service/src/request_handling/exceptions.dart';
+import 'package:github/github.dart' as github;
+import 'package:github/github.dart';
```
digiter, line 12: These two imports can be merged.

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -17,22 +22,34 @@ import 'buildbucket.dart';
...
     Status.infraFailure
   };
 
-  /// Returns a map of the BuildBucket builds for a given [repositoryName]
+  /// Sets the appengine [log] used by this class to log debug and error
+  /// messages. This method has to be called before any other method in this
+  /// class.
+  void setLogger(Logging log) {
+    this.log = log;
+  }
```
digiter, line 46: Nit: such logic usually sits in the constructor or a function named `init`.

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -224,27 +265,131 @@ class LuciBuildService {
...
+      slug,
+      checkRunEvent.checkRun.name,
+      pr.head.sha,
+    );
+    userData['check_suite_id'] = checkRunEvent.checkRun.checkSuite.id;
+    userData['check_run_id'] = githubCheckRun.id;
+    userData['repo_owner'] = slug.owner;
+    userData['repo_name'] = slug.name;
+    userData['user_agent'] = 'flutter-cocoon';
+    userData['retries'] = 1;
```
keyonghan, line 314: It is a little bit confusing with initialization 1 in a rescheduleXXX function without any doc explanation.

Is this the initialization value for any new run or any reschedule? If yes, the following `retries` is meant for re-runs of any new run or any reschedule run?

People may easily confuse the reschedule with retries. Some doc. would be helpful.

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -214,7 +255,7 @@ class LuciBuildService {
   Future<bool> rescheduleBuild({
     @required String commitSha,
     @required String builderName,
-    @required push_message.Build build,
+    @required push_message.BuildPushMessage buildPushMessage,
     @required int retries,
   }) async {
     if (retries >= config.luciTryInfraFailureRetries) {
```
keyonghan, line 260: shall this `retries` be removed and use `userData['retries']` instead in the condition?

app_dart/lib/src/service/luci_build_service.dart
godofredoc: The reason for this is that the logger depends on the AppEngine context which only exists when the handler is running. I couldn't find any other way to pass the logger without major refactorings.

app_dart/lib/src/service/luci_build_service.dart
godofredoc: Done!

app_dart/lib/src/request_handlers/luci_status.dart
godofredoc: Updated.

app_dart/lib/src/service/luci_build_service.dart
godofredoc: This will be removed on https://github.com/flutter/flutter/issues/60942 to avoid breaking functionality.

app_dart/lib/src/service/luci_build_service.dart
godofredoc: Added docs. Actually retries will be removed as devs will be able to rerun builds manually if something fails.

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -359,6 +359,10 @@ class LuciBuildService {
     userData['repo_owner'] = slug.owner;
     userData['repo_name'] = slug.name;
     userData['user_agent'] = 'flutter-cocoon';
+    // Retries were used to auto re-run builds when they failed with infra
+    // failure. Now with github checks api support automated retries won't be
+    // needed anymore and will be removed:
+    // https://github.com/flutter/flutter/issues/60942.
```
keyonghan, line 365: Nit: you may want to follow the TODO format.

godofredoc: Done!

PR #830, opened 2020-07-07
Remove retries functionality.

godofredoc:
In preparation to land github checks api functionality we are removing
the auto retries.

Bug:
  https://github.com/flutter/flutter/issues/60942


app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -148,7 +148,10 @@ class LuciBuildService {
         bucket: 'try',
         builder: builder,
       );
-      final Map<String, dynamic> userData = <String, dynamic>{'retries': 0};
+      final Map<String, dynamic> userData = <String, dynamic>{};
+      userData['repo_owner'] = slug.owner;
+      userData['repo_name'] = slug.name;
+      userData['user_agent'] = 'flutter-cocoon';
```
digiter, line 154: Nit: inline the assignments to the construction of the map.

keyonghan:
Considering devicelab tasks, we are maintaining auto retries with a second chance to get rid of flaky cases. Shall we keep that here as well?

app_dart/lib/src/service/luci_build_service.dart
godofredoc: Done!

> Considering devicelab tasks, we are maintaining auto retries with a second chance to get rid of flaky cases. Shall we keep that here as well?

Many of these tasks take over 45 mins to run, having the retries makes them non deterministic and it is really difficult to find the flakiness. Additionally PROD tasks do not have retries and are very stable with a couple of exceptions,

PR #834, opened 2020-07-08
Only process check updates if messages come from a supported repo.

godofredoc:
The implementation of the checks api will be one repo at a time and we
need to support repos with checks api support and the ones no yet
migrated. This PR checks is the status message comes with all the
expected fields to decide whether to process checks api updates or not.

Bug:

### DIFF
--- a/app_dart/lib/src/request_handlers/luci_status.dart
+++ b/app_dart/lib/src/request_handlers/luci_status.dart
@@ -46,7 +46,10 @@ class LuciStatusHandler extends RequestHandler<Body> {
   /// Creates an endpoint for listening to LUCI status updates.
   const LuciStatusHandler(
     Config config,
-    this.buildBucketClient, {
+    this.buildBucketClient,
+    this.luciBuildService,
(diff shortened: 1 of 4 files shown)


app_dart/lib/src/request_handlers/luci_status.dart
```diff
@@ -94,11 +93,28 @@ class LuciStatusHandler extends RequestHandler<Body> {
...
-    await githubChecksService.updateCheckStatus(
-      buildPushMessage,
-      luciBuildService,
-      slug,
-    );
+    final Map<String, dynamic> userData =
+        jsonDecode(buildPushMessage.userData) as Map<String, dynamic>;
+    if (userData != null &&
+        userData.containsKey('repo_owner') &&
+        userData.containsKey('repo_owner')) {
```
keyonghan, line 100: Duplicated or check `repo_name`?

godofredoc: Nice catch I meant repo_name. Updated!

PR #836, opened 2020-07-09
Use pull request to trigger builds instead of CheckSuiteEvent.

godofredoc:
PR information is not populated on the CheckSuiteEvents and PR
information is needed to trigger the build bucket builds.

Bug:

### DIFF
--- a/app_dart/lib/src/datastore/cocoon_config.dart
+++ b/app_dart/lib/src/datastore/cocoon_config.dart
@@ -456,7 +456,7 @@ class Config {
     return RepositorySlug('flutter', repoName);
   }
 
-  bool checksSupportedRepo(RepositorySlug slug) {
+  bool isChecksSupportedRepo(RepositorySlug slug) {
     return checksSupportedRepos.contains('${slug.owner}/${slug.name}');
   }
 }
--- a/app_dart/lib/src/datastore/cocoon_config.dart
+++ b/app_dart/lib/src/datastore/cocoon_config.dart
@@ -40,6 +40,11 @@ class Config {
     'packages',
   };
 
+  /// List of Github presubmit supported repos.
+  static const Set<String> checksSupportedRepos = <String>{
+    'flutter/cocoon',
+  };
+
   @visibleForTesting
   static const String configCacheName = 'config';
 
@@ -449,6 +454,10 @@ class Config {
       return null;
     }
     return RepositorySlug('flutter', repoName);
+  }
+
+  bool checksSupportedRepo(RepositorySlug slug) {
+    return checksSupportedRepos.contains('${slug.owner}/${slug.name}');
   }
 }
 
--- a/app_dart/lib/src/foundation/github_checks_util.dart
+++ b/app_dart/lib/src/foundation/github_checks_util.dart
@@ -32,11 +32,11 @@ class GithubChecksUtil {
   Future<github.CheckSuite> getCheckSuite(
     github.GitHub gitHubClient,
     github.RepositorySlug slug,
-    int checkSuitId,
+    int checkSuiteId,
   ) async {
     return gitHubClient.checks.checkSuites.getCheckSuite(
       slug,
-      checkSuiteId: checkSuitId,
+      checkSuiteId: checkSuiteId,
     );
   }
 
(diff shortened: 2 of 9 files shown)


app_dart/lib/src/foundation/github_checks_util.dart
```diff
@@ -29,6 +29,17 @@ class GithubChecksUtil {
     );
   }
 
+  Future<github.CheckSuite> getCheckSuite(
+    github.GitHub gitHubClient,
+    github.RepositorySlug slug,
+    int id,
```
keyonghan, line 35: nit: checkSuiteId.

app_dart/lib/src/datastore/cocoon_config.dart
```diff
@@ -450,6 +455,10 @@ class Config {
     }
     return RepositorySlug('flutter', repoName);
   }
+
+  bool checksSupportedRepo(RepositorySlug slug) {
```
keyonghan, line 459: nit: isChecksSupportedRepo.

app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -80,13 +80,6 @@ class GithubWebhook extends RequestHandler<Body> {
         case 'pull_request':
           await _handlePullRequest(stringRequest);
           break;
-        case 'check_suite':
-          final CheckSuiteEvent checkSuiteEvent = CheckSuiteEvent.fromJson(
-            jsonDecode(stringRequest) as Map<String, dynamic>,
-          );
-          await githubChecksService.handleCheckSuite(
-              checkSuiteEvent, luciBuildService);
-          break;
```
keyonghan, line 89: With these lines removed, shall this test be removed: https://github.com/flutter/cocoon/blob/master/app_dart/test/request_handlers/github_webhook_test.dart#L1513

keyonghan: Also the handleCheckSuite function: https://github.com/flutter/cocoon/blob/master/app_dart/lib/src/service/github_checks_service.dart#L43

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -153,14 +153,13 @@ class LuciBuildService {
         'repo_name': slug.name,
         'user_agent': 'flutter-cocoon',
       };
-      if (checkSuiteEvent != null) {
-        final github.CheckRun checkRun =
-            await githubClient.checks.checkRuns.createCheckRun(
-          checkSuiteEvent.repository.slug(),
-          name: builder,
-          headSha: commitSha,
+      if (checkSuiteEvent != null || config.checksSupportedRepo(slug)) {
```
keyonghan, line 156: `checkSuiteEvent` does not seem to be available any more.

app_dart/lib/src/foundation/github_checks_util.dart
godofredoc: Done

app_dart/lib/src/datastore/cocoon_config.dart
godofredoc: Done

app_dart/lib/src/request_handlers/github_webhook.dart
godofredoc: The test was already removed in this CL. I'd like to keep the method in github_checks_service, the correct workflow should use it but there is a bug in the apis that forced us to use the pull request instead. If the apis get fixed at some point then we can get back to the initial implementation.

app_dart/lib/src/service/luci_build_service.dart
godofredoc: It is available is just that we are workaround it because of a failure in the checks api.

app_dart/lib/src/foundation/github_checks_util.dart
```diff
@@ -29,6 +29,17 @@ class GithubChecksUtil {
     );
   }
 
+  Future<github.CheckSuite> getCheckSuite(
+    github.GitHub gitHubClient,
+    github.RepositorySlug slug,
+    int checkSuitId,
```
keyonghan, line 35: Typo: checkSuiteId.

godofredoc: Updated.

PR #838, opened 2020-07-09
Using one shared cloudbuild yaml for both prod and stage build dashboard

keyonghan:
Instead of using fixed project values in two yaml files, this PR uses a shared variable instead to service two projects.


cloudbuild.yaml
```diff
@@ -5,17 +5,17 @@
...
-  - name: gcr.io/flutter-dashboard/flutter
+  - name: gcr.io/$PROJECT_ID/flutter
     entrypoint: '/bin/bash'
     args: ['cloud_build/app_build.sh']
 
   # Deploy a new version to google cloud.
   - name: 'gcr.io/cloud-builders/gcloud'
     entrypoint: '/bin/bash'
-    args: ['cloud_build/deploy.sh', 'flutter-dashboard', '$SHORT_SHA']
+    args: ['cloud_build/deploy.sh', '$PROJECT_ID', '$SHORT_SHA']
```
godofredoc, line 20: How are we providing PROJECT_ID?

keyonghan: It's one of the default substitutions provided by cloud build: https://cloud.google.com/cloud-build/docs/configuring-builds/substitute-variable-values

PR #841, opened 2020-07-10
Remove "whitelist" terminology in Cocoon

CaseyHillers:
Switch `WhitelistedAccount` to `AllowedAccount` based off recommended practices. I originally had it as `AllowlistedAccount`, but dropped it as I don't think "listed" gave much value.

Public documentation available at https://developers.google.com/style/inclusive-documentation#features-and-users (googlers see go/avoid-whitelist)

If this looks good, I can duplicate the existing entities over to the new table before submitting. After deployed and verified, I can remove the `WhitelistedAccount` entities.

There's only a few accounts, so this change is relatively low risk.


CaseyHillers:
I created the `AllowedAccount` entity in Datastore, and duplicated the accounts over to that.

PR #842, opened 2020-07-11
GitHub Bot Improvements

christopherfujino:
This PR makes the following bot improvements:

1. Does **not** reset release candidate PRs to target master branch; fixes https://github.com/flutter/flutter/issues/55900
1. Auto-closes PRs that try to merge from {stable,beta,dev} with a comment pointing the user toward "Tree Hygiene" link in wiki; fixes https://github.com/flutter/flutter/issues/60386
1. Replaces a lot of hard-coded uses of the string `'master'` in the codebase to use a constant variable.

### DIFF
--- a/app_dart/lib/src/datastore/cocoon_config.dart
+++ b/app_dart/lib/src/datastore/cocoon_config.dart
@@ -107,14 +107,25 @@ class Config {
   Future<String> get githubOAuthToken => _getSingleValue('GitHubPRToken');
 
   String get wrongBaseBranchPullRequestMessage =>
-      'This pull request was opened against a branch other than _master_. '
-      'Since Flutter pull requests should not normally be opened against '
-      'branches other than master, I have changed the base to master. If this '
-      'was intended, you may modify the base back to {{branch}}. See the '
-      '[Release Process](https://github.com/flutter/flutter/wiki/Release-process) '
-      'for information about how other branches get updated.\n\n'
+      'This pull request was opened against a branch other than '
+      '_${kDefaultBranchName}_. Since Flutter pull requests should not '
+      'normally be opened against branches other than master, I have changed '
+      'the base to master. If this was intended, you may modify the base back '
+      'to {{branch}}. See the [Release Process]'
+      '(https://github.com/flutter/flutter/wiki/Release-process) for '
+      'information about how other branches get updated.\n\n'
       '__Reviewers__: Use caution before merging pull requests to branches other '
       'than master, unless this is an intentional hotfix/cherrypick.';
+
+  String wrongHeadBranchPullRequestMessage(String branch) =>
+      'This pull request is trying merge the branch $branch, which is the name '
+      'of a release branch. This is usually a mistake. See '
+      '[Tree Hygiene](https://github.com/flutter/flutter/wiki/Tree-hygiene) '
+      'for detailed instructions on how to contribute to the Flutter project. '
+      'In particular, ensure that before you start coding, you create your '
+      'feature branch off of _${kDefaultBranchName}_.\n\n'
+      'This PR has been closed. If you are sure you want to merge $branch, you '
+      'may re-open this issue.';
 
   Future<String> get webhookKey => _getSingleValue('WebhookKey');
 
--- a/app_dart/lib/src/datastore/cocoon_config.dart
+++ b/app_dart/lib/src/datastore/cocoon_config.dart
@@ -25,6 +25,9 @@ import '../service/access_client_provider.dart';
 import '../service/bigquery.dart';
 import '../service/github_service.dart';
 
+/// Name of the default git branch.
+const String kDefaultBranchName = 'master';
+
 class Config {
   Config(this._db, this._cache) : assert(_db != null);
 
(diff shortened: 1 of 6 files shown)


app_dart/lib/src/datastore/cocoon_config.dart
```diff
@@ -106,16 +106,27 @@ class Config {
...
-      'against a branch other than _master_. Since Flutter pull requests should '
-      'not normally be opened against branches other than master, I have changed '
-      'the base to master. If this was intended, you may modify the base back to '
-      '{{branch}}. See the [Release Process]'
-      '(https://github.com/flutter/flutter/wiki/Release-process) for information '
-      'about how other branches get updated.\n\n'
+  String get wrongBaseBranchPullRequestMessage =>
+      'This pull request was opened against a branch other than '
+      '_${kDefaultBranchName}_. Since Flutter pull requests should not '
+      'normally be opened against branches other than master, I have changed '
```
dnfield, line 112: s/master/$kDefaultBranchName

dnfield: And line below this

app_dart/lib/src/datastore/cocoon_config.dart
```diff
@@ -106,16 +106,27 @@ class Config {
...
+      'the base to master. If this was intended, you may modify the base back '
+      'to {{branch}}. See the [Release Process]'
+      '(https://github.com/flutter/flutter/wiki/Release-process) for '
+      'information about how other branches get updated.\n\n'
       '__Reviewers__: Use caution before merging pull requests to branches other '
       'than master, unless this is an intentional hotfix/cherrypick.';
```
dnfield, line 118: And here :)

app_dart/lib/src/foundation/utils.dart
```diff
@@ -24,6 +24,9 @@ Duration twoSecondLinearBackoff(int attempt) {
   return const Duration(seconds: 2) * (attempt + 1);
 }
 
+/// Name of the default git branch.
+const String kDefaultBranchName = 'master';
```
dnfield, line 28: nit: maybe this belongs in config rather than here

app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -440,21 +441,51 @@ class GithubWebhook extends RequestHandler<Body> {
...
+    if (pr.base.ref == kDefaultBranchName) {
+      return;
+    }
+    final RegExp candidateTest = RegExp(r'flutter-\d+\.\d+-candidate\.\d+');
+    if (pr.base.ref == pr.head.ref &&
+        candidateTest.hasMatch(pr.base.ref) &&
+        candidateTest.hasMatch(pr.head.ref)) {
+      // This is most likely a release branch
+      return;
+    }
```
dnfield, line 478: This will leave PRs alone for anyone who opens a patch against a release branch. 

Can we at least have this check that the author is a member of the Flutter org? I know it's possible to get that with the GraphQL API, it's probably possible with the rest API as well.  The graphQL query for it is here: https://github.com/flutter/cocoon/blob/master/app_dart/lib/src/request_handlers/check_for_waiting_pull_requests_queries.dart#L40

dnfield: My only concern is that the reason this was opened in the first place is that people opened seemingly helpful pull requests against the wrong branch, and they would sometimes get merged because a reviewer didn't notice which branch it was for.

christopherfujino: I can if you feel strongly about it, but I doubt someone would accidentally open a PR from a release candidate branch on their fork to an upstream release candidate branch of the exact same name. I can't think of an easier way to do this than:

1. `git checkout flutter-1.20-candidate.7`
2. maybe make commit some local changes
3. `git push -u origin flutter-1.20-candidate.7`
4. In GitHub, open a PR.
5. Change the base branch to be `flutter-1.20-candidate.7`.

christopherfujino: However, I guess we should still comment on the issue, even if we don't rebase.

christopherfujino: I'm playing with this endpoint, I can see the flutter org at https://api.github.com/users/dnfield/orgs, but I get an empty array when I GET https://api.github.com/users/christopherfujino/orgs

dnfield: Anyone who wants to see their desired commit hotfixed into a release branch could do that, especially if they're reading our public documentation about how to do it :) 

I don't think it would happen often, but that's part of the issue. If we just comment that's probably good enough.

app_dart/lib/src/foundation/utils.dart
christopherfujino: what do you know, it already is: https://github.com/flutter/cocoon/blob/master/app_dart/lib/src/datastore/cocoon_config.dart#L167

app_dart/lib/src/datastore/cocoon_config.dart
christopherfujino: done.

app_dart/lib/src/datastore/cocoon_config.dart
christopherfujino: done.

app_dart/lib/src/foundation/utils.dart
christopherfujino: done.

app_dart/lib/src/request_handlers/github_webhook.dart
christopherfujino: If that's their intention, the bot specifically says to rebase it back to what it was previously if you're doing a hotfix, and the user would assume this is part of the procedure and change the base branch (I think the hotfix docs even have that instruction). The bot isn't really enforcing policy, but warning of a possible (probably?) error. To that end, the bot should certainly still comment (and comments are cheap :) ).

christopherfujino: added a comment

dnfield: Comment is fine, don't worry about org if it's a pain here.

dnfield:
Tests are failing - maybe forgot to save a file before commiting?

app_dart/lib/src/request_handlers/github_webhook.dart
christopherfujino: I'll look into it another time. It seems easy to implement, but I'm not sure why my account isn't showing up as a member of Flutter.

PR #843, opened 2020-07-14
Fix TaskOverlay golden test flakes

CaseyHillers:
Two of the tests needed a `tester.pump()` after pumping the widget to allow for the precache to populate. When those tests were run first, they were not loading the asset in time.

# Issues
Fixes https://github.com/flutter/flutter/issues/58833

### DIFF
--- a/app_flutter/test/widgets/task_overlay_test.dart
+++ b/app_flutter/test/widgets/task_overlay_test.dart
@@ -84,8 +84,7 @@ void main() {
     expect(find.text(expectedTaskInfoString), findsNothing);
     expect(find.text(expectedTask.reservedForAgentId), findsNothing);
 
-    // TODO(chillers): Fix flakiness of precacheTaskIcons. https://github.com/flutter/flutter/issues/58833
-    // await expectLater(find.byType(MaterialApp), matchesGoldenFile('task_overlay_test.normal_overlay_closed.png'));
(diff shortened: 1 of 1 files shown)


app_flutter/test/widgets/task_overlay_test.dart
```diff
@@ -253,17 +249,16 @@ void main() {
         ),
       ),
     );
+    await tester.pump();
 
     expect(find.text(expectedTaskInfoString), findsNothing);
-    // TODO(chillers): Fix flakiness of precacheTaskIcons. https://github.com/flutter/flutter/issues/58833
-    // await expectLater(find.byType(MaterialApp), matchesGoldenFile('task_overlay_test.nondevicelab_closed.png'));
+    await expectLater(find.byType(MaterialApp), matchesGoldenFile('task_overlay_test.nondevicelab_closed.png'));
```
keyonghan, line 255: Did a quick test locally, after adding `await tester.pump()`, it seems to still throw the diff error.
Anything else to change to get it passed?

CaseyHillers: Can you post the log? These tests should only be run on Linux. I have a followup PR that will skip these tests on non-linux platforms.

CaseyHillers: See https://github.com/flutter/flutter/issues/60720 for running golden tests only on linux.

keyonghan: Oh, i see. I ran it in Mac.

PR #845, opened 2020-07-15
Update cloudbuild configuration

keyonghan:
Cloud build has default 10 mins timeout setting. However our build usually takes longer than 10 mins, causing build timeout frequently. Here we extend the build timeout to be 20 mins.

Additionally, we add `_GAE_PROMTE` variable to easily control the `gcloud app deploy` from trigger side in cloud console.


cloud_build/deploy.sh
```diff
@@ -6,5 +6,5 @@
 # Deploy a new flutter dashboard version to google cloud.
 
 pushd app_dart > /dev/null
```
godofredoc, line 8: use set -e at the top of these scripts as they are they will continue even if one command fails.

keyonghan: Added.

PR #847, opened 2020-07-16
Make author avatars meet a11y text contrast on dashboard

Piinks:
This change makes the text contrast of the author avatars meet a11y guidelines. 

Prior to change:

Screen Shot 2020-07-16 at 11 39 54 AM

With change (dark and light themes):

Screen Shot 2020-07-16 at 12 05 24 PM

Screen Shot 2020-07-16 at 12 30 42 PM

PR #848, opened 2020-07-16
Formatting for app_flutter

Piinks:
Re-formats app_flutter to pass pre-submit dartfmt checks.

Un-blocks https://github.com/flutter/cocoon/pull/847

### DIFF
--- a/app_flutter/lib/agent_dashboard_page.dart
+++ b/app_flutter/lib/agent_dashboard_page.dart
@@ -30,10 +30,12 @@ class AgentDashboardPage extends StatelessWidget {
 
   static const String routeName = '/agents';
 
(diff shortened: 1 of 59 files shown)


keyonghan:
https://github.com/flutter/cocoon/blob/master/test_utilities/bin/flutter_test_runner.sh#L26 seems to be intended with option `--line-length=120` for app_flutter.

Piinks:
> https://github.com/flutter/cocoon/blob/master/test_utilities/bin/flutter_test_runner.sh#L26 seems to be intended with option `--line-length=120` for app_flutter.

Yeah, its strange, I can't tell why it won't pass for #847

PR #851, opened 2020-07-17
Unify analysis and formatting for repository

Piinks:
This change primarily updates the valid formatting for the repository, and secondarily applies it across the board.

### Points of interest:

- Added `curly_braces_in_flow_control_structures` to:
  - `app_dart/analysis_options.yaml`
  - `app_flutter/analysis_options.yaml`
This fixes a contradiction with `dartfmt`, which is also enforced in this repository. Prior to this change, `dartfmt` will put a control statement and its body on a single line, causing analysis to fail. Requiring curly braces solves this and alerts dev while writing rather than dartfmt creating an issue later on.

- Updated the enforced `dartfmt` to include `--line-length=120` across the whole repository:
  - `test_utilities/bin/dart_test_runner.sh`
Rather than 80 here, 120 there, chose unified line length for everything.

- Added `format.sh`
Executable from the root of the repository so no one will have to wonder which formatting is correct. :)

The rest of the changed files are the implementation of these updates.


format.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+# Copyright 2019 The Flutter Authors. All rights reserved.
```
keyonghan, line 2: 2020?

format.sh
```diff
@@ -0,0 +1,8 @@
...
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+set -e
+
+dartfmt -n --line-length=120 .
```
keyonghan, line 8: Nit: a new line at the end.

format.sh
Piinks: Whoops! 😅

Hixie:
re always_put_control_body_on_new_line, one option is to require braces around the bodies.

FWIW, personally I find having control flow on the same line as the `if` makes it significantly more likely that I'll misread the control flow and thus end up with more bugs.

Piinks:
> re always_put_control_body_on_new_line, one option is to require braces around the bodies.
> 
> FWIW, personally I find having control flow on the same line as the `if` makes it significantly more likely that I'll misread the control flow and thus end up with more bugs.

Agreed! I also prefer the separation. The contradiction between analysis and dartfmt is incredibly frustrating though.
There is an analysis option to require curly braces. Although a comment beside it says it is untested. Should we give that a go instead?

Piinks:
I think the curly braces option may be a better bet. That way you get the analyzer warning to add brackets while actually writing the code.
I think the frustration for me came from writing the code, running dartfmt and then later getting analysis errors due to dartfmt.

PR #861, opened 2020-07-22
Solve corner case that summaryMarkdown is null when `updateCheckRun` in Github

keyonghan:
This fixes issue: https://github.com/flutter/flutter/issues/62056


app_dart/lib/src/service/github_checks_service.dart
```diff
@@ -128,7 +128,7 @@ class GithubChecksService {
     if (status == github.CheckRunStatus.completed && failedStatesSet.contains(conclusion)) {
       final Build build =
           await luciBuildService.getBuildById(buildPushMessage.build.id, fields: 'id,builder,summaryMarkdown');
-      output = github.CheckRunOutput(title: checkRun.name, summary: build.summaryMarkdown);
+      output = github.CheckRunOutput(title: checkRun.name, summary: build.summaryMarkdown ?? 'No summaryMarkdown');
```
godofredoc, line 131: s/No/Empty?

keyonghan: Done,

PR #863, opened 2020-07-23
Access Luci builders via http access

keyonghan:
This is to finish https://github.com/flutter/flutter/issues/61769.

### DIFF
--- a/app_dart/lib/src/datastore/cocoon_config.dart
+++ b/app_dart/lib/src/datastore/cocoon_config.dart
@@ -321,22 +321,6 @@ class Config {
     return supportedRepos.contains(repositoryName);
   }
 
-  Future<RepositorySlug> repoNameForBuilder(String builderName) async {
-    final List<Map<String, dynamic>> builders = await luciTryBuilders;
-    final Map<String, dynamic> builderConfig = builders.firstWhere(
-      (Map<String, dynamic> builder) => builder['name'] == builderName,
-      orElse: () => <String, String>{'repo': ''},
-    );
-    final String repoName = builderConfig['repo'] as String;
-    // If there is no builder config for the builderName then we
-    // return null. This is to allow the code calling this method
-    // to skip changes that depend on builder configurations.
-    if (repoName.isEmpty) {
-      return null;
-    }
-    return RepositorySlug('flutter', repoName);
-  }
-
   bool isChecksSupportedRepo(RepositorySlug slug) {
     return checksSupportedRepos.contains('${slug.owner}/${slug.name}');
   }
--- a/app_dart/lib/src/datastore/cocoon_config.dart
+++ b/app_dart/lib/src/datastore/cocoon_config.dart
@@ -69,15 +69,10 @@ class Config {
     return String.fromCharCodes(cacheValue).split(',');
   }
 
+  // Gets supported luci builders based on github http request.
   Future<List<Map<String, dynamic>>> _getLuciBuilders(String bucket) async {
-    final Uint8List cacheValue = await _cache.getOrCreate(
-      configCacheName,
-      'luci${bucket}Builders',
-      createFn: () => getBuilders(Providers.freshHttpClient, loggingService, twoSecondLinearBackoff, bucket),
-      ttl: configCacheTtl,
-    );
-
-    final String builderContent = String.fromCharCodes(cacheValue);
+    final String builderContent =
+        await getBuilders(Providers.freshHttpClient, loggingService, twoSecondLinearBackoff, bucket);
     final Map<String, dynamic> builderMap = json.decode(builderContent) as Map<String, dynamic>;
     final List<dynamic> builderList = builderMap['builders'] as List<dynamic>;
     return builderList.map((dynamic builder) => builder as Map<String, dynamic>).toList();
--- a/app_dart/lib/src/datastore/cocoon_config.dart
+++ b/app_dart/lib/src/datastore/cocoon_config.dart
@@ -69,6 +69,20 @@ class Config {
     return String.fromCharCodes(cacheValue).split(',');
   }
 
+  Future<List<Map<String, dynamic>>> _getLuciBuilders(String bucket) async {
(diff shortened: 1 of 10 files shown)


app_dart/lib/src/datastore/cocoon_config.dart
```diff
@@ -69,6 +69,20 @@ class Config {
     return String.fromCharCodes(cacheValue).split(',');
   }
 
+  Future<List<Map<String, dynamic>>> _getLuciBuilders(String bucket) async {
```
godofredoc, line 72: Docs?

app_dart/lib/src/datastore/cocoon_config.dart
```diff
@@ -69,6 +69,20 @@ class Config {
...
+    final Uint8List cacheValue = await _cache.getOrCreate(
```
godofredoc, line 73: Should we bypass the cache? this will force people to wait up to 15 mins to see the change.

app_dart/lib/src/datastore/cocoon_config.dart
godofredoc, line 72: Do we also need the repo?

app_dart/lib/src/foundation/utils.dart
```diff
@@ -25,9 +25,9 @@ Duration twoSecondLinearBackoff(int attempt) {
   return const Duration(seconds: 2) * (attempt + 1);
 }
 
-Future<List<String>> loadBranches(
-    HttpClientProvider branchHttpClientProvider, Logging log, GitHubBackoffCalculator gitHubBackoffCalculator) async {
-  const String path = '/flutter/cocoon/master/app_dart/dev/branches.txt';
+Future<String> loadContents(HttpClientProvider branchHttpClientProvider, Logging log,
```
godofredoc, line 28: remoteFileContent?

app_dart/lib/src/foundation/utils.dart
```diff
@@ -25,9 +25,9 @@ Duration twoSecondLinearBackoff(int attempt) {
...
+    GitHubBackoffCalculator gitHubBackoffCalculator, String filename) async {
+  final String path = '/flutter/cocoon/master/app_dart/dev/$filename';
```
godofredoc, line 30: Shouldn't this be a path on the target repo?

app_dart/lib/src/foundation/utils.dart
```diff
@@ -42,28 +42,28 @@ Future<List<String>> loadBranches(
...
+        log.error('Attempt to download $filename failed:\n$error\n$stackTrace');
       }
       await Future<void>.delayed(gitHubBackoffCalculator(attempt));
     }
   } finally {
     client.close(force: true);
   }
   log.error('GitHub not responding; giving up');
-  return <String>['master'];
+  return null;
```
godofredoc, line 58: Isn't null the default?

app_dart/lib/src/foundation/utils.dart
```diff
@@ -42,28 +42,28 @@ Future<List<String>> loadBranches(
...
 }
 
 Future<Uint8List> getBranches(
     HttpClientProvider branchHttpClientProvider, Logging log, GitHubBackoffCalculator gitHubBackoffCalculator) async {
-  final List<String> branches = await loadBranches(branchHttpClientProvider, log, gitHubBackoffCalculator);
-
+  String content = await loadContents(branchHttpClientProvider, log, gitHubBackoffCalculator, 'branches.txt');
```
godofredoc, line 63: Docs for this method?

app_dart/lib/src/foundation/utils.dart
```diff
@@ -81,3 +81,11 @@ Future<RepositorySlug> repoNameForBuilder(List<Map<String, dynamic>> builders, S
   }
   return RepositorySlug('flutter', repoName);
 }
+
+Future<Uint8List> getBuilders(HttpClientProvider branchHttpClientProvider, Logging log,
```
godofredoc, line 85: Docs?

keyonghan: Added.

app_dart/lib/src/foundation/utils.dart
keyonghan: If http request is successful, then it returns the content. Otherwise it returns null.

app_dart/lib/src/foundation/utils.dart
keyonghan: what do you mean by target repo?

app_dart/lib/src/foundation/utils.dart
keyonghan: done.

app_dart/lib/src/datastore/cocoon_config.dart
keyonghan: SGTM. Updated.

app_dart/lib/src/datastore/cocoon_config.dart
keyonghan: Do we change that frequently? If this is needed, I would prefer to support that in a separate PR.

keyonghan: Added.

godofredoc: sounds good to me to add in a different PR, the use case for this is that config files will now live in their own repositories. E.g engine configs will be pulled from the engine repo and sdk from flutter repo. This way those test will be branched out along with the repository code.

\cc @christopherfujino

app_dart/lib/src/foundation/utils.dart
godofredoc: A path in github.com/flutter/engine/dev/../.. or github.com/flutter/flutter/dev/../..

keyonghan: Gonna support repo in a separate PR, this will then be updated accordingly.

PR #867, opened 2020-07-24
[agent] Close iOS systeam dialogs in health check

digiter:
- The dialogs often cause test flakiness and performance regressions, therefore we'd like to close them before every test execution.
- This change will run the single XCUITest in the https://github.com/digiter/infra-dialog repo. The test takes about 16 secs to build and run.
- I have merged the infra-dialog repo with cocoon in https://github.com/flutter/cocoon/pull/868

Bug: https://github.com/flutter/flutter/issues/54792

PR #868, opened 2020-07-24
[agent] Check in the infra-dialog Xcode project

digiter:
- This project contains a XCUITest that could be applied to dismiss iOS system dialogs.
- Most of this change are boilerplates that I'm not familiar with, please advice on how to reduce them.
- The real content is "infra_dialogUITests.swift" and a few settings that make it compatible with older versions of iOS, e.g. iOS 9.0.
- https://github.com/flutter/cocoon/pull/867 will integrate it as an agent health check.

Bug: https://github.com/flutter/flutter/issues/54792


jmagman:
You can delete the dummy app since you're not actually loading it in the test.  
@digiter Mind if I push a commit to your fork?  It'll be easier than doing a code review.

digiter:
@jmagman Thanks, that would be great! Or maybe create a PR to cocoon if that's easier.

jmagman:
@digiter I pushed the change to your fork, commit is https://github.com/flutter/cocoon/pull/868/commits/1dbaef5a6bde58ab5741f4f9242379b2fbc70708.

digiter:
Thanks a lot @jmagman !

jmagman:
@digiter I confirmed the test runs and the springboard "app" is loaded, but I wasn't able to actually test closing a real dialog.  Can you check my changes didn't regress that?

digiter:
I tested on a physical iphone, yes, it still can close a real dialog! @jmagman

PR #874, opened 2020-07-28
[agent] Overwrites code signing config of infra-dialog

digiter:
If an existing code singing config exists on a test bed, we should use that one in order to build the infra-dialog project.

Bug: flutter/flutter#54792

### DIFF
--- a/agent/lib/src/health.dart
+++ b/agent/lib/src/health.dart
@@ -150,7 +150,7 @@ Future<HealthCheckResult> removeCachedData(
 /// The dialogs often cause test flakiness and performance regressions.
 @visibleForTesting
 Future<HealthCheckResult> closeIosDialog(
-    {ProcessManager pm = const LocalProcessManager(), DeviceDiscovery discovery}) async {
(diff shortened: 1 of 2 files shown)


agent/lib/src/health.dart
```diff
@@ -166,9 +166,15 @@ Future<HealthCheckResult> closeIosDialog(
     await unlockKeyChain();
     // Runs the single XCUITest in infra-dialog.
     await inDirectory(dialogDir, () async {
-      String command =
-          'xcrun xcodebuild -project infra-dialog.xcodeproj -scheme infra-dialog -destination id=${d.deviceId} test';
-      Process proc = await pm.start(command.split(' '), workingDirectory: dialogDir.path);
+      List<String> command =
+          'xcrun xcodebuild -project infra-dialog.xcodeproj -scheme infra-dialog -destination id=${d.deviceId} test'.split(' ');
+      // Overwrites code signing config when that exists.
```
godofredoc, line 171: Can you please add more info on why we need to overwrite the code signing config?

digiter: Added.

PR #875, opened 2020-07-28
Add supported Luci builders for cocoon repo

keyonghan:
Related issue: flutter/flutter#62429


keyonghan:
> Can we add a readme file with information about the structure of the folder, the expected format of the json file and a test that validates the json validity and content?

README added.
Didn't find an easy way to test json validity from pre-submit tests, but added corresponding logic to check json format in `utils`, and added tests in `utiles_test.dart` to cover that.

dev/README.md
```diff
@@ -0,0 +1,19 @@
...
+    "builders":[
+        {
+            "name":"xxx",
+            "repo":"cocoon"
+        }
+    ]
+}
+```
+This file will be mainly used in [`flutter/cocoon`](https://github.com/flutter/cocoon)
+to update luci task statuses of cocoon pre-submit tests.
```
godofredoc, line 18: To trigger LUCI presubmit tasks?

dev/cocoon_try_builders.json
```diff
@@ -0,0 +1,8 @@
+{
```
godofredoc, line 1: For the test I was thinking something like a dart script that receives a the path of the json file as a flag and validates it has the correct content and the file is parseable.

keyonghan:
> Is this a breaking change? do we need to land the files in the different repositories first?

No, this is not.
Existing cocoon is still reading build coinfigs from `/app_dart/dev/luci_try_builders.json`.

dev/README.md
keyonghan: Updated.

dev/cocoon_try_builders.json
keyonghan: I have added a separate dart file to validate the json contents: `validate_json.dart`, and included it in the README.

godofredoc: > I have added a separate dart file to validate the json contents: `validate_json.dart`, and included it in the README.

Great, maybe not in this PR but can we run the validation in presubmit?

keyonghan: I will try adding that in a separate PR.

PR #879, opened 2020-07-29
Show commit message in CommitBox

tvolkert:
Fixes https://github.com/flutter/flutter/issues/43107


app_flutter/lib/widgets/commit_box.dart
```diff
@@ -143,3 +162,42 @@ class CommitOverlayContents extends StatelessWidget {
...
+  _HyperlinkState createState() => _HyperlinkState();
+}
+
+class _HyperlinkState extends State<Hyperlink> {
+  bool hover = false;
+
+  @override
+  Widget build(BuildContext context) {
+    return MouseRegion(
+      onEnter: (PointerEnterEvent _) => setState(() => hover = true),
```
Hixie, line 187: i thought we were going to create ourselves a Hover class that had two children, one when hovered and one when not, or a HoverBuilder class with a builder that takes a boolean, or something. Did we never get around to that? cc @gspencergoog  @dkwingsmt

app_flutter/lib/widgets/commit_box.dart
```diff
@@ -111,26 +114,42 @@ class CommitOverlayContents extends StatelessWidget {
...
+                              onPressed: _openGithub,
+                            ),
+                          ),
+                          const SizedBox(height: 8),
+                          AnimatedDefaultTextStyle(
+                            style: theme.textTheme.bodyText2.copyWith(
+                              color: theme.textTheme.caption.color,
+                            ),
+                            duration: kThemeChangeDuration,
+                            child: SelectableText('${commit.message} ${commit.message} ${commit.message}'),
```
Hixie, line 143: three times?

app_flutter/lib/service/dev_cocoon.dart
```diff
@@ -121,6 +121,7 @@ class DevelopmentCocoonService implements CocoonService {
       ..key = (RootKey()..child = (Key()..name = '$commitTimestamp'))
       ..author = _authors[author]
       ..authorAvatarUrl = 'https://avatars2.githubusercontent.com/u/${2148558 + author}?v=4'
+      ..message = 'Fake commit message'
```
CaseyHillers, line 124: Since the backend does not send protos, appengine_cocoon need's to be updated to parse message in the json

app_flutter/lib/widgets/commit_box.dart
tvolkert: Oops - I was testing wrapping to see what it looked like.  Fixed.

app_flutter/lib/service/dev_cocoon.dart
tvolkert: Good catch - done

PR #880, opened 2020-07-29
Enable flutter luci try by default

keyonghan:
SDK luci try jobs are triggered by labeling `CQ+1`. This PR removes the `CQ+1` label to enable Luci try jobs by default for SDK.

This is part of SDK migration to Luci project.

### DIFF
--- a/app_dart/lib/src/request_handlers/github_webhook.dart
+++ b/app_dart/lib/src/request_handlers/github_webhook.dart
@@ -22,9 +22,6 @@ import '../request_handling/exceptions.dart';
 import '../request_handling/request_handler.dart';
 import '../service/buildbucket.dart';
 
-/// List of repos that require CQ+1 label.
-const Set<String> kNeedsCQLabelList = <String>{};
-
 /// List of repos that require check for golden triage.
 const Set<String> kNeedsCheckGoldenTriage = <String>{'flutter/flutter'};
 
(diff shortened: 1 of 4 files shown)


app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -23,7 +23,7 @@ import '../request_handling/request_handler.dart';
 import '../service/buildbucket.dart';
 
 /// List of repos that require CQ+1 label.
-const Set<String> kNeedsCQLabelList = <String>{'flutter/flutter'};
+const Set<String> kNeedsCQLabelList = <String>{};
```
godofredoc, line 26: We won't need the label anymore, would you mind removing it?

keyonghan: label removed, and the corresponding logic/tests.

app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -134,33 +131,13 @@ class GithubWebhook extends RequestHandler<Body> {
         await _scheduleIfMergeable(pullRequestEvent);
         break;
       case 'labeled':
```
godofredoc, line 133: we still need a break here.

keyonghan: Added. Thanks!

PR #882, opened 2020-07-30
Fix bug whereby we weren't capturing the commit message

tvolkert:
https://github.com/flutter/flutter/issues/43107


keyonghan:
Shall we also add `message` here https://github.com/flutter/cocoon/blob/master/app_dart/test/request_handlers/refresh_github_commits_test.dart#L58 to get it tested?

tvolkert:
Done

PR #886, opened 2020-07-31
Add json content validation to presubmit

keyonghan:
Recently supported LUCI try builders configuration has been available in cocoon repo in a JSON file. This PR validates the json format in presubmit tasks.


dev/validate_json.dart
```diff
@@ -3,22 +3,30 @@
...
-Future<bool> main(List<String> args) async {
+Future<void> main(List<String> args) async {
   final String jsonString = await File(args[0]).readAsString();
   Map<String, dynamic> decodedJson;
-  bool decodedStatus = true;
+  final List<String> messages = <String>[];
   try {
     decodedJson = json.decode(jsonString) as Map<String, dynamic>;
     final List<dynamic> builders = decodedJson['builders'] as List<dynamic>;
     if (builders == null) {
```
digiter, line 27: The current implementation doesn't check the schema of a builder.

A better way is adding a @JsonSerializable() class representing the schema of the json, then parse the json with that class. We need to parse the json in Cocoon backend code anyway.

keyonghan: Good point! Have added logic to check builder schema.

dev/pubspec.yaml
```diff
@@ -0,0 +1,17 @@
+name: dev_check
+description: Script to check licenses.
```
godofredoc, line 2: Update description?

dev/pubspec.lock
```diff
@@ -0,0 +1,369 @@
+# Generated by pub
```
godofredoc, line 1: Do we really need all these packages to validate a json file or we need an updated file?

dev/pubspec.yaml
keyonghan: updated.

dev/pubspec.lock
keyonghan: removed unnecessary dependencies, and re-ran `pub get`. Not much change though.

PR #887, opened 2020-08-01
Move the github client auth closer to where it will be used.

godofredoc:
Github client is now using short lived tokens and in some cases when we
need to validate is builds exist and cancel them the token expired in
the middle of the request. We are moving the github client creation and
authentication to the place it will be used instead of creating at the
beginning of the method.

Bug:
  https://github.com/flutter/flutter/issues/62233

### DIFF
--- a/app_dart/lib/src/service/luci_build_service.dart
+++ b/app_dart/lib/src/service/luci_build_service.dart
@@ -128,7 +128,6 @@ class LuciBuildService {
     assert(prNumber != null);
     assert(commitSha != null);
     assert(slug != null);
-    final github.GitHub githubClient = await config.createGitHubClient(slug.owner, slug.name);
     if (!config.githubPresubmitSupportedRepo(slug.name)) {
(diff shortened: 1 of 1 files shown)


app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -158,6 +157,7 @@ class LuciBuildService {
     }
 
     final List<Request> requests = <Request>[];
+    final github.GitHub githubClient = await config.createGitHubClient(slug.owner, slug.name);
```
digiter, line 160: Can we check if a token is expired?

godofredoc: We can if we create a wrapper class for the github client but most likely not needed. The token is generated per request and in this particular case the problem is we were creating the client, sending some RPCs and after we got the responses from the RPCs we were using the token. With this change the token will be be used immediately after it was created.

PR #877, opened 2020-07-29
Store commit message in the `Checklist` Datastore & BigQuery

tvolkert:
This will allow us to show the commit message in the overlay
card of the build dashboard that pops up when the user clicks
on the commit.

https://github.com/flutter/flutter/issues/43107

PR #889, opened 2020-08-03
Truncate commit message at 1500 chars.

godofredoc:
Commit messages can be longer than 1500 chars however the limit in the
datastore field is 1500 chars. We are truncating commit messages when
they are longer than 1500 chars.

Bug:
  https://github.com/flutter/flutter/issues/62789


tvolkert:
Is the 1500 limit new?  How did older commits like https://github.com/flutter/flutter/commit/f06ee8d land successfully?

keyonghan:
This may be due to the recent PR (https://github.com/flutter/cocoon/pull/877) saving commit message to datastore.

tvolkert:
I had ruled that out because flutter/flutter@f06ee8d landed after that Cocoon change.... until I just realized that  flutter/flutter@f06ee8d was never added to the dashboard!  Thanks @keyonghan 🙂

PR #892, opened 2020-08-04
Refactor and Update flutter-gold check

Piinks:
This updates and refactors the functionality of the flutter-dashboard bot as it relates to the flutter-gold check.
- Fixes https://github.com/flutter/flutter/issues/62682
  - Bot was broken by switch from status to check
  - follows the same pattern of `check_for_waiting_pull_requests` to query checkSuites, have not found a way to confirm yet
  - updated tests
- Removes gold logic from the `github_webhook`
  - This is no longer applicable
  - updated tests
- Updates the config to include flutter-gold Strings for status and notifications
  - updated fake config and relevant tests

### DIFF
--- a/app_dart/lib/src/model/appengine/github_gold_status_update.dart
+++ b/app_dart/lib/src/model/appengine/github_gold_status_update.dart
@@ -21,6 +21,12 @@ class GithubGoldStatusUpdate extends Model {
     id = key?.id;
   }
 
+  // The flutter-gold status cannot report a `failure` status
+  // due to auto-rollers. This is why we hold a `pending` status
+  // when there are image changes. This provides the opportunity
+  // for images to be triaged, and the auto-roller to proceed.
+  // For more context, see: https://github.com/flutter/flutter/issues/48744
+
   static const String statusCompleted = 'success';
 
   static const String statusRunning = 'pending';
--- a/app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
+++ b/app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
@@ -106,12 +106,7 @@ class PushGoldStatusToGithub extends ApiRequestHandler<Body> {
         }
       }
 
-      if (runsGoldenFileTests ||
-        // Adding this short-circuit to cut down GitHub API requests. If we are
-        // already failing, or still pending on Cirrus, no need to query for
-        // LUCI status.
-        !(cirrusCheckStatuses.any(kCirrusInProgressStates.contains) ||
-        cirrusCheckStatuses.any(kCirrusFailedStates.contains))) {
+      if (runsGoldenFileTests) {
         // Make sure we account for any running Luci builds
         final GraphQLClient gitHubGraphQLClient = await config.createGitHubGraphQLClient();
         final List<String> luciIncompleteChecks = <String>[];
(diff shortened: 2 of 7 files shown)


app_dart/lib/src/model/appengine/github_gold_status_update.dart
```diff
@@ -21,6 +21,12 @@ class GithubGoldStatusUpdate extends Model {
     id = key?.id;
   }
 
+  // The flutter-gold status cannot report a `failure` status
+  // due to auto-rollers. This is why we hold a `pending` status
+  // when there are image changes. This provides the opportunity
+  // for images to be triaged, and the auto-roller to proceed.
+  // For more context, see: https://github.com/flutter/flutter/issues/48744
+
```
Piinks, line 29: I added this reference since I forgot and almost refactored to include a fail state.

app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
```diff
@@ -261,24 +259,57 @@ class PushGoldStatusToGithub extends ApiRequestHandler<Body> {
...
+const String pullRequestChecksQuery = r'''
+query ChecksForPullRequest($sPullRequest: int!) {
+  repository(owner: "flutter", name: "flutter") {
+    pullRequest(number: $sPullRequest) {
+      commits(last: 1) {
+        nodes {
+          commit {
+            # (appId: 64368) == flutter-dashboard. We only care about
+            # flutter-dashboard checks.
+            checkSuites(last: 1, filterBy: {appId: 64368}) {
```
Piinks, line 313: I have not been able to run this in the github graphql explorer. I based it off of the check for waiting pull requests pattern. How can I check this?

app_dart/test/request_handlers/push_gold_status_to_github_test.dart
```diff
@@ -978,6 +938,33 @@ QueryResult createCirrusQueryResult(List<dynamic> statuses, String branch) {
   );
 }
 
+QueryResult createGithubQueryResult(List<dynamic> statuses) {
```
Piinks, line 941: I have not been able to confirm this format yet.

app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
```diff
@@ -108,31 +108,49 @@ class PushGoldStatusToGithub extends ApiRequestHandler<Body> {
...
+              .cast<Map<String, dynamic>>();
+        }
+        checkRuns = checkRuns ?? <Map<String, dynamic>>[];
+        log.debug('Found luci checks: $checkRuns');
+        for (Map<String, dynamic> checkRun in checkRuns) {
+          final String name = checkRun['name'] as String;
+          if (checkRun['status'] != null && checkRun['status'].toUpperCase() != 'COMPLETED') {
+            luciIncompleteChecks.add(name);
+          } else if (checkRun['conclusion'] != null && checkRun['conclusion'].toUpperCase() != 'SUCCESS') {
+            luciIncompleteChecks.add(name);
```
godofredoc, line 132: Seems like most of this code is duplicated from check_for_waiting_pull_requests. Would it be possible to move it to github_checks_service and reuse it in both places?

app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
godofredoc: Unfortunately there is no way to test it with the explorer as the data from preview apis is not available. What I did was to create a temporary handler  call it with post and print the result from the query.

app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
Piinks: Definitely! Let me see what I can extract. The `check_for_waiting_pull_requests` queries for all PRs with the right label, while here we're looking up a specific PR. I might be able to just use the original query to get what I need.

Piinks: I think I'll open a separate PR to move the logic to `github_checks_service` and then update here.

PR #893, opened 2020-08-05
Save engine build status to bigquery

keyonghan:
Framework build status has been collected to monitor the tree status. It would be useful to track the build status of Engine as well.  
By saving status to BigQuery, we can further show its trending chart in datastudio.

\cc @pcsosinski

### DIFF
--- a/app_dart/lib/src/request_handlers/push_engine_status_to_github.dart
+++ b/app_dart/lib/src/request_handlers/push_engine_status_to_github.dart
@@ -5,7 +5,6 @@
 import 'dart:async';
 
 import 'package:github/github.dart';
-import 'package:googleapis/bigquery/v2.dart';
 import 'package:meta/meta.dart';
 
 import '../datastore/cocoon_config.dart';
--- a/app_dart/lib/src/request_handlers/push_engine_status_to_github.dart
+++ b/app_dart/lib/src/request_handlers/push_engine_status_to_github.dart
@@ -41,10 +41,10 @@ class PushEngineStatusToGithub extends ApiRequestHandler<Body> {
 
   @override
   Future<Body> get() async {
-    //if (authContext.clientContext.isDevelopmentEnvironment) {
+    if (authContext.clientContext.isDevelopmentEnvironment) {
       // Don't push GitHub status from the local dev server.
-    //  return Body.empty;
-    //}
+      return Body.empty;
+    }
 
     final LuciService luciService = luciServiceProvider(this);
     final Map<LuciBuilder, List<LuciTask>> luciTasks = await luciService.getRecentTasks(repo: 'engine');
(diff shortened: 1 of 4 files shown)


app_dart/lib/src/request_handlers/push_engine_status_to_github.dart
```diff
@@ -99,4 +102,30 @@ class PushEngineStatusToGithub extends ApiRequestHandler<Body> {
     }
     return GithubBuildStatusUpdate.statusSuccess;
   }
+
+  Future<void> _insertBigquery(String buildStatus) async {
+    // Define const variables for [BigQuery] operations.
```
godofredoc, line 107: Is this a method that can be abstracted and moved to a service or utility class?

app_dart/lib/src/request_handlers/push_engine_status_to_github.dart
```diff
@@ -99,4 +102,30 @@ class PushEngineStatusToGithub extends ApiRequestHandler<Body> {
...
+    const String projectId = 'flutter-dashboard';
+    const String dataset = 'cocoon';
+    const String table = 'EngineBuildStatus';
```
godofredoc, line 110: Any reason to have a separate table for engine? can we just use the same one as sdk?

app_dart/lib/src/request_handlers/push_engine_status_to_github.dart
```diff
@@ -99,4 +102,30 @@ class PushEngineStatusToGithub extends ApiRequestHandler<Body> {
...
+
+    final TabledataResourceApi tabledataResourceApi = await config.createTabledataResourceApi();
+    final List<Map<String, Object>> requestRows = <Map<String, Object>>[];
+
+    requestRows.add(<String, Object>{
+      'json': <String, Object>{
+        'Timestamp': DateTime.now().millisecondsSinceEpoch,
```
godofredoc, line 117: Does this need to be UTC?

app_dart/lib/src/request_handlers/push_engine_status_to_github.dart
keyonghan: Moved it to utils, and the corresponding test.

app_dart/lib/src/request_handlers/push_engine_status_to_github.dart
keyonghan: If using existing table for sdk, we need to add a new column to distinguish sdk and engine. However Bigquery does not support adding new column on the fly. That means we need to recreate a new table, and then import existing sdk data to the new table.

app_dart/lib/src/request_handlers/push_engine_status_to_github.dart
keyonghan: We have been using now() for all cocoon time related values. I know GoB commit time, luci bigquery timestamp are using UTC. 
But this should not be an issue. We can easily adjust the time when we present the results in dashboard.

app_dart/lib/src/request_handlers/push_engine_status_to_github.dart
godofredoc: I'm assuming we'll need to do it sooner or later when adding a new repo. The question is do we do it now where we have a single table or we move the effort to merge two tables later?

keyonghan: Now we are using the same table to both sdk and engine, and it is easy to extend to other repos as well.

PR #894, opened 2020-08-05
Remove json validation logic & use generic name

keyonghan:
Related https://github.com/flutter/cocoon/pull/886, https://github.com/flutter/cocoon/pull/891, https://github.com/flutter/flutter/issues/62429

This PR removes existing json validation scripts, and use generic name instead.
This PR also disables the cocoon pre-submit check of json validation, which runs the above scripts.

Instead, we will use recipe coverage to validate json format in luci pre-submit tasks.


dev/README.md
```diff
@@ -2,7 +2,7 @@ This directory contains resources that the Flutter team uses during
 the development of cocoon.
 
 ## Luci builder file
-`cocoon_try_builders.json` contains the supported luci try builders 
+`try_builders.json` contains the supported luci try builders
```
godofredoc, line 5: Is the change on the recipes already running the test?

keyonghan: The cocoon recipe is running the json check test now.
With this PR removing the json check codes, the cocoon recipe will stop running the check without any change.
Then I will submit a cl to enable json check from cocoon recipe side by calling the json util.

PR #897, opened 2020-08-06
Reland flutter-gold refactor

Piinks:
The flutter-gold refactor was reverted due to there being `int!` instead of `Int!` in the query string.

### DIFF
--- a/app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
+++ b/app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
@@ -106,12 +106,7 @@ class PushGoldStatusToGithub extends ApiRequestHandler<Body> {
         }
       }
 
-      if (runsGoldenFileTests ||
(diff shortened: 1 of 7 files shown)


app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
```diff
@@ -275,10 +273,57 @@ class PushGoldStatusToGithub extends ApiRequestHandler<Body> {
...
+    for (GraphQLError error in result.errors) {
+      log.error(error.toString());
+    }
+    throw const BadRequestException('GraphQL query failed');
+  }
+  return result.data as Map<String, dynamic>;
+}
+
+const String pullRequestChecksQuery = r'''
+query ChecksForPullRequest($sPullRequest: Int!) {
```
Piinks, line 305: This is what was changed, `int!` -> `Int!`

PR #904, opened 2020-08-11
Read prod configurations from config file.

godofredoc:
Prod configuration values were hardcoded in different places. This PR
reads those configs from a single place.

Bug:
  https://github.com/flutter/flutter/issues/63256

### DIFF
--- a/app_dart/lib/src/request_handlers/refresh_github_commits.dart
+++ b/app_dart/lib/src/request_handlers/refresh_github_commits.dart
@@ -6,6 +6,7 @@ import 'dart:async';
 import 'dart:convert';
 import 'dart:io';
 
+import 'package:cocoon_service/src/service/luci.dart';
 import 'package:gcloud/db.dart';
 import 'package:github/github.dart';
 import 'package:googleapis/bigquery/v2.dart';
@@ -198,10 +199,12 @@ class RefreshGithubCommits extends ApiRequestHandler<Body> {
     final List<Task> tasks = <Task>[
       // These built-in tasks are not listed in the manifest.
       newTask('cirrus', 'cirrus', <String>['can-update-github'], false, 0),
-      newTask('mac_bot', 'chromebot', <String>['can-update-chromebots'], false, 0),
-      newTask('linux_bot', 'chromebot', <String>['can-update-chromebots'], false, 0),
-      newTask('windows_bot', 'chromebot', <String>['can-update-chromebots'], false, 0),
     ];
+
+    final List<LuciBuilder> prodBuilders = await LuciBuilder.getProdBuilders('flutter', config);
+    for (LuciBuilder builder in prodBuilders) {
+      tasks.add(newTask(builder.name, 'chromebot', <String>['can-update-github'], false, 0));
+    }
 
     final YamlMap yaml = await _loadDevicelabManifest(sha);
     final Manifest manifest = Manifest.fromJson(yaml);
--- a/app_dart/lib/src/request_handlers/refresh_github_commits.dart
+++ b/app_dart/lib/src/request_handlers/refresh_github_commits.dart
@@ -203,7 +203,7 @@ class RefreshGithubCommits extends ApiRequestHandler<Body> {
 
     final List<LuciBuilder> prodBuilders = await LuciBuilder.getProdBuilders('flutter', config);
     for (LuciBuilder builder in prodBuilders) {
-      tasks.add(newTask(builder.name, 'chromebot', <String>['can-update-github'], false, 0));
+      tasks.add(newTask(builder.name, 'chromebot', <String>['can-update-github'], builder.flaky, 0));
     }
 
     final YamlMap yaml = await _loadDevicelabManifest(sha);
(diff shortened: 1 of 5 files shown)


app_dart/lib/src/request_handlers/refresh_github_commits.dart
```diff
@@ -198,11 +199,13 @@ class RefreshGithubCommits extends ApiRequestHandler<Body> {
...
       // These built-in tasks are not listed in the manifest.
       newTask('cirrus', 'cirrus', <String>['can-update-github'], false, 0),
-      newTask('mac_bot', 'chromebot', <String>['can-update-chromebots'], false, 0),
-      newTask('linux_bot', 'chromebot', <String>['can-update-chromebots'], false, 0),
-      newTask('windows_bot', 'chromebot', <String>['can-update-chromebots'], false, 0),
     ];
 
+    final List<LuciBuilder> prodBuilders = await LuciBuilder.getProdBuilders('flutter', config);
+    for (LuciBuilder builder in prodBuilders) {
+      tasks.add(newTask(builder.name, 'chromebot', <String>['can-update-github'], false, 0));
```
keyonghan, line 206: shall we add and use builder.flaky instead? since flaky is being added here https://github.com/flutter/flutter/pull/63492/files

godofredoc: Done.

keyonghan:
`flaky` needs to be added for `prod` as well in fake_cocoon_config: https://github.com/flutter/cocoon/blob/master/app_dart/test/src/datastore/fake_cocoon_config.dart#L205

`refresh_chromebot_status_test` and `push_engine_status_to_github_test` need to call the above fake function.

app_dart/lib/src/service/luci.dart
```diff
@@ -186,6 +187,10 @@ class LuciBuilder {
   @JsonKey(required: true, disallowNullValue: true)
   final String repo;
 
+  /// The name of the repository for which this builder runs.
```
keyonghan, line 190: update doc?

app_dart/lib/src/service/luci.dart
```diff
@@ -186,6 +187,10 @@ class LuciBuilder {
...
+  @JsonKey(required: true, disallowNullValue: true)
+  final bool flaky;
```
keyonghan, line 192: JsonSerialization needs update as well: https://github.com/flutter/cocoon/tree/master/app_dart#json

godofredoc:
> `flaky` needs to be added for `prod` as well in fake_cocoon_config: https://github.com/flutter/cocoon/blob/master/app_dart/test/src/datastore/fake_cocoon_config.dart#L205
> 
> `refresh_chromebot_status_test` and `push_engine_status_to_github_test` need to call the above fake function.

Not sure what you mean by prod? if it is the file the pending change already contains the flaky property. Added the flaky property to the fake_config.

keyonghan:
> > `flaky` needs to be added for `prod` as well in fake_cocoon_config: https://github.com/flutter/cocoon/blob/master/app_dart/test/src/datastore/fake_cocoon_config.dart#L205
> > `refresh_chromebot_status_test` and `push_engine_status_to_github_test` need to call the above fake function.
> 
> Not sure what you mean by prod? if it is the file the pending change already contains the flaky property. Added the flaky property to the fake_config.

By `prod` here I meant adding `flaky` property for the `bucket==prod` case. I guess this is what you have done.

app_dart/lib/src/service/luci.dart
godofredoc: Done

app_dart/lib/src/service/luci.dart
godofredoc: Done

godofredoc:
> `flaky` needs to be added for `prod` as well in fake_cocoon_config: https://github.com/flutter/cocoon/blob/master/app_dart/test/src/datastore/fake_cocoon_config.dart#L205
> 
> `refresh_chromebot_status_test` and `push_engine_status_to_github_test` need to call the above fake function.

yeah, this is confusing because only flutter repo is using prod_builders.json file.

keyonghan:
> > `flaky` needs to be added for `prod` as well in fake_cocoon_config: https://github.com/flutter/cocoon/blob/master/app_dart/test/src/datastore/fake_cocoon_config.dart#L205
> > `refresh_chromebot_status_test` and `push_engine_status_to_github_test` need to call the above fake function.
> 
> yeah, this is confusing because only flutter repo is using prod_builders.json file.

A condition check may help:
```
if  (bucket=='prod' and repo='flutter') {
  return builder_with_flaky_flag
}
```
All other cases remain unchanged.

godofredoc:
> > > `flaky` needs to be added for `prod` as well in fake_cocoon_config: https://github.com/flutter/cocoon/blob/master/app_dart/test/src/datastore/fake_cocoon_config.dart#L205
> > > `refresh_chromebot_status_test` and `push_engine_status_to_github_test` need to call the above fake function.
> > 
> > 
> > yeah, this is confusing because only flutter repo is using prod_builders.json file.
> 
> A condition check may help:
> 
> ```
> if  (bucket=='prod' and repo='flutter') {
>   return builder_with_flaky_flag
> }
> ```
> 
> All other cases remain unchanged.

I updated that case to include flaky flag, but should get rid of those in the future as they are not used.

godofredoc:
Any other comments? I'd like to land this change today

PR #907, opened 2020-08-12
Enable class BranchLuciBuilder as map key by overriding hashCode & ==

keyonghan:
Cocoon has been using `BranchLuciBuilder` as a map key when refreshing chromebot status. When multiple runs happen for a commit `sha`, the map is expected to update the value of existing key of `branchLuciBuilder`. But it treats the key (which shares same values) as a new key, failing to update with real LUCI statuses.

This PR fixes the above issue by overriding `hashCode` and operator `==`.

### DIFF
--- a/app_dart/test/service/luci_test.dart
+++ b/app_dart/test/service/luci_test.dart
@@ -10,13 +10,15 @@ void main() {
   BranchLuciBuilder branchLuciBuilder1;
   BranchLuciBuilder branchLuciBuilder2;
 
-  test('listCommits decodes all relevant fields of each commit', () async {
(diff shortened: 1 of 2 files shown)


app_dart/test/service/luci_test.dart
```diff
@@ -0,0 +1,22 @@
+// Copyright 2020 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:cocoon_service/src/service/luci.dart';
+
+import 'package:test/test.dart';
+
+void main() {
```
godofredoc, line 9: Did we forget the real change? seems like only tests were uploaded

keyonghan: Could you refresh the page. The main change is in the first commit.

PR #909, opened 2020-08-12
Add logic to check if luci try builders are enabled

keyonghan:
This is part of https://github.com/flutter/flutter/issues/63437

### DIFF
--- a/app_dart/lib/src/foundation/utils.dart
+++ b/app_dart/lib/src/foundation/utils.dart
@@ -85,6 +85,8 @@ Future<RepositorySlug> repoNameForBuilder(List<Map<String, dynamic>> builders, S
 }
 
 /// Gets supported luci builders based on [bucket] and [repo] via GitHub http request.
+/// 
+/// Only `enabled` luci builders will be returned.
 Future<List<Map<String, dynamic>>> getRepoBuilders(HttpClientProvider branchHttpClientProvider, Logging log,
     GitHubBackoffCalculator gitHubBackoffCalculator, String bucket, String repo) async {
   final String filePath = repo == 'engine' ? '$repo/master/ci/dev/' : '$repo/master/dev/';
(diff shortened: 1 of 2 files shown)


keyonghan:
> Do we need to update the try json files in the different repos first?

We can do this in parallel. `where((Map<String, dynamic> element) => (element['enabled'] as bool) ?? true)` considers the null case, replacing with default `true`.

app_dart/lib/src/foundation/utils.dart
```diff
@@ -95,7 +97,10 @@ Future<List<Map<String, dynamic>>> getRepoBuilders(HttpClientProvider branchHttp
   Map<String, dynamic> builderMap;
   builderMap = json.decode(builderContent) as Map<String, dynamic>;
   final List<dynamic> builderList = builderMap['builders'] as List<dynamic>;
-  return builderList.map((dynamic builder) => builder as Map<String, dynamic>).toList();
+  return builderList
+      .map((dynamic builder) => builder as Map<String, dynamic>)
+      .where((Map<String, dynamic> element) => (element['enabled'] as bool) ?? true)
+      .toList();
```
godofredoc, line 103: Do we need to add the property to the model class?

app_dart/test/foundation/utils_test.dart
```diff
@@ -118,11 +120,11 @@ void main() {
         lucBuilderHttpClient = FakeHttpClient();
         log = FakeLogging();
       });
-      test('returns luci builders', () async {
+      test('returns enabled luci builders', () async {
         lucBuilderHttpClient.request.response.body = luciBuilders;
         final List<Map<String, dynamic>> builders =
             await getRepoBuilders(() => lucBuilderHttpClient, log, (int attempt) => Duration.zero, 'try', 'cocoon');
-        expect(builders.length, 2);
+        expect(builders.length, 1);
```
godofredoc, line 127: Can we also add a test for the validation of the model class?

app_dart/lib/src/foundation/utils.dart
keyonghan: The model class is for prod builders usage only. We don't need `enabled` flag for prod builders, right?

godofredoc: > The model class is for prod builders usage only. We don't need `enabled` flag for prod builders, right?

Good point I though we were using the model class for try too.

PR #910, opened 2020-08-13
[flutter-gold] Add web tests to Cirrus build checks

Piinks:
With framework tests in the process of being migrated to LUCI, `framework` shards are in the process of being removed from Cirrus builds. This change adds `web` shards to the conditional Cirrus check for flutter-gold.

### DIFF
--- a/app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
+++ b/app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
@@ -98,8 +98,8 @@ class PushGoldStatusToGithub extends ApiRequestHandler<Body> {
             '${pr.head.sha}: $taskName ($status)');
 
         cirrusCheckStatuses.add(status);
-        if (taskName.contains('framework')) {
-          // Any pull request that runs a framework shard runs golden file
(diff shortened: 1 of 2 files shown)


Piinks:
> LGTM
> 
> Since we are moving web test to luci (soon) as well, how will the gold status check be by then?

Good point. I'm not familiar with the exact timeline or new naming conventions since the LUCI builds are still in the process of being migrated/sharded. Do we know if the migrated LUCI builds will have similar (framework, web) naming conventions? If so I can just transfer the same logic when we remove the Cirrus checks here.

Piinks:
Started a discussion here: https://discord.com/channels/608014603317936148/608021351567065092/743876278658007041

Piinks:
Going to refactor a bit more to account for the rest of the luci migration

godofredoc:
> framework, web
We are planning to complete the migration before the end of next week. The names will the current shard name prefixed by platform.

Here is an example: https://devops-console-oss.corp.google.com/flutter/flutter/+/master:dev/prod_builders.json

Piinks:
@godofredoc confirmed it is ok to just remove all the cirrus logic now for this check as the relevant tests are now running on luci. I will re-open as a new PR.

PR #912, opened 2020-08-14
Give authors with same initial differently colored avatars

goderbauer:
Authors that have the same first initial are now shown with avatars in different colors to make it easier to differentiate the commits of "goderbauer" and "godofredo" in the dashboard with one glance :P


Piinks:
Can you update the test cases for the a11y contrast test? The color choice was changed recently to make sure it was readable.
https://github.com/flutter/cocoon/blob/master/app_flutter/test/widgets/accessibility_test.dart

goderbauer:
@CaseyHillers Updated goldens
@Piinks added some longer names to the a11y test

PTAL

goderbauer:
It is unclear to me why the cocoon check is failing: https://ci.chromium.org/p/flutter/builders/try/Cocoon/1212?

There's no error message in the log. Does anybody know what's going on there?

Piinks:
> It is unclear to me why the cocoon check is failing: https://ci.chromium.org/p/flutter/builders/try/Cocoon/1212?
> 
> There's no error message in the log. Does anybody know what's going on there?

If you scroll to the bottom to the failure of step 22, you can slick stdout to see the error message

goderbauer:
There's no failure message in there: 

https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket.appspot.com/8871698349697998480/+/steps/app_flutter/0/stdout

But @godofredoc pointed out that I may need to run dartfmt. Let's see if that fixes things.

Piinks:
Oh that's strange. Are you logged in? I see:
Screen Shot 2020-08-17 at 11 04 51 AM

goderbauer:
@Piinks I see the same thing you're showing in the screenshot. None of that looks like a failure, though...

goderbauer:
OK, dartfmt fixed the issue... We should probably improve https://github.com/flutter/cocoon/blob/954881096116592d91b65dc8c550624fe097a70a/test_utilities/bin/flutter_test_runner.sh#L26 to print a more actionalbe failure message if dartfmt fails instead of just printing file names without context.

PR #913, opened 2020-08-18
save files listed in result[detailFiles] in a GCS bucket

flar:
Note that the CI part of these changes have not been tested. The commented-out code in runner.dart was used to simulate what might happen during a CI run by using a local agent run.

### DIFF
--- a/agent/lib/src/agent.dart
+++ b/agent/lib/src/agent.dart
@@ -159,7 +159,8 @@ class Agent {
 
   Future<void> saveDetailFile(String taskKey, String detailFilename) async {
     await uploadLogChunk(taskKey, '\n\nUploading task details from $detailFilename');
-    await cpFileToGcs(detailFilename, 'gs://flutter-dashboard-task-detail/$taskKey/');
+    final String dirName = taskKey ?? 'testing';
+    await cpFileToGcs(detailFilename, 'gs://flutter-dashboard-task-detail/$dirName/');
   }
 
   Future<String> getAuthenticationStatus() async {
--- a/agent/lib/src/agent.dart
+++ b/agent/lib/src/agent.dart
@@ -158,7 +158,7 @@ class Agent {
   }
 
   Future<void> saveDetailFile(String taskKey, String detailFilename) async {
-    await uploadLogChunk(taskKey, '\n\nUploading task details from $detailFilename');
+    await uploadLogChunk(taskKey, '\n\nUploading task details from $detailFilename\n');
     final String dirName = taskKey ?? 'testing';
     await cpFileToGcs(detailFilename, 'gs://flutter-dashboard-task-detail/$dirName/');
   }
--- a/agent/lib/src/agent.dart
+++ b/agent/lib/src/agent.dart
@@ -157,6 +157,11 @@ class Agent {
     });
   }
 
+  Future<void> saveDetailFile(String taskKey, String detailFilename) async {
+    await uploadLogChunk(taskKey, '\n\nUploading task details from $detailFilename');
+    await cpFileToGcs(detailFilename, 'gs://flutter-dashboard-task-detail/$taskKey/');
+  }
+
   Future<String> getAuthenticationStatus() async {
     return (await _cocoon('get-authentication-status'))['Status'] as String;
   }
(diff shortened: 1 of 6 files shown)


agent/lib/src/agent.dart
```diff
@@ -157,6 +157,11 @@ class Agent {
     });
   }
 
+  Future<void> saveDetailFile(String taskKey, String detailFilename) async {
+    await uploadLogChunk(taskKey, '\n\nUploading task details from $detailFilename');
+    await cpFileToGcs(detailFilename, 'gs://flutter-dashboard-task-detail/$taskKey/');
```
godofredoc, line 162: @keyonghan do you remember the problem with gsutil concurrency when called from dart? I remember it was blocking the agents and had to revert.

agent/lib/src/utils.dart
```diff
@@ -649,3 +649,20 @@ Future<Null> unlockKeyChain() async {
         canFail: false, silent: true);
   }
 }
+
+Future<Null> cpFileToGcs(String file, String gsBucket) async {
+  if (Platform.isLinux || Platform.isMacOS) {
+    await eval('gsutil.py', ['cp', file, gsBucket],
```
godofredoc, line 655: where is gsutil.py coming from? is this depot_tools gsutil?

flar: I took this code from here: https://github.com/flutter/cocoon/compare/master...keyonghan:save_log_to_gcs

agent/lib/src/agent.dart
keyonghan: I don't remember we ever merged PRs with gsutil. Last time I quitted the gsutil PR is due to not working from Windows bot, and stackdriver was chosen as an alternative for viewing task log.

agent/lib/src/utils.dart
keyonghan: I tried https://cloud.google.com/storage/docs/gsutil_install#mac, and it worked on agent bots.
To make agents work, gsutil needs to be installed beforehand, and also we need to copy over the authentication (key & .boto).

agent/lib/src/utils.dart
```diff
@@ -649,3 +649,20 @@ Future<Null> unlockKeyChain() async {
         canFail: false, silent: true);
   }
 }
+
+Future<Null> cpFileToGcs(String file, String gsBucket) async {
+  if (Platform.isLinux || Platform.isMacOS) {
```
digiter, line 654: Do we want to support windows?

flar: Windows would fall into the else clause which runs the command a different way. Is that a problem?

I'm going by the previous experiences of @keyonghan

agent/lib/src/utils.dart
flar: Are those installation and authentication steps already part of creating a new agent? What work is needed for that outside of this PR?

agent/lib/src/agent.dart
flar: Ah, so the "else" clause in `cpFileToGcs()` doesn't capture Windows?

What's the plan to move forward for Windows support? This PR would break the Windows agents if your claim is true.

agent/lib/src/utils.dart
flar: On the other hand, in another comment @keyonghan mentions that the previous PR upon which this is based failed on Windows agents...

flar:
@godofredoc @digiter what is the outcome of the conversations that you started? Are there still unresolved issues?

@keyonghan do you have any more information for any of those conversations?

agent/lib/src/utils.dart
keyonghan: (Sorry missed this comment!)
No, they are not part of the agent at this moment. We need to install/authenticate with extra work.

I would suggest test on a specific agent first to make sure CI works as expected before we submit this change and synchronize across all agents. Also, it would be good to compare test execution time before and after this change.

> What work is needed for that outside of this PR?
Install the gsutil; generate `.boto` file and copy over to the agent, also the key json file; (i remembered I have already installed on some Mac and Linux bots), etc.

Please check the commands/steps I sent earlier, and let me know if you need help access the devicelab agents.

godofredoc:
Closing as outdated.

PR #914, opened 2020-08-18
Fix calculation of engine github status.

godofredoc:
The calculation of the engine github status was failing when the first
task succeded and subsequent tasks failed.

Bug:
  https://github.com/flutter/flutter/issues/64061


godofredoc:
> Are we getting in `tasks` any pending tasks?

Potentially yes but pending tasks should not affect the state of the tree.

PR #917, opened 2020-08-19
Enable filtering builders based on changed files in PRs

keyonghan:
This is to solve https://github.com/flutter/flutter/issues/62841

### DIFF
--- a/app_dart/lib/src/foundation/utils.dart
+++ b/app_dart/lib/src/foundation/utils.dart
@@ -103,14 +103,31 @@ Future<List<Map<String, dynamic>>> getRepoBuilders(HttpClientProvider branchHttp
       .toList();
 }
 
-/// Returns [builder] when any file falls in any pre-set dir.
+/// Returns a LUCI [builder] list that covers changed [files].
+///
+/// [builders]: enabled luci builders.
+/// [files]: changed files in corresponding PRs.
+///
+/// [builder] is with format:
+/// {
+///   "name":"yyy",
+///   "repo":"flutter",
+///   "taskName":"zzz",
+///   "enabled":true,
+///   "run_if":["a/b/", "c/d/**"]
+/// }
+///
+/// [file] is based on repo root: `a/b/c.dart`.
 Future<List<Map<String, dynamic>>> getFilteredBuilders(List<Map<String, dynamic>> builders, List<String> files) async {
   final List<Map<String, dynamic>> filteredBuilders = <Map<String, dynamic>>[];
   for (Map<String, dynamic> builder in builders) {
     final List<String> dirs = List<String>.from((builder['run_if'] as List<dynamic>) ?? <String>['']);
     for (String dir in dirs) {
+      dir = dir.replaceAll('**', '[a-zA-Z_\/]?');
+      dir = dir.replaceAll('*', '[a-zA-Z_\/]*');
       // If a file is found within a pre-set dir, the builder needs to run. No need to check further.
-      if (files.any((String file) => file.startsWith(dir))) {
+      final RegExp regExp = RegExp('^$dir');
+      if (dir.isEmpty || files.any((String file) => regExp.hasMatch(file))) {
         filteredBuilders.add(builder);
         break;
       }
--- a/app_dart/lib/src/foundation/utils.dart
+++ b/app_dart/lib/src/foundation/utils.dart
@@ -103,6 +103,22 @@ Future<List<Map<String, dynamic>>> getRepoBuilders(HttpClientProvider branchHttp
       .toList();
 }
 
+/// Returns [builder] when any file falls in any pre-set dir.
+Future<List<Map<String, dynamic>>> getFilteredBuilders(List<Map<String, dynamic>> builders, List<String> files) async {
+  final List<Map<String, dynamic>> filteredBuilders = <Map<String, dynamic>>[];
+  for (Map<String, dynamic> builder in builders) {
+    final List<String> dirs = List<String>.from((builder['run_if'] as List<dynamic>) ?? <String>['']);
+    for (String dir in dirs) {
+      // If a file is found within a pre-set dir, the builder needs to run. No need to check further.
+      if (files.any((String file) => file.startsWith(dir))) {
+        filteredBuilders.add(builder);
+        break;
+      }
+    }
+  }
+  return filteredBuilders;
+}
+
 Future<void> insertBigquery(
     String tableName, Map<String, dynamic> data, TabledataResourceApi tabledataResourceApi, Logging log) async {
   // Define const variables for [BigQuery] operations.
--- a/app_dart/lib/src/foundation/utils.dart
+++ b/app_dart/lib/src/foundation/utils.dart
@@ -87,6 +87,7 @@ Future<RepositorySlug> repoNameForBuilder(List<Map<String, dynamic>> builders, S
 /// Gets supported luci builders based on [bucket] and [repo] via GitHub http request.
 ///
 /// Only `enabled` luci builders will be returned.
+// TODO(keyonghan): update non-type-safe Map<String, dynamic> to @JsonSeriablizable annotated class. https://github.com/flutter/flutter/issues/64286
 Future<List<Map<String, dynamic>>> getRepoBuilders(HttpClientProvider branchHttpClientProvider, Logging log,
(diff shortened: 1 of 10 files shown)


app_dart/lib/src/foundation/utils.dart
```diff
@@ -103,6 +103,22 @@ Future<List<Map<String, dynamic>>> getRepoBuilders(HttpClientProvider branchHttp
       .toList();
 }
 
+/// Returns [builder] when any file falls in any pre-set dir.
```
godofredoc, line 106: Update? given that functions in this library are made for reusability let's add complete docs. E.g parameters received, return value and an example.

app_dart/lib/src/foundation/utils.dart
```diff
@@ -103,6 +103,22 @@ Future<List<Map<String, dynamic>>> getRepoBuilders(HttpClientProvider branchHttp
...
+Future<List<Map<String, dynamic>>> getFilteredBuilders(List<Map<String, dynamic>> builders, List<String> files) async {
+  final List<Map<String, dynamic>> filteredBuilders = <Map<String, dynamic>>[];
+  for (Map<String, dynamic> builder in builders) {
+    final List<String> dirs = List<String>.from((builder['run_if'] as List<dynamic>) ?? <String>['']);
+    for (String dir in dirs) {
```
godofredoc, line 111: How do we support * and **?

app_dart/lib/src/foundation/utils.dart
```diff
@@ -103,6 +103,22 @@ Future<List<Map<String, dynamic>>> getRepoBuilders(HttpClientProvider branchHttp
...
+      // If a file is found within a pre-set dir, the builder needs to run. No need to check further.
+      if (files.any((String file) => file.startsWith(dir))) {
```
godofredoc, line 113: How do we differentiate from a full path and wildcards, e.g:

/a/b file in a
and
/a/bc/a ?

Should we implement the functionality using regexes? e.g * replaced with '[a-zA-Z_]*' and ** as[a-zA-Z_/]?

app_dart/lib/src/service/github_service.dart
```diff
@@ -99,4 +99,25 @@ class GithubService {
         ..head = (PullRequestHead()..sha = commit['head']['sha'] as String);
     }).toList();
   }
+
+  // Returns changed files of [slug] and [prNumber].
+  // https://developer.github.com/v3/pulls/#list-pull-requests-files
+  Future<List<String>> listFiles(RepositorySlug slug, int prNumber) async {
```
godofredoc, line 105: Isn't this list of files already available in the PR event?

app_dart/test/foundation/utils_test.dart
```diff
@@ -180,5 +181,36 @@ void main() {
...
+
+      test('returns builders when run_if is null', () async {
+        files = <String>['a/b', 'c/d'];
+        builders = (json.decode('[{"name": "abc", "repo": "def", "task_name": "ghi", "flaky": true}]') as List<dynamic>)
+            .cast<Map<String, dynamic>>();
+        final List<Map<String, dynamic>> result = await getFilteredBuilders(builders, files);
+        expect(result, builders);
+      });
+
+      test('returns builders when run_if matches files', () async {
```
godofredoc, line 206: we'll need tests for the wildcards * and **

app_dart/lib/src/foundation/utils.dart
keyonghan: Existing cirrus seems to use **  only, do we have the use case of a single *?

app_dart/lib/src/foundation/utils.dart
keyonghan: Updated.

app_dart/lib/src/foundation/utils.dart
keyonghan: add * support.

app_dart/lib/src/foundation/utils.dart
keyonghan: updated.

app_dart/lib/src/service/github_service.dart
keyonghan: Doesn't seem so. Based on `CheckSuiteEvent`, we can track down to `CheckSuite`, then `PullRequest`, where `changedFilesCount` is available, but there is no `changedFiles` though.

app_dart/test/foundation/utils_test.dart
keyonghan: added.

app_dart/lib/src/foundation/utils.dart
```diff
@@ -103,6 +103,39 @@ Future<List<Map<String, dynamic>>> getRepoBuilders(HttpClientProvider branchHttp
...
+/// {
+///   "name":"yyy",
+///   "repo":"flutter",
+///   "taskName":"zzz",
+///   "enabled":true,
+///   "run_if":["a/b/", "c/d/**"]
+/// }
+///
+/// [file] is based on repo root: `a/b/c.dart`.
+Future<List<Map<String, dynamic>>> getFilteredBuilders(List<Map<String, dynamic>> builders, List<String> files) async {
```
digiter, line 121: It would be really helpful to fix the unreadable, non-type-safe Map<String, dynamic> everywhere with a `@JsonSerializable` annotated class.

godofredoc: +1 a list of map of maps is prone to errors.

app_dart/lib/src/foundation/utils.dart
```diff
@@ -103,6 +103,39 @@ Future<List<Map<String, dynamic>>> getRepoBuilders(HttpClientProvider branchHttp
...
+/// [builders]: enabled luci builders.
+/// [files]: changed files in corresponding PRs.
+///
+/// [builder] is with format:
+/// {
+///   "name":"yyy",
+///   "repo":"flutter",
+///   "taskName":"zzz",
+///   "enabled":true,
+///   "run_if":["a/b/", "c/d/**"]
```
godofredoc, line 117: What about single files?

app_dart/lib/src/foundation/utils.dart
godofredoc: should we call it glob instead of dirs? dirs is confusing as it can be a file a folder or a glob, glob matches all of them.

app_dart/test/foundation/utils_test.dart
```diff
@@ -180,5 +181,56 @@ void main() {
...
+            (json.decode('[{"name": "abc", "repo": "def", "task_name": "ghi", "flaky": true, "run_if": ["a/**"]}]')
+                    as List<dynamic>)
+                .cast<Map<String, dynamic>>();
+        files = <String>['a/b', 'c/d'];
+        final List<Map<String, dynamic>> result = await getFilteredBuilders(builders, files);
+        expect(result, builders);
+      });
+
+      test('returns builders when run_if matches files with both * and **', () async {
+        builders =
```
godofredoc, line 226: We need one more test for file and folder with the same name "a" and a/b/c. E.g a run_if = ["a"] does not match a files changes "a/b/c"

app_dart/lib/src/foundation/utils.dart
keyonghan: Agree. Updating to use annotated class will affect quite some files. Created TODO (https://github.com/flutter/flutter/issues/64286) to update that in a separate PR.

app_dart/lib/src/foundation/utils.dart
keyonghan: Replaced.

app_dart/lib/src/foundation/utils.dart
keyonghan: The case of single files has been covered. Please refer to the test below. Updated the doc to include a case of a single file and a case of single *

app_dart/test/foundation/utils_test.dart
keyonghan: Added.

keyonghan:
> LGTM, let's land and monitor to ensure builders are triggered correctly.

Will keep an eye on this tomorrow morning. 
With https://github.com/flutter/flutter/pull/64180, new PRs in flutter repo should be triggered based on `run_if` after a new deployment.

PR #918, opened 2020-08-19
[app_dart] Ensure chromebot tasks have valid format in datastore

digiter:
Prevents isFlaky and other required fields be null in datastore.

Bug: https://github.com/flutter/flutter/issues/64114

### DIFF
--- a/app_dart/lib/src/model/appengine/task.dart
+++ b/app_dart/lib/src/model/appengine/task.dart
@@ -4,7 +4,6 @@
 
 import 'package:gcloud/db.dart';
 import 'package:json_annotation/json_annotation.dart';
-import 'package:meta/meta.dart';
 
 import 'commit.dart';
 import 'key_converter.dart';
@@ -52,14 +51,13 @@ class Task extends Model {
     this.createTimestamp,
     this.name,
     this.isFlaky,
-  )   : assert(commitKey != null),
-        assert(createTimestamp != null),
+  )   : assert(createTimestamp != null),
         assert(name != null),
         assert(isFlaky != null),
         timeoutInMinutes = 0,
         requiredCapabilities = <String>['can-update-github'],
         stageName = 'chromebot' {
-    final Key key = commitKey.append(Task);
+    final Key key = commitKey?.append(Task);
     parentKey = key?.parent;
     id = key?.id;
     status = Task.statusNew;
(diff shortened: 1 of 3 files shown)


app_dart/lib/src/request_handlers/refresh_github_commits.dart
```diff
@@ -203,7 +203,7 @@ class RefreshGithubCommits extends ApiRequestHandler<Body> {
 
     final List<LuciBuilder> prodBuilders = await LuciBuilder.getProdBuilders('flutter', config);
     for (LuciBuilder builder in prodBuilders) {
-      tasks.add(newTask(builder.taskName, 'chromebot', <String>['can-update-github'], builder.flaky, 0));
+      tasks.add(Task.chromebot(commitKey, createTimestamp, builder.taskName, builder.flaky));
```
keyonghan, line 206: maybe only one-line change `builder.flaky` => `builder.flay??false`?
If a builder is missing `isFlaky` flag, we will treat it as `false` by default.

app_dart/lib/src/model/appengine/task.dart
```diff
@@ -45,6 +45,24 @@ class Task extends Model {
...
+  // Represents a LUCI build.
+  Task.chromebot(
+    this.commitKey,
+    this.createTimestamp,
+    this.name,
+    this.isFlaky,
+  )   : assert(createTimestamp != null),
+        assert(name != null),
+        assert(isFlaky != null),
+        timeoutInMinutes = 0,
```
godofredoc, line 57: does 0 means unlimited?

app_dart/lib/src/request_handlers/refresh_github_commits.dart
digiter: What is the default value of `flaky`? It's not documented in https://github.com/flutter/flutter/blob/master/dev/README.md

app_dart/lib/src/model/appengine/task.dart
digiter: "The timeout of the task, or zero if the task has no timeout"

app_dart/lib/src/request_handlers/refresh_github_commits.dart
keyonghan: Now the `isFlaky` flag is required by cocoon backend. If for the bad case people forget or have typos, the default value should be `false`. @godofredoc 
Since `isFlaky` is treated as non-optional, not documented on the README.

digiter: OK

PR #919, opened 2020-08-19
Reland deleted logic to calculate tree status and add logging.

godofredoc:
We are not sure yet what is causing the tree to status to be green when
there are failures. We are reverting the previous logic, using only
builds of master for the calculation and adding logs to identify the
cause of the problem when it happens again.

Bug:
  https://github.com/flutter/flutter/issues/64061

### DIFF
--- a/app_dart/lib/src/request_handlers/push_engine_status_to_github.dart
+++ b/app_dart/lib/src/request_handlers/push_engine_status_to_github.dart
@@ -100,10 +100,21 @@ class PushEngineStatusToGithub extends ApiRequestHandler<Body> {
 
   String _getLatestStatus(List<LuciTask> tasks) {
     for (LuciTask task in tasks) {
-      if (task.status == Task.statusFailed) {
-        return GithubBuildStatusUpdate.statusFailure;
+      if (task.ref != 'refs/heads/master') {
+        log.debug('Skipping ${task.status} from commit ${task.commitSha} ref ${task.ref} builder ${task.builderName}');
+        continue;
+      }
+      switch (task.status) {
+        case Task.statusFailed:
+          log.debug('Using ${task.status} from commit ${task.commitSha} ref ${task.ref} builder ${task.builderName}');
+          return GithubBuildStatusUpdate.statusFailure;
+        case Task.statusSucceeded:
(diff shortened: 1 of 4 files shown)


christopherfujino:
@godofredoc is this the reland of a reverted change?

christopherfujino:
> @godofredoc is this the reland of a reverted change?

Ok never mind, I remember this code.

app_dart/lib/src/request_handlers/push_engine_status_to_github.dart
```diff
@@ -100,10 +100,21 @@ class PushEngineStatusToGithub extends ApiRequestHandler<Body> {
...
+        log.debug('Skipping ${task.status} from commit ${task.commitSha} ref ${task.ref} builder ${task.builderName}');
+        continue;
+      }
+      switch (task.status) {
+        case Task.statusFailed:
+          log.debug('Using ${task.status} from commit ${task.commitSha} ref ${task.ref} builder ${task.builderName}');
+          return GithubBuildStatusUpdate.statusFailure;
+        case Task.statusSucceeded:
+          log.debug('Using ${task.status} from commit ${task.commitSha} ref ${task.ref} builder ${task.builderName}');
+          return GithubBuildStatusUpdate.statusSuccess;
```
christopherfujino, line 113: Don't we still want to loop through everything in `tasks` if the first one is green?

godofredoc: The current logic is very inefficient:

1.- get the list of prod builders
2.- get the last 40 build for each builder
3.- run each builder through _getLatestStatus

The current logic will get the first pass or fail for each builder and set a failure status if one builder is red

godofredoc: Ideally later on we may want to refactor this logic to calculate per commit instead of per builders.

christopherfujino: Ahhhh, I see, that makes sense. Can you add a dartdoc to this method, explicitly calling out that it takes in the tasks for a particular builder. LGTM!

PR #922, opened 2020-08-24
Read try builder config file from commit rather than TOT  if there is a config change

keyonghan:
This is to fix: https://github.com/flutter/flutter/issues/64107

### DIFF
--- a/app_dart/lib/src/datastore/cocoon_config.dart
+++ b/app_dart/lib/src/datastore/cocoon_config.dart
@@ -71,8 +71,18 @@ class Config {
     return String.fromCharCodes(cacheValue).split(',');
   }
 
-  Future<List<LuciBuilder>> getRepoLuciBuilders(String bucket, String repo) async {
-    return await getRepoBuilders(Providers.freshHttpClient, loggingService, twoSecondLinearBackoff, bucket, repo);
+  Future<List<LuciBuilder>> getLuciBuilders(String bucket, String repo, {String commitSha}) async {
+    return await getRepoBuilders(Providers.freshHttpClient, loggingService, twoSecondLinearBackoff, bucket, repo,
+        commitSha: commitSha);
+  }
+
+  // Trigger builds via updated try builder config file in commit rather than TOT if any changes.
+  Future<List<LuciBuilder>> getLuciTryBuilders(List<String> files, String repo, String commitSha) async {
+    if (files.any((String file) => file.endsWith('try_builders.json'))) {
+      return await getLuciBuilders('try', repo, commitSha: commitSha);
+    } else {
+      return await getLuciBuilders('try', repo);
+    }
   }
 
   Future<String> _getSingleValue(String id) async {
--- a/app_dart/lib/src/datastore/cocoon_config.dart
+++ b/app_dart/lib/src/datastore/cocoon_config.dart
@@ -71,17 +71,12 @@ class Config {
     return String.fromCharCodes(cacheValue).split(',');
   }
 
-  /// Returns `try` or `prod` LUCI builders from TOT for [repo].
-  Future<List<LuciBuilder>> luciBuilders(String bucket, String repo) async {
-    return await getLuciBuilders(Providers.freshHttpClient, loggingService, twoSecondLinearBackoff, bucket, repo);
-  }
-
-  /// Returns LUCI try builders based on try_builders.json config file in the corresponding [commitSha], and
-  /// also based on filtering changed files in [prNumber] via property [run_if] in each config.
-  Future<List<LuciBuilder>> luciTryBuilders(String commitSha, RepositorySlug slug, int prNumber) async {
-    final GithubService githubService = await createGithubService(slug.owner, slug.name);
-    return await getLuciTryBuilders(
-        githubService, Providers.freshHttpClient, twoSecondLinearBackoff, loggingService, slug, prNumber, commitSha);
+  // Returns LUCI builders based on [bucket].
+  Future<List<LuciBuilder>> luciBuilders(String bucket, String repo, {String commitSha, int prNumber}) async {
+    final GithubService githubService = await createGithubService('flutter', repo);
+    return await getLuciBuilders(githubService, Providers.freshHttpClient, twoSecondLinearBackoff, loggingService,
+        RepositorySlug('flutter', repo), bucket,
+        prNumber: prNumber, commitSha: commitSha);
   }
 
   Future<String> _getSingleValue(String id) async {
--- a/app_dart/lib/src/datastore/cocoon_config.dart
+++ b/app_dart/lib/src/datastore/cocoon_config.dart
@@ -71,8 +71,9 @@ class Config {
     return String.fromCharCodes(cacheValue).split(',');
   }
 
-  // Returns LUCI builders based on [bucket].
-  Future<List<LuciBuilder>> luciBuilders(String bucket, String repo, {String commitSha, int prNumber}) async {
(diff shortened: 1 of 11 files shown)


app_dart/test/datastore/cocoon_config_test.dart
```diff
@@ -32,4 +35,25 @@ void main() {
...
+    FakeConfig config;
+    setUp(() {
+      config = FakeConfig();
+    });
+    test('returns updated builders when there is a config file change in commit', () async {
+      final List<String> files = <String>['dev/try_builders.json'];
+      final List<LuciBuilder> luciBuilders = await config.getLuciTryBuilders(files, 'cocoon', 'test');
+
+      expect(luciBuilders, <LuciBuilder>[const LuciBuilder(name: 'test', repo: 'cocoon', flaky: false)]);
+    });
```
digiter, line 49: It's hard to understand the test expectations without understanding fake_cocoon_config.dart.

keyonghan: Thanks for the comment. Added `cocoonTryBuilderValueTot`, and `cocoonTryBuilderValueCommit` to make test self-readable.

app_dart/lib/src/datastore/cocoon_config.dart
```diff
@@ -71,8 +71,18 @@ class Config {
     return String.fromCharCodes(cacheValue).split(',');
   }
 
-  Future<List<LuciBuilder>> getRepoLuciBuilders(String bucket, String repo) async {
```
digiter, line 74: Cocoon config is designed to wrap operations to Cocoon datastore. These functions have nothing to do with the datastore, therefore it's confusing to see them.

app_dart/lib/src/foundation/utils.dart
```diff
@@ -89,8 +89,9 @@ Future<RepositorySlug> repoNameForBuilder(List<LuciBuilder> builders, String bui
 ///
 /// Only `enabled` luci builders will be returned.
 Future<List<LuciBuilder>> getRepoBuilders(HttpClientProvider branchHttpClientProvider, Logging log,
```
digiter, line 91: The logic "fetch builders -> list files -> get filtered builders" is duplicated repetitively 3 times in luci_build_service and github_status_service. Since we always use filtered builders, it's better to move the logic into this function.

Bonus point is this can simply read `X_builders.json` from `commitSha` (including when it's `master`), regardless whether the PR contains a change to  `X_builders.json`.

app_dart/lib/src/datastore/cocoon_config.dart
keyonghan: Cocoon config has now been extended to support functions not limited to datastore, including github service, gold status refresh, etc. Nevertheless, all these are configs related to cocoon.
It might be better to move `cocoon_config.dart` out of existing `datastore` folder to avoid confusion, or rename `datastore`. 
Created issue https://github.com/flutter/flutter/issues/64666 to track.

app_dart/lib/src/foundation/utils.dart
keyonghan: Moved all logics here as suggested (renamed function to `getLuciTryBuilders`), and enabled default reading from `commitSha` for all Try builders.

app_dart/lib/src/foundation/utils.dart
```diff
@@ -104,6 +106,24 @@ Future<List<LuciBuilder>> getRepoBuilders(HttpClientProvider branchHttpClientPro
       .toList();
 }
 
+/// First gets luci try builders from any commit.
+///
+/// Returns luci try builders by filtering `run_if` property over changed files.
+Future<List<LuciBuilder>> getLuciTryBuilders(
```
digiter, line 112: It's confusing to see both `getLuciBuilders` and `getLuciTryBuilders`, especially `getLuciBuilders` can return try builders as well. Consider merging these two functions.

keyonghan: Yeah, thought about this earlier, combining logics would add the complexity to the function. But as you pointed out, combining should clear out any confusion. Just combined logics to a single function `getLuciBuilders`.

app_dart/lib/src/datastore/cocoon_config.dart
keyonghan: `cocoon_config.dart` is now the central place to manage all configs related to `cocoon` backend server, not limited to the `Cocoon` repo. 
[`githubClient`](https://github.com/flutter/cocoon/blob/master/app_dart/lib/src/datastore/cocoon_config.dart#L263), [`cirrusGraphQLClient`](https://github.com/flutter/cocoon/blob/master/app_dart/lib/src/datastore/cocoon_config.dart#L290), etc. are similarly servicing other repos.

When you talk about complexity, I guess you are referring to the length or function numbers. If that is the case, we may need to consider to split the whole config file to sub ones.

keyonghan:
@digiter Could you please let me know if you still have any other comment? I would like to land this soon to address the P1 issue: https://github.com/flutter/flutter/issues/64107.

app_dart/lib/src/datastore/cocoon_config.dart
godofredoc: Let's move the discussion to the bug. Ideally we still want to have a single object that we can pass around to get the configurations. However we can do better on splitting the functionality as luci_configs, github_configs, etc.

app_dart/lib/src/foundation/utils.dart
```diff
@@ -88,12 +89,13 @@ Future<RepositorySlug> repoNameForBuilder(List<LuciBuilder> builders, String bui
...
-    GitHubBackoffCalculator gitHubBackoffCalculator, String bucket, String repo) async {
-  final String filePath = repo == 'engine' ? '$repo/master/ci/dev/' : '$repo/master/dev/';
+Future<List<LuciBuilder>> getLuciBucketBuilders(HttpClientProvider luciHttpClientProvider, Logging log,
+    GitHubBackoffCalculator gitHubBackoffCalculator, String bucket, String repo,
+    {String commitSha = 'master'}) async {
+  final String filePath = repo == 'engine' ? '$repo/$commitSha/ci/dev/' : '$repo/$commitSha/dev/';
   final String fileName = bucket == 'try' ? 'try_builders.json' : 'prod_builders.json';
   String builderContent =
-      await remoteFileContent(branchHttpClientProvider, log, gitHubBackoffCalculator, '/flutter/$filePath$fileName');
+      await remoteFileContent(luciHttpClientProvider, log, gitHubBackoffCalculator, '/flutter/$filePath$fileName');
```
godofredoc, line 98: Why we need an specific http client for luci?

app_dart/lib/src/foundation/utils.dart
```diff
@@ -104,6 +106,27 @@ Future<List<LuciBuilder>> getRepoBuilders(HttpClientProvider branchHttpClientPro
       .toList();
 }
 
+/// Returns LUCI builders based on [bucket].
```
godofredoc, line 109: How is this different from the previous function? from docs they look like the same. Can we update docs or even better merge the two functions?

keyonghan: Merged. It is more concise now.

app_dart/lib/src/foundation/utils.dart
keyonghan: This is a name change/update (the old one copying from elsewhere). We need http client to request json try configs.

godofredoc (approved):

PR #926, opened 2020-08-25
Save LUCI task to BigQuery for flakiness analysis

keyonghan:
This is part of https://github.com/flutter/flutter/issues/64090

Existing Cocoon saves only DeviceLab tasks to BigQuery. This PR saves LUCI tasks to BigQuery as well when tasks finish.

### DIFF
--- a/app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
+++ b/app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
@@ -90,8 +90,31 @@ class RefreshChromebotStatus extends ApiRequestHandler<Body> {
           update.builderName = builder.name;
           update.luciBucket = 'luci.flutter.prod';
           await datastore.insert(<Task>[update]);
+          // Save luci task record to BigQuery only when task finishes.
+          if (update.status == Task.statusFailed || update.status == Task.statusSucceeded) {
+            await _insertBigquery(update);
+          }
         }
       }
     }
+  }
+
+  Future<void> _insertBigquery(Task task) async {
+    const String bigqueryTableName = 'Task';
+    final Map<String, dynamic> bigqueryData = <String, dynamic>{
+      'ID': task.commitKey.id,
+      'CreateTimestamp': task.createTimestamp,
+      'StartTimestamp': task.startTimestamp,
+      'EndTimestamp': task.endTimestamp,
+      'Name': task.name,
+      'Attempts': task.attempts,
+      'IsFlaky': task.isFlaky,
+      'TimeoutInMinutes': task.timeoutInMinutes,
+      'RequiredCapabilities': task.requiredCapabilities,
+      'ReservedForAgentID': task.reservedForAgentId,
+      'StageName': task.stageName,
+      'Status': task.status,
+    };
+    await insertBigquery(bigqueryTableName, bigqueryData, await config.createTabledataResourceApi(), log);
   }
 }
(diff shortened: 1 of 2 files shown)


app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
```diff
@@ -90,8 +90,31 @@ class RefreshChromebotStatus extends ApiRequestHandler<Body> {
...
+          // Save luci task record to BigQuery only when task finishes.
+          if (update.status == Task.statusFailed || update.status == Task.statusSucceeded) {
+            await _insertBigquery(update);
+          }
         }
       }
     }
   }
+
+  Future<void> _insertBigquery(Task task) async {
```
godofredoc, line 102: should this be a generic function that works for devicelab, luci, etc? if it is let's move to a utilities class and reuse instead if rewrite here.

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
```diff
@@ -90,8 +90,31 @@ class RefreshChromebotStatus extends ApiRequestHandler<Body> {
           update.builderName = builder.name;
           update.luciBucket = 'luci.flutter.prod';
           await datastore.insert(<Task>[update]);
+          // Save luci task record to BigQuery only when task finishes.
+          if (update.status == Task.statusFailed || update.status == Task.statusSucceeded) {
```
godofredoc, line 94: Are those two the only accepted values?

app_dart/test/request_handlers/refresh_chromebot_status_test.dart
```diff
@@ -168,9 +174,42 @@ void main() {
...
+    test('save data to BigQuery when task finishes', () async {
+      final Commit commit = Commit(key: config.db.emptyKey.append(Commit, id: 'abc'), sha: 'abc');
+      final Task task = Task(key: commit.key.append(Task, id: 123), commitKey: commit.key, status: Task.statusNew);
+      config.tabledataResourceApi = tabledataResourceApi;
+      config.db.values[commit.key] = commit;
+      config.db.values[task.key] = task;
+
+      final Map<BranchLuciBuilder, Map<String, List<LuciTask>>> luciTasks =
+          Map<BranchLuciBuilder, Map<String, List<LuciTask>>>.fromIterable(
+              await LuciBuilder.getProdBuilders('flutter', config),
```
digiter, line 186: Does this make http requests? That would make tests flaky.

keyonghan: No. It is using `FakeConfig`, which will not make http requests.

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
keyonghan: Yes, we want to insert record only when task finishes with a fail or success.
There are status change from `New` to `In Process`, for which we do not intend to save the record (BigQuery does not support update, and it simply insert another record.

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
keyonghan: The generic function is `insertBigquery` defined in `utils.dart`. Here the `_insertBigquery` only prepares the data to be injected to the generic function.

app_dart/test/request_handlers/refresh_chromebot_status_test.dart
```diff
@@ -236,8 +277,9 @@ void main() {
     });
 
     test('update task status for non master branch', () async {
+      config.tabledataResourceApi = tabledataResourceApi;
```
digiter, line 280: Nit: consider moving this line to setUp().

keyonghan: Done.

PR #929, opened 2020-08-28
This is to ensure slow devices are being discovered.

godofredoc:
Bug:
  https://github.com/flutter/flutter/issues/64753

### DIFF
--- a/agent/lib/src/adb.dart
+++ b/agent/lib/src/adb.dart
@@ -295,6 +295,9 @@ class IosDeviceDiscovery implements DeviceDiscovery {
 
   @override
   Future<List<Device>> discoverDevices({int retriesDelayMs = 10000}) async {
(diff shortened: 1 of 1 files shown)


agent/lib/src/adb.dart
```diff
@@ -295,6 +295,9 @@ class IosDeviceDiscovery implements DeviceDiscovery {
 
   @override
   Future<List<Device>> discoverDevices({int retriesDelayMs = 10000}) async {
+    // Warm up device list. This is to ensure slow devices are discovered.
+    // https://github.com/flutter/flutter/issues/64753
+    await eval('xcrun', ['xcdevice', 'list', '--timeout', '10']);
```
jmagman, line 300: 🤷‍♀️  Let's try it.

PR #930, opened 2020-08-28
Update endTimestamp for luci task when saving to bigquery

keyonghan:
This is a follow-up of https://github.com/flutter/cocoon/pull/926.

Existing logic keeps saving 0 as endTimestamp, but we need to have this value to distinguish different records of same `sha` and same `name`.


keyonghan:
It turns out `Attempts` is good enough.

PR #931, opened 2020-09-01
Update app_flutter to match the latest dartfmt rules

jason-simmons:


flar:
Members of the infra team have other ideas, like not doing format testing during a customer test run. I'll wait for someone on that team to approve this or suggest another approach.

Tagging @christopherfujino

christopherfujino:
> Members of the infra team have other ideas, like not doing format testing during a customer test run. I'll wait for someone on that team to approve this or suggest another approach.
> 
> Tagging @christopherfujino

I'm not part of the infra team, and I wasn't suggesting not landing this. However, I find the cocoon analyzer failure puzzling.

keyonghan:
Cocoon has a test to enforce `dartfmt --line-length=120`..
https://github.com/flutter/cocoon/blob/master/test_utilities/bin/flutter_test_runner.sh#L26

flar:
> Cocoon has a test to enforce `dartfmt --line-length=120`..
> https://github.com/flutter/cocoon/blob/master/test_utilities/bin/flutter_test_runner.sh#L26

I don't believe this was failing the 120 character rule, the line that was reformatted was only 86 characters long.

Was there another change that happened other than line length?

Or perhaps there is a bug in the new dartfmt that ignored the requested 120 char length and went with its own 80 char length instead?

jason-simmons:
The Dart SDK team may have more info.  But I suspect that dartfmt changed its rules for formatting nested conditional expressions.

flar:
I just applied this diff manually and reran the `dartfmt` command (with my local version that hasn't been updated) and it rejected the diff. This means that this change to `progress_button.dart` must be accompanied by the roll to the `dartfmt` command.

flar:
Here is the Dart change that is now enforcing this:

https://github.com/dart-lang/dart_style/pull/945

flar:
I found that this formatting will pass the old formatter because I guess it wants to preserve comments:

```
+          onPressed: _busy // dartfmt will soon require this new formatting
+              ? null
+              : widget.onPressed != null // dartfmt will soon require this new formatting
+                  ? _handlePressed
+                  : null,
```

Oddly, aligning the comments made it complain because it wanted the comments to be right after the code they were commenting... :(

flar:
Suggestion from @jakemac53 

```
Note that instead of running dartfmt from the sdk,  you _could_ globally activate the package and run it like `pub global run dart_style:format .`

that would allow you to run a consistent version (if you activated a specific version)
```

flar:
Have we double-checked that it passes the new formatter with the comments? I only ran the old formatter on it.

jason-simmons:
Yes - this works with the latest dartfmt

PR #932, opened 2020-09-01
Make sure the task overlay appears in the visible window bounds

dnfield:
Addresses the first part of https://github.com/flutter/flutter/issues/65030

Eliminates a TODO related to that

Updates goldens.

I did not reuse the framework's positionDependentBox because that centers things like they're a tooltip - the logic here is a little simpler and stays a bit closer to current behavior.


Hixie (approved):
LGTM

app_flutter/lib/widgets/task_overlay.dart
```diff
@@ -15,6 +17,68 @@ import 'progress_button.dart';
...
+  }) {
+    assert(size != null);
+    assert(childSize != null);
+    assert(target != null);
+    const double margin = 10.0;
+    const double verticalOffset = TaskBox.cellSize * .9;
+
+    // VERTICAL DIRECTION
+    final bool fitsBelow = target.dy + verticalOffset + childSize.height <= size.height - margin;
+    // final bool fitsAbove = target.dy - verticalOffset - childSize.height >= margin;
```
Hixie, line 40: commented out code

dnfield: Deleted it

PR #934, opened 2020-09-02
[app_dart] Add dimensions and priority to ScheduleBuildRequest model

digiter:
In order to schedule devicelab builds based on capabilities, we need to add the dimensions field. Also add priority for potential future usage.

Test: I have tested end to end.

Bug: https://github.com/flutter/flutter/issues/65046

### DIFF
--- a/app_dart/lib/src/model/luci/buildbucket.dart
+++ b/app_dart/lib/src/model/luci/buildbucket.dart
@@ -398,10 +398,6 @@ class ScheduleBuildRequest extends JsonBody {
   ///
   /// A set of entries with the same key defines a new or replaces an existing
   /// dimension with the same key.
-  /// If the config does not allow overriding/adding the dimension, the request
-  /// will fail with InvalidArgument error code.
-  ///
-  /// After merging, dimensions with empty value will be excluded.
   ///
   /// A dimension expiration must be a multiple of 1min.
   final List<RequestedDimension> dimensions;
--- a/app_dart/lib/src/model/luci/buildbucket.dart
+++ b/app_dart/lib/src/model/luci/buildbucket.dart
@@ -398,8 +398,6 @@ class ScheduleBuildRequest extends JsonBody {
   ///
   /// A set of entries with the same key defines a new or replaces an existing
   /// dimension with the same key.
-  ///
-  /// A dimension expiration must be a multiple of 1min.
   final List<RequestedDimension> dimensions;
 
   // If not zero, overrides swarming task priority.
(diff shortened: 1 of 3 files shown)


app_dart/lib/src/model/luci/buildbucket.dart
```diff
@@ -392,6 +394,16 @@ class ScheduleBuildRequest extends JsonBody {
...
 
+  /// Overrides default dimensions defined by builder config or template build.
+  ///
+  /// A set of entries with the same key defines a new or replaces an existing
+  /// dimension with the same key.
+  final List<RequestedDimension> dimensions;
+
+  // If not zero, overrides swarming task priority.
+  // See also Build.infra.swarming.priority.
+  final int priority;
```
godofredoc, line 405: Should we have the default 30 here? Ideally we should not change this as changing priorities can potentially make all the builders perform worst than without changing the default priority.

app_dart/lib/src/model/luci/buildbucket.dart
```diff
@@ -650,3 +662,25 @@ enum Trinary {
   @JsonValue('UNSET')
   unset,
 }
+
+/// A requested dimension. Looks like StringPair, but also has an expiration.
+@JsonSerializable(includeIfNull: false)
+class RequestedDimension extends JsonBody {
```
godofredoc, line 668: Isn't this basically a map?

app_dart/test/model/buildbucket_test.dart
```diff
@@ -61,4 +61,25 @@ void main() {
...
+        project: 'flutter',
+        bucket: 'try',
+        builder: 'fake_builder',
+      ),
+      properties: <String, String>{
+        'git_url': 'https://github.com/flutter/flutter',
+        'git_ref': 'refs/pull/63834/head',
+      },
+      dimensions: <RequestedDimension>[RequestedDimension(key: 'a', value: 'b', expiration: '120s')],
+      priority: 100,
```
godofredoc, line 77: why expiration would be a dimension?

app_dart/lib/src/model/luci/buildbucket.dart
digiter: We can leave this field unset if there's no need to overwrite the default.

app_dart/lib/src/model/luci/buildbucket.dart
digiter: Kind of, this is defined in https://chromium.googlesource.com/infra/luci/luci-go/+/master/buildbucket/proto/common.proto.

app_dart/test/model/buildbucket_test.dart
digiter: See RequestedDimension in https://chromium.googlesource.com/infra/luci/luci-go/+/master/buildbucket/proto/common.proto

app_dart/lib/src/model/luci/buildbucket.dart
godofredoc: How does the string gets translated to the proto value?

godofredoc (approved):

PR #939, opened 2020-09-09
[build_dashboard] Fix negative times in task overlay

CaseyHillers:
## Description
- Fix negative times that occurred when tasks was enqueue or running
  - Instead calculate the time a task has been running or queuing if this case is encountered
- Clean up task overlay code 
  - Remove unused field of task.reason
  - Use task constants

## Tests
- Added tests for when tasks are enqueue and running

## Issues

Fixes https://github.com/flutter/flutter/issues/60031


dnfield (approved):
LGTM

PR #941, opened 2020-09-16
[app_dart] Fix Dart 2.10 issue and log commands run in test_utilities

CaseyHillers:
Tip of tree is failing due to Dart 2.10 code landing. There's a missing field in the HttpClientRequest fake.

https://github.com/flutter/flutter/issues/63269 requires updating some of app_flutter render object usage.

Additionally, I enabled on the bash scripts logging the commands run so it's easier to reproduce issues thrown in LUCI.

### DIFF
--- a/app_flutter/lib/widgets/lattice.dart
+++ b/app_flutter/lib/widgets/lattice.dart
@@ -277,18 +277,18 @@ class _LatticeBodyElement extends RenderObjectElement implements _LatticeDelegat
   }
 
   @override
-  void insertChildRenderObject(RenderObject child, _Coordinate slot) {
+  void insertRenderObjectChild(RenderObject child, _Coordinate slot) {
     renderObject.placeChild(slot, child);
   }
 
   @override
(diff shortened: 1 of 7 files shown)


app_flutter/lib/widgets/lattice.dart
```diff
@@ -545,8 +545,7 @@ class _RenderLatticeBody extends RenderBox {
     newChildParentData.coordinate = newCoordinate;
   }
 
-  void removeChild(RenderBox child) {
-    final _Coordinate coordinate = _childrenByRenderBox[child];
```
tvolkert, line 549: You probably don't need `_childrenByRenderBox` anymore.

tvolkert (approved):

app_flutter/lib/widgets/lattice.dart
CaseyHillers: Thanks for pointing this out! Removed.

PR #943, opened 2020-09-17
Do not automatically deflake flaky tasks

CaseyHillers:
Flaky tasks do not block the tree, but we still invest capacity to try and deflake them.

While prototyping the MILO redesign, I noticed this could be an issue with the iOS capacity.

Column of flaky ios tasks retried


keyonghan (approved):
LGTM

PR #944, opened 2020-09-17
[app_flutter] Catch google sign in platform exceptions

CaseyHillers:
Catch issues in the Google Sign in plugin so they do not crash the app's sign in functionality.

Re-enable silent sign in.

Update dependencies to ensure latest sign in plugin is used.

## Issues

https://github.com/flutter/flutter/issues/65912

### DIFF
--- a/app_flutter/lib/service/google_authentication.dart
+++ b/app_flutter/lib/service/google_authentication.dart
@@ -24,7 +24,7 @@ class GoogleSignInService extends ChangeNotifier {
 
     try {
       _googleSignIn.signInSilently();
-    } on PlatformException catch(error) {
+    } on PlatformException catch (error) {
       debugPrint(error.message);
     }
   }
@@ -45,12 +45,12 @@ class GoogleSignInService extends ChangeNotifier {
   final GoogleSignIn _googleSignIn;
 
   /// Whether or not the application has been signed in to.
-  /// 
+  ///
   /// If the plugin fails, default to unauthenticated.
   Future<bool> get isAuthenticated {
     try {
-       return _googleSignIn.isSignedIn();
-    } on PlatformException catch(error) {
+      return _googleSignIn.isSignedIn();
+    } on PlatformException catch (error) {
       debugPrint(error.message);
     }
     return Future<bool>.value(false);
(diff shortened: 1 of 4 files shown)


app_flutter/lib/service/google_authentication.dart
```diff
@@ -18,8 +22,11 @@ class GoogleSignInService extends ChangeNotifier {
       notifyListeners();
     });
 
-    // TODO(chillers): Re-enable when flakiness is fixed. https://github.com/flutter/flutter/issues/52338
-    // _googleSignIn.signInSilently();
+    try {
+      _googleSignIn.signInSilently();
+    } on PlatformException catch (error) {
+      debugPrint(error.message);
+    }
```
ditman, line 29: (You can manually verify this works by opening cocoon in an incognito window with cookies disabled, it should throw the idpiframe PlatformException here)

app_flutter/lib/service/google_authentication.dart
```diff
@@ -62,12 +78,20 @@ class GoogleSignInService extends ChangeNotifier {
...
   /// Initiate the Google Sign In process.
   Future<void> signIn() async {
-    user = await _googleSignIn.signIn();
-    notifyListeners();
+    try {
+      user = await _googleSignIn.signIn();
+      notifyListeners();
+    } on PlatformException catch (error) {
+      debugPrint(error.message);
+    }
```
ditman, line 86: This can be tested by closing the sign in popup (and waiting a few ms until the script picks it up)

ditman (commented):
Note: logging only `error.message` might not be enough, I'd recommend to log as well: `error.code`, which contains the actual error code from the JS SDK. `error.details` contains a link to the docs that explain what `code` means.

See [this](https://github.com/flutter/plugins/blob/master/packages/google_sign_in/google_sign_in_web/lib/google_sign_in_web.dart#L111-L114) and [this](https://github.com/flutter/plugins/blob/master/packages/google_sign_in/google_sign_in_web/lib/google_sign_in_web.dart#L136-L139).

CaseyHillers:
Thanks, added logging the error code too.

ditman (approved):
Let's see those logs!

PR #945, opened 2020-09-17
Fix flutter-gold for stale PRs

Piinks:
This patches a bug where the flutter-gold check is stuck on old PRs.
Gold no longer caches image results >20 days old, so we need to account for that in the check.
This could also be helpful in cleaning up old PRs that aren't active. :) 

Maybe the bot should just close them if they have not been update in so long? Not sure.

I also added `v1` to the endpoints as part of the new versioning Gold is doing. :)

Fixes https://github.com/flutter/flutter/issues/65979

### DIFF
--- a/app_dart/lib/src/datastore/cocoon_config.dart
+++ b/app_dart/lib/src/datastore/cocoon_config.dart
@@ -169,8 +169,8 @@ class Config {
       'this pull request.';
 
   String get flutterGoldStalePR => 'This pull request has not been updated in a '
-    'while. Please update this pull request to receive results from Gold, or '
-    'close it.';
+      'while. Please update this pull request to receive results from Gold, or '
+      'close it.';
 
   String get flutterGoldDraftChange => 'This pull request has been changed to a '
       'draft. The currently pending flutter-gold status will not be able '
--- a/app_dart/lib/src/datastore/cocoon_config.dart
+++ b/app_dart/lib/src/datastore/cocoon_config.dart
@@ -168,6 +168,10 @@ class Config {
   String get flutterGoldChanges => 'Image changes have been found for '
       'this pull request.';
 
+  String get flutterGoldStalePR => 'This pull request has not been updated in a '
+    'while. Please update this pull request to receive results from Gold, or '
+    'close it.';
+
   String get flutterGoldDraftChange => 'This pull request has been changed to a '
       'draft. The currently pending flutter-gold status will not be able '
       'to resolve until a new commit is pushed or the change is marked ready for '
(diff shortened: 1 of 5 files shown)


Piinks:
FYI @kjlubick :)

app_dart/lib/src/datastore/cocoon_config.dart
```diff
@@ -168,6 +168,10 @@ class Config {
   String get flutterGoldChanges => 'Image changes have been found for '
       'this pull request.';
 
+  String get flutterGoldStalePR => 'This pull request has not been updated in a '
+      'while. Please update this pull request to receive results from Gold, or '
+      'close it.';
+
```
Piinks, line 174: Not sure of the wording here. Suggestions welcome. :)

godofredoc: Sounds good to me but @Hixie may have comments.

app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
```diff
@@ -53,6 +53,20 @@ class PushGoldStatusToGithub extends ApiRequestHandler<Body> {
     final List<GithubGoldStatusUpdate> statusUpdates = <GithubGoldStatusUpdate>[];
     log.debug('Beginning Gold checks...');
     await for (PullRequest pr in gitHubClient.pullRequests.list(slug)) {
+      // Check when this PR was last updated. Gold does not keep results after
+      // >20 days. If a PR has gone stale, we should draw attention to it to be
+      // updated or closed.
+      final DateTime updatedAt = pr.updatedAt;
```
godofredoc, line 59: Do we need to take into account timezones and/or UTC in this logic?

app_dart/test/request_handlers/push_gold_status_to_github_test.dart
```diff
@@ -143,12 +144,13 @@ void main() {
...
 
-      PullRequest newPullRequest(int number, String sha, String baseRef, {bool draft = false}) {
+      PullRequest newPullRequest(int number, String sha, String baseRef, {bool draft = false, DateTime updated}) {
         return PullRequest()
           ..number = 123
           ..head = (PullRequestHead()..sha = 'abc')
           ..base = (PullRequestHead()..ref = baseRef)
-          ..draft = draft;
+          ..draft = draft
+          ..updatedAt = updated ?? DateTime.now();
```
godofredoc, line 153: Same comment as above UTC/timezone?

app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
kjlubick: FYI: The Gold API returns timestamps in UTC (with a 'Z' suffix).

app_dart/test/request_handlers/push_gold_status_to_github_test.dart
Piinks: I changed the check to always convert the DateTimes to UTC for the comparison. Good catch, thanks!

godofredoc (approved):

PR #948, opened 2020-09-21
add filtering options to build dashboard

flar:
I've spent many an annoying minute or two trying to find a particular task on the dashboard so I thought I would add some filtering options. I populated the filter with a few options that I thought might be useful, but am open to suggestions on things that we might want to be filtering on in practice.

Fixes https://github.com/flutter/flutter/issues/52370


app_flutter/lib/build_dashboard_page.dart
```diff
@@ -48,6 +97,10 @@ class BuildDashboardPage extends StatelessWidget {
           title: statusTable[_buildState.isTreeBuilding],
           backgroundColor: colorTable[_buildState.isTreeBuilding],
           actions: <Widget>[
+            FlatButton(
+              child: const Icon(Icons.sort),
```
CaseyHillers, line 101: nit: Icons add vagueness to the dashboard, and it's preferred to use text (e.g. 'Filter')

app_flutter/lib/widgets/task_grid.dart
```diff
@@ -20,7 +20,9 @@ import 'task_overlay.dart';
 ///
 /// If there's no data for [TaskGrid], it shows [CircularProgressIndicator].
 class TaskGridContainer extends StatelessWidget {
-  const TaskGridContainer({Key key}) : super(key: key);
+  const TaskGridContainer({Key key, this.filterNotifier}) : super(key: key);
+
+  final ValueNotifier<TaskGridFilter> filterNotifier;
```
CaseyHillers, line 25: dartdoc

app_flutter/lib/widgets/task_grid.dart
```diff
@@ -47,12 +49,367 @@ class TaskGridContainer extends StatelessWidget {
...
   }
 }
 
+@immutable
+class TaskGridFilter {
+  TaskGridFilter({
+    RegExp taskFilter,
+    RegExp authorFilter,
+    RegExp messageFilter,
+    this.showAndroid,
```
CaseyHillers, line 65: With the ongoing migration to LUCI, this should be clarified to devicelabAndroid along with iOS and windows.

app_flutter/lib/widgets/task_grid.dart
```diff
@@ -47,12 +49,367 @@ class TaskGridContainer extends StatelessWidget {
...
+      ],
+    );
+  }
+
+  TableRow _makeTextFilterRow(String label, TextEditingController controller, void onChanged(RegExp newValue)) {
+    return _makeRow(label,
+      TextField(
+        controller: controller,
+        decoration: const InputDecoration(
+          hintText: '(regular expression)',
```
CaseyHillers, line 360: This should be an example regular expression

app_flutter/lib/widgets/task_grid.dart
```diff
@@ -47,12 +49,367 @@ class TaskGridContainer extends StatelessWidget {
...
+  TaskGridFilter get filter => widget.filterNotifier.value;
+  TextEditingController taskController;
+  TextEditingController authorController;
+  TextEditingController messageController;
+
+  void _setFilter(TaskGridFilter newFilter) {
+    widget.filterNotifier.value = newFilter;
+  }
+
+  void _newTaskFilter(RegExp newRegExp) {
```
CaseyHillers, line 306: This seems verbose. Can these new functions be condensed into a single function?

app_flutter/lib/widgets/task_grid.dart
```diff
@@ -72,6 +430,8 @@ class TaskGrid extends StatefulWidget {
   /// Reference to the build state to perform actions on [TaskMatrix], like rerunning tasks.
   final BuildState buildState;
 
+  final ValueNotifier<TaskGridFilter> filterNotifier;
```
CaseyHillers, line 433: dartdoc?

app_flutter/lib/widgets/task_grid.dart
```diff
@@ -81,6 +441,23 @@ class _TaskGridState extends State<TaskGrid> {
   // lattice matrix each time the task grid has to update, regardless of whether
   // we've received new data or not.
 
+  TaskGridFilter filter;
```
CaseyHillers, line 444: dartdoc?

app_flutter/lib/widgets/task_grid.dart
```diff
@@ -47,12 +49,367 @@ class TaskGridContainer extends StatelessWidget {
...
+        showAndroid == other.showAndroid &&
+        showIos == other.showIos &&
+        showWindows == other.showWindows &&
+        showCirrus == other.showCirrus &&
+        showLuci == other.showLuci
+    ;
+  }
+}
+
+class TaskGridFilterWidget extends StatefulWidget {
```
CaseyHillers, line 259: The filter logic should be moved into `logic/task_grid_filter.dart`

CaseyHillers (commented):
Thank you for the PR, this is a great improvement! Just some comments to align it with the reset of the codebase. Missing dart docs and tests.

app_flutter/lib/build_dashboard_page.dart
flar: I guess I prefer them because they don't require localization, but I guess our entire build environment requires English.

In the future, I can see adding alternate sorting capabilities as well as filtering. Would "settings" be a better name? (icon?)

app_flutter/lib/widgets/task_grid.dart
flar: I was thinking about that. I'll see what I can do.

I was also thinking about making a generic filter object with arbitrary filters so we could do the same thing with other dashboards.

(FWIW, I have an alternate implementation of the benchmark dashboard in the [benchmark utilities repo](https://github.com/flar/flutter_benchmark_utils/blob/master/README.md) that also has filtering...)

app_flutter/lib/build_dashboard_page.dart
CaseyHillers: SGTM. A settings icon is pretty universal and would be okay (similar to the hamburger menu icon). We could also move the branch dropdown selection into that settings area.

app_flutter/lib/widgets/task_grid.dart
CaseyHillers: It's great to see the benchmark utilities repo! If you want to have it replace the AngularDart dashboard let me know. It's trivial to setup a new page that has all the data needed, I just never got around to building graph widgets that could show the data.

app_flutter/lib/build_dashboard_page.dart
flar: I switched to the settings icon and moved the branch selector in there as well.

app_flutter/lib/widgets/task_grid.dart
flar: I solved this one by re-using the map constructor as a copyWithMap() method so I don't need a per-attribute updater.

app_flutter/lib/widgets/task_grid.dart
flar: I guess I don't understand the configuration of stages and how it relates to the migration. I have left this alone for now while I worked on the other suggestions.

I'm also a little wary of having long property names.

app_flutter/lib/widgets/task_grid.dart
flar: I thought the hints were descriptive rather than exemplary...?

app_flutter/lib/widgets/task_grid.dart
flar: I've broken this out into a FilterPropertySheet widget that can layout and edit arbitrary filtering properties and the TaskGridFilter now implements that "interface". A lot of the redundancy here went away as well at the expense of some generic field get/set functions. I might revisit that in the future to see if I can get it simplified some more.

app_flutter/lib/widgets/task_grid.dart
flar: Done, along with creating filter_property_sheet.dart in the widgets directory as well.

app_flutter/lib/widgets/task_grid.dart
flar: Made private as it is just a cached copy for triggering state changes.

flar:
One error I was seeing that I still need to resolve - I was getting exceptions when I was focused to one of the TextFields and I hit the TAB key. I'm guessing I need to manage the focus traversal somehow, but my focus mojo is weak.

app_flutter/lib/widgets/task_grid.dart
CaseyHillers: My concern is that a user won't know that this dashboard is built in Dart, and it's expecting an ECMAScript formatted regex. So when they try their python regex it fails. Additionally, if they know regex they might not know what advanced options there are.

Can we have the helper/placeholder text be an example and the hint text be something like `ECMAScript/Javascript Regular Expression`?

app_flutter/lib/widgets/task_grid.dart
CaseyHillers: Stages are a source of technical debt when there were several different infras powering Flutter. Once everything is moved to LUCI we'll be able to remove the staging logic.

Piinks:
cc/ @gspencergoog re: the focus issue you mentioned @flar
Here too is the related issue this will fix: https://github.com/flutter/flutter/issues/52370 Thank you for this!

gspencergoog:
@flar What were the exceptions?

flar:
@gspencergoog here is what is displayed on the terminal. I have a feeling this is because I am using a number of independently created TextEditingControllers that are not in a Material environment that maintains focus, but these things are mostly magic to me...

```
══╡ EXCEPTION CAUGHT BY SERVICES LIBRARY ╞══════════════════════════════════════════════════════════
The following assertion was thrown during a platform message callback:
Assertion failed:
file:///Users/jimgraham/Development/flutter/packages/flutter/lib/src/rendering/viewport.dart:103:15
rect != null
is not true

When the exception was thrown, this was the stack:
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/errors.dart 212:49  throw_
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/errors.dart 24:3    assertFailed
packages/flutter/src/rendering/viewport.dart 103:19                           new
packages/app_flutter/widgets/lattice.dart 114:12                              getOffsetToReveal
packages/flutter/src/widgets/scroll_position.dart 650:26                      ensureVisible
packages/flutter/src/widgets/scrollable.dart 314:29                           ensureVisible
packages/flutter/src/widgets/focus_traversal.dart 45:14                       _focusAndEnsureVisible
packages/flutter/src/widgets/focus_traversal.dart 398:9                       [_moveFocus]
packages/flutter/src/widgets/focus_traversal.dart 227:39                      next
packages/flutter/src/widgets/focus_manager.dart 1125:46                       nextFocus
packages/flutter/src/widgets/focus_traversal.dart 1844:18                     invoke
packages/flutter/src/widgets/actions.dart 396:22                              invokeAction
packages/flutter/src/widgets/actions.dart 639:65                              invoke
packages/flutter/src/widgets/shortcuts.dart 344:15                            handleKeypress
packages/flutter/src/widgets/shortcuts.dart 486:20                            [_handleOnKey]
packages/flutter/src/widgets/focus_manager.dart 1625:43                       [_handleRawKeyEvent]
packages/flutter/src/services/raw_keyboard.dart 616:53                        _handleKeyEvent
dart-sdk/lib/_internal/js_dev_runtime/patch/async_patch.dart 84:54            runBody
dart-sdk/lib/_internal/js_dev_runtime/patch/async_patch.dart 123:12           _async
packages/flutter/src/services/raw_keyboard.dart 578:34                        [_handleKeyEvent]
packages/flutter/src/services/platform_channel.dart 74:49                     <fn>
dart-sdk/lib/_internal/js_dev_runtime/patch/async_patch.dart 84:54            runBody
dart-sdk/lib/_internal/js_dev_runtime/patch/async_patch.dart 123:12           _async
packages/flutter/src/services/platform_channel.dart 73:58                     <fn>
packages/flutter/src/services/binding.dart 283:33                             handlePlatformMessage
dart-sdk/lib/_internal/js_dev_runtime/patch/async_patch.dart 84:54            runBody
dart-sdk/lib/_internal/js_dev_runtime/patch/async_patch.dart 123:12           _async
packages/flutter/src/services/binding.dart 274:37                             handlePlatformMessage
lib/_engine/engine/window.dart 789:13                                         _invoke3
lib/_engine/engine/window.dart 444:5                                          invokeOnPlatformMessage
lib/_engine/engine/keyboard.dart 127:12                                       [_handleHtmlEvent]
lib/_engine/engine/keyboard.dart 39:7                                         <fn>
════════════════════════════════════════════════════════════════════════════════════════════════════
```

app_flutter/lib/widgets/task_grid.dart
flar: I believe this is resolved in a recent commit.

flar:
I believe that I fixed the focus problems which were mainly due to the lattice/_FakeViewport classes creating bogus reveal objects. I also automatically grab the focus into the first editable field and capture the focus traversal within the settings dialog before recursing to the surrounding lattice.

flar:
The only issues remaining AFAICT:

- Should we deal with runaway commit loading when the filter is so narrow that only a couple of commits match (eg. when they search on a commit hash)?

- What kind of testing should I add. I had to disable the branch test because the branch dropdown was moved. Should I create new tests that find it in the dialog?

flar:
I had commented out the test for the branch dropdown, but I was able to put it back with new code to open the settings property sheet so it could be found properly.

app_flutter/lib/build_dashboard_page.dart
```diff
@@ -48,37 +139,21 @@ class BuildDashboardPage extends StatelessWidget {
...
-              },
-              items: _buildState.branches.map<DropdownMenuItem<String>>(
-                (String value) {
-                  return DropdownMenuItem<String>(
-                    value: value,
-                    child: Text(value),
-                  );
-                },
-              ).toList(),
+            FlatButton(
```
CaseyHillers, line 142: Is there a difference between this and `IconButton`?

app_flutter/lib/logic/task_grid_filter.dart
```diff
@@ -0,0 +1,232 @@
...
+
+import '../logic/qualified_task.dart';
+import '../widgets/filter_property_sheet.dart';
+
+/// A filter object for controlling which entries are visible in the Build dashboard grid
+/// of tasks. This filter object can filter on a number of properties of the tasks including
+/// the name of the task, information on the commit that generated the tasks, and platform
+/// stage type toggles.
+class TaskGridFilter extends FilterPropertySource {
+  TaskGridFilter();
```
CaseyHillers, line 19: nit: Is there a use case where a Map is not passed to construct this?

app_flutter/lib/logic/task_grid_filter.dart
```diff
@@ -0,0 +1,232 @@
...
+    for (final ValueFilterProperty<dynamic> property in _allProperties.values) {
+      property.reset();
+    }
+  }
+
+  /// Modifies this filter based on the values in the map. If the map is null, no changes are
+  /// made. All keys in the [valueMap] must match one of the field names of the filter properties.
+  void applyMap(Map<String, String> valueMap) {
+    if (valueMap != null) {
+      for (final MapEntry<String, String> mapEntry in valueMap.entries) {
```
CaseyHillers, line 38: `valueMap?.entries`?

app_flutter/lib/logic/task_grid_filter.dart
```diff
@@ -0,0 +1,232 @@
...
+        ..[_authorProperty.fieldName] = _authorProperty
+        ..[_messageProperty.fieldName] = _messageProperty
+        ..[_hashProperty.fieldName] = _hashProperty
+        ..[_androidProperty.fieldName] = _androidProperty
+        ..[_iosProperty.fieldName] = _iosProperty
+        ..[_windowsProperty.fieldName] = _windowsProperty
+        ..[_cirrusProperty.fieldName] = _cirrusProperty
+        ..[_luciProperty.fieldName] = _luciProperty;
+
+  // TODO(flar): Decide whether or not to expose these fields or do everything through the
```
CaseyHillers, line 69: Removed commented out code?

app_flutter/lib/logic/task_grid_filter.dart
```diff
@@ -0,0 +1,232 @@
...
+  final RegExpFilterProperty _messageProperty =
+      RegExpFilterProperty(fieldName: 'messageFilter', label: 'Commit Message');
+  final RegExpFilterProperty _hashProperty = RegExpFilterProperty(fieldName: 'hashFilter', label: 'Commit Hash');
+  final BoolFilterProperty _androidProperty = BoolFilterProperty(fieldName: 'showAndroid', label: 'Android');
+  final BoolFilterProperty _iosProperty = BoolFilterProperty(fieldName: 'showIos', label: 'iOS');
+  final BoolFilterProperty _windowsProperty = BoolFilterProperty(fieldName: 'showWindows', label: 'Windows');
+  final BoolFilterProperty _cirrusProperty = BoolFilterProperty(fieldName: 'showCirrus', label: 'Cirrus');
+  final BoolFilterProperty _luciProperty = BoolFilterProperty(fieldName: 'showLuci', label: 'Luci');
+
+  // [_allProperties] is a LinkedHashMap so we can trust its iteration order
```
CaseyHillers, line 55: Should private variables be referenced in dart docs?

app_flutter/lib/widgets/filter_property_sheet.dart
```diff
@@ -0,0 +1,330 @@
+// Copyright 2020 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter/foundation.dart';
+import 'package:flutter/material.dart';
+
+/// Bridge between the [FilterPropertySheet] and a filter object that has a number of
```
CaseyHillers, line 8: This is great documentation!

app_flutter/lib/widgets/filter_property_sheet.dart
```diff
@@ -0,0 +1,330 @@
...
+          defaultVerticalAlignment: TableCellVerticalAlignment.middle,
+          columnWidths: const <int, TableColumnWidth>{
+            0: IntrinsicColumnWidth(),
+            1: FixedColumnWidth(300.0),
+          },
+          children: widget.propertySource.sheetLayout.map(_makeTableRow).toList(),
+        ),
+      ],
+    );
+  }
```
CaseyHillers, line 329: I would like to see a test that toggles one of the filters, and checks that it has been applied. It doesn't need to be hooked up to a TaskGrid, but it should verify that this widget can update the filter datastructure.

app_flutter/test/build_dashboard_page_test.dart
```diff
@@ -54,6 +54,24 @@ void main() {
     buildState.dispose();
   });
 
+  testWidgets('shows settings button', (WidgetTester tester) async {
```
CaseyHillers, line 57: This only tests the default filter, it should have a test when given a filter that we expect to change the dashboard.

CaseyHillers (commented):
Looks great! Just some comments on clean up and tests to add to ensure this functionality doesn't get broken 😄

app_flutter/lib/build_dashboard_page.dart
flar: I wasn't aware of IconButton. Fixed.

app_flutter/lib/logic/task_grid_filter.dart
flar: If you want a default/uncustomized filter.

I used to have a path that made use of that, but it isn't around any longer. I suppose you could use the map constructor with a null or empty map, but I figured something like this class should always support a default constructor as a general rule of thumb.

app_flutter/lib/logic/task_grid_filter.dart
flar: You can't iterate on null. Is there an `in?` operator?

app_flutter/lib/logic/task_grid_filter.dart
flar: There is no documentation there. That is a code comment.

app_flutter/lib/logic/task_grid_filter.dart
flar: There is an implicit question there, but I take it you side on the "not" part of whether these accessors make sense?

app_flutter/test/build_dashboard_page_test.dart
flar: This doesn't even test the filter at all. It just verifies that the icon button is in the tree somewhere so that the property sheet can be brought up.

I'll take this and the previous comment to mean "add some tests that verify that filter properties are affected by the property sheet". It might also make sense to add some tests to test the filter itself independent of display.

flar: I added a golden test to verify the look with the default filter. Since the test shows all boxes where text characters might be, I'm not sure it has the ability to show the difference with non-default values.

I also added a whole suite of tests to verify the properties and actions of the `TaskGridFilter` object.

app_flutter/lib/widgets/filter_property_sheet.dart
flar: I added a whole file full of tests for `TaskGridFilter`. I didn't apply it to the grid, but I did isolated applications against mock `CommitStatus` and `QualifiedTask` objects.

app_flutter/lib/widgets/filter_property_sheet.dart
flar: Thanks!

app_flutter/lib/logic/task_grid_filter.dart
flar: I ended up putting them in (uncommenting this section) as they helped a lot for testing.

app_flutter/lib/logic/task_grid_filter.dart
flar: This constructor is also used frequently in the tests.

app_flutter/test/build_dashboard_page_test.dart
flar: I just added another test that builds the TaskGridContainer with a variety of filters and checks the size of the generated grid. This should help determine that the filters are being applied correctly by the widgets. Note that the size of the grid depends on the reproducibility of the testing fake grid generator, but that has to be consistent for a lot of tests to be valid.

CaseyHillers: Sorry, GitHub comment was in a weird place, but looks like you got the message!

app_flutter/lib/logic/task_grid_filter.dart
CaseyHillers: Ah I didn't know. I've been spending too much in python. I was naive in thinking that it would just work.

app_flutter/lib/logic/task_grid_filter.dart
CaseyHillers: SGTM

CaseyHillers (approved):
LGTM! Thanks for working on this, this is a great improvement! 🥳

PR #949, opened 2020-09-21
[build-dashboard] Revert get-status refresh rate back to original timing

CaseyHillers:
This reverts back to the AngularDart dashboard's timing.

This was set to 10 seconds due to most users being on networks without data caps. With everyone working from home, the app is downloading ~200 KB every 10 seconds.

This should reduce the amount of repainting happening, and reduce the flickering in the GitHub avatars.


keyonghan (approved):
LGTM.
You may want to manually trigger a cocoon build to pick this up immediately.

PR #953, opened 2020-09-21
[build-dashboard] Pass width and height to web image

CaseyHillers:
Fixes a small issue where images were being initialized with their default constraints. This caused issues in the commit overlays where the image would push everything off.


keyonghan (approved):
LGTM

CaseyHillers:
Thanks for the quick review Keyong!

PR #954, opened 2020-09-22
add dev branch to branches.txt

christopherfujino:
I think we need a cherry pick on the dev branch to fix a devicelab regression, so this will let us run devicelab tests on it.

### DIFF
--- a/app_dart/dev/branches.txt
+++ b/app_dart/dev/branches.txt
@@ -1,4 +1,5 @@
 flutter-1.20-candidate.7
 flutter-1.21-candidate.9
 flutter-1.22-candidate.12
+flutter-1.23-candidate.4
 master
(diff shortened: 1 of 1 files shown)


app_dart/dev/branches.txt
```diff
@@ -1,4 +1,5 @@
 flutter-1.20-candidate.7
 flutter-1.21-candidate.9
 flutter-1.22-candidate.12
+flutter-1.23-candidate.4
```
godofredoc, line 4: Do we also need to remove branches?

christopherfujino: We can remove 1.21, lemme do that now

christopherfujino: done, ptal

godofredoc (approved):

PR #958, opened 2020-09-23
Add verbose flags to flutter config and doctor

jmagman:
Investigation for flaky `config` call. https://github.com/flutter/flutter/issues/65413
Add `--verbose` to doctor for good measure (will let us see tethered device OS version).


christopherfujino (approved):
LGTM

PR #959, opened 2020-09-23
[build_dashboard] Update sorting to show flaky tasks after failed tasks

CaseyHillers:
- Normalize the weights to prevent the first few rows skewing the weights
- Normalize flakiness weight to be within [0.2, 0.3] so it consistently shows after failures

## Issues

Fixes https://github.com/flutter/flutter/issues/62084

## Tests

It looks like the test coverage for the algorithm was moved to be covered by the goldens, let me know if we want to cover this with unit tests


app_flutter/lib/widgets/task_grid.dart
```diff
@@ -148,15 +153,21 @@ class _TaskGridState extends State<TaskGrid> {
           final QualifiedTask qualifiedTask = QualifiedTask.fromTask(task);
           if (commitCount <= 25) {
             double score = 0.0;
+            // Reruns take up extra infra capacity and should be prioritized.
             if (task.attempts > 1) {
-              score += 1.0;
+              score += 0.3;
```
keyonghan, line 158: Do we want to distinguish success and failure for the case of attempts>1, like `failure` with a slightly higher score?

CaseyHillers: That's how it works as it also adds the weights for those statuses 😅 

Do you think this would be more readable if I generated a bunch of different task categories and hard coded their weights instead of having these modifiers? In this case, `failure-rerun` and `success-rerun`?

Instead of all these extra checks, the weighting algorithm would just be `score = _statusScores[status];`. It would create a much larger setup to define all these, but the weight would be clearly defined.

keyonghan: Ah.. sorry missed the status part. I think this is fine. A doc may be good to explicitly list the order of different cases.

keyonghan (approved):
LGTM with nit.

app_flutter/lib/widgets/task_grid.dart
jmagman: > Do you think this would be more readable if I generated a bunch of different task categories and hard coded their weights instead of having these modifiers? In this case, `failure-rerun` and `success-rerun`?

Yes, that would make this much more readable.  Right now I'm not really sure what the magic numbers are (why 0.3 here?)

CaseyHillers: Added!

PR #960, opened 2020-09-24
Fix scope for stale PR alert

Piinks:
This changes the scope of PRs that the bot will alert for stale Gold results.
It also buffs up the message a bit to be clearer and more helpful. 

Fixes https://github.com/flutter/flutter/issues/66478


jmagman (approved):
LGTM, thank you!

PR #962, opened 2020-10-01
Save the one-off scripts for extracting cirrus metrics

digiter:
We will use these metrics to compare the performance between Cirrus and LUCI.

Bug: https://github.com/flutter/flutter/issues/64844


keyonghan (commented):
Thanks for uploading these codes. A README wiki page would be helpful.

oneoff/cirrus_stats/README.md
```diff
@@ -0,0 +1,33 @@
...
+* The size of local db: 408 MB.
+* The number of build records in BigQuery: 39,574
+* The number of task records in BigQuery: 1,036,471
+
+## Usage
+
+Here's a brief example of how to run these scripts.
+
+```
+go test -v ./model/
```
keyonghan, line 28: Did a quick check. It popped up below message, is this expected?

```
fork/exec /var/folders/yj/8c2z2b4x5_98ln143ynmc9_h00nqb9/T/go-build134062014/b001/exe/extract: operation not permitted

Santa

This application has been blocked from executing
because its trustworthiness cannot be determined.
```

digiter: It's a google internal security ware, see go/santa.

keyonghan (approved):
Trying all codes need to be unblocked from santa, which takes up to a day to get exception. I would assume these codes work. LGTM.

oneoff/cirrus_stats/README.md
```diff
@@ -0,0 +1,33 @@
...
+go test -v ./graphql/
+go run extract.go
```
keyonghan, line 30: Running `extract.go` hit
```
go run extract.go
2020/10/01 12:09:56 Error Creating Dir: "./tmp/badger_db": mkdir ./tmp/badger_db: no such file or directory
exit status 1
```
I tried creating a dir named `badger_db`, and also a file, both do not help..

Anything else do I miss here?

digiter: You will need to create `tmp` dir first.

PR #963, opened 2020-10-03
Fix status update condition in refresh-chromebot-status API

keyonghan:
There is a bug in existing `refresh-chromebot-status` API: the logic compares with datastore status with the oldest LUCI build's status. This will cause confusion when re-run is triggered.

This PR updates the logic to use latest build's status to do comparison and then update the datastore and dashboard.


jonahwilliams (approved):
LGTM

PR #971, opened 2020-10-15
Add verbose logging to test utilities

CaseyHillers:
This will make it easier to view logs from infra on cocoon. For example, https://github.com/flutter/flutter/issues/67681 has logs that get collated into one line which ends up being hard to parse.

### DIFF
--- a/test_utilities/bin/dart_test_runner.sh
+++ b/test_utilities/bin/dart_test_runner.sh
@@ -30,7 +30,7 @@ pub global run tuneup check
 # Only try tests if test folder exist.
 if [ -d 'test' ]; then
   echo "############# tests ###########"
-  pub run test --test-randomize-ordering-seed=random
+  pub run test --test-randomize-ordering-seed=random --reporter expanded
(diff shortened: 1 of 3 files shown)


keyonghan:
Will this verbose change temporally for debugging, or long-term? Will the verbose logging be saved to datastore/stackdriver as well? If debugging that specific test, a local trial is preferred.

CaseyHillers:
> Will this verbose change temporally for debugging, or long-term?
I think this is a good long term change. Currently, we can't act much on flake issues as there's no information in the logs other than the failing test. This reduces the work needed to jump in and start fixing some of Cocoon's flakes.

Ideally, Flutter will support the expanded flag which will reduce the noise (https://github.com/flutter/flutter/issues/68648)

> Will the verbose logging be saved to datastore/stackdriver as well?

No, this is just for Cocoon CI which isn't stored in Datastore or Stackdriver.

keyonghan (approved):
LGTM

PR #975, opened 2020-10-21
Selectable raw health details and agent name

jmagman:
Make the agent name selectable so it can be copied:
Screen Shot 2020-10-20 at 5 23 08 PM

Also raw health details so the IP can be easily copied.
Screen Shot 2020-10-20 at 5 27 21 PM

See also #662


CaseyHillers (approved):
LGTM, thanks for the improvement!

PR #979, opened 2020-10-27
[app_flutter] Remove authorized agent UI

CaseyHillers:
This tends to be accidentally clicked by users. With the migration to LUCI, we'll remove this to remove a source of issues with the cocoon agents.

Most of this is removing authorize agent code. There is one change where I needed to update the snackbar duration to be the correct variable.


keyonghan (approved):
LGTM. Thanks for the quick fix.

PR #981, opened 2020-10-28
[app_dart] Task statuses can be updated without task keys

CaseyHillers:
Migrating the DeviceLab to LUCI requires updating Cocoon to work with what LUCI knows. LUCI runs will not know the task key they are running. Instead, they know the git sha, git branch, and task name. This updates the request handler to process those to find corresponding task.

# Issue

# Tests

Added a unit test for this use case

### DIFF
--- a/app_dart/lib/src/request_handlers/update_task_status.dart
+++ b/app_dart/lib/src/request_handlers/update_task_status.dart
@@ -31,10 +31,13 @@ class UpdateTaskStatus extends ApiRequestHandler<UpdateTaskStatusResponse> {
 
   final DatastoreServiceProvider datastoreProvider;
 
-  static const String taskKeyParam = 'TaskKey';
+  static const String gitBranchParam = 'Branch';
+  static const String gitShaParam = 'CommitSha';
   static const String newStatusParam = 'NewStatus';
   static const String resultsParam = 'ResultData';
   static const String scoreKeysParam = 'BenchmarkScoreKeys';
+  static const String taskKeyParam = 'TaskKey';
+  static const String taskNameParam = 'TaskName';
 
   /// const variables for [BigQuery] operations
   static const String projectId = 'flutter-dashboard';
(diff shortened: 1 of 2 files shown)


app_dart/test/request_handlers/update_task_status_test.dart
```diff
@@ -120,5 +120,31 @@ void main() {
       expect(task.status, 'Failed');
       expect(task.attempts, 1);
     });
+
+    test('non-task key requests can update tasks', () async {
```
keyonghan, line 124: one more test for cases where no task is found?

app_dart/lib/src/request_handlers/update_task_status.dart
```diff
@@ -21,6 +21,20 @@ import '../request_handling/body.dart';
 import '../request_handling/exceptions.dart';
 import '../service/datastore.dart';
 
+/// Endpoint for task runners to update Cocoon with test run information.
+///
+/// This handlers requires (1) task identifier and (2) task status information.
```
keyonghan, line 26: nit: handler

app_dart/lib/src/request_handlers/update_task_status.dart
```diff
@@ -111,6 +117,58 @@ class UpdateTaskStatus extends ApiRequestHandler<UpdateTaskStatusResponse> {
...
+
+  /// Retrieve [Task] from [DatastoreService] when given [gitShaParam], [gitBranchParam], and [taskNameParam].
+  ///
+  /// This is used when the DeviceLab test runner is uploading results to Cocoon for runs on LUCI.
+  /// LUCI does not know the [Key] assigned to task when scheduling the build, but Cocoon can
+  /// lookup the task based on these key values.
+  ///
+  /// To lookup the value, we construct the ancestor key, which corresponds to the [Commit].
+  /// Then we query the tasks with that ancestor key and filter for the one that has the corresponding name.
+  Future<Task> _getTaskFromNamedParams(DatastoreService datastore) async {
```
keyonghan, line 152: add checkRequiredParameters for [gitShaParam], [gitBranchParam], [taskNameParam] when the task is from luci.

app_dart/test/request_handlers/update_task_status_test.dart
CaseyHillers: Good catch, added.

app_dart/lib/src/request_handlers/update_task_status.dart
CaseyHillers: Added check for this when from luci, otherwise verify taskKeyParam was passed.

keyonghan (approved):
LGTM

PR #983, opened 2020-11-03
[app_dart] Add swarming authentication provider

CaseyHillers:
Add a `SwarmingAuthenticationProvider` that is a child of the existing `AuthenticationProvider`. This provider only authenticates for Cocoon agents and LUCI swarming bot service accounts.

This ensures a given `RequestHandler` can only be updated through these methods (enforcing an ACL). For now, I added this to the `update-task-status`, but we should do a greater security review to audit endpoints that should only be used by automated systems.

Added LUCI swarming authentication which reuses the existing Google authentication.

# Issues

# Tests

- Added tests that ensure this child class does not grant the same permissions as the parent
- Added tests for LUCI auth being passed to id token

### DIFF
--- a/app_dart/lib/src/request_handling/authentication.dart
+++ b/app_dart/lib/src/request_handling/authentication.dart
@@ -22,7 +22,7 @@ import 'exceptions.dart';
 
 /// Class capable of authenticating [HttpRequest]s.
 ///
-/// There are four types of authentication this class supports:
+/// There are three types of authentication this class supports:
 ///
 ///  1. If the request has the `'Agent-ID'` HTTP header set to the ID of the
 ///     Cocoon agent making the request and the `'Agent-Auth-Token'` HTTP
@@ -52,9 +52,6 @@ import 'exceptions.dart';
 ///
 ///     User accounts are only authorized if the user is either a "@google.com"
 ///     account or is an [AllowedAccount] in Cocoon's Datastore.
-/// 
-///  4. If the request has the `Service-Account-Token` HTTP header, the token
-///     will be authenticated against LUCI.
 ///
 /// If none of the above authentication methods yield an authenticated
 /// request, then the request is unauthenticated, and any call to
@@ -63,6 +60,7 @@ import 'exceptions.dart';
 /// See also:
 ///
 ///  * <https://cloud.google.com/appengine/docs/standard/python/reference/request-response-headers>
+// TODO(chillers): Remove (1) when DeviceLab has migrated to LUCI, https://github.com/flutter/flutter/projects/151#card-47536851
 @immutable
 class AuthenticationProvider {
   const AuthenticationProvider(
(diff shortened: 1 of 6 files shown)


keyonghan:
why not expanding existing `AuthenticationProvider`, but creating a separate `SwarmingAuthenticationProvider`?

CaseyHillers:
> why not expanding existing `AuthenticationProvider`, but creating a separate `SwarmingAuthenticationProvider`?

1. This improves the security of `update-task-status` so only automated systems can push changes. Right now, any dashboard user can push changes.
2. `AuthenticationProvider` has 4 separate logic branches that are hard to follow to ensure authentication. Adding a 5th is somewhat trivial, but it adds more complexity. It gets hard to ensure good test coverage of this provider when it has so many different paths.

app_dart/lib/src/request_handling/authentication.dart
```diff
@@ -60,6 +60,7 @@ import 'exceptions.dart';
 /// See also:
 ///
 ///  * <https://cloud.google.com/appengine/docs/standard/python/reference/request-response-headers>
+// TODO(chillers): Remove (1) when DeviceLab has migrated to LUCI, https://github.com/flutter/flutter/projects/151#card-47536851
```
keyonghan, line 63: With the same logic being implemented in `SwarmingAuthenticationProvider`, can we delete this part now?

app_dart/lib/src/request_handling/swarming_authentication.dart
```diff
@@ -0,0 +1,135 @@
...
+      return await authenticateIdToken(swarmingToken, clientContext: clientContext, log: log);
+    }
+
+    throw const Unauthenticated('Request rejected due to not from LUCI or Cocoon agent');
+  }
+
+  // This method is expensive (run time of ~1,500ms!). If the server starts
+  // handling any meaningful API traffic, we should move request processing
+  // to dedicated isolates in a pool.
+  static bool _compareHashAndPassword(List<int> serverAuthTokenHash, String clientAuthToken) {
```
keyonghan, line 125: move to a common script so both `SwarmingAuthenticationProvider` and `Authentication` can share?

CaseyHillers: Done. Made public in `AuthenticationProvider`, with plan to remove after devicelab luci migration is complete.

app_dart/lib/src/request_handling/authentication.dart
CaseyHillers: `/api/append-log`, `/api/get-authentication-status`, and `'/api/update-agent-health'` still use the `AuthenticationProvider`, so not yet. I'll investigate what endpoints require what auth methods, and update in a separate PR.

My expectation is for this to be done after the devicelab to LUCI migration as we will be able to remove several request handlers.

keyonghan: SGTM

keyonghan (approved):
LGTM

PR #985, opened 2020-11-03
Add retries to buildbucket tries.

godofredoc:
Sometimes the buildbucket requests are raising exceptions and github
status stays on pending forever.

Bug: https://github.com/flutter/flutter/issues/69589

### DIFF
--- a/app_dart/lib/src/service/luci_build_service.dart
+++ b/app_dart/lib/src/service/luci_build_service.dart
@@ -10,6 +10,7 @@ import 'package:cocoon_service/src/model/github/checks.dart';
 import 'package:cocoon_service/src/request_handling/exceptions.dart';
 import 'package:github/github.dart' as github;
 import 'package:meta/meta.dart';
+import 'package:retry/retry.dart';
 
 import '../../cocoon_service.dart';
 import '../model/appengine/service_account_info.dart';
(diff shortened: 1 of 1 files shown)


app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -204,7 +205,14 @@ class LuciBuildService {
         ),
       );
     }
-    await buildBucketClient.batch(BatchRequest(requests: requests));
+    const RetryOptions r = RetryOptions(maxAttempts: 3);
+    await r.retry(
+      () async {
+        await buildBucketClient.batch(BatchRequest(requests: requests));
+      },
+      retryIf: (Exception e) => e is BuildBucketException,
```
CaseyHillers, line 213: This will only do retries with in a 2 second span. Should we modify `delayFactor` to be longer? Like 1s, 10s, 30s?

CaseyHillers (approved):
LGTM

keyonghan (approved):
LGTM

app_dart/lib/src/service/luci_build_service.dart
godofredoc: Added a 2 sec delayFactor.

PR #986, opened 2020-11-03
add cirrus config to build docker image on dockerfile diffs

christopherfujino:
Currently, Cirrus will auto-detect the Dockerfile in this repo and build it at every commit. This built docker image is not used. This change will explicitly build the docker image only on dockerfile diffs.


godofredoc (approved):

CaseyHillers (approved):
LGTM - Thank you for including the documentation link!

PR #989, opened 2020-11-06
Add retries with backoff to the call of updateCheckRun.

godofredoc:
This is because github apis frequently reject the access token and we
need to generate a new one and retry.

Bug:
  https://github.com/flutter/flutter/issues/69248


keyonghan (approved):
LGTM

CaseyHillers (approved):

PR #990, opened 2020-11-06
Specify pubsub topic use when rescheduling prod builds.

yusufm:
This PR adds the new pubsub topic to use when cocoon reschedules a prod build. This is part of the solution for https://github.com/flutter/flutter/issues/69621. We will need to also specify the notification to happen when the task is initially scheduled by luci.


keyonghan (approved):
LGTM

godofredoc (approved):

PR #994, opened 2020-11-10
[app_dart] Add access token authentication for swarming tasks

CaseyHillers:
Switches `SwarmingAuthenticationProvider` id token authentication to access token.

Make some members in `AuthenticationProvider` public to remove code duplication in `SwarmingAuthenticationProvider`.

# Issues

# Tests

Switched tests from id token to access token equivalents.

Check that only the LUCI prod account is authenticated.

### DIFF
--- a/app_dart/lib/src/request_handling/swarming_authentication.dart
+++ b/app_dart/lib/src/request_handling/swarming_authentication.dart
@@ -3,18 +3,18 @@
 // found in the LICENSE file.
 
 import 'dart:async';
+import 'dart:convert';
 import 'dart:io';
 
 import 'package:appengine/appengine.dart';
 import 'package:cocoon_service/cocoon_service.dart';
+import 'package:cocoon_service/src/model/google/token_info.dart';
 import 'package:gcloud/db.dart';
-import 'package:googleapis/oauth2/v2.dart';
-import 'package:http/http.dart';
 import 'package:meta/meta.dart';
 
+import '../foundation/providers.dart';
 import '../foundation/typedefs.dart';
 import '../model/appengine/agent.dart';
-import '../model/appengine/service_account_info.dart';
 
 import 'exceptions.dart';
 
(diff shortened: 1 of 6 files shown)


app_dart/lib/src/request_handling/swarming_authentication.dart
```diff
@@ -111,9 +100,57 @@ class SwarmingAuthenticationProvider extends AuthenticationProvider {
...
+  /// validated against a different endpoint. We only validate access tokens if they belong
+  /// to a DeviceLab service account.
+  ///
+  /// If LUCI auth adds id tokens, we can switch to that and remove this.
+  Future<AuthenticatedContext> authenticateAccessToken(String accessToken,
+      {ClientContext clientContext, Logging log}) async {
+    // Authenticate as a signed-in Google account via OAuth id token.
+    final HttpClient client = httpClientProvider();
+    try {
+      final HttpClientRequest verifyTokenRequest = await client.getUrl(Uri.https(
```
CaseyHillers, line 122: Using the Google OAuth apis was malforming the access token request, causing a 400 bad request to return from the API. I scratched it and went for this approach as it's similar to id tokens.

app_dart/lib/src/request_handling/swarming_authentication.dart
```diff
@@ -111,9 +100,57 @@ class SwarmingAuthenticationProvider extends AuthenticationProvider {
...
 
     throw const Unauthenticated('Request rejected due to not from LUCI or Cocoon agent');
   }
+
+  /// Authenticate [accessToken] against Google OAuth 2 API.
+  ///
+  /// Access tokens are the legacy authentication strategy for Google OAuth, where ID tokens
+  /// are the new technique to use. LUCI auth only generates access tokens, and must be
+  /// validated against a different endpoint. We only validate access tokens if they belong
+  /// to a DeviceLab service account.
```
godofredoc, line 114: Should this be updated to also cover the prod luci service account?

app_dart/lib/src/request_handling/swarming_authentication.dart
godofredoc: Is this an error with the APIs or the client libraries making the request?

app_dart/lib/src/request_handling/swarming_authentication.dart
CaseyHillers: Done

app_dart/lib/src/request_handling/swarming_authentication.dart
CaseyHillers: The client libraries as the the request looks to be putting all possible query parameters in. I'm not sure who owns that library, but to unblock for now I went with this approach.

godofredoc: Actually the question was to decide if we need to open a bug for the client library but not a blocker for this PR.

godofredoc (approved):

app_dart/lib/src/request_handling/swarming_authentication.dart
CaseyHillers: I'll create a reproduction and file an issue.

PR #995, opened 2020-11-10
[app_dart] Speed up dev deploy script

CaseyHillers:
Add a `--skip-build` flag that will reuse the existing build files in `app_dart` (will not try to build the Angular or Flutter projects which can add a few minutes). This is useful for backend only changes.

Add `--quiet` to gcloud deploy so user input is not required.

### DIFF
--- a/app_dart/dev/deploy.dart
+++ b/app_dart/dev/deploy.dart
@@ -88,7 +88,7 @@ Future<bool> _checkDependencies() async {
 Future<bool> _buildAngularDartApp() async {
   if (_skipBuild) {
     stdout.writeln('Skipping the build of Angular app');
-    return Future<bool>.value(true);
+    return true;
   }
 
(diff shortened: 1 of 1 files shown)


yusufm (approved):
LGTM.

Quick question, wouldn't you want to return a bool, rather than Future<bool> to indicate the returned future is completed with that value?

CaseyHillers:
> Quick question, wouldn't you want to return a bool, rather than Future to indicate the returned future is completed with that value?

Yes, changed. Since this function is marked async, Dart will convert a bool returned into `Future<bool>`. I didn't see these were marked `async` so went the long way round :-)

keyonghan:
shall we update the readme as well?

CaseyHillers:
> shall we update the readme as well?

The app_dart README points to using the `--help` flag to learn more. I added documentation there for this flag.

PR #996, opened 2020-11-10
[app_flutter] Update gitignore with latest from flutter/flutter

CaseyHillers:
With the support for MacOS, Linux, and Windows desktops there's a few files that stay in staging due to Cocoon not having an updated gitignore with these platforms.

### DIFF
--- a/app_flutter/.gitignore
+++ b/app_flutter/.gitignore
@@ -1,6 +1,5 @@
 # Miscellaneous
 *.class
-*.lock
 *.log
 *.pyc
 *.swp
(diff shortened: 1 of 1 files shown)


yusufm (approved):

app_flutter/.gitignore
```diff
@@ -1,5 +1,6 @@
 # Miscellaneous
 *.class
+*.lock
```
dnfield, line 3: Shouldn't we keep pubspec locks for application?

CaseyHillers: Good catch! I didn't recognize it without pubspec in front of it

dnfield (approved):

PR #997, opened 2020-11-11
Remove handlers to update cirrus prod tasks.

godofredoc:
Handlers to update cirrus tasks are not used anymore and are getting
removed to cleanup the source code.


yusufm (approved):
Can't  app_dart/lib/src/request_handlers/refresh_cirrus_status.dart be deleted if the handler being removed?

godofredoc:
> art/lib/src/request_handlers/refresh_cirrus_status.dart be deleted if the handler bein

It has a couple of models that are still used in the waiting_for_tree_to_go_green handler. Those will be removed when we update that part.

CaseyHillers (approved):

PR #998, opened 2020-11-11
[app_dart] Add logging in swarming auth

CaseyHillers:
I'm having a hard time to debugging what's going wrong in the devicelab tests trying to upload results on LUCI. Adding logs to see what's happening. Right now the test runner is returning an empty response body, so I want to double check that we even get to authenticating the service account code path.


keyonghan (approved):
LGTM

CaseyHillers:
Filed https://github.com/flutter/flutter/issues/70312 for the flake

PR #1000, opened 2020-11-12
[app_dart] Fix update-task-status query filter with extra space

CaseyHillers:
Datastore in production differs slightly from our test service. It cannot have a space at the end of the filter, otherwise an error is thrown.

# Issue

https://logs.chromium.org/logs/flutter/led/chillers_google.com/226e07e7f862d86628e3b6b73e7b54061ff3caea821ec3915b83c298f8d60937/+/steps/run_web_benchmarks_canvaskit/0/stdout

500 on devicelab luci task due to this filter.

# Test

Added assert check to test datastore service to prevent future issues like this.


keyonghan (approved):
LGTM

PR #1001, opened 2020-11-12
Stop adding cirrus task to datastore when refreshing new github commits

keyonghan:
This is a follow up of https://github.com/flutter/cocoon/pull/997


jmagman (approved):
LGTM

PR #1002, opened 2020-11-13
[app_dart] update-task-status only updates chromebot tasks when given LUCI information

CaseyHillers:
Give more informative messages when errors occur.

Only update chromebot tasks when given the LUCI information. This is needed for the DeviceLab migration to LUCI as there will be some period of time where a chromebot and devicelab task both have the same name.

# Tests

- Added check for Cocoon task exists but not a LUCI one
- Cocoon and LUCI task
- There's an existing one for only a LUCI task

### DIFF
--- a/app_dart/lib/src/request_handlers/update_task_status.dart
+++ b/app_dart/lib/src/request_handlers/update_task_status.dart
@@ -160,13 +160,12 @@ class UpdateTaskStatus extends ApiRequestHandler<UpdateTaskStatusResponse> {
     final String taskName = requestData[taskNameParam];
     final Query<Task> query = datastore.db.query<Task>(ancestorKey: commitKey)..filter('name =', taskName);
     final List<Task> tasks = await query.run().toList();
-    if (tasks.length == 1) {
-      return tasks.first;
+    if (tasks.length != 1) {
+      log.warn('Found ${tasks.length} entries for $taskName');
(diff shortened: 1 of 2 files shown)


CaseyHillers:
I forgot to push to a separate branch, but these are related so i'll keep this PR.

keyonghan:
The logic here is based on: devicelab builders are already defined in https://devops-console-oss.corp.google.com/flutter/flutter/+/master:dev/prod_builders.json, so that `refresh-github-commits` api can insert another set of records related to luci.

CaseyHillers:
@keyonghan this is concerned with the migration period. During migration we will have devicelab tasks running on both Cocoon agent and LUCI, so there will be duplicate entries in Datastore when searching by taskName. This also allows us to validate LUCI against Cocoon before making the final switch.

keyonghan:
If this is landed without updating prod json, and if you want to run devicelab tasks and update task status from LUCI side (with the existing devicelab agent tasks running), it seems going to throw exception all the time.

Unless you do not plan to update task status from LUCI side for now?

CaseyHillers:
> Unless you do not plan to update task status from LUCI side for now?

Yes, I have a CL to change the recipe to use a separate property for uploading metrics. [CL here](https://flutter-review.googlesource.com/c/recipes/+/8644)

When this PR and that CL land, all we need to do is add the devicelab tasks to the prod builders. There is no expected downtime / red time for this to occur.

keyonghan (approved):
LGTM

PR #1005, opened 2020-11-16
[nnbd] Move repository to its own Flutter project

CaseyHillers:
With NNBD, Cocoon has reached some issues with being compatible with Flutter. As a temporary workaround, we've been able to move to the beta channel to not cause the breakage.

This PR moves the repository dashboard that is based on a Flutter web fork into its own Flutter project.

## app
- Remove repository logic
- Remove unused json objects (used by old build dashboard)
- Remove json serializable option that was removed
- Remove Flutter for web fork dependencies
- Upgrade pubspec to be nullsafe compatible
- Fixed Angular dart nullability issue

## repo_dashboard
- Copy of the repository files here
- Ran `flutter create`
- Skipped a test that requires some refactoring of the network image code
- Switch use of `dart:html` http calls to `package:http`
- Switch use of `dart:html` localstorage to `package:localstorage`
- Switch use of `dart:html` window.open to `package:url_launcher`
- Switch use of `dart:html` refresh to an HTML attribute
- Switch `print` calls to `debugPrint` to minimize noise in tests
- Fix lints

## app_dart
- To prevent build artifact overlap, repository is moved to `/repository/`, added redirect

## build + test
- `dev/deployment.dart` calls the scripts in `cloudbuild` to reduce build logic duplication
- Deployment: Added repo_dashboard logic to `cloudbuild`
- CI: Add repository to `tests.yaml` based on existing Flutter testing

# Issue

https://github.com/flutter/flutter/issues/70428

# Preview

http://testchillers.flutter-dashboard.appspot.com/repository/index.html

# Future work

I worked on fixing the repo dashboard to work on Flutter, but didn't implement best practices. Additionally, for trivial test failures I disabled them. These skipped tests were because the widget is now off screen due to using a mobile layout instead of the default browser size.

Next steps are to start picking this apart and importing it into the main flutter dashboard app.


CaseyHillers:
\cc @christopherfujino I have this running and can clean it up today.

My TODO is:
1. Run license adder to repository project
2. Add repository project to Cocoon CI
3. Add repository project to deploy process
4. Fix the roll test as it fails hard with `dart:html` imports

christopherfujino (commented):
While we're moving things can we change the repository names? they're very confusing. maybe something like:

app -> benchmarks_dashboard
repository -> repo_dashboard
app_dart -> server
app_flutter -> frontend

CaseyHillers:
> While we're moving things can we change the repository names?

I'm happy to follow up in a separate PR with these renames (outside of repository). Let's keep the scope of this PR to the repository dashboard.

christopherfujino:
> > While we're moving things can we change the repository names?
> 
> I'm happy to follow up in a separate PR with these renames (outside of repository). Let's keep the scope of this PR to the repository dashboard.

makes sense. good chance renaming them will cause an integration test in another repo to fail :P

app/pubspec.yaml
```diff
@@ -1,42 +1,25 @@
 name: cocoon
-version: 0.0.2
 homepage: https://github.com/flutter/cocoon
+publish_to: none
```
christopherfujino, line 3: nice lol

christopherfujino (approved):
LGTM. Thanks for making this change!

CaseyHillers:
@jmagman would you mind reviewing the dashboard changes?

CaseyHillers:
I still need to update the licenses to match the expected format (using `addlicense` wasn't sufficient).

Fix a small issue where going to http://testchillers.flutter-dashboard.appspot.com/repository/#/ 404's (should redirect to the index.html)

jmagman (commented):
Let's delete all the `repo_dashboard/*` platform directories we know don't work (desktop, right?)

CaseyHillers:
> Let's delete all the repo_dashboard/* platform directories we know don't work (desktop, right?)

Done.

CaseyHillers:
@keyonghan would you mind reviewing the deploy, cloudbuild, and bin/server.dart changes?

app_dart/bin/server.dart
```diff
@@ -142,13 +142,15 @@ Future<void> main() async {
         final int trimIndex = min(queryIndex, anchorIndex);
         final String filePath = request.uri.path.substring(0, trimIndex);
 
-        const List<String> indexRedirects = <String>['/build.html'];
-        if (indexRedirects.contains(filePath)) {
+        const Map<String, String> redirects = <String, String>{
+          '/build.html': '/#/build',
+          '/repository': '/repository/index.html',
+          '/repository/': '/repository/index.html',
```
keyonghan, line 148: Do we need these two options? Existing cocoon seems to only reference /repository.html. Or I missed somewhere.

keyonghan (approved):
LGTM with nit.

app_dart/bin/server.dart
CaseyHillers: Yes. The Flutter app rewrites `/repository/index.html` to `/repository/#/`. On a reload, Cocoon see this as `/repository` and 404s. Ideally, we'd do something similar to below to just return the `index.html`, but it adds more complexity.

CaseyHillers:
Thanks all for the reviews on such a large PR!

PR #1007, opened 2020-11-17
[app_dart] update-task-status uses builder name instead of task name

CaseyHillers:
## Description

LUCI tasks in Cocoon use the builder name for unique look up. This matches that logic.

In the instance of running `web_benchmarks_html` devicelab test, Cocoon stored it as `linux_web_benchmarks_html`. This would cause a 500 issue when retrieving the task to update. Instead, if we search for the task by the builder name, we can uniquely identify it.

## Related Issues

## Tests

Updated tests to use builder name instead of task name.


keyonghan (approved):
LGTM

PR #1008, opened 2020-11-18
[agent] Clean log args

CaseyHillers:
Clean up some of the output in the devicelab logs.

# Issues

fxb/64761

### DIFF
--- a/agent/lib/src/utils.dart
+++ b/agent/lib/src/utils.dart
@@ -258,7 +258,11 @@ Future<Null> getFlutterAt(String revision) async {
 Future<Process> startProcess(String executable, List<String> arguments,
     {Map<String, String> env, bool silent: false}) async {
   final List<String> logSafeArguments = arguments ?? <String>[];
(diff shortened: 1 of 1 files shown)


agent/lib/src/utils.dart
```diff
@@ -257,7 +257,9 @@ Future<Null> getFlutterAt(String revision) async {
 
 Future<Process> startProcess(String executable, List<String> arguments,
     {Map<String, String> env, bool silent: false}) async {
-  String command = '$executable ${arguments?.join(" ") ?? ""}';
+  final List<String> logSafeArguments = arguments ?? <String>[];
+  logSafeArguments.removeWhere((String arg) => arg.contains('cloud-auth-token'));
```
jonahwilliams, line 261: maybe create a const set excludeList in case we need to add more?

CaseyHillers: Done

jonahwilliams (approved):
LGTM

PR #1009, opened 2020-11-18
[app_dart] Use LuciBuilder when creating chromebot tasks

CaseyHillers:
Use `LuciBuilder` to populate the chromebot tasks. The information in `prod_builders.json` is enough to populate some extra values on creation, mainly `builderName`.

# Issues

Fixes https://github.com/flutter/flutter/issues/69839 - `builderName` was not available on new chromebot tasks

https://github.com/flutter/flutter/issues/66191 - `builderName` is not initially populated on LUCI tasks, and can cause race condition failures depending on when datastore refreshed from luci

# Test

Updated chromebot task to expect correct values.

### DIFF
--- a/app_dart/lib/src/model/appengine/task.dart
+++ b/app_dart/lib/src/model/appengine/task.dart
@@ -56,6 +56,7 @@ class Task extends Model {
     assert(builder.flaky != null);
     return Task(
       attempts: 1,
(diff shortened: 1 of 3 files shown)


keyonghan (approved):
LGTM

yusufm (approved):

PR #1011, opened 2020-11-19
[agent] Null check before checking for excluded keyword

CaseyHillers:
Follow up from https://github.com/flutter/cocoon/pull/1008


godofredoc (approved):

CaseyHillers:
Landing as this is blocked with new lints from latest beta roll of Flutter. Fix is in https://github.com/flutter/cocoon/pull/1010

PR #1014, opened 2020-12-01
Preserve LUCI infra failures in Cocoon

CaseyHillers:
LUCI publishes this status, and would be helpful to differentiate infra failures on the dashboard versus test failures. This also allows us to reduce tree closures due to infra causes

This PR takes LUCI builds that were infra failures and keeps the status. Previously, they were marked as failures.

# Issue

https://github.com/flutter/flutter/issues/70331

# Test

Added table test for the app_dart changes

Flutter app already has a similar table test that uses the map to validate expectations

### DIFF
--- a/app_dart/lib/src/model/appengine/task.dart
+++ b/app_dart/lib/src/model/appengine/task.dart
@@ -93,6 +93,7 @@ class Task extends Model {
   /// The list of legal values for the [status] property.
   static const List<String> legalStatusValues = <String>[
     statusNew,
+    statusInfraFailure,
     statusInProgress,
     statusSucceeded,
     statusFailed,
--- a/app_dart/lib/src/model/appengine/task.dart
+++ b/app_dart/lib/src/model/appengine/task.dart
@@ -73,6 +73,9 @@ class Task extends Model {
   /// The task is yet to be run.
   static const String statusNew = 'New';
 
+  /// The task failed to run due to an unexpected issue.
+  static const String statusInfraFailure = 'Infra Failure';
+
   /// The task is currently running.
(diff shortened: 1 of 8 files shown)


keyonghan:
Do we want to keep the tree open if the test fails with infra failure? If yes, are we 100% sure the `infra failure` is indeed not test related? IIRC, we had cases where some `infra failure` was caused by test issues.

CaseyHillers:
> Do we want to keep the tree open if the test fails with infra failure? If yes, are we 100% sure the `infra failure` is indeed not test related? IIRC, we had cases where some `infra failure` was caused by test issues.

The existing build status logic will ignore infra failures similar to what's done for running tasks. It will pull from an earlier result instead.

To keep the PR small, I just updated the status. I want to follow up with another PR to auto-retry infra failures.

In general, I think we will need to improve the recipes to help assert test failures are shown red and everything else is purple. This is a start so we can start filing issues against misclassified builds.

keyonghan:
Yeah, distinguishing infra failures from test failures should be our next target. Before we have a clear solution, I would prefer to keep the tree closed to avoid missing potential hidden issues. Rather than the other way around, to file bugs when we see regressions. At least at this moment. 
What do you think @godofredoc ?

But an auto retry for `infra failures` should be good.

CaseyHillers:
@keyonghan the follow up PR will reschedule infra failures. The existing tree status policy will close the tree as any incomplete rerunning tasks close the tree.

Since it's hard to do dependent PRs in git, is it okay to land this and I can prioritize that PR this week?

keyonghan (approved):
LGTM.

PR #1015, opened 2020-12-01
[app_flutter] Fix golden tests on windows

CaseyHillers:
Adding flutter/cocoon back to customer_tests fails because of the windows tests. We should use the path library instead of hardcoding how the path is generated.

# Issues

https://github.com/flutter/flutter/issues/65045


Piinks (approved):
LGTM

PR #1016, opened 2020-12-02
Implement device doctor tool

keyonghan:
As part of devicelab migration to LUCI, we need to support device health check and management in LUCI.

This PR creates a device management tool `device_doctor` to do pre health check before running task, clean up after task finishes, and restart the device when task fails or device is identified as unhealthy.

Related https://github.com/flutter/flutter/issues/66193

PR #1018, opened 2020-12-04
Create the basic structure of device doctor tool

keyonghan:
https://github.com/flutter/cocoon/pull/1016 is way too big to make it difficult to review (sorry for that 😢 ).

Here come the divided pieces, and this is the first one for basic structure. It addressed some comments from the above PR. Please help take another look.

What's included in this PR:
- main.dart to include only args check and initialization
- includes only ios device case for simplicity
- adds basic utils-support for current logic.
- creates fake ios device for testing
- adds corresponding tests

More PRs to come to finish up the whole tool.

Related flutter/flutter#66193


device_doctor/bin/main.dart
```diff
@@ -0,0 +1,62 @@
+// Copyright 2020 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:io';
+
+import 'package:args/args.dart';
+
+import 'package:device_doctor/src/config.dart';
```
CaseyHillers, line 9: Prefer relative imports - https://dart-lang.github.io/linter/lints/prefer_relative_imports.html

device_doctor/bin/main.dart
```diff
@@ -0,0 +1,62 @@
...
+
+const String actionFlag = 'action';
+const String deviceOSFlag = 'deviceOS';
```
CaseyHillers, line 12: nit: CLI args in the Flutter tooling follows the style of `--device-os`

device_doctor/bin/main.dart
```diff
@@ -0,0 +1,62 @@
...
+  if (!_checkArgs(parser, args)) {
+    stdout.write('\nRequired flags:\n'
+        '--$actionFlag Supported actions are healthcheck and recovery.\n'
+        '--$deviceOSFlag Supported device OS: android and ios.\n');
+    exit(1);
+  }
+
+  Config.initialize(_deviceOS);
+
+  // Healthcheck and recovery will be implemented here. PR to be continued.
```
CaseyHillers, line 40: Is there an issue we can track to?

device_doctor/bin/main.dart
```diff
@@ -0,0 +1,62 @@
...
+Future<void> main(List<String> args) async {
+  final ArgParser parser = ArgParser();
+  parser
+    ..addOption('$actionFlag')
+    ..addOption('$deviceOSFlag')
+    ..addFlag('$helpFlag');
+
+  if (!_checkArgs(parser, args)) {
+    stdout.write('\nRequired flags:\n'
+        '--$actionFlag Supported actions are healthcheck and recovery.\n'
```
CaseyHillers, line 33: Arg parser can be used to generate this dynamically. The devicelab test runner does it that way - https://github.com/flutter/flutter/blob/master/dev/devicelab/bin/run.dart

device_doctor/lib/src/config.dart
```diff
@@ -0,0 +1,40 @@
...
+// found in the LICENSE file.
+
+import 'package:meta/meta.dart';
+
+import 'device.dart';
+
+Config _config;
+Config get config => _config;
+
+/// Configures device operating system based on injected decice OS.
```
CaseyHillers, line 12: nit: decice -> device

device_doctor/lib/src/config.dart
```diff
@@ -0,0 +1,40 @@
...
+///
+/// This determines the [DeviceDiscovery].
+class Config {
```
CaseyHillers, line 15: This doesn't need to be addressed this PR, but we should consider passing a device ip instead of assuming there is only ever going to be one device attached.

device_doctor/lib/src/device.dart
```diff
@@ -0,0 +1,92 @@
...
+enum DeviceOperatingSystem { android, ios }
+
+/// Discovers available devices and chooses one to work with.
+abstract class DeviceDiscovery {
+  factory DeviceDiscovery() {
+    switch (config.deviceOperatingSystem) {
+      case DeviceOperatingSystem.ios:
+        return IosDeviceDiscovery();
+      default:
+        throw StateError('Unsupported device operating system: {config.deviceOperatingSystem}');
```
CaseyHillers, line 23: nit: is this missing $? ${config.deviceOperatingSystem}

device_doctor/lib/src/device.dart
```diff
@@ -0,0 +1,92 @@
...
+    }
+  }
+
+  /// Lists all available devices' IDs.
+  Future<List<Device>> discoverDevices({int retriesDelayMs = 10000});
```
CaseyHillers, line 28: Swarming bots are usually associated with one DUT. Is there a use case for Flutter to support multiple devices from one bot?

device_doctor/lib/src/device.dart
```diff
@@ -0,0 +1,92 @@
...
+
+  /// Checks the health of the available devices.
+  Future<Map<String, HealthCheckResult>> checkDevices();
+
+  /// Recovers the device.
+  Future<void> recoverDevice();
```
CaseyHillers, line 34: The rest of the DeviceDiscovery API talks about devices, but this references one device. How do we know which device gets recovered?

device_doctor/lib/src/device.dart
```diff
@@ -0,0 +1,92 @@
...
+}
+
+/// A proxy for one specific device.
+abstract class Device {
```
CaseyHillers, line 38: This API seems too specific to implement beyond phones. Would `initialize``recover`, and `collect_logs` make more sense? Some of these methods make more sense as part of a child interface like `PhoneDevice`.

device_doctor/lib/src/device.dart
```diff
@@ -0,0 +1,92 @@
...
+  ///
+  /// Assumes the device doesn't have a secure unlock pattern.
+  Future<void> unlock();
+
+  /// Recovers the device back to a healthy state.
+  Future<void> recover();
+}
+
+/// Result of a health check for a specific parameter.
+class HealthCheckResult {
```
CaseyHillers, line 70: Health checks seem out of scope for this device file.

device_doctor/lib/src/process_helper.dart
```diff
@@ -0,0 +1,65 @@
...
+    _runningProcesses.remove(procInfo);
+  });
+
+  return proc;
+}
+
+/// Executes a command and returns its standard output as a String.
+///
+/// Standard error is redirected to the current process' standard error stream.
+Future<String> eval(String executable, List<String> arguments,
```
CaseyHillers, line 53: I hesitate from `eval` as it's more associated with dynamic code calling, but here we know what we're calling. `execCommand`?

device_doctor/lib/src/utils.dart
```diff
@@ -0,0 +1,84 @@
...
+  final int _level;
+  final String name;
+
+  static const LogLevel debug = const LogLevel._(0, 'DEBUG');
+  static const LogLevel info = const LogLevel._(1, 'INFO');
+  static const LogLevel warning = const LogLevel._(2, 'WARN');
+  static const LogLevel error = const LogLevel._(3, 'ERROR');
+}
+
+abstract class Logger {
```
CaseyHillers, line 28: https://pub.dev/packages/logging

CaseyHillers (commented):
Thank you for splitting this up!

We should add https://github.com/flutter/flutter/blob/master/analysis_options.yaml here to better match the Flutter style guide.

device_doctor/lib/src/config.dart
```diff
@@ -0,0 +1,40 @@
...
+  Config({
+    @required this.deviceOperatingSystem,
+  });
+
+  static void initialize(String deviceOS) {
```
jonahwilliams, line 20: Consider not using a global variable for passing around configuration. Its easy to track now when this project is small but will become difficult to refactor as it grows larger

device_doctor/lib/src/device.dart
jonahwilliams, line 28: Might be more dart-y if this accepted a `Duration` object instead

device_doctor/lib/src/ios_device.dart
```diff
@@ -0,0 +1,90 @@
...
+import 'package:meta/meta.dart';
+
+import 'device.dart';
+import 'process_helper.dart';
+
+/// IOS implementation of [DeviceDiscovery].
+///
+/// Discovers available ios devices and chooses one to work with.
+class IosDeviceDiscovery implements DeviceDiscovery {
+  factory IosDeviceDiscovery() {
```
jonahwilliams, line 17: What is the goal of having this be a singleton?

device_doctor/lib/src/ios_device.dart
```diff
@@ -0,0 +1,90 @@
...
+  IosDeviceDiscovery._();
+
+  @visibleForTesting
+  IosDeviceDiscovery.testing();
+
+  static IosDeviceDiscovery _instance;
+
+  @override
+  Future<List<Device>> discoverDevices({int retriesDelayMs = 10000}) async {
+    final List<String> iosDeviceIds = LineSplitter.split(await deviceListOutput()).toList();
```
jonahwilliams, line 30: you don't need to convert it to a List to map it, consider:

```
LineSplitter.split(await deviceListOutput())
  .map((String id) => IosDevice(deviceId: id))
  .toList()
```

device_doctor/lib/src/utils.dart
```diff
@@ -0,0 +1,84 @@
+// Copyright 2020 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:io';
+
+import 'package:process/process.dart';
+
+/// Virtual current working directory, which affect functions, such as [exec].
+String cwd = Directory.current.path;
```
jonahwilliams, line 10: If you use package path it has a `path.current` getter.

device_doctor/lib/src/utils.dart
```diff
@@ -0,0 +1,84 @@
...
+
+ProcessManager processManager = LocalProcessManager();
```
jonahwilliams, line 12: Same comment as elsewhere: globals are going to make this code difficult to refactor later

device_doctor/lib/src/utils.dart
jonahwilliams: Right, I would either use package logging, or a much simpler logging interface

device_doctor/bin/main.dart
keyonghan: That seems for files located within lib?

device_doctor/bin/main.dart
keyonghan: updated.

device_doctor/bin/main.dart
keyonghan: Thanks for this info. Updated the parser with `help`.

device_doctor/bin/main.dart
keyonghan: Added https://github.com/flutter/flutter/issues/66193

device_doctor/lib/src/config.dart
keyonghan: Done.

device_doctor/lib/src/config.dart
keyonghan: Yeah, I need to check swarming bot_config to see what we can use here, or maybe device id. Will include that once confirmed.

device_doctor/lib/src/config.dart
keyonghan: Updated to use local variable instead.

device_doctor/lib/src/device.dart
keyonghan: Good catch! Updated.

device_doctor/lib/src/device.dart
keyonghan: We currently only support one device on one bot, but may support more to one in the future when needed.

keyonghan: Updated to `Duration`.

device_doctor/lib/src/device.dart
keyonghan: Ah, a typo here. It should be `recoverDevices`. The implementation will recover all devices.
Will generalize this when `device id/ip` is available.

device_doctor/lib/src/device.dart
keyonghan: The device here is meant for phone device. Updated the doc to reflect this.

device_doctor/lib/src/device.dart
keyonghan: Moved to utils.

device_doctor/lib/src/ios_device.dart
keyonghan: This is mirrored over from existing cocoon/agent. My guess is to ensure only one `IosDeviceDiscovery` instance. @yjbanov may have a better idea?

device_doctor/lib/src/ios_device.dart
keyonghan: Thanks. Updated.

device_doctor/lib/src/process_helper.dart
keyonghan: As this is just the initial step of the tool, we will have more general cases on the way, and we also will implement exec as well.

device_doctor/lib/src/utils.dart
keyonghan: replaced.

device_doctor/lib/src/utils.dart
keyonghan: Changed to use local variables instead.

device_doctor/lib/src/utils.dart
keyonghan: Updated as suggested. Thanks!

device_doctor/lib/src/device.dart
CaseyHillers: A host can run multiple swarming bots, but a swarming bot should only be connected to one DUT. Since this is a swarming bot utility, shouldn't this be phrased for a single device?

The counterexample I listed was moreso if there is a test that requires more than one phone. I've seen those set ups for swarming where they created specialized pools to support those use cases.

device_doctor/lib/src/device.dart
CaseyHillers: My use case is how can we generalize this to also work with Fuchsia devices later on. With this approach, we'll need to further abstract this class when we do add that support.

device_doctor/lib/src/process_helper.dart
CaseyHillers: https://unix.stackexchange.com/questions/296838/whats-the-difference-between-eval-and-exec

What would be the difference between the exec and eval in this project?

keyonghan: See https://github.com/flutter/cocoon/blob/master/agent/lib/src/utils.dart#L304 and https://github.com/flutter/cocoon/blob/master/agent/lib/src/utils.dart#L336

CaseyHillers: Good point. Returning ProcessResult should deduplicate that logic

keyonghan: Will consider consolidating when `exec` is called in ongoing PRs.

device_doctor/lib/src/device.dart
keyonghan: Talked offline, will leave the class as is now. May update/extend the class when we get to the point to support fuchsia devices.

device_doctor/lib/src/config.dart
```diff
@@ -0,0 +1,36 @@
+// Copyright 2020 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:meta/meta.dart';
```
godofredoc, line 5: Do we actually need this class?  we can potentially directly create the classes in the factory method using the flags.

device_doctor/lib/src/device.dart
```diff
@@ -0,0 +1,65 @@
...
+import 'config.dart';
+import 'ios_device.dart';
+import 'utils.dart';
+
+/// Operating system on the devices that this host is configured to test.
+enum DeviceOperatingSystem { android, ios }
+
+/// Discovers available devices and chooses one to work with.
+abstract class DeviceDiscovery {
+  factory DeviceDiscovery(Config config) {
```
godofredoc, line 16: We can use the flags directly in calls to this factory method.

device_doctor/lib/src/ios_device.dart
```diff
@@ -0,0 +1,90 @@
...
+
+  @override
+  final String deviceId;
+
+  // The methods below are stubs for now. They will need to be expanded.
+  // We currently do not have a way to lock/unlock iOS devices. So we assume the
+  // devices are already unlocked. For now we'll just keep them at minimum
+  // screen brightness so they don't drain battery too fast.
+
+  @override
```
godofredoc, line 67: The Device interface should not define all the methods below if they are not supported in ios only common methods.

device_doctor/lib/src/utils.dart
```diff
@@ -0,0 +1,45 @@
...
+  BuildFailedError(this.message);
+
+  final String message;
+
+  @override
+  String toString() => message;
+}
+
+/// Result of a health check for a specific parameter.
+class HealthCheckResult {
```
godofredoc, line 23: Does this class belong to a different file? health.dart maybe?

device_doctor/test/src/fake_ios_device.dart
```diff
@@ -0,0 +1,29 @@
+// Copyright 2020 The Flutter Authors. All rights reserved.
```
godofredoc, line 1: Why a fake rather than a mock?

device_doctor/bin/main.dart
```diff
@@ -0,0 +1,66 @@
...
+    exit(1);
+  }
+
+  final Config config = Config(deviceOS: _deviceOS);
+  DeviceDiscovery(config);
+
+  // TODO(keyonghan): Implement healthcheck and recovery, https://github.com/flutter/flutter/issues/66193.
+}
+
+bool _checkArgs(ArgParser parser, List<String> args) {
```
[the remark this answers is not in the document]
keyonghan: Good point! Have implemented `callback` to handle this.

device_doctor/lib/src/config.dart
keyonghan: Based on our current logic, this config class is not needed. Updated to use flag directly as suggested.

device_doctor/lib/src/device.dart
keyonghan: Replaced.

device_doctor/lib/src/ios_device.dart
keyonghan: Removed the ones not common in ios.

device_doctor/lib/src/utils.dart
keyonghan: Moved to health.dart instead.

device_doctor/test/src/fake_ios_device.dart
keyonghan: This is used to test function `discoverDevices`, which calls another function `deviceListOutput`.
If we mock `discoverDevices` directly, we will lose testing logic of this function.
Instead if we use fake, `deviceListOut` can be manipulated so that we can test the logic of `discoverDevices`. This is how we do here.

godofredoc (approved):
LGTM but will defer the final approval to @CaseyHillers

device_doctor/lib/src/ios_device.dart
```diff
@@ -0,0 +1,78 @@
...
+  Future<bool> isAwake() async => true;
+
+  @override
+  Future<bool> isAsleep() async => false;
+
+  @override
+  Future<void> unlock() async {}
```
CaseyHillers, line 74: nit: In general, we should throw an unimplemented error on no-ops because we haven't finished the code

CaseyHillers (approved):
LGTM!

device_doctor/bin/main.dart
```diff
@@ -0,0 +1,46 @@
...
+Future<void> main(List<String> args) async {
+  final ArgParser parser = ArgParser();
+  parser
+    ..addOption('$actionFlag', help: 'Supported actions are healthcheck and recovery.', callback: (String value) {
+      if (value == null || !supportedOptions.contains(value)) {
+        throw FormatException('Invalid value for option --action: $value');
+      }
+    })
+    ..addOption('$deviceOSFlag', help: 'Supported device OS: android and ios.', callback: (String value) {
+      if (value == null || !supportedDeviceOS.contains(value)) {
```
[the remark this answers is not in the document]
keyonghan: Good catch. The null check is not necessary. Removed.

device_doctor/lib/src/ios_device.dart
keyonghan: The following PR is intended to implement corresponding actions. Will leave it there.

PR #1019, opened 2020-12-04
[app_flutter] Sort infra failures in grid after failures

CaseyHillers:
https://github.com/flutter/cocoon/pull/1014 forgot to weight this new status, so they are given a weight of 0. This will show them after failures.


keyonghan (approved):
LGTM

PR #1022, opened 2020-12-08
[app_dart] Check git branch on LUCI update-task-status

CaseyHillers:
This will return a more descriptive error on LUCI devicelab runs instead of a general check in the test runner.

# Issues

https://github.com/flutter/flutter/issues/71749

# Tests

Added check for bad request on a branch that does not exist in cocoon config

### DIFF
--- a/app_dart/lib/src/request_handlers/update_task_status.dart
+++ b/app_dart/lib/src/request_handlers/update_task_status.dart
@@ -184,7 +184,9 @@ class UpdateTaskStatus extends ApiRequestHandler<UpdateTaskStatusResponse> {
     final String gitBranch = requestData[gitBranchParam] as String;
     final List<String> flutterBranches = await config.flutterBranches;
     if (!flutterBranches.contains(gitBranch)) {
-      throw BadRequestException('$gitBranch is not a flutter/flutter branch');
+      throw BadRequestException('Failed to find flutter/flutter branch: $gitBranch\n'
+          'If this is a valid branch, '
+          'see https://github.com/flutter/cocoon/tree/master/app_dart#branching-support-for-flutter-repo');
     }
     final String id = 'flutter/flutter/$gitBranch/${requestData[gitShaParam]}';
     final Key commitKey = datastore.db.emptyKey.append(Commit, id: id);
--- a/app_dart/lib/src/request_handlers/update_task_status.dart
+++ b/app_dart/lib/src/request_handlers/update_task_status.dart
@@ -155,7 +155,7 @@ class UpdateTaskStatus extends ApiRequestHandler<UpdateTaskStatusResponse> {
   /// To lookup the value, we construct the ancestor key, which corresponds to the [Commit].
   /// Then we query the tasks with that ancestor key and search for the one that matches the builder name.
   Future<Task> _getTaskFromNamedParams(DatastoreService datastore) async {
-    final Key commitKey = _constructCommitKey(datastore);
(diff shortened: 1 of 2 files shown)


app_dart/lib/src/request_handlers/update_task_status.dart
```diff
@@ -178,8 +178,15 @@ class UpdateTaskStatus extends ApiRequestHandler<UpdateTaskStatusResponse> {
...
   /// Construct the Datastore key for [Commit] that is the ancestor to this [Task].
-  Key _constructCommitKey(DatastoreService datastore) {
-    final String id = 'flutter/flutter/${requestData[gitBranchParam]}/${requestData[gitShaParam]}';
+  ///
+  /// Throws [BadRequestException] if the given git branch does not exist in [CocoonConfig].
+  Future<Key> _constructCommitKey(DatastoreService datastore) async {
+    final String gitBranch = requestData[gitBranchParam] as String;
+    final List<String> flutterBranches = await config.flutterBranches;
+    if (!flutterBranches.contains(gitBranch)) {
+      throw BadRequestException('$gitBranch is not a flutter/flutter branch');
```
christopherfujino, line 187: how does this exception get surfaced? Will this register as a failed task?

CaseyHillers: Yes, red task. The test runner will output this error message as the reason why the task failed. After the migration, this should only be possibly encountered on new release branches.

christopherfujino: Cool. Can you update the exception message to explicitly state that `'$gitBranch is not defined in https://github.com/flutter/cocoon/blob/master/app_dart/dev/branches.txt'`?

CaseyHillers: Good point! Added, but there's a bit more steps (like flush the cache) so I linked to https://github.com/flutter/cocoon/tree/master/app_dart#branching-support-for-flutter-repo

I thought about pinning to a commit sha, but it got really long. Do you think a flutter.dev/go/ link would be appropriate?

app_dart/lib/src/request_handlers/update_task_status.dart
```diff
@@ -178,8 +178,17 @@ class UpdateTaskStatus extends ApiRequestHandler<UpdateTaskStatusResponse> {
...
-    final String id = 'flutter/flutter/${requestData[gitBranchParam]}/${requestData[gitShaParam]}';
+  ///
+  /// Throws [BadRequestException] if the given git branch does not exist in [CocoonConfig].
+  Future<Key> _constructCommitKey(DatastoreService datastore) async {
+    final String gitBranch = requestData[gitBranchParam] as String;
+    final List<String> flutterBranches = await config.flutterBranches;
+    if (!flutterBranches.contains(gitBranch)) {
+      throw BadRequestException('Failed to find flutter/flutter branch: $gitBranch\n'
+          'If this is a valid branch, '
+          'see https://github.com/flutter/cocoon/tree/master/app_dart#branching-support-for-flutter-repo');
```
christopherfujino, line 189: :+1:

christopherfujino (approved):
LGTM

app_dart/lib/src/request_handlers/update_task_status.dart
christopherfujino: I think linking to the master tree is probably better than pinning it. And if we change the anchor name, the link still sends you to the readme.

PR #1024, opened 2020-12-09
Add build script for device_doctor

keyonghan:
This creates the build script to build device_doctor. Currently it supports linux, mac, and windows.

Tests will be added from flutter/infra, and recipe side once this is in.

### DIFF
--- a/device_doctor/.gitignore
+++ b/device_doctor/.gitignore
@@ -1,0 +1,12 @@
+# CIPD and Dart-SDK binaries
+.cipd/
+dart-sdk/
+
+# Build outputs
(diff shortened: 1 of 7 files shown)


CaseyHillers (commented):
I missed this in the first review, but can you add device_doctor to the test_utilities so it's run in CI? It looks like we're missing test runs.

keyonghan:
> I missed this in the first review, but can you add device_doctor to the test_utilities so it's run in CI? It looks like we're missing test runs.

Added.

CaseyHillers (approved):

PR #1026, opened 2020-12-09
Add device_doctor ios management logic

keyonghan:
This PR implements basic health check:  device accessible, and recovery: restart and dismiss dialogs.


device_doctor/bin/main.dart
```diff
@@ -5,6 +5,7 @@
 import 'package:args/args.dart';
 
 import 'package:device_doctor/src/device.dart';
+import 'package:device_doctor/src/ios_device.dart';
```
CaseyHillers, line 8: Consider exporting what's needed from src, similar to how app_dart/lib/cocoon_service.dart works.

https://dart.dev/guides/language/effective-dart/usage#do-use-relative-paths-when-importing-libraries-within-your-own-packages-lib-directory

device_doctor/bin/main.dart
```diff
@@ -14,9 +15,15 @@ const List<String> supportedDeviceOS = <String>['ios', 'android'];
...
+String _action;
 String _deviceOS;
 
-/// Manage healthcheck and recovery for devices.
+/// Manage `healthcheck` and `recovery` for devices.
+///
+/// For `healthcheck`, if no device is found or any health check fails an stderr will be logged,
+/// and an exception will be thrown.
+///
+/// For `recovery`, device will do cleanup, reboot, and then dismiss any dialog for `ios` device.
```
CaseyHillers, line 26: > then dismiss any dialog for `ios` device

Could this be generalized to initialize device? If there's future dialogs that need to be closed for android, this doc would fall out of date.

device_doctor/lib/src/health.dart
```diff
@@ -2,20 +2,72 @@
...
+      command.add("PROVISIONING_PROFILE_SPECIFIER=${pl.environment['FLUTTER_XCODE_PROVISIONING_PROFILE_SPECIFIER']}");
+    }
+    Process proc = await pm.start(command, workingDirectory: dialogDir.path);
+    logger.info('Executing: $command');
+    // Discards stdout and stderr as they are too large.
+    await proc.stdout.drain<Object>();
+    await proc.stderr.drain<Object>();
+    int exitCode = await proc.exitCode;
+    if (exitCode != 0) {
+      fail('Command "xcrun xcodebuild -project infra-dialog..." failed with exit code $exitCode.');
```
CaseyHillers, line 49: The command should be moved into a variable so changes made to one place propagate to all calls.

device_doctor/lib/src/health.dart
```diff
@@ -26,3 +78,25 @@ class HealthCheckResult {
...
 }
+
+/// Check healthiness for discovered devices.
+///
+/// If any failed check, an explanation message will be sent to stdout and
+/// an exception will be thrown.
+Future<void> healthcheck(Map<String, List<HealthCheckResult>> deviceChecks) async {
+  if (deviceChecks.isEmpty) {
+    stderr.writeln('No healthy device is available');
+    throw StateError('No healthy device is available');
```
CaseyHillers, line 89: The bot config should already write the error when it is thrown, why are we also logging it?

device_doctor/test/src/health_test.dart
```diff
@@ -0,0 +1,77 @@
...
+    test('failed', () async {
+      Process proc = FakeProcess(123);
+      when(pm.start(any, workingDirectory: anyNamed('workingDirectory'))).thenAnswer((_) => Future.value(proc));
+
+      expect(
+        closeIosDialog(pm: pm),
+        throwsA(TypeMatcher<BuildFailedError>()),
+      );
+    });
+  });
```
CaseyHillers, line 55: What about a test where the tool cannot be found?

device_doctor/lib/src/utils.dart
```diff
@@ -18,3 +21,33 @@ class BuildFailedError extends Error {
   @override
   String toString() => message;
 }
+
+/// Creates a directory from the given path, or multiple path parts by joining
+/// them using OS-specific file path separator.
+Directory dir(String thePath,
+    [String part2, String part3, String part4, String part5, String part6, String part7, String part8]) {
```
CaseyHillers, line 28: Spread operator would clean this up.

https://dart.dev/guides/language/language-tour#spread-operator

device_doctor/test/src/health_test.dart
keyonghan: Added.

device_doctor/lib/src/utils.dart
keyonghan: This is to align with: https://github.com/dart-lang/path/blob/master/lib/path.dart#L257

device_doctor/lib/src/health.dart
keyonghan: Added for testing. Removed.

device_doctor/lib/src/health.dart
keyonghan: `command` is already there. Have included variable `command` in the message.

device_doctor/bin/main.dart
keyonghan: Updated to make it more general.

device_doctor/bin/main.dart
keyonghan: Updated.

device_doctor/lib/src/health.dart
```diff
@@ -17,10 +17,11 @@ import 'utils.dart';
 Future<HealthCheckResult> closeIosDialog(
     {ProcessManager pm = const LocalProcessManager(),
     String deviceId,
-    platform.Platform pl = const platform.LocalPlatform()}) async {
-  Directory dialogDir = dir(path.dirname(Platform.script.path), 'tool', 'infra-dialog');
+    platform.Platform pl = const platform.LocalPlatform(),
+    String infraDialog = 'infra-dialog'}) async {
```
CaseyHillers, line 21: nit: if you add a comma after `infra-dialog',` it will format the args to each be on their own line

CaseyHillers (approved):
LGTM!

device_doctor/lib/src/health.dart
keyonghan: reformatted. Thanks.

PR #1027, opened 2020-12-09
[app_dart] Pass branch when rescheduling prod builds

CaseyHillers:
Fixes https://github.com/flutter/flutter/issues/72007

### DIFF
--- a/app_dart/lib/src/service/luci_build_service.dart
+++ b/app_dart/lib/src/service/luci_build_service.dart
@@ -416,6 +416,7 @@ class LuciBuildService {
         'user_agent': const <String>['luci-scheduler'],
       },
       properties: <String, String>{
+        'branch': 'refs/heads/$branch',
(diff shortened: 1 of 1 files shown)


app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -416,6 +416,7 @@ class LuciBuildService {
         'user_agent': const <String>['luci-scheduler'],
       },
       properties: <String, String>{
+        'branch': 'refs/heads/$branch',
```
godofredoc, line 419: would this be 1.23-xyz or dev|beta|stable?

CaseyHillers: 1.23-xyz, this is for Cocoon, not the flutter tool

godofredoc (approved):

PR #1028, opened 2020-12-10
make github bot wrong merge base test more lenient

christopherfujino:
The former release process advised that branches on a user's mirror should be named the same as the release branch they were going to merge into. The new process advises prefixing their user branch with `cherrypicks-`. Thus, make this test in the github bot more lenient, just requiring both base ref and head ref to match the regex, and not requiring them to be exactly the same.

### DIFF
--- a/app_dart/lib/src/request_handlers/github_webhook.dart
+++ b/app_dart/lib/src/request_handlers/github_webhook.dart
@@ -333,7 +333,7 @@ class GithubWebhook extends RequestHandler<Body> {
       return;
     }
     final RegExp candidateTest = RegExp(r'flutter-\d+\.\d+-candidate\.\d+');
-    if (pr.base.ref == pr.head.ref && candidateTest.hasMatch(pr.head.ref)) {
+    if (candidateTest.hasMatch(pr.base.ref) && candidateTest.hasMatch(pr.head.ref)) {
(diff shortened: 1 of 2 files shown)


CaseyHillers (commented):
Can you add a test to prevent regressions?

christopherfujino:
> Can you add a test to prevent regressions?

will do.

christopherfujino:
> Can you add a test to prevent regressions?

I updated the existing test so that it will fail without my code change, but pass with it.

Also, when I did a `pub get` with `pub` from the latest flutter sdk, it updated the `pubspec.lock` dart sdk constraint. should i check this in?

CaseyHillers:
> Also, when I did a `pub get` with `pub` from the latest flutter sdk, it updated the `pubspec.lock` dart sdk constraint. should i check this in?

Go for it, or feel free to upload as a separate PR.

CaseyHillers (approved):

christopherfujino:
> > Also, when I did a `pub get` with `pub` from the latest flutter sdk, it updated the `pubspec.lock` dart sdk constraint. should i check this in?
> 
> Go for it, or feel free to upload as a separate PR.

Ok, I'll check it in in a follow-up PR.

PR #1031, opened 2020-12-11
Create bat script for device_doctor windows build

keyonghan:
Existing build.sh works for all platforms, but LUCI bot is more comfortable with a `.bat` file for windows.

This PR creates a corresponding `.bat` build file for windows, parallel to the `build.sh` for linux and mac.

### DIFF
--- a/device_doctor/tool/build.bat
+++ b/device_doctor/tool/build.bat
@@ -1,0 +1,29 @@
+@ECHO off
+REM Copyright 2020 The Flutter Authors. All rights reserved.
+REM Use of this source code is governed by a BSD-style license that can be
+REM found in the LICENSE file.
+
+REM Checks if cipd command is available.
+FOR /F "tokens=*" %%g IN ('where cipd') do (SET CIPD=%%g)
+IF %ERRORLEVEL% NEQ 0 (
+        ECHO "Please install CIPD (available from depot_tools) and add to path first.";
+        EXIT
+)
+
+REM `path` is \path\to\device_doctor\tool\
+REM `DIR` is \path\to\device_doctor\tool
+SET path=%~dp0
+SET DIR=%path:~0,-1%
+%CIPD% ensure --ensure-file %path%\ensure_file_windows -root %DIR%
+
+REM `BUILD_DIR` is \path\to\device_doctor
+for %%a in (%DIR:~0,-1%) do set "BUILD_DIR=%%~dpa"
+PUSHD %BUILD_DIR%
+if exist %BUILD_DIR%\build (
+        ECHO "Please remove the build directory before proceeding"
+        EXIT
+)
+MKDIR %BUILD_DIR%\build
+
+tool\dart-sdk\bin\pub.bat get
+tool\dart-sdk\bin\dart2native.bat bin\main.dart -o build\device_doctor.exe
--- a/device_doctor/tool/build.bat
+++ b/device_doctor/tool/build.bat
@@ -2,7 +2,6 @@
 :: Use of this source code is governed by a BSD-style license that can be
 :: found in the LICENSE file.
 
-@ECHO off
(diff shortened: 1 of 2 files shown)


CaseyHillers (commented):
If the bash script works on windows, why are we making a batch version? Duplicating logic adds to the maintenance burden.

keyonghan:
> If the bash script works on windows, why are we making a batch version? Duplicating logic adds to the maintenance burden.

I tried on devicelab windows bots, where the `sh.exe` exists under `C:\Program Files\Git\bin`, which we need to pre-locate before running the script. But I am not sure how `sh.exe` is managed from LUCI windows side.

More importantly, the `.bat` seems is the way to follow from LUCI recipe side.

device_doctor/tool/build.bat
```diff
@@ -1,8 +1,8 @@
-@ECHO off
-REM Copyright 2020 The Flutter Authors. All rights reserved.
-REM Use of this source code is governed by a BSD-style license that can be
-REM found in the LICENSE file.
+:: Copyright 2020 The Flutter Authors. All rights reserved.
+:: Use of this source code is governed by a BSD-style license that can be
+:: found in the LICENSE file.
 
+@ECHO off
```
CaseyHillers, line 5: Why echo off? I find there to be an advantage that magic scripts like this print out their process. This makes it easier for developers to understand why a failure may have occurred based on CI output.

CaseyHillers (approved):
LGTM w/ comment

device_doctor/tool/build.bat
keyonghan: Turn echo on.

device_doctor/tool/build.bat
```diff
@@ -1,8 +1,8 @@
-@ECHO off
-REM Copyright 2020 The Flutter Authors. All rights reserved.
-REM Use of this source code is governed by a BSD-style license that can be
-REM found in the LICENSE file.
+:: Copyright 2020 The Flutter Authors. All rights reserved.
```
godofredoc, line 1: Why do we require scripts?

keyonghan: You mean why we have build.bat/build.sh? They are used to build the tool.

device_doctor/tool/build.bat
```diff
@@ -0,0 +1,28 @@
...
+IF %ERRORLEVEL% NEQ 0 (
+        ECHO "Please install CIPD (available from depot_tools) and add to path first.";
+        EXIT
+)
+
+REM `path` is \path\to\device_doctor\tool\
+REM `DIR` is \path\to\device_doctor\tool
+SET path=%~dp0
+SET DIR=%path:~0,-1%
+%CIPD% ensure --ensure-file %path%\ensure_file_windows -root %DIR%
```
godofredoc, line 16: ensure file should be generic for the different platforms why we need something specific for windows?

device_doctor/tool/build.sh
```diff
@@ -6,7 +6,7 @@
 # Fetches corresponding dart sdk from CIPD for different platforms, builds
 # an executable binary of device_doctor to `build` folder.
 #
-# This currently supports linux, mac and windows.
+# This currently supports linux and mac.
```
godofredoc, line 9: If we implement this logic in the recipe we don't need to branch out scripts for different platforms. How are we getting the dart sdk?

device_doctor/tool/build.bat
keyonghan: Linux and mac are using the `stable` version, but `dart2native` is not available in `stable` yet for windows.

device_doctor/tool/build.sh
keyonghan: Dart sdk is defined in the ensure_file, obtained via cipd.

PR #1032, opened 2020-12-11
Revert "[app_dart] Pass branch when rescheduling prod builds"

CaseyHillers:
Reverts flutter/cocoon#1027

This breaks in prod. flutter/flutter#72007

We need to switch how the recipe pulls the branch (use gitiles instead of the property)


keyonghan (approved):
LGTM

PR #1033, opened 2020-12-11
[app_dart] Sanitize update task status input

CaseyHillers:
The backend is receiving string input with one byte difference from what's in datastore. This cleans the hidden whitespace to match what's in datastore.

Reverts back to the key lookup instead of the brute force scan.

# Issue

Fixes https://github.com/flutter/flutter/issues/71749

# Test

Added unit test with hidden whitespace from the prod example

https://ci.chromium.org/ui/p/flutter/builders/prod/Linux%20web_benchmarks_canvaskit/76/overview


godofredoc (approved):

christopherfujino (approved):
Nice catch

PR #1036, opened 2020-12-14
Add device_doctor android management logic

keyonghan:
This PR implements basic health check: device accessible, and recovery: restart and dismiss dialogs for android devices.


device_doctor/bin/main.dart
```diff
@@ -56,3 +74,23 @@ Future<void> main(List<String> args) async {
       break;
   }
 }
+
+/// Different properties need to be defined for corresponding devices:
+///
+///   ios: `idevice_id`, `idevicediagnostics`
+///   android: `adb`
+void _validateProperties(String deviceOS, Map<String, dynamic> properties) {
```
CaseyHillers, line 82: Properties looks like an implementation detail that should be in the device class, not the CLI tooling.

device_doctor/lib/src/android_device.dart
```diff
@@ -0,0 +1,195 @@
...
+
+import 'device.dart';
+import 'process_helper.dart';
+import 'health.dart';
+import 'utils.dart';
+
+/// Constant battery health values returned from Android Battery Manager.
+///
+/// https://developer.android.com/reference/android/os/BatteryManager.html
+class AndroidBatteryHealth {
```
CaseyHillers, line 19: Does iOS have a similar API? If so, we should see if we can generalize this.

device_doctor/lib/src/android_device.dart
```diff
@@ -0,0 +1,195 @@
...
+
+  // Parses information about a device. Example:
+  //
+  // 015d172c98400a03       device usb:340787200X product:nakasi model:Nexus_7 device:grouper
+  static final RegExp _kDeviceRegex = RegExp(r'^(\S+)\s+(\S+)(.*)');
+
+  static AndroidDeviceDiscovery _instance;
+
+  Future<String> deviceListOutput() async {
+    return eval(properties['adb'], <String>['devices', '-l'], canFail: false).timeout(Duration(seconds: 15));
```
CaseyHillers, line 47: Duration should be stored in a variable to prevent this being a magic number

device_doctor/lib/src/android_device.dart
```diff
@@ -0,0 +1,195 @@
...
+  }
+
+  Future<List<String>> deviceListOutputWithRetries(Duration retriesDelayMs) async {
```
CaseyHillers, line 50: nit: `retriesDelayMs` -> `retriesDelay` since it's a Duration

device_doctor/lib/src/android_device.dart
```diff
@@ -0,0 +1,195 @@
...
+      maxAttempts: 3,
+      delayFactor: retriesDelayMs,
+    );
+    return await r.retry(
+      () async {
+        String result = await deviceListOutput();
+        return result.trim().split('\n');
+      },
+      retryIf: (Exception e) => e is TimeoutException,
+      //onRetry: (Exception e) => killAdbServer(),
```
CaseyHillers, line 61: Should this be implemented?

device_doctor/lib/src/android_device.dart
```diff
@@ -0,0 +1,195 @@
...
+    );
+  }
+
+  void killAdbServer() async {
```
CaseyHillers, line 65: nit: ADB related calls are implementation details to this class and should be hidden

device_doctor/lib/src/android_device.dart
```diff
@@ -0,0 +1,195 @@
...
+    if (Platform.isWindows) {
+      await killAllRunningProcessesOnWindows('adb');
+    } else {
+      await eval(properties['adb'], <String>['kill-server'], canFail: false);
+    }
+  }
+
+  @override
+  Future<List<Device>> discoverDevices({Duration retriesDelayMs = const Duration(seconds: 10)}) async {
```
CaseyHillers, line 74: nit: same here, `retriesDelayMs` -> `retriesDelay`

device_doctor/lib/src/android_device.dart
```diff
@@ -0,0 +1,195 @@
...
+      }
+      checks.add(check);
+      results['android-device-${device.deviceId}'] = checks;
+    }
+    return results;
+  }
+
+  @override
+  Future<void> recoverDevices() async {
+    for (Device device in await discoverDevices()) {
```
CaseyHillers, line 121: Should this be inherited from the parent class?

device_doctor/lib/src/process_helper.dart
```diff
@@ -67,3 +69,50 @@ Future<String> eval(String executable, List<String> arguments,
...
+  while (true) {
+    final pids = runningProcessesOnWindows(processName);
+    if (pids.isEmpty) {
+      return;
+    }
+    for (String pid in pids) {
+      processManager.runSync(<String>['taskkill', '/pid', pid, '/f']);
+    }
+    // Killed processes don't release resources instantenously.
+    await Future<void>.delayed(Duration(seconds: 1));
```
CaseyHillers, line 91: nit: Move to variable to prevent this being a magic number.

device_doctor/lib/src/process_helper.dart
```diff
@@ -67,3 +69,50 @@ Future<String> eval(String executable, List<String> arguments,
...
+/// Splits [from] into lines and selects those that contain [pattern].
+Iterable<String> grep(Pattern pattern, {@required String from}) {
+  return from.split('\n').where((String line) {
+    return line.contains(pattern);
+  });
+}
+
+Future<void> killAllRunningProcessesOnWindows(String processName) async {
+  ProcessManager processManager = LocalProcessManager();
+  while (true) {
```
CaseyHillers, line 82: What if a different user owns the process, and device doctor cannot kill it? The tool will hang here trying to do the impossible. We should limit the tries on this loop

device_doctor/lib/src/process_helper.dart
```diff
@@ -67,3 +69,50 @@ Future<String> eval(String executable, List<String> arguments,
 
   return output.trimRight();
 }
+
+/// Splits [from] into lines and selects those that contain [pattern].
+Iterable<String> grep(Pattern pattern, {@required String from}) {
```
CaseyHillers, line 74: Can you add tests for these util functions?

device_doctor/test/src/fake_android_device.dart
```diff
@@ -0,0 +1,101 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:collection/collection.dart';
+
+import 'package:device_doctor/src/android_device.dart';
+
+class FakeAndroidDeviceDiscovery extends AndroidDeviceDiscovery {
```
CaseyHillers, line 11: Shouldn't the tests be using the real implementation, but injecting the process manager to AndroidDevice?

device_doctor/bin/main.dart
keyonghan: Moved to device.

device_doctor/lib/src/android_device.dart
keyonghan: This is legacy code from agent. We may not need it since all phones are connected to hosts with full battery all the time. Just keep it here for potential future use. For ios case, I would leave the implementation when needed.

device_doctor/lib/src/android_device.dart
keyonghan: Used a variable instead.

device_doctor/lib/src/android_device.dart
keyonghan: Done

device_doctor/lib/src/android_device.dart
keyonghan: Yes!

device_doctor/lib/src/android_device.dart
keyonghan: This needs to be implemented in fake_android_device for testing purpose.

device_doctor/lib/src/android_device.dart
keyonghan: Done

device_doctor/lib/src/android_device.dart
keyonghan: the parent class is defined as an abstract class.

device_doctor/lib/src/process_helper.dart
keyonghan: Done.

device_doctor/lib/src/process_helper.dart
keyonghan: Set tries limit 3.

device_doctor/lib/src/process_helper.dart
keyonghan: Added some tests for functions needing logic checks.

device_doctor/test/src/fake_android_device.dart
keyonghan: This is testing the real implementation by faking the very low level function `deviceListOutput` (migrated from earlier agent).
Injecting `process manager` will need a non-trivial change to existing logics. I would prefer to keep it as of now.

device_doctor/lib/src/device.dart
```diff
@@ -4,22 +4,36 @@
...
 import 'health.dart';
 import 'ios_device.dart';
+import 'utils.dart';
 
 /// Discovers available devices and chooses one to work with.
 abstract class DeviceDiscovery {
   factory DeviceDiscovery(String deviceOs) {
     switch (deviceOs) {
       case 'ios':
+        for (String property in supportedIosProperties) {
```
godofredoc, line 17: Where is supportedIosProperties coming from and why we need this logic?

device_doctor/lib/src/device.dart
```diff
@@ -4,22 +4,36 @@
 
 import 'dart:async';
 
+import 'android_device.dart';
 import 'health.dart';
 import 'ios_device.dart';
+import 'utils.dart';
 
 /// Discovers available devices and chooses one to work with.
 abstract class DeviceDiscovery {
```
godofredoc, line 13: Do we really need this abstract factory when we only support two devices?

device_doctor/lib/src/device.dart
```diff
@@ -4,22 +4,36 @@
...
+        }
+        return AndroidDeviceDiscovery();
       default:
         throw StateError('Unsupported device operating system: $deviceOs');
     }
   }
 
   /// Lists all available devices' IDs.
-  Future<List<Device>> discoverDevices({Duration retriesDelayMs = const Duration(seconds: 10)});
+  Future<List<Device>> discoverDevices({Duration retriesDelay = const Duration(seconds: 10)});
```
godofredoc, line 36: retriesDelaySeconds?

device_doctor/lib/src/device.dart
```diff
@@ -39,11 +53,6 @@ abstract class Device {
   /// Whether the device is asleep.
   Future<bool> isAsleep();
```
godofredoc, line 54: is this and the method above supported in ios?

device_doctor/lib/src/ios_device.dart
```diff
@@ -27,12 +28,12 @@ class IosDeviceDiscovery implements DeviceDiscovery {
   static IosDeviceDiscovery _instance;
 
   @override
-  Future<List<Device>> discoverDevices({Duration retriesDelayMs = const Duration(seconds: 10)}) async {
+  Future<List<Device>> discoverDevices({Duration retriesDelay = const Duration(seconds: 10)}) async {
```
godofredoc, line 31: RetriesDelaySeconds?

device_doctor/lib/src/ios_device.dart
```diff
@@ -73,13 +74,10 @@ class IosDevice implements Device {
   @override
   Future<bool> isAsleep() async => false;
```
godofredoc, line 75: rm is we are hardcoding the value?

device_doctor/lib/src/device.dart
keyonghan: `supportedIosProperties` is defined in utils.dart, including `idevice_id`, idevicediagnostics`. These are the utilities we need for healthcheck and recovery.
From the bot_config when calling device_doctor, we need to inject these properties so that the device_doctor tool know where they are located.

device_doctor/lib/src/process_helper.dart
```diff
@@ -6,12 +6,13 @@ import 'dart:async';
...
 import 'dart:io';
 
+import 'package:meta/meta.dart';
 import 'package:path/path.dart' as path;
 import 'package:process/process.dart';
 
 import 'utils.dart';
 
-List<ProcessInfo> _runningProcesses = <ProcessInfo>[];
+final RegExp _whitespace = RegExp(r'\s+');
```
godofredoc, line 15: Docs why we need this?

device_doctor/lib/src/process_helper.dart
```diff
@@ -67,3 +61,56 @@ Future<String> eval(String executable, List<String> arguments,
...
 }
+
+/// Splits [from] into lines and selects those that contain [pattern].
+Iterable<String> grep(Pattern pattern, {@required String from}) {
+  return from.split('\n').where((String line) {
+    return line.contains(pattern);
+  });
+}
+
+Future<bool> killAllRunningProcessesOnWindows(String processName, {ProcessManager processManager}) async {
```
godofredoc, line 72: Can we move this to a different file? this seems like host related functionality rather than device or generic.

device_doctor/lib/src/process_helper.dart
```diff
@@ -67,3 +61,56 @@ Future<String> eval(String executable, List<String> arguments,
...
+  processManager ??= LocalProcessManager();
+  // Avoid endless loop when a process from a different use exists, and fails
+  // to get killed every try.
+  int tries = 3;
```
godofredoc, line 76: use retry package instead of implementing your own logic?

device_doctor/lib/src/process_helper.dart
```diff
@@ -67,3 +61,56 @@ Future<String> eval(String executable, List<String> arguments,
...
+      processManager.runSync(<String>['taskkill', '/pid', pid, '/f']);
+    }
+    // Killed processes don't release resources instantenously.
+    const Duration delay = Duration(seconds: 1);
+    await Future<void>.delayed(delay);
+  }
+  return false;
+}
+
+List<String> runningProcessesOnWindows(String processName, {ProcessManager processManager}) {
```
godofredoc, line 93: Same as above, different file?

godofredoc: Also do we really need this from a device_doctor tool? or should the tool be renamed to something else to account for host related functionality?

device_doctor/lib/src/utils.dart
```diff
@@ -7,6 +7,17 @@ import 'dart:io';
 import 'package:logging/logging.dart';
 import 'package:path/path.dart' as path;
 
+/// Properties used in device_doctor tool.
```
godofredoc, line 10: Can we document that this tool requires adb and idevice_id and remove this complexity?

device_doctor/lib/src/utils.dart
```diff
@@ -7,6 +7,17 @@ import 'dart:io';
...
+///
+/// {
+///   "adb": "/path/to/adb",
+///   "idevice_id": "/path/to/idevice_id",
+/// }
+Map<String, dynamic> properties;
+
+const List<String> supportedIosProperties = <String>['idevice_id', 'idevicediagnostics'];
```
godofredoc, line 18: Docs what are this used for?

device_doctor/lib/src/ios_device.dart
keyonghan: Done

device_doctor/lib/src/utils.dart
keyonghan: Removed this logic.

device_doctor/lib/src/utils.dart
keyonghan: Removed.
I will document this in a following PR which focus on the REAMED.

device_doctor/lib/src/process_helper.dart
keyonghan: These two functions are used to serve the device health check. They are needed when we first fail to list devices and need a retry
I believe these logics have been added to cover some corner cases.

device_doctor/lib/src/process_helper.dart
keyonghan: retry packages is suitable for retry logics with exception. Here we simply keep killing process if exist.

device_doctor/lib/src/process_helper.dart
keyonghan: Moved device related functionalities to `utils.dart`, and renamed this one to `host_utils.dart`.

device_doctor/lib/src/ios_device.dart
keyonghan: removed.

device_doctor/lib/src/device.dart
keyonghan: Removed.

device_doctor/lib/src/device.dart
keyonghan: Done.

device_doctor/lib/src/device.dart
keyonghan: The abstract class was used to also support `vm` case when no device is there in old cocoon agent. I keep it here to support android and ios. Casey mentioned potential fuchsia device case in the future.
I can refactor to avoid this class if you see any problems.

device_doctor/lib/src/android_device.dart
CaseyHillers: If it's not used, we should remove it :-) Less code is easier to maintain, and worst case we can look up old git refs to view the agent code.

device_doctor/lib/src/android_device.dart
CaseyHillers: In that case, it sounds like you want a dedicated Adb class that you can create a fake for

device_doctor/lib/src/android_device.dart
CaseyHillers: Does it need to be abstract? If there's common code, we can put it there. Other methods can be required to be implemented.

device_doctor/test/src/android_device_test.dart
```diff
@@ -0,0 +1,139 @@
...
+      expect(devices.length, equals(1));
+      expect(devices[0].deviceId, equals('ZY223JQNMR'));
+    });
+
+    test('deviceDiscovery retries', () async {
+      StringBuffer sb = new StringBuffer();
+      sb.writeln('List of devices attached');
+      sb.writeln('ZY223JQNMR      device');
+      deviceDiscovery.outputs = <dynamic>[new TimeoutException('a'), new TimeoutException('b'), sb.toString()];
+      List<Device> devices = await deviceDiscovery.discoverDevices(retriesDelaySeconds: const Duration(seconds: 1));
```
CaseyHillers, line 40: Avoid delays in tests

device_doctor/test/src/fake_android_device.dart
CaseyHillers: Okay I see. In this case, we have AndroidDeviceDiscovery not able to be tested. I would be more comfortable with a refactor to allow testing the processes similar to how Flutter tools does it, but I defer to you to make that call.

device_doctor/lib/src/device.dart
CaseyHillers: Durations can be minutes, hours, and days too. We shouldn't be specifying the time unit in the name. What about `retryDuration`?

device_doctor/test/src/fake_android_device.dart
keyonghan: Injected, and have added a couple of more tests for utils.

device_doctor/test/src/android_device_test.dart
keyonghan: Done.

device_doctor/lib/src/android_device.dart
keyonghan: Yes. The class contains other functions as well which don't share common codes.

device_doctor/lib/src/android_device.dart
keyonghan: Have refactored class to test the whole logic by injecting progress manager, hiding related functionsl

device_doctor/lib/src/android_device.dart
keyonghan: SGTM to remove. Done.

device_doctor/lib/src/device.dart
keyonghan: Here, and also in this tool, we are using seconds, which makes it clear about the delay. Any doc suggests not specifying the unit?

CaseyHillers: I can't find anything specific in style guides, but searching the Flutter repo shows this is followed there: https://cs.opensource.google/search?q=Duration&sq=&ss=flutter%2Fflutter

device_doctor/test/src/fake_android_device.dart
CaseyHillers: Thanks!

CaseyHillers (approved):

godofredoc (approved):

device_doctor/lib/src/device.dart
keyonghan: Looking through list of flutter repo, I agree that it makes sense to avoid the unit in the name. Updated it with suggested `retryDuration`.

PR #1041, opened 2020-12-23
Allow skipping goldens on Flutter tests

CaseyHillers:
customer_tests should not be running goldens, as there's no migration strategy. This was surfaced by https://github.com/flutter/flutter/pull/70160#issuecomment-744351890

This allows passing `--update-goldens` in customer_tests to the test runner script.


Piinks:
I don't know that we should make this change. Following up on discord: https://discord.com/channels/608014603317936148/697942891020157058/791428584278589510

Also for reference: https://github.com/flutter/flutter/issues/71437

PR #1043, opened 2020-12-24
Remove the sign_in_button placeholder top padding

xu-baolin:
https://github.com/flutter/flutter/pull/70160
This PR update the `PopupMenuButton` that the text will be automatically centered in the vertical direction, no need to use `Padding`, and it has been communicated with customers.
cc @Piinks @CaseyHillers


CaseyHillers (approved):
LGTM!

It looks like you will need to update the goldens with this PR. Do you have a linux machine you can to do that? Otherwise, I can push to your remote.

xu-baolin:
@CaseyHillers Thank you, my workstation is Windows, so I will need your help with new goldens.
Before the update, we seem to have to discuss the migration steps again.

https://github.com/flutter/flutter/pull/70160#issuecomment-742832181
This is the migration step given by @Piinks. 
Now because this padding does not seem to be applicable, we must modify the cocoon and flutter framework at the same time to smoothly complete the migration, is it right?

xu-baolin:
@CaseyHillers Thanks for your help.

There is another thing I want to consult with your opinion. 
I want to temporarily close this test case(['SignInButton shows avatar when authenticated'](https://github.com/flutter/cocoon/blob/1f316c8fe0a759a2de5692c5b31df6f40fd85907/app_flutter/test/widgets/sign_in_button_test.dart#L76)) because it still blocks https://github.com/flutter/flutter/pull/70160 landing.
I will update the test after the change landing.
CC @Piinks

CaseyHillers:
> I will update the test after the change landing.

Are you proposing:
1. Contribute PR to skip this test case
2. Update flutter/tests with the Cocoon version from (1)
3. Land the impacted change in flutter/flutter
4. Re-enable the test case and update the goldens in a PR
5. Update flutter/tests with the Cocoon version from (4)

If so and this can be done in the next few weeks, SGTM. Please create an issue for tracking.

PR #1044, opened 2021-01-05
Update device doctor tool action to `properties`

keyonghan:
Device doctor tool is using action `properties` in the logic, but the argument is still based on `propertycheck`. This is to fix that.

### DIFF
--- a/device_doctor/bin/main.dart
+++ b/device_doctor/bin/main.dart
@@ -11,7 +11,7 @@ import 'package:device_doctor/device_doctor.dart';
 const String actionFlag = 'action';
 const String deviceOSFlag = 'device-os';
 const String helpFlag = 'help';
-const List<String> supportedOptions = <String>['healthcheck', 'recovery', 'propertycheck'];
+const List<String> supportedOptions = <String>['healthcheck', 'recovery', 'properties'];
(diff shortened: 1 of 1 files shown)


device_doctor/bin/main.dart
```diff
@@ -11,7 +11,7 @@ import 'package:device_doctor/device_doctor.dart';
 const String actionFlag = 'action';
 const String deviceOSFlag = 'device-os';
 const String helpFlag = 'help';
-const List<String> supportedOptions = <String>['healthcheck', 'recovery', 'propertycheck'];
+const List<String> supportedOptions = <String>['healthcheck', 'recovery', 'properties'];
```
CaseyHillers, line 14: nit: Moving these to static variables would allow for reuse, eg `kHealthCheckCommand`

CaseyHillers (approved):
LGTM

device_doctor/bin/main.dart
keyonghan: IIRC, `static` is used to implement class-wide variables and methods. Will extend this to more general like a const file when needed.

CaseyHillers:
Just something to consider, package:args allows you to create a
CommandRunner https://pub.dev/packages/args#dispatching-commands

PR #1045, opened 2021-01-06
Updated references to obsolete Material button widgets

HansMuller:
Updated references to the obsolete Material button widgets:

- FlatButton becomes TextButton
- RaisedButton becomes ElevatedButton
- OutlineButton becomes OutlinedButton

These changes will enable the obsolete widgets to be deprecated: https://github.com/flutter/flutter/pull/73352

More information about the evolution of these classes:

- flutter.dev/go/material-button-migration-guide
- flutter.dev/go/material-button-system-updates

The default appearance of the new button widgets match the current Material spec and are noticeably different from the old ones. For example the home page now looks like this:

Screen Shot 2021-01-06 at 10 59 21 AM


CaseyHillers (approved):
LGTM

`dartfmt -n --line-length=120 app_flutter` should fix the lint issues in CI

PR #1050, opened 2021-01-12
Show which tasks are failing in the status bar

gspencergoog:
When I was gardener last week, I was frustrated several times because Cocoon would say that a task was failing, but all of the tiles on the screen were either green or yellow, so I had to wait a while until something turned red to find out what was broken.

This adds information to the status display that shows which tasks are failing when the build is failing, so that even before any indication is given about the failure, you can tell that it failed.

The main changes in this PR are to add a new query to the server `failed-tasks`, and to display those 
in the UI in `build_dashboard_page.dart`.

One alternative to this approach (or an additional thing that could be done) would be to introduce another color for "failed once, but still in progress" tasks (orange?) that would show up right away.

[This is my first submission to Cocoon in a while, so I can't remember if we do things like check in updated pubspec.lock files, or regenerate generated files.  I did update those, but I can revert.]

### DIFF
--- a/app_dart/lib/src/datastore/cocoon_config.dart
+++ b/app_dart/lib/src/datastore/cocoon_config.dart
@@ -238,14 +238,15 @@ class Config {
     return ServiceAccountCredentials.fromJson(json.decode(rawValue));
   }
 
-  /// The names of autoroller accounts for the repositories.
+  /// The user IDs of autoroller accounts for the repositories.
   ///
   /// These accounts should not need reviews before merging. See
   /// https://github.com/flutter/flutter/wiki/Autorollers
-  Set<String> get rollerAccounts => const <String>{
-        'skia-flutter-autoroll',
-        'engine-flutter-autoroll',
-        'dependabot[bot]',
+  /// Find the ID number via https://api.github.com/users/<username>
+  Set<int> get rollerAccounts => const <int>{
+        37626415, // skia-flutter-autoroll
+        42042535, // engine-flutter-autoroll
+        49699333, // dependabot[bot]
       };
 
   Future<String> generateJsonWebToken() async {
--- a/app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
+++ b/app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
@@ -191,12 +191,12 @@ class CheckForWaitingPullRequests extends ApiRequestHandler<Body> {
       if (utcDate.add(const Duration(hours: 1)).isAfter(DateTime.now().toUtc())) {
         continue;
       }
-      final String author = pullRequest['author']['login'] as String;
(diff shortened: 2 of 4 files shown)


app_dart/lib/src/service/build_status_provider.dart
```diff
@@ -81,14 +59,58 @@ class BuildStatusService {
               /// the build status to fail.
               tasksInProgress[task.name] = false;
             } else if (_isFailed(task) || _isRerunning(task)) {
-              return BuildStatus.failed;
+              failedTasks?.add(task.name);
```
CaseyHillers, line 62: This stops the algorithm from failing fast. Do we know what the performance implications would be?

app_flutter/lib/state/build.dart
```diff
@@ -41,7 +41,10 @@ class BuildState extends ChangeNotifier {
 
   /// Whether or not flutter/flutter currently passes tests.
   bool get isTreeBuilding => _isTreeBuilding;
-  bool _isTreeBuilding;
+  bool _isTreeBuilding = true;
```
CaseyHillers, line 44: Should this be left null?

CaseyHillers (commented):
Thank you Greg for working on this, and the other clean up to Cocoon!

Instead of creating a new handler, have you tried adding this to the existing `/api/public/build-status`?

gspencergoog (commented):
I could move the failed tasks into the build status query, but I kept them separate because it costs extra to build the list of all failing tasks, so I only wanted to do that when it was failing. Of course, the extra round trip could very well be more expensive than the collection of the tasks, so I'm happy to combine them if you think it would be better.

app_flutter/lib/state/build.dart
CaseyHillers: Wouldn't this remove the loading animation for the tree status?

Would explicitly checking `_isTreeBuilding == false` work for the use case below?

CaseyHillers (commented):
> I could move the failed tasks into the build status query, but I kept them separate because it costs extra to build the list of all failing tasks, so I only wanted to do that when it was failing. Of course, the extra round trip could very well be more expensive than the collection of the tasks, so I'm happy to combine them if you think it would be better.

I lean towards merging them into the same endpoint as the most expensive operation is the Datastore query. Running the same query twice for failing trees seems unnecessary. If there's performance issues, there's other places we can improve to get this query faster.

app_flutter/lib/state/build.dart
gspencergoog: Ahh, I didn't consider that.  Yes, I can check below instead.

gspencergoog:
OK, I'll merge the failing tasks into the build status query.

gspencergoog:
I accidentally submitted this off of my fork's master.  I'll submit a new PR off of a branch.

PR #1052, opened 2021-01-12
Show which tasks are failing in the status bar

gspencergoog:
When I was gardener last week, I was frustrated several times because Cocoon would say that a task was failing, but all of the tiles on the screen were either green or yellow, so I had to wait a while until something turned red to find out what was broken.

This adds information to the status display that shows which tasks are failing when the build is failing, so that even before any indication is given about the failure, you can tell that it failed.

The main changes in this PR are to add failed task information to the server response for build-status, and to display those failures in the UI in build_dashboard_page.dart. To do this, I created a BuildResponseProto so that it could return more than just a bool.

One alternative to this approach (or an additional thing that could be done) would be to introduce another color for "failed once, but still in progress" tasks (orange?) that would show up right away.

(I originally accidentally submitted a PR off my fork's master, this is now on a branch in my fork. The original was https://github.com/flutter/cocoon/pull/1050)


gspencergoog:
Still working to address previous comments in https://github.com/flutter/cocoon/pull/1050

gspencergoog:
OK, @CaseyHillers please take a look: this has the failed tasks as part of the build status response.

app_dart/lib/src/service/build_status_provider.dart
```diff
@@ -56,14 +56,15 @@ class BuildStatusService {
       branch: branch,
     ).toList();
     if (statuses.isEmpty) {
-      return BuildStatus.failed;
+      return BuildStatus.success();
```
CaseyHillers, line 59: Should this still be a failure? If Datastore is failing to return the tasks, this is an infra issue that should close the tree.

CaseyHillers (approved):
LGTM when comment is addressed

app_dart/lib/src/service/build_status_provider.dart
gspencergoog: Thanks, that was just an typo.  Fixed.

gspencergoog:
OK, so once this is committed, do I have to deploy it, or will it get deployed automagically?

CaseyHillers:
> OK, so once this is committed, do I have to deploy it, or will it get deployed automagically?

Automagically :-)

The flutter-dashboard GCP project deploys master 8 AM every weekday.

CaseyHillers:
The license check failure seems to be caused by autogenerated files from `flutter build` wiping the licenses. This is probably just a change in behavior where we need to update the gitignore with the latest from flutter/flutter.

gspencergoog:
No, it's because this is the first time that a generated `.pbenum.dart` file has been added. I fixed the licenses script.

PR #1054, opened 2021-01-13
Auto rerun framework Mac builders with infra failure

keyonghan:
This is a workaround for issue https://github.com/flutter/flutter/issues/68322.

An alternative is to filter `-9 recode` via `Build` property `summaryMarkdown`. However `-9 recode` does not show up in parent `summaryMarkdown` for builders with shards, but exists in sub-builders `summaryMarkdown`. This makes it difficult to correctly filter corresponding builders for rerun.

Since most existing `infra failure` Mac builders are related to `-9 recode` issue, and it doesn't hurt to rerun the small portion of other `infra failure` builders, here we are rerunning all Mac builders which fail with `infra failures`. 

This PR focuses only on framework repo. We need to do the same for engine in a separate PR.

### DIFF
--- a/app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
+++ b/app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
@@ -95,11 +95,8 @@ class RefreshChromebotStatus extends ApiRequestHandler<Body> {
         final Task update = datastoreTask.task;
         update.status = latestLuciTask.status;
 
-        if (_shouldRerunTask(latestLuciTask, update.attempts)) {
-          await luciBuildService.rescheduleProdBuild(
-            commitSha: commitSha,
-            builderName: latestLuciTask.builderName,
-          );
+        if (await luciBuildService.checkRerunBuilder(
+            commitSha: commitSha, luciTask: latestLuciTask, taskAttempts: update.attempts)) {
           update.status = Task.statusNew;
           update.attempts += 1;
         }
@@ -114,15 +111,6 @@ class RefreshChromebotStatus extends ApiRequestHandler<Body> {
         }
       }
     }
-  }
-
-  /// Auto-rerun Mac builders with `infra failure`.
-  ///
-  /// This is a workaround for -9 retcode issue: https://github.com/flutter/flutter/issues/68322.
-  bool _shouldRerunTask(LuciTask luciTask, int attempts) {
-    return luciTask.builderName.contains('Mac') &&
-        luciTask.status == Task.statusInfraFailure &&
-        attempts < config.maxTaskRetries;
   }
 
   Future<void> _insertBigquery(Task task) async {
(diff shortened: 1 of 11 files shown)


dnfield (approved):
LGTM

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
```diff
@@ -80,27 +83,48 @@ class RefreshChromebotStatus extends ApiRequestHandler<Body> {
...
     }
   }
 
+  /// Auto-rerun Mac builders with `infra failure`.
+  ///
+  /// This is a workaround for -9 retcode issue: https://github.com/flutter/flutter/issues/68322.
+  bool _shouldRerunTask(LuciTask luciTask, int attempts) {
+    return luciTask.builderName.contains('Mac') &&
+        luciTask.status == Task.statusInfraFailure &&
+        attempts < config.maxTaskRetries;
```
godofredoc, line 125: I'd limit to one retry only. Also how do we know that we are explicitly re-running failures with exit code -9

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
```diff
@@ -80,27 +83,48 @@ class RefreshChromebotStatus extends ApiRequestHandler<Body> {
...
+          update.attempts += 1;
+        }
+
+        update.buildNumberList = buildNumberList;
+        update.builderName = builder.name;
+        update.luciBucket = 'luci.flutter.prod';
+        await datastore.insert(<Task>[update]);
+        // Save luci task record to BigQuery only when task finishes.
+        if (update.status == Task.statusFailed || update.status == Task.statusSucceeded) {
+          await _insertBigquery(update);
```
godofredoc, line 113: can we move the retry logic to a service class out of the handler.

keyonghan: Moved to `luci_build_service`, so that the following up engine change can share the code. Also added more unit tests.

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
keyonghan: > I'd limit to one retry only. 
Done.
> Also how do we know that we are explicitly re-running failures with exit code -9
Please refer to the PR description. We are rerunning Mac `infra-failure` builds since it covers all `-9` ones, and there is not an easy way to specifically filter out `-9` ones.

godofredoc: What is the plan to prevent hiding real flakes that report themselves as infra_failures? additionally  what is the plan to only retry things that we understand and already triaged? as it is this PR will be adding a lot of technical debt, we can land it if this is temporary and we already have a plan on how to make the retries more specific and configurable.

keyonghan: Yeah, it may hide real flakes if we rerun all infra failure builds.
To avoid potential technical debt, I updated the rerun logic to make it more specific for our -9 use case:
1) rerun all Mac builders (without shards) failing with exact `-9 retcode` in `summaryMarkdown`
2) rerun Mac builders with shards failing with `summaryMarkdown` equal "Step('display builds.build(s) failed') (retcode: 1)".

godofredoc: Awesome, thanks!

godofredoc (approved):

PR #1055, opened 2021-01-15
Fix update of failed tasks list

gspencergoog:
Fixed the status string so that it is calculated inside of the animated builder instead of outside of it.


CaseyHillers (approved):
LGTM!

PR #1056, opened 2021-01-15
Add command line options for switching data source.

gspencergoog:
I added a couple of simple command line options to allow you to run a debug version with production data, and a release version with a local server. They're `--use-production-service` and `--no-use-production-service`. They default to production on release and fake on debug, as before.

In order for the command line options to be passed to the Dart main, I updated the platform shims to the latest versions (I didn't touch web, because I was too afraid that would break something), which accounts for the bulk of this change. You can run the local app this way with `flutter run -a --use-production-service`, for instance. Sadly, for now this only works on non-web platforms.

The only substantive changes in this PR are in app_flutter/lib


app_flutter/lib/main.dart
```diff
@@ -15,9 +16,16 @@ import 'state/index.dart';
 import 'widgets/now.dart';
 import 'widgets/state_provider.dart';
 
-void main() {
+void main(List<String> args) {
   final GoogleSignInService authService = GoogleSignInService();
-  final CocoonService cocoonService = CocoonService();
+  bool useProductionService = kReleaseMode;
+  if (args.contains('--use-production-service')) {
```
CaseyHillers, line 22: Can you document these in `app_flutter/README.md`?

CaseyHillers (approved):
LGTM!

app_flutter/lib/main.dart
gspencergoog: Yes, sure.  I did that, and added a usage message and "--help" option.

PR #1057, opened 2021-01-15
Follow up of luci auto rerurn

keyonghan:
This PR makes two changes:

1. `task.attempts` includes the first run, so here we minus 1 when making rerun trials check.
2. Add `infra failure` check for rerun. There is a corner case: https://ci.chromium.org/ui/p/flutter/builders/prod/Mac%20tool_tests/1389/overview, which has same `summaryMarkdown` as the one with shards -9.

### DIFF
--- a/app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
+++ b/app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
@@ -95,8 +95,9 @@ class RefreshChromebotStatus extends ApiRequestHandler<Body> {
         final Task update = datastoreTask.task;
         update.status = latestLuciTask.status;
 
+        /// Use `update.attempts - 1` as the `retries` to skip the initial run.
         if (await luciBuildService.checkRerunBuilder(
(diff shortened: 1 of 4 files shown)


app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
```diff
@@ -96,7 +96,7 @@ class RefreshChromebotStatus extends ApiRequestHandler<Body> {
         update.status = latestLuciTask.status;
 
         if (await luciBuildService.checkRerunBuilder(
-            commitSha: commitSha, luciTask: latestLuciTask, taskAttempts: update.attempts)) {
+            commitSha: commitSha, luciTask: latestLuciTask, taskAttempts: update.attempts - 1)) {
```
dnfield, line 99: Not: add a comment explaining why -1

dnfield (approved):

app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
keyonghan: Added.

PR #1059, opened 2021-01-16
Allow 'flutter config --no-analytics' to fail

jmagman:
Investigation for flaky `flutter config` call. Allow the command to fail to see if we actually get some real logging.  Not sure why there's no [stdout logging](https://github.com/flutter/cocoon/pull/659) for when this fails.  

https://github.com/flutter/flutter/issues/65413

### DIFF
--- a/agent/lib/src/utils.dart
+++ b/agent/lib/src/utils.dart
@@ -516,7 +516,7 @@ Future<Null> getFlutter(String revision) async {
     await exec('git', ['checkout', revision]);
   });
 
-  await flutter('config', options: ['--no-analytics', '--verbose']);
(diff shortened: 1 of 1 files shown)


CaseyHillers (approved):
LGTM. Should this be added to LUCI as well? https://cs.opensource.google/search?q=%22flutter%20config%22&ss=flutter%2Frecipes

jmagman:
> LGTM. Should this be added to LUCI as well? https://cs.opensource.google/search?q=%22flutter%20config%22&ss=flutter%2Frecipes

Maybe, if it also flakes there.  I was investigating large_image_changer_perf_ios as a top devicelab flake https://github.com/flutter/flutter/issues/60217#issuecomment-761265061

PR #1061, opened 2021-01-19
Upgrade gcloud to ^0.7.3

liyuqian:
This unblocks https://github.com/flutter/cocoon/pull/1060.

PR #1060, opened 2021-01-19
Add a script to manually transfer Cocoon metrics

liyuqian:
Some part of this PR's code is from https://github.com/flutter/cocoon/pull/788.

Related issue: https://github.com/flutter/flutter/issues/73872

PR #1062, opened 2021-01-19
Let Cocoon write into metrics center

liyuqian:
This needs https://github.com/flutter/cocoon/pull/1061 and the `metrics_center` change in `pubspec.yaml` from https://github.com/flutter/cocoon/pull/1060.

Once this PR is successfully deployed, we should be able to cut the dependency on Tong's desktop.

### DIFF
--- a/app_dart/lib/src/datastore/cocoon_config.dart
+++ b/app_dart/lib/src/datastore/cocoon_config.dart
@@ -16,6 +16,7 @@ import 'package:googleapis_auth/auth.dart';
 import 'package:graphql/client.dart' hide Cache;
 import 'package:http/http.dart' as http;
 import 'package:meta/meta.dart';
+import 'package:metrics_center/metrics_center.dart';
 
 import '../../cocoon_service.dart';
 import '../foundation/providers.dart';
@@ -337,6 +338,12 @@ class Config {
   Future<GithubService> createGithubService(String owner, String repository) async {
     final GitHub github = await createGitHubClient(owner, repository);
     return GithubService(github);
+  }
+
+  Future<FlutterDestination> createMetricsDestination() async {
+    return await FlutterDestination.makeFromCredentialsJson(
+      (await deviceLabServiceAccount).toJson(),
+    );
   }
 
   bool githubPresubmitSupportedRepo(String repositoryName) {
--- a/app_dart/lib/src/datastore/cocoon_config.dart
+++ b/app_dart/lib/src/datastore/cocoon_config.dart
@@ -340,6 +340,12 @@ class Config {
     return GithubService(github);
   }
 
+  /// Return a [FlutterDestination] (subclass of [MetricsDestination]) for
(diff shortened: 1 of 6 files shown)


liyuqian:
@keyonghan : I think this is now ready for review as https://github.com/flutter/engine/pull/23767 has been successful on the engine side.

app_dart/lib/src/datastore/cocoon_config.dart
```diff
@@ -333,6 +334,12 @@ class Config {
     return GithubService(github);
   }
 
+  Future<FlutterDestination> createMetricsDestination() async {
```
keyonghan, line 337: can we add a doc briefly explain Metrics/flutter destination?

app_dart/lib/src/request_handlers/update_task_status.dart
```diff
@@ -122,6 +127,35 @@ class UpdateTaskStatus extends ApiRequestHandler<UpdateTaskStatusResponse> {
     return UpdateTaskStatusResponse(task);
   }
 
+  Future<void> _writeToMetricsCenter(
```
keyonghan, line 130: ditto.

keyonghan (approved):
LGTM with doc nits.

app_dart/lib/src/datastore/cocoon_config.dart
liyuqian: Done.

app_dart/lib/src/request_handlers/update_task_status.dart
liyuqian: Done.

liyuqian (commented):
Thanks for the review! Docs added :)

app_dart/lib/src/request_handlers/update_task_status.dart
```diff
@@ -105,7 +105,12 @@ class UpdateTaskStatus extends ApiRequestHandler<UpdateTaskStatusResponse> {
       await _insertBigquery(commit, task);
     }
 
-    await _writeToMetricsCenter(resultData, scoreKeys, commit, task);
+    try {
+      await _writeToMetricsCenter(resultData, scoreKeys, commit, task);
+    } catch (e) {
```
keyonghan, line 110: From the log, the error is `DetailedApiRequestError`, do you think we can use the exact error?

app_dart/lib/src/request_handlers/update_task_status.dart
```diff
@@ -105,7 +105,12 @@ class UpdateTaskStatus extends ApiRequestHandler<UpdateTaskStatusResponse> {
...
+      log.info(e.toString());
```
keyonghan, line 111: can we use log.error?

liyuqian: Sure, done.

PR #1064, opened 2021-01-21
Write healthcheck and properties to file instead of using stdout/stderr

keyonghan:
This is to write query results to a local file, avoiding issues from unexpected stdout/stderr messages.

Part of https://github.com/flutter/flutter/issues/74363

### DIFF
--- a/device_doctor/lib/src/health.dart
+++ b/device_doctor/lib/src/health.dart
@@ -82,21 +82,18 @@ class HealthCheckResult {
 }
 
 /// Check healthiness for discovered devices.
-///
-/// If any failed check, an explanation message will be sent to stdout and
-/// an exception will be thrown.
-Future<void> healthcheck(Map<String, List<HealthCheckResult>> deviceChecks) async {
+Future<Map<String, String>> healthcheck(Map<String, List<HealthCheckResult>> deviceChecks) async {
+  final Map<String, String> healthcheckMap = <String, String>{};
   if (deviceChecks.isEmpty) {
-    stderr.writeln('No healthy device is available');
+    healthcheckMap['Attached device'] = 'No device is available';
   }
   for (String deviceID in deviceChecks.keys) {
     List<HealthCheckResult> checks = deviceChecks[deviceID];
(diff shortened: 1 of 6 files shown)


yusufm (approved):

device_doctor/lib/src/health.dart
```diff
@@ -82,21 +82,18 @@ class HealthCheckResult {
...
 /// Check healthiness for discovered devices.
-///
-/// If any failed check, an explanation message will be sent to stdout and
-/// an exception will be thrown.
-Future<void> healthcheck(Map<String, List<HealthCheckResult>> deviceChecks) async {
+Future<Map<String, String>> healthcheck(Map<String, List<HealthCheckResult>> deviceChecks) async {
+  final Map<String, String> healthcheckMap = <String, String>{};
   if (deviceChecks.isEmpty) {
-    stderr.writeln('No healthy device is available');
+    healthcheckMap['Attached device'] = 'No device is available';
```
yusufm, line 88: can you make the key and value be constants here too, like you did for other keys. And the key be a little more machine friendly, like:

attached_device

?

device_doctor/lib/src/utils.dart
```diff
@@ -11,6 +11,8 @@ import 'package:process/process.dart';
 
 import 'dart:convert' show utf8;
 
+const String kDevicePropertiesFilename = '.properties';
+const String kDeviceFailedHealthcheckFilename = '.failedhealthcheck';
```
yusufm, line 15: It might be easier to handle a .healthcheck file whose contents are if it succeeded or failed, and any additional info in future, rather than a negative file, which is specific for a certain case only.

yusufm (approved):
LGTM
Added a couple of comments, that might be good to consider for this or a future CL.

device_doctor/lib/src/utils.dart
keyonghan: Created a follow up PR: https://github.com/flutter/cocoon/pull/1065/files to include status and details.

PR #1066, opened 2021-01-22
Use Dart dev instead of stable version for cocoon deployment

keyonghan:
This is to fix: https://github.com/flutter/flutter/issues/74430 to unblock Cocoon auto-build.


dnfield (approved):

liyuqian (approved):
LGTM

PR #1068, opened 2021-01-25
Fix buildbucket grpc errors.

godofredoc:
The results extraction process runs a single query to retrieve 40 builds
for +200 builders making the grpc service to fail with 500 errors. This
pr changes the logic to send several requests for <50 builders and
aggregates the data at the end.

Bug: https://github.com/flutter/flutter/issues/74536

### DIFF
--- a/app_dart/lib/src/service/luci.dart
+++ b/app_dart/lib/src/service/luci.dart
@@ -21,6 +21,10 @@ import 'buildbucket.dart';
 part 'luci.g.dart';
 
 const int _maxResults = 40;
+
+/// The batch size used to query buildbucket service.
+const int _buildersBatchSize = 50;
+
 const Map<Status, String> luciStatusToTaskStatus = <Status, String>{
   Status.unspecified: Task.statusInProgress,
   Status.scheduled: Task.statusInProgress,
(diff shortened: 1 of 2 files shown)


app_dart/lib/src/service/luci.dart
```diff
@@ -63,7 +66,23 @@ class LuciService {
   }) async {
     assert(requireTaskName != null);
     final List<LuciBuilder> builders = await LuciBuilder.getProdBuilders(repo, config);
-    final Iterable<Build> builds = await getBuilds(repo, requireTaskName, builders);
+    final List<Build> builds = <Build>[];
+
+    // Get builds data is batches of 50 builders also add retries to the operations.
```
keyonghan, line 71: typo `in`?

app_dart/lib/src/service/luci.dart
```diff
@@ -63,7 +66,23 @@ class LuciService {
...
+    const RetryOptions r = RetryOptions(maxAttempts: 3);
+    for (int j = 0; j < builders.length; j += 50) {
```
keyonghan, line 73: can we use a variable like `buildBatchSize` instead of fixed 50 here and 49 below?

app_dart/lib/src/service/luci.dart
```diff
@@ -102,7 +121,23 @@ class LuciService {
...
+      await r.retry(
+        () async {
+          final Iterable<Build> partialBuilds = await getBuilds(repo, requireTaskName, partialBuilders);
+          builds.addAll(partialBuilds);
+        },
+        retryIf: (Exception e) => e is BuildBucketException,
+      );
+      // Wait in between requests to prevent rate limiting.
+      final Random random = Random();
+      await Future<dynamic>.delayed(Duration(seconds: random.nextInt(10)));
```
keyonghan, line 139: Nit: move this logic to a shared function?

keyonghan (approved):
LGTM with nits.

app_dart/lib/src/service/luci.dart
godofredoc: @keyonghan Actually I see the above function called from a single place would it be better to just remove it and updated https://cs.opensource.google/flutter/cocoon/+/master:app_dart/lib/src/request_handlers/push_engine_status_to_github.dart;l=51?q=getRecentTasks&ss=flutter%2Fcocoon to use the branched version.

keyonghan: Update `push_engine_status_to_github` may not be an easy change.
`getRecentTasks` returns builds of a builder in descending order, which makes it easy to get the tot status for engine.
whereas `getBranchRecentTasks` returns a BranchLuciBuilder based map, which is then based on commit, and then LUCI tasks. The tot status of each builder needs a second calculation.

christopherfujino (commented):
So the problem was the size of the request, rather than rate limiting?

app_dart/lib/src/service/luci.dart
godofredoc: This part of the code was removed.

app_dart/lib/src/service/luci.dart
godofredoc: done!

app_dart/lib/src/service/luci.dart
godofredoc: Refactored to a method.

godofredoc:
> So the problem was the size of the request, rather than rate limiting?

Potentially the size of the request, the size of the response and rate limiting with the new implementation.

keyonghan (approved):
LGTM

PR #1071, opened 2021-01-26
Run manual retries with higher priority to open the tree faster.

godofredoc:
Manual retries are usually waiting for ~30 mins as they are added to the
end of the queue of tasks for the current commit. We are changing the
priorities to ensure retried tasks are not waiting for too long in the
queue.

Bug: https://github.com/flutter/flutter/issues/74692


jmagman (commented):
Love the idea, but I don't know the details about the magic number 29, so someone who knows this better than I do should review.

keyonghan (approved):
LGTM

christopherfujino (approved):
LGTM

PR #1072, opened 2021-01-26
Auto rerun engine Mac builders with infra failure

keyonghan:
This is a workaround for issue flutter/flutter#68322: auto rerun Mac builders failing with -9 retcode.

This is for post-submit builders only.

### DIFF
--- a/app_dart/lib/src/request_handlers/push_engine_status_to_github.dart
+++ b/app_dart/lib/src/request_handlers/push_engine_status_to_github.dart
@@ -17,18 +17,21 @@ import '../request_handling/body.dart';
 import '../service/buildbucket.dart';
 import '../service/datastore.dart';
 import '../service/luci.dart';
+import '../service/luci_build_service.dart';
 
 @immutable
 class PushEngineStatusToGithub extends ApiRequestHandler<Body> {
   const PushEngineStatusToGithub(
     Config config,
-    AuthenticationProvider authenticationProvider, {
+    AuthenticationProvider authenticationProvider,
+    this.luciBuildService, {
     @visibleForTesting LuciServiceProvider luciServiceProvider,
     @visibleForTesting DatastoreServiceProvider datastoreProvider,
   })  : luciServiceProvider = luciServiceProvider ?? _createLuciService,
         datastoreProvider = datastoreProvider ?? DatastoreService.defaultProvider,
         super(config: config, authenticationProvider: authenticationProvider);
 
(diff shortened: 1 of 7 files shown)


app_dart/lib/src/request_handlers/push_engine_status_to_github.dart
```diff
@@ -50,18 +53,18 @@ class PushEngineStatusToGithub extends ApiRequestHandler<Body> {
...
 
-    String latestStatus = GithubBuildStatusUpdate.statusSuccess;
+    String status = GithubBuildStatusUpdate.statusSuccess;
     for (List<LuciTask> tasks in luciTasks.values) {
-      latestStatus = _getLatestStatus(tasks);
-      if (latestStatus == GithubBuildStatusUpdate.statusFailure) {
-        break;
+      final String latestStatus = await _getLatestStatus(tasks);
+      if (status == GithubBuildStatusUpdate.statusSuccess && latestStatus == GithubBuildStatusUpdate.statusFailure) {
+        status = GithubBuildStatusUpdate.statusFailure;
```
godofredoc, line 60: Why are we overriding status?

app_dart/lib/src/request_handlers/push_engine_status_to_github.dart
```diff
@@ -112,6 +115,11 @@ class PushEngineStatusToGithub extends ApiRequestHandler<Body> {
         case Task.statusFailed:
         case Task.statusInfraFailure:
           log.debug('Using ${task.status} from commit ${task.commitSha} ref ${task.ref} builder ${task.builderName}');
+          await luciBuildService.checkRerunBuilder(
```
godofredoc, line 118: Let's log that we are retrying the task.

godofredoc (approved):
LGTM with some nits.

app_dart/lib/src/request_handlers/push_engine_status_to_github.dart
keyonghan: Added to `luci_build_service` when rerun is actually needed.

app_dart/lib/src/request_handlers/push_engine_status_to_github.dart
keyonghan: The status will only be overridden once: when latest build of any builder returns `failure`. Once overridden from `success` to `failure`, it will not be updated any more.

PR #1076, opened 2021-02-02
Fix engine luci auto rerun by setting logger

keyonghan:
This is a follow up of https://github.com/flutter/cocoon/pull/1072

Also increased the retry limit to 2.

Related issue: https://github.com/flutter/flutter/issues/68322

PR #1078, opened 2021-02-03
Stop device_doctor writing results to stdout

keyonghan:
This is a post cleanup step following https://github.com/flutter/cocoon/pull/1064, which starts writing results to a local file.

Related issue: https://github.com/flutter/flutter/issues/74363

### DIFF
--- a/device_doctor/lib/src/android_device.dart
+++ b/device_doctor/lib/src/android_device.dart
@@ -118,12 +118,14 @@ class AndroidDeviceDiscovery implements DeviceDiscovery {
     Map<String, String> properties = <String, String>{};
     if (devices.isEmpty) {
       await writeToFile(json.encode(properties), _outputFilePath);
+      logger.info('No device is available.');
       return properties;
     }
     properties = await getDeviceProperties(devices[0], processManager: processManager);
     final String propertiesJson = json.encode(properties);
 
     await writeToFile(propertiesJson, _outputFilePath);
+    logger.info('Properties for deviceID ${devices[0].deviceId}: $propertiesJson');
     return properties;
   }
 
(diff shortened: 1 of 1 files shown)


keyonghan:
Test was fixed in https://github.com/flutter/cocoon/pull/1076, waiting for that PR to land first.

device_doctor/lib/src/android_device.dart
```diff
@@ -117,7 +117,6 @@ class AndroidDeviceDiscovery implements DeviceDiscovery {
     final String propertiesJson = json.encode(properties);
 
     await writeToFile(propertiesJson, _outputFilePath);
-    stdout.write(propertiesJson);
```
yusufm, line 120: Would printing this out to stdout still be helpful? Maybe as a log message?

yusufm (approved):
LGTM, with a comment for consideration.

device_doctor/lib/src/android_device.dart
keyonghan: Since writeTofile does exactly the same thing as the stdout.write, and from the bot_config side, we are not utilizing the stdout any more, this line doesn't help much.
Would prefer to do the cleanup now. If we need this at any point in the future, we can always add it back.

godofredoc: Maybe a logging.info? if the properties are unexpected we won't have any way to verify the problem.

keyonghan: We only have a client side call to the device_doctor, namely no server is running. I am not sure where the logging.info will be logged to?

From bot_config side, if the process fails, we catch the [os_err](https://chrome-internal.googlesource.com/infradata/config/+/refs/heads/master/configs/chromium-swarm/scripts/flutter_mac_android_staging.py#157); if the file parse fails, we catch [io_err](https://chrome-internal.googlesource.com/infradata/config/+/refs/heads/master/configs/chromium-swarm/scripts/flutter_mac_android_staging.py#175) and [value_err](https://chrome-internal.googlesource.com/infradata/config/+/refs/heads/master/configs/chromium-swarm/scripts/flutter_mac_android_staging.py#179). Do we still miss anything?

godofredoc: The reasoning behind printing or logging the output is that if something goes unexpectedly we can still see the output from /var/log/system or log system in mac even after the file was deleted from the bot.

godofredoc (approved):
LGTM after adding logging the data.

device_doctor/lib/src/android_device.dart
keyonghan: Added. 
Also fixed a corner case of the getDeviceProperties.

godofredoc:
Awesome, thanks! LGTM.

PR #1079, opened 2021-02-04
Add always visible scrollbars to build page

shihaohong:
I found it difficult to use the trackpad to traverse the build page. This PR adds always visible scrollbars for both vertical and horizontal scrolling.


CaseyHillers (approved):
LGTM! Goldens may need to be updated

CaseyHillers:
I filed https://github.com/flutter/flutter/issues/75373 as it looks like Cocoon try is failing due to an unrelated error

shihaohong:
We should wait until https://github.com/flutter/flutter/issues/75373 is resolved right?

PR #1082, opened 2021-02-08
Add cert check and device pair check to device doctor for ios device

keyonghan:
This adds certificate check and device pair check for ios device.

Related issue: flutter/flutter#74960


device_doctor/lib/src/ios_device.dart
```diff
@@ -66,7 +68,7 @@ class IosDeviceDiscovery implements DeviceDiscovery {
     }
   }
 
-  Future<HealthCheckResult> _keychainUnlockCheck({ProcessManager processManager}) async {
+  Future<HealthCheckResult> keychainUnlockCheck({ProcessManager processManager}) async {
```
jmagman, line 71: Should these checks have `@visibleForTesting` decoration?

jmagman (approved):
LGTM!

device_doctor/lib/src/ios_device.dart
```diff
@@ -76,6 +78,39 @@ class IosDeviceDiscovery implements DeviceDiscovery {
     }
     return healthCheckResult;
   }
+
+  Future<HealthCheckResult> certCheck({ProcessManager processManager}) async {
+    HealthCheckResult healthCheckResult;
+    try {
+      final String certCheckResult =
+          await eval('security', <String>['find-identity', '-p', 'codesigning', '-v'], processManager: processManager);
+      if (certCheckResult.contains('Apple Development: Flutter Devicelab') &&
```
godofredoc, line 89: Would it make sense to validate using the certificate code rather the names?

device_doctor/lib/src/ios_device.dart
```diff
@@ -76,6 +78,39 @@ class IosDeviceDiscovery implements DeviceDiscovery {
...
+          certCheckResult.contains('1 valid identities found')) {
+        healthCheckResult = HealthCheckResult.success(kCertCheckKey);
+      } else {
+        healthCheckResult = HealthCheckResult.failure(kCertCheckKey, certCheckResult);
+      }
+    } on BuildFailedError catch (error) {
```
godofredoc, line 95: What if the command execution fails? would this catch that error?

device_doctor/lib/src/ios_device.dart
```diff
@@ -76,6 +78,39 @@ class IosDeviceDiscovery implements DeviceDiscovery {
...
+    HealthCheckResult healthCheckResult;
+    try {
+      final String devicePairCheckResult =
+          await eval('idevicepair', <String>['validate'], processManager: processManager);
+      if (devicePairCheckResult.contains('SUCCESS')) {
+        healthCheckResult = HealthCheckResult.success(kDevicePairCheckKey);
+      } else {
+        healthCheckResult = HealthCheckResult.failure(kDevicePairCheckKey, devicePairCheckResult);
+      }
+    } on BuildFailedError catch (error) {
```
godofredoc, line 112: Is BuildFailedError the only exception raised by eval?

device_doctor/test/src/ios_device_test.dart
```diff
@@ -42,31 +44,119 @@ void main() {
...
+    test('Keychain unlock check - success', () async {
+      when(processManager.start(any, workingDirectory: anyNamed('workingDirectory')))
+          .thenAnswer((_) => Future.value(process));
+      process = FakeProcess(0);
+      HealthCheckResult healthCheckResult = await deviceDiscovery.keychainUnlockCheck(processManager: processManager);
+      expect(healthCheckResult.succeeded, true);
     });
 
-    test('checkDevices with unhealthy device - unlock keychain failed', () async {
+    test('Keychain unlock check - exeption', () async {
```
godofredoc, line 82: nit s/exeption/exception

device_doctor/test/src/ios_device_test.dart
```diff
@@ -42,31 +44,119 @@ void main() {
...
+      sb.writeln('hijklmn');
+      output = <List<int>>[utf8.encode(sb.toString())];
+      process = FakeProcess(0, out: output);
+      HealthCheckResult healthCheckResult = await deviceDiscovery.certCheck(processManager: processManager);
+      expect(healthCheckResult.succeeded, false);
+      expect(healthCheckResult.name, kCertCheckKey);
+      expect(healthCheckResult.details, sb.toString().trim());
+    });
+
+    test('Cert check - exeption', () async {
```
godofredoc, line 118: nit here and below exception?

device_doctor/lib/src/ios_device.dart
keyonghan: Good point. Added.

device_doctor/test/src/ios_device_test.dart
```diff
@@ -42,31 +44,119 @@ void main() {
...
+      sb.writeln('abcdefg');
+      output = <List<int>>[utf8.encode(sb.toString())];
+      process = FakeProcess(0, out: output);
+      HealthCheckResult healthCheckResult = await deviceDiscovery.devicePairCheck(processManager: processManager);
+      expect(healthCheckResult.succeeded, false);
+      expect(healthCheckResult.name, kDevicePairCheckKey);
+      expect(healthCheckResult.details, sb.toString().trim());
+    });
+
+    test('Device pair check - exeption', () async {
```
godofredoc, line 152: Can we add a test with more than one valid certificate?

device_doctor/lib/src/ios_device.dart
keyonghan: Do we have the case where certificate code is changed while we maintain the certificate name?

device_doctor/lib/src/ios_device.dart
keyonghan: both the command execution failure and the exitcode not equal to 0 will be returned as BuildFailedError.

device_doctor/lib/src/ios_device.dart
keyonghan: Yes. We have wrapped any error there as BuildFailedError.

device_doctor/test/src/ios_device_test.dart
keyonghan: Done

device_doctor/test/src/ios_device_test.dart
keyonghan: Added.

device_doctor/test/src/ios_device_test.dart
keyonghan: Done

godofredoc (approved):

PR #1085, opened 2021-02-13
[app_dart] Track GitHub API quota usage

CaseyHillers:
* Add `/api/public/github-rate-limit-status` to track all supported Flutter repo API usage
  * Add 1 minute cron job to push to BigQuery (metric collection)
* Clean up of imports/exports in `bin/server.dart`

### DIFF
--- a/app_dart/lib/src/request_handlers/github_rate_limit_status.dart
+++ b/app_dart/lib/src/request_handlers/github_rate_limit_status.dart
@@ -16,17 +16,19 @@ import '../service/github_service.dart';
 ///
 /// This endpoint pushes data to BigQuery for metric collection to analyze usage over time.
 @immutable
-class GithubQuotaStatus extends RequestHandler<Body> {
-  const GithubQuotaStatus(
+class GithubRateLimitStatus extends RequestHandler<Body> {
+  const GithubRateLimitStatus(
     Config config,
   ) : super(config: config);
 
   @override
   Future<Body> get() async {
     final List<Map<String, dynamic>> totalQuotaUsage = <Map<String, dynamic>>[];
+
+    /// Pull quota usage for all supported repos.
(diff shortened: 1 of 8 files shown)


app_dart/lib/src/request_handlers/github_rate_limit_status.dart
```diff
@@ -0,0 +1,42 @@
...
+
+import 'package:meta/meta.dart';
+
+import '../datastore/cocoon_config.dart';
+import '../foundation/utils.dart';
+import '../request_handling/body.dart';
+import '../request_handling/request_handler.dart';
+import '../service/github_service.dart';
+
+/// Endpoint to collect the current GitHub API quota usage of Cocoon on all supported Flutter repos.
```
godofredoc, line 15: Can we add documentation on the format of the data being collected? I'm particularly interested on how are we managing overlapping windows.

CaseyHillers: Added documentation on the BQ entry format. Overlapping windows is managed by locking this behind a CacheRequestHandler to ensure its only run once a minute.

In general, BQ documentation in Cocoon is lacking. I created https://github.com/flutter/flutter/issues/75959 to track.

godofredoc: Do we still need to align the windows with the server? usually the APIs provide the necessary information in the form of headers.

godofredoc (approved):

app_dart/lib/src/request_handlers/github_rate_limit_status.dart
CaseyHillers: There's an explicit rate limit api call that's being used here. I opted for this as it was simpler and didn't see a need for monitoring more fine grained than a minute.

PR #1086, opened 2021-02-16
Override path_provider

stuartmorgan:
`localstorage` depends on `path_provider`'s stable version, but due to breaking changes in FFI the stable version `path_provider_windows` (and thus `path_provider`) no longer work with the latest version of Dart. This overrides that transitive dependency with one that is compatible with with Dart ToT.


zanderso:
https://github.com/flutter/flutter/issues/76129

zanderso (approved):

CaseyHillers (approved):

PR #1087, opened 2021-02-16
[app_dart] Create scheduler service

CaseyHillers:
Refactors logic in `refresh_github_commits` into a service

# Issues

# Future work
* Trigger adding tasks from github webhooks
  * Remove the cron method
* Move retry logic into this service

### DIFF
--- a/app_dart/lib/src/request_handlers/refresh_github_commits.dart
+++ b/app_dart/lib/src/request_handlers/refresh_github_commits.dart
@@ -4,7 +4,6 @@
 
 import 'dart:async';
 
-import 'package:cocoon_service/src/service/scheduler.dart';
 import 'package:gcloud/db.dart';
 import 'package:github/github.dart';
 import 'package:meta/meta.dart';
@@ -21,6 +20,7 @@ import '../request_handling/authentication.dart';
 import '../request_handling/body.dart';
 import '../service/datastore.dart';
 import '../service/github_service.dart';
+import '../service/scheduler.dart';
 
 /// Queries GitHub for the list of recent commits according to different branches,
 /// and creates corresponding rows in the cloud datastore and the BigQuery for any commits
--- a/app_dart/lib/src/request_handlers/refresh_github_commits.dart
+++ b/app_dart/lib/src/request_handlers/refresh_github_commits.dart
@@ -50,7 +50,12 @@ class RefreshGithubCommits extends ApiRequestHandler<Body> {
     final GithubService githubService = await config.createGithubService(slug.owner, slug.name);
     final DatastoreService datastore = datastoreProvider(config.db);
 
-    final Scheduler scheduler = Scheduler(config: config, datastore: datastore, httpClient: httpClientProvider(), gitHubBackoffCalculator: gitHubBackoffCalculator, log: log);
+    final Scheduler scheduler = Scheduler(
+        config: config,
+        datastore: datastore,
+        httpClient: httpClientProvider(),
+        gitHubBackoffCalculator: gitHubBackoffCalculator,
+        log: log);
 
     for (String branch in await config.flutterBranches) {
       final List<Commit> lastProcessedCommit = await datastore.queryRecentCommits(limit: 1, branch: branch).toList();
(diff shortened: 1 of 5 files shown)


app_dart/lib/src/service/scheduler.dart
```diff
@@ -0,0 +1,224 @@
...
+import '../request_handling/exceptions.dart';
+import '../service/luci.dart';
+import 'datastore.dart';
+
+/// Scheduler service to validate all commits to supported Flutter repositories.
+///
+/// Scheduler responsibilties include:
+///   1. Tracking commits in Cocoon
+///   2. Ensuring commits are validated (via scheduling tasks against commits)
+///   3. Retry mechanisms for taskss
```
keyonghan, line 28: nit: typo

app_dart/lib/src/service/scheduler.dart
```diff
@@ -0,0 +1,224 @@
...
+
+    await _uploadToBigQuery(commit);
+  }
+
+  /// Return subset of [commits] not stored in Datastore.
+  Future<List<Commit>> _getNewCommits(List<Commit> commits) async {
+    final List<Commit> newCommits = <Commit>[];
+    // Ensure commits are sorted from newest to oldest
+    commits.sort((Commit a, Commit b) => a.timestamp.compareTo(b.timestamp));
+    print(commits);
```
keyonghan, line 81: Remove?

app_dart/lib/src/service/scheduler.dart
```diff
@@ -0,0 +1,224 @@
...
+    });
+
+    try {
+      await datastore.withTransaction<void>((Transaction transaction) async {
+        transaction.queueMutations(inserts: tasks);
+        await transaction.commit();
+        log.debug('Committed ${tasks.length} new tasks for commit ${commit.sha}');
+      });
+    } catch (error) {
+      log.error('Failed to add commit ${commit.sha}: $error');
```
keyonghan, line 146: This is for tasks update.

app_dart/lib/src/service/scheduler.dart
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    await _uploadToBigQuery(commit);
+  }
+
+  /// Return subset of [commits] not stored in Datastore.
+  Future<List<Commit>> _getNewCommits(List<Commit> commits) async {
+    final List<Commit> newCommits = <Commit>[];
+    // Ensure commits are sorted from newest to oldest
+    commits.sort((Commit a, Commit b) => a.timestamp.compareTo(b.timestamp));
```
keyonghan, line 80: list is already sorted.

app_dart/lib/src/request_handlers/refresh_github_commits.dart
```diff
@@ -75,186 +72,31 @@ class RefreshGithubCommits extends ApiRequestHandler<Body> {
...
-
-      if (newCommits.isEmpty) {
-        // Nothing to do.
-        continue;
-      }
-      log.debug('Found ${newCommits.length} new commits for branch $branch on GitHub');
-
-      // Save [Commit] to BigQuery and create [Task] in Datastore.
-      await _saveData(newCommits, datastore);
+      final List<Commit> recentCommits = await _getRecentCommits(commits, datastore, branch);
```
keyonghan, line 75: Move this logic to scheduler as well to save both time and space (processing `RepositoryCommit` directly)

app_dart/lib/src/service/scheduler.dart
CaseyHillers: This is mostly for tests. I removed documentation. In the next PR, I want to switch this from a list to a single commit and will be irrelevant.

app_dart/lib/src/request_handlers/refresh_github_commits.dart
CaseyHillers: I left this code here as in the follow up PR I want to remove it completely (move to a single commit from a webhook). SG?

keyonghan (approved):
LGTM

godofredoc (commented):
Given that this is new functionality should it be implemented as a separate service detached from cocoon?  ideally this should be a microservice with a well defined api, implementing and then migrating later is more complicated because we need to consider backward compatibility, etc.

PR #1092, opened 2021-02-18
Add adb power service check to for Android device health check

keyonghan:
This adds adb power service health check for android device: make sure the power service is available before reserving/running tests.

Related: https://github.com/flutter/flutter/issues/76130

### DIFF
--- a/device_doctor/lib/src/android_device.dart
+++ b/device_doctor/lib/src/android_device.dart
@@ -96,6 +96,7 @@ class AndroidDeviceDiscovery implements DeviceDiscovery {
     for (AndroidDevice device in await discoverDevices(processManager: processManager)) {
       final List<HealthCheckResult> checks = <HealthCheckResult>[];
       checks.add(HealthCheckResult.success('device_access'));
+      checks.add(await adbPowerServiceCheck(processManager: processManager));
       results['android-device-${device.deviceId}'] = checks;
     }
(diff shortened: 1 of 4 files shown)


device_doctor/lib/src/android_device.dart
```diff
@@ -151,6 +152,23 @@ class AndroidDeviceDiscovery implements DeviceDiscovery {
...
     }
   }
+
+  @visibleForTesting
+  Future<HealthCheckResult> adbPowerServiceCheck({ProcessManager processManager}) async {
+    HealthCheckResult healthCheckResult;
+    try {
+      final String result =
+          await eval('adb', <String>['shell', 'dumpsys', '-l', '|', 'grep', 'power\$'], processManager: processManager);
+      if (result.trim() == 'power') {
```
godofredoc, line 162: Why we need to check for a string, is the exit code not enough in this situation?

keyonghan: I am not sure if there are cases where other services are available, whereas the `power` service is not up. For this case, it will return 0, but with null value.

keyonghan: Updated to run `adb shell dympsys power` directly. This way, it relies on exit code to determine success or failure.

godofredoc: Awesome, thanks!

godofredoc (approved):

PR #1093, opened 2021-02-19
[app_dart] Reland scheduler service

CaseyHillers:
See the last commit for the diff between https://github.com/flutter/cocoon/pull/1087

The issue appears to have been the age was backwards in the scheduler. It was only looking to add older commits that did not exist in Datastore. Instead, it should add newer commits that do not exist in Datastore. Simply, larger timestamp commits should get added to datastore.

# Issues

# Future work
* Trigger adding tasks from github webhooks
  * Remove the cron method
* Move retry logic into this service

### DIFF
--- a/app_dart/lib/src/request_handlers/refresh_github_commits.dart
+++ b/app_dart/lib/src/request_handlers/refresh_github_commits.dart
@@ -4,7 +4,6 @@
 
 import 'dart:async';
 
-import 'package:cocoon_service/src/service/scheduler.dart';
 import 'package:gcloud/db.dart';
 import 'package:github/github.dart';
 import 'package:meta/meta.dart';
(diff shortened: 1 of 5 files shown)


keyonghan (commented):
Can we validate the effectiveness by sharing a portion of traffic before merge?

CaseyHillers:
I pushed this with a 50/50 split of traffic. This is now pushing new commits:

Log query:
```
"Found 1 new commits on GitHub"
resource.type="gae_app"
resource.labels.module_id="default"
resource.labels.version_id="testchillers"
```

Logs:
```
Retrieved 2 commits from GitHub
Found 1 new commits on GitHub
Committed commit a46139a2cef8144444a3e13a3970e9a267304b91
Committed 193 new tasks for commit a46139a2cef8144444a3e13a3970e9a267304b91 
```

keyonghan (approved):
LGTM

CaseyHillers:
\cc @yusufm as you'll be on call when this rolls out Monday morning. If the build dashboard is not updating, this can be reverted.

PR #1094, opened 2021-02-19
Add developer mode health check for android device

keyonghan:
This add Android device developer mode health check: make sure it is on before reserving tests.

Related: https://github.com/flutter/flutter/issues/76165

### DIFF
--- a/device_doctor/lib/src/android_device.dart
+++ b/device_doctor/lib/src/android_device.dart
@@ -167,12 +167,16 @@ class AndroidDeviceDiscovery implements DeviceDiscovery {
   }
 
   @visibleForTesting
+  /// The health check for Android device developer mode.
+  /// 
+  /// Developer mode `on` is expected for a healthy Android device.
   Future<HealthCheckResult> developerModeCheck({ProcessManager processManager}) async {
     HealthCheckResult healthCheckResult;
     try {
       final String result = await eval(
           'adb', <String>['shell', 'settings', 'get', 'global', 'development_settings_enabled'],
           processManager: processManager);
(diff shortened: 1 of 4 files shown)


device_doctor/lib/src/android_device.dart
```diff
@@ -164,6 +165,24 @@ class AndroidDeviceDiscovery implements DeviceDiscovery {
     }
     return healthCheckResult;
   }
+
+  @visibleForTesting
```
godofredoc, line 169: Can we add some docs on what is this checking?

device_doctor/lib/src/android_device.dart
```diff
@@ -164,6 +165,24 @@ class AndroidDeviceDiscovery implements DeviceDiscovery {
...
+  Future<HealthCheckResult> developerModeCheck({ProcessManager processManager}) async {
+    HealthCheckResult healthCheckResult;
+    try {
+      final String result = await eval(
+          'adb', <String>['shell', 'settings', 'get', 'global', 'development_settings_enabled'],
+          processManager: processManager);
+      if (result == '1') {
```
godofredoc, line 176: Can you please add a comment describing this is the actual output command not the exit code and 1 means enabled?

godofredoc (approved):
LGTM after fixing some nits

device_doctor/lib/src/android_device.dart
keyonghan: Added.

device_doctor/lib/src/android_device.dart
keyonghan: Added.

PR #1095, opened 2021-02-22
Revert "Revert "Override path_provider""

gw280:
Reverts flutter/cocoon#1089


zanderso (approved):
This is still needed as not enough of `dart:ffi` deprecations were reverted to totally avoid the need to roll this dependency forward.

zanderso:
It looks like this is failing on small golden diffs.

gw280:
This is because we want to land https://github.com/flutter/cocoon/pull/1091 which updates the baselines but we can't right now due to the ffi failures fixed by this revert. I've updated this branch to also pull in the rebaselined images so we can atomically land them.

gw280:
Closing as https://github.com/flutter/cocoon/pull/1091 ended up being identical

PR #1096, opened 2021-02-23
Add swarming user auto login check for Mac bots

keyonghan:
This adds health check for Mac bot swarming user auto login.
Though the check is on host level, we are adding it to device doctor tool to make it possible to quarantine bots when failing.

This is based on assumption that `swarming` user has been setup auto login for all Mac bots running devicelab tasks.

Related issue: https://github.com/flutter/flutter/issues/76091

### DIFF
--- a/device_doctor/lib/src/mac.dart
+++ b/device_doctor/lib/src/mac.dart
@@ -10,6 +10,9 @@ import 'health.dart';
 import 'utils.dart';
 
 /// The health check for Mac swarming user auto login.
+///
+/// The `swarming` user auto login is required for Flutter desktop tests which need to run
+/// GUI application on Mac.
 Future<HealthCheckResult> userAutoLoginCheck({ProcessManager processManager}) async {
   HealthCheckResult healthCheckResult;
   try {
--- a/device_doctor/lib/src/mac.dart
+++ b/device_doctor/lib/src/mac.dart
@@ -1,0 +1,30 @@
+// Copyright 2020 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:async';
+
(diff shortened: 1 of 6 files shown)


yusufm (approved):

device_doctor/lib/src/mac.dart
```diff
@@ -0,0 +1,30 @@
...
+// found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:process/process.dart';
+
+import 'health.dart';
+import 'utils.dart';
+
+/// The health check for Mac swarming user auto login.
```
godofredoc, line 12: Can we expand this comment to tell that this is a requirement to run GUI application on mac?

device_doctor/test/src/ios_device_test.dart
```diff
@@ -52,11 +52,12 @@ void main() {
       Map<String, List<HealthCheckResult>> results = await deviceDiscovery.checkDevices(processManager: processManager);
       expect(results.keys.length, equals(1));
       expect(results.keys.toList()[0], 'ios-device-abcdefg');
-      expect(results['ios-device-abcdefg'].length, equals(4));
+      expect(results['ios-device-abcdefg'].length, equals(5));
```
godofredoc, line 55: Can we make this more specific? like validating a certain check returns true and some other false? as it is this type of test requires changes every time a new test check is added/removed and the indexes do not tell very much about what check is being executed.

device_doctor/lib/src/mac.dart
keyonghan: Updated.

device_doctor/test/src/ios_device_test.dart
keyonghan: Expanded to include each test info.

godofredoc (approved):

PR #1097, opened 2021-02-23
Update Cocoon golden images

jason-simmons:


zanderso (approved):

zanderso:
Does this need to be rebased, or are the failing tests expected to fail due to the updates?

jason-simmons:
This is expected.  The presubmit should pass after the engine->framework roll lands.

zanderso:
I believe the engine -> framework roll that this needs has landed.

PR #1102, opened 2021-03-04
Add uninstall appllcations to iOS device recovery

keyonghan:
Uninstall applications when task fails for iOS device tasks.

This is to prevent application installation failure caused by using different signing certificate from previous installed application.

Related issue: https://github.com/flutter/flutter/issues/76896

### DIFF
--- a/device_doctor/lib/src/ios_device.dart
+++ b/device_doctor/lib/src/ios_device.dart
@@ -127,7 +127,46 @@ class IosDevice implements Device {
 
   @override
   Future<void> recover() async {
-    // Restarts the device first.
-    await eval('idevicediagnostics', <String>['restart']);
+    await restart_device();
+    await uninstall_applications();
+  }
+
+  /// Restart iOS device.
+  @visibleForTesting
+  Future<bool> restart_device({ProcessManager processManager}) async {
+    processManager ??= LocalProcessManager();
+    try {
+      await eval('idevicediagnostics', <String>['restart'], processManager: processManager);
+    } on BuildFailedError catch (error) {
+      logger.severe('device restart fails: $error');
+      return false;
+    }
+    return true;
+  }
+
+  /// Uninstall applications from a device.
+  ///
+  /// This is to prevent application installation failure caused by using different siging
+  /// certificate from previous installed application.
+  /// Issue: https://github.com/flutter/flutter/issues/76896
+  @visibleForTesting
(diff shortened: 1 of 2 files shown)


device_doctor/lib/src/ios_device.dart
```diff
@@ -127,7 +127,46 @@ class IosDevice implements Device {
 
   @override
   Future<void> recover() async {
-    // Restarts the device first.
-    await eval('idevicediagnostics', <String>['restart']);
+    await restart_device();
+    await uninstall_applications();
```
godofredoc, line 131: Should we uninstall first and then restart?

device_doctor/lib/src/ios_device.dart
```diff
@@ -127,7 +127,46 @@ class IosDevice implements Device {
...
+  }
+
+  /// Restart iOS device.
+  @visibleForTesting
+  Future<bool> restart_device({ProcessManager processManager}) async {
+    processManager ??= LocalProcessManager();
+    try {
+      await eval('idevicediagnostics', <String>['restart'], processManager: processManager);
+    } on BuildFailedError catch (error) {
+      logger.severe('device restart fails: $error');
```
godofredoc, line 141: should we also sdterr or is it included on the exception?

device_doctor/lib/src/ios_device.dart
```diff
@@ -127,7 +127,46 @@ class IosDevice implements Device {
...
+
+  /// Uninstall applications from a device.
+  ///
+  /// This is to prevent application installation failure caused by using different siging
+  /// certificate from previous installed application.
+  /// Issue: https://github.com/flutter/flutter/issues/76896
+  @visibleForTesting
+  Future<bool> uninstall_applications({ProcessManager processManager}) async {
+    processManager ??= LocalProcessManager();
+    final String result = await eval('ideviceinstaller', <String>['-l'], processManager: processManager);
```
godofredoc, line 155: what if eval fails?

keyonghan: Added a try catch block.

device_doctor/lib/src/ios_device.dart
keyonghan: Added stderr as well.

device_doctor/lib/src/ios_device.dart
keyonghan: Switched.

godofredoc (approved):

PR #1104, opened 2021-03-06
Add submit-queue check to the list of not related checks.

godofredoc:
This is to avoid the githubbot from removing the waiting for tree to go
green label.

Bug: https://github.com/flutter/flutter/issues/77401


Hixie (approved):
LGTM
cc @stuartmorgan

PR #1107, opened 2021-03-10
[app_dart] Add description to pubspec.yaml

CaseyHillers:
This is required for other Dart packages to pull Cocoon as a remote dependency. Which comes up when deploying a microservice dependent on it.


christopherfujino (approved):
LGTM

PR #1108, opened 2021-03-11
Enable test case('SignInButton shows avatar when authenticated')

xu-baolin:
@Piinks @CaseyHillers 

Fixes https://github.com/flutter/flutter/issues/73527

I will rework the related changes and re-enable the test case.


CaseyHillers (approved):
LGTM. Thank you for centering it!

PR #1109, opened 2021-03-11
[app_dart] Schedule prod tasks from GitHub webhooks

CaseyHillers:
Given [PullRequest], add it to Cocoon so tasks can be scheduled against it.

# Issues
https://github.com/flutter/flutter/issues/76140 - Move from cron method to webhooks
https://github.com/flutter/flutter/issues/77385 - Webhook method will no longer have this issue. The GitHub refresh cron doesn't dynamically add commits

### DIFF
--- a/app_dart/lib/src/request_handlers/github_webhook.dart
+++ b/app_dart/lib/src/request_handlers/github_webhook.dart
@@ -5,20 +5,22 @@
 import 'dart:async';
 import 'dart:convert';
 
-import 'package:cocoon_service/src/model/github/checks.dart';
-import 'package:cocoon_service/src/service/github_checks_service.dart';
-import 'package:cocoon_service/src/service/luci_build_service.dart';
 import 'package:crypto/crypto.dart';
 import 'package:github/github.dart';
 import 'package:github/hooks.dart';
 import 'package:meta/meta.dart';
 
 import '../datastore/cocoon_config.dart';
-
+import '../foundation/providers.dart';
+import '../model/github/checks.dart';
 import '../request_handling/body.dart';
 import '../request_handling/exceptions.dart';
 import '../request_handling/request_handler.dart';
 import '../service/buildbucket.dart';
+import '../service/datastore.dart';
+import '../service/github_checks_service.dart';
+import '../service/luci_build_service.dart';
+import '../service/scheduler.dart';
 
 /// List of repos that require check for labels and tests.
 const Set<String> kNeedsCheckLabelsAndTests = <String>{'flutter/flutter', 'flutter/engine'};
@@ -27,12 +29,24 @@ final RegExp kEngineTestRegExp = RegExp(r'tests?\.(dart|java|mm|m|cc)$');
 
 @immutable
 class GithubWebhook extends RequestHandler<Body> {
-  const GithubWebhook(Config config, this.buildBucketClient, this.luciBuildService, this.githubChecksService)
+  GithubWebhook(Config config, {
+    @required this.buildBucketClient,
+    @visibleForTesting Scheduler schedulerValue,
+    this.luciBuildService,
+    this.githubChecksService})
       : assert(buildBucketClient != null),
+        scheduler = schedulerValue ?? Scheduler(
+          config: config,
+          datastore: DatastoreService.defaultProvider(config.db),
+          httpClient: Providers.freshHttpClient(),
+        ),
         super(config: config);
 
   /// A client for querying and scheduling LUCI Builds.
   final BuildBucketClient buildBucketClient;
+
+  /// Cocoon scheduler to trigger tasks against changes from GitHub.
+  final Scheduler scheduler;
 
   /// LUCI service class to communicate with buildBucket service.
   final LuciBuildService luciBuildService;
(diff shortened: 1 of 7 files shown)


christopherfujino:
do you mean the PR merged webhook?

app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -27,13 +29,26 @@ final RegExp kEngineTestRegExp = RegExp(r'tests?\.(dart|java|mm|m|cc)$');
 
 @immutable
 class GithubWebhook extends RequestHandler<Body> {
-  const GithubWebhook(Config config, this.buildBucketClient, this.luciBuildService, this.githubChecksService)
+  GithubWebhook(Config config,
+      {@required this.buildBucketClient,
+      @visibleForTesting Scheduler schedulerValue,
```
christopherfujino, line 34: TIL you can apply `@visibleForTesting` to a param

CaseyHillers:
> do you mean the PR merged webhook?

Yes. This means commits can either be scheduled via the cron or or the GitHub webhook for when PRs are closed.

After I validate the webhook approach, I'll remove the cron method.

app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -103,6 +119,9 @@ class GithubWebhook extends RequestHandler<Body> {
             pr.head.sha,
             'Pull request closed',
           );
+        } else {
+          // Merged pull requests can be added to CI.
+          await scheduler.addPullRequest(pr);
```
christopherfujino, line 124: Should we be listening for merged PRs, or commits?

CaseyHillers: I don't think there's a commit webhook event - https://docs.github.com/en/developers/webhooks-and-events/webhook-events-and-payloads

The PR event provides a lot of data, and it provides the release branch info. I'll add a test to verify release PRs.

CaseyHillers: Offline conversation. There is a commit webhook called push - https://docs.github.com/en/developers/webhooks-and-events/webhook-events-and-payloads#push

It looks like the GitHub library doesn't support that event yet, so I'm going to look into adding it

CaseyHillers: I looked into the push events on the archive table, and they're not the solution. We need to support this event + push events to get 100% coverage. From the last 90 days, push events made up 0% of commits that need to be tested by infra.

I'm going to follow up and modify the cron job from every minute to every hour, and have it check for any missing commits. Right now it only adds newer, missing commits.

CaseyHillers: Filed https://github.com/flutter/flutter/issues/78351 for tracking the PushEvent addition

app_dart/lib/src/service/scheduler.dart
```diff
@@ -56,6 +65,45 @@ class Scheduler {
...
+      branch: branch,
+      key: key,
+      // The field has a max length of 1500 so ensure the commit message is not longer.
+      message: truncate(pr.title, 1490, omission: '...'),
+      repository: repo,
+      sha: sha,
+      timestamp: pr.mergedAt.millisecondsSinceEpoch,
+    );
+
+    if (await _commitExistsInDatastore(mergedCommit)) {
```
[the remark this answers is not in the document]
CaseyHillers: Yes. Since commits can be scheduled via (1) cron job and (2) webhook triggers we should ensure we haven't already added the commit. GitHub will retry commits if cocoon sends a non-2** response, so we should handle this case gracefully.

app_dart/lib/src/request_handlers/github_webhook.dart
christopherfujino: We sync'ed up offline again, and triggering based on PR merged even SGTM

app_dart/lib/src/service/scheduler.dart
```diff
@@ -90,6 +138,11 @@ class Scheduler {
     return newCommits;
   }
 
+  /// Whether [Commit] already exists in [datastore].
```
christopherfujino, line 141: can you expand on this doc, explaining that this is cocoon's source of truth for builds it has scheduled, and that these builds could be scheduled via webhook OR cron job?

app_dart/test/src/utilities/mocks.dart
```diff
@@ -27,6 +28,8 @@ class MockTabledataResourceApi extends Mock implements TabledataResourceApi {}
 
 class MockAccessTokenService extends Mock implements AccessTokenService {}
 
+class MockScheduler extends Mock implements Scheduler {}
```
christopherfujino, line 31: not a blocker for this PR, but `mockito` might be a blocker to migrating `app_dart` to null-safety. FYI https://github.com/dart-lang/mockito/blob/master/NULL_SAFETY_README.md

christopherfujino (approved):
nit about expanding dartdoc for `_commitExistsInDatastore`, otherwise LGTM

app_dart/lib/src/service/scheduler.dart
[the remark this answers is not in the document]
keyonghan: That makes sense. I guess we can still guarantee the order of commits with these two scheduling logics, right?

app_dart/lib/src/service/scheduler.dart
CaseyHillers: Added!

app_dart/test/src/utilities/mocks.dart
CaseyHillers: Ack, thanks!

PR #1112, opened 2021-03-17
[app_dart] Switch scheduler to httpClientProvider + cleanup

CaseyHillers:
- Scheduler needs to pass httpClientProvider instead of httpClient due to timeout issues where the client is closed
- Use common remote file utility for devicelab manifest
- Update remote file utility to catch all possible places where an error can occur
- Set timeout of get request on the remote file utility

# Issues

Fixes https://github.com/flutter/flutter/issues/78479

### DIFF
--- a/app_dart/lib/src/foundation/utils.dart
+++ b/app_dart/lib/src/foundation/utils.dart
@@ -36,10 +36,9 @@ Future<String> remoteFileContent(HttpClientProvider branchHttpClientProvider, Lo
   try {
     // TODO(keyonghan): apply retry logic here to simply, https://github.com/flutter/flutter/issues/52427
     for (int attempt = 0; attempt < 3; attempt++) {
-      final HttpClientRequest clientRequest = await client.getUrl(url);
-
       try {
-        final HttpClientResponse clientResponse = await clientRequest.close();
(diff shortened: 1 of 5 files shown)


app_dart/lib/src/foundation/utils.dart
```diff
@@ -36,10 +36,9 @@ Future<String> remoteFileContent(HttpClientProvider branchHttpClientProvider, Lo
   try {
     // TODO(keyonghan): apply retry logic here to simply, https://github.com/flutter/flutter/issues/52427
     for (int attempt = 0; attempt < 3; attempt++) {
-      final HttpClientRequest clientRequest = await client.getUrl(url);
-
       try {
-        final HttpClientResponse clientResponse = await clientRequest.close();
+        final HttpClientRequest clientRequest = await client.getUrl(url).timeout(const Duration(seconds: 5));
+        final HttpClientResponse clientResponse = await clientRequest.close().timeout(const Duration(seconds: 5));
```
keyonghan, line 41: Nit: inject the timeout as an optional parameter with a default value?

keyonghan (approved):
LGMT with nit.

app_dart/lib/src/foundation/utils.dart
CaseyHillers: Done, and added dartdoc

PR #1113, opened 2021-03-18
Add retries to the call of getCheckRun and updateCheckStatus

keyonghan:
This is to prevent errors when calling `/api/luci-status-handler` due to `GitHub Error` and `SocketException`.

Related issues: https://github.com/flutter/flutter/issues/78555, https://github.com/flutter/flutter/issues/78561

### DIFF
--- a/app_dart/lib/src/foundation/github_checks_util.dart
+++ b/app_dart/lib/src/foundation/github_checks_util.dart
@@ -3,6 +3,7 @@
 // found in the LICENSE file.
 
 import 'dart:core';
+import 'dart:io';
 
 import 'package:cocoon_service/src/model/github/checks.dart';
 import 'package:github/github.dart' as github;
@@ -70,7 +71,7 @@ class GithubChecksUtil {
         detailsUrl: detailsUrl,
         output: output,
       );
-    }, retryIf: (Exception e) => e is github.GitHubError);
+    }, retryIf: (Exception e) => e is github.GitHubError || e is SocketException);
   }
 
   Future<github.CheckRun> getCheckRun(
@@ -91,7 +92,7 @@ class GithubChecksUtil {
         slug,
         checkRunId: id,
       );
-    }, retryIf: (Exception e) => e is github.GitHubError);
+    }, retryIf: (Exception e) => e is github.GitHubError || e is SocketException);
   }
 
   /// Sends a request to github checks api to create a new [CheckRun] associated
--- a/app_dart/lib/src/foundation/github_checks_util.dart
+++ b/app_dart/lib/src/foundation/github_checks_util.dart
@@ -74,14 +74,24 @@ class GithubChecksUtil {
   }
 
   Future<github.CheckRun> getCheckRun(
-    github.GitHub gitHubClient,
+    Config cocoonConfig,
     github.RepositorySlug slug,
     int id,
   ) async {
-    return gitHubClient.checks.checkRuns.getCheckRun(
-      slug,
-      checkRunId: id,
+    const RetryOptions r = RetryOptions(
+      maxAttempts: 3,
+      delayFactor: Duration(seconds: 2),
     );
+    return r.retry(() async {
+      final github.GitHub gitHubClient = await cocoonConfig.createGitHubClient(
+        slug.owner,
+        slug.name,
+      );
+      return await gitHubClient.checks.checkRuns.getCheckRun(
+        slug,
+        checkRunId: id,
+      );
+    }, retryIf: (Exception e) => e is github.GitHubError);
   }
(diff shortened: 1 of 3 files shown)


app_dart/lib/src/request_handlers/luci_status.dart
```diff
@@ -96,17 +97,34 @@ class LuciStatusHandler extends RequestHandler<Body> {
...
+  /// The `githubChecksService.updateCheckStatus` has been hitting `SocketException` occasionally.
+  /// Retry logic is to mitigate this issue: https://github.com/flutter/flutter/issues/74611.
+  Future<void> _updateCheckStatus(BuildPushMessage buildPushMessage, RepositorySlug slug) async {
+    const RetryOptions r = RetryOptions(
+      maxAttempts: 3,
+      delayFactor: Duration(seconds: 2),
+    );
+    await r.retry(
+      () async {
+        await githubChecksService.updateCheckStatus(
```
godofredoc, line 118: Should githubChecksService.updateCheckStatus handle the retries? that way any code using githubChecksService.updateCheckStatus will get the benefit.

app_dart/lib/src/foundation/github_checks_util.dart
```diff
@@ -74,14 +74,24 @@ class GithubChecksUtil {
   }
 
   Future<github.CheckRun> getCheckRun(
-    github.GitHub gitHubClient,
+    Config cocoonConfig,
```
godofredoc, line 77: Do we need tests to validate the retries are actually happening?

keyonghan: Have searched around, but didn't find an easy way to mock the corresponding function (`getCheckRun`), which is located in a private class...  Not sure if you got any chance to look at this when implementing other functions like `updateCheckRun`?
If not, I may prefer to skip it here as I have spent quite a while on this, and this is not a blocking thing.

app_dart/lib/src/request_handlers/luci_status.dart
keyonghan: Retry logic has been available in `githubChecksUtil.updateCheckRun` when hitting  `GitHubError`.
Here the `SocketException` happens when calling `githubChecksService.updateCheckStatus`, but not the sub function inside. 
`githubChecksUtil.getCheckRun` is another sub function called inside, and the retry is added there as well so that other codes can benefit as well.

app_dart/lib/src/foundation/github_checks_util.dart
keyonghan: Another potential blocking thing on the test is: we need to return different values for the same mock call, but it is not supported by default via dart mockito (though Java supports that by default). It seems doable if we return same type of values (https://github.com/dart-lang/mockito/issues/260), but here we need to return a combination of an exception and a valid value.

app_dart/lib/src/request_handlers/luci_status.dart
godofredoc: I'm not sure I'm following. Given that the retry block is wrapping a single method and that method is already handling retries, wouldn't be easier/more reusable to just add SocketException to the list of retried exceptions?

keyonghan: It is not clear which step of `githubChecksUtil.updateCheckRun` throws the `SocketException`. From log, it just throws one line error
```
SocketException: OS Error: Connection timed out, errno = 110, address = api.github.com, port = 40232
```
I am concerned that simply appending `SocketException` to the retried exception may not catch the error. Logs in https://github.com/flutter/flutter/issues/78555 makes me feel the `SocketException` happens elsewhere, otherwise it should catch the stacktrace?

godofredoc: updateCheckRun has only two statements and the full method body is wrapped up in a retry. All this additional logic is equivalent to adding Socket exception here: https://cs.opensource.google/flutter/cocoon/+/master:app_dart/lib/src/foundation/github_checks_util.dart;l=73?q=GithubChecksUtil&ss=flutter

keyonghan: Sorry I meant `githubChecksService.updateCheckStatus` from luci_status...

keyonghan: seems only `updateCheckStatus` and ``getCheckRun` are related to github api call, as you suggested, I appended the `SocketException` to both functions instead of retry the call from `luci_status.dart`.
We can monitor how it goes with this change first before implementing retries in the upper call.

godofredoc (approved):

PR #1115, opened 2021-03-19
[app_dart] Configurable supported repos for scheduler

CaseyHillers:
Only `flutter/flutter` has prod tasks that are supported by `Scheduler`. The GitHub webhooks are trying to schedule commits against unsupported repos (like `flutter/engine`) which is causing an increase in failures.

This doesn't affect presubmit tasks. When the scheduler is more fleshed out, this will include both types of builds.

### DIFF
--- a/app_dart/lib/src/service/scheduler.dart
+++ b/app_dart/lib/src/service/scheduler.dart
@@ -104,6 +104,11 @@ class Scheduler {
   }
 
   Future<void> _addCommit(Commit commit) async {
+    if (!Config.schedulerSupportedRepos.contains(commit.repository)) {
+      log.debug('Skipping ${commit.id} as repo is not supported');
+      return;
+    }
+
     final List<Task> tasks = await _getTasks(commit);
     try {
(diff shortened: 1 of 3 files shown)


app_dart/lib/src/service/scheduler.dart
```diff
@@ -104,6 +104,11 @@ class Scheduler {
   }
 
   Future<void> _addCommit(Commit commit) async {
+    if (!Config.schedulerSupportedRepos.contains(commit.repository)) {
+      log.debug('Skipping ${commit.id} as repo is not supported');
```
keyonghan, line 108: Nit: use $repo instead.

keyonghan (approved):
LGTM with nit.

app_dart/lib/src/service/scheduler.dart
CaseyHillers: `commit.id` includes `$repository/$branch/$sha` so that's why I favored it

christopherfujino: makes sense, LGTM

PR #1117, opened 2021-03-22
[app_dart] Cache static files for 30 minutes

CaseyHillers:
Often we don't need to do an IO operation as these files rarely change (at most, once a weekday). Additionally, changes to the Flutter apps average once a week. This should speed up the overall load time of the dashboards.

### DIFF
--- a/app_dart/README.md
+++ b/app_dart/README.md
@@ -176,6 +176,12 @@ For more options run:
 $ dart dev/deploy.dart --help
 ```
 
+##### Changes not propagated
+
+Cocoon caches static assets for 30 minutes. This will delay releases until the
+cache is reset. Breaking changes will require wiping the memorystore instance
+from the Google Cloud dashboard.
+
 ### Branching support for flutter repo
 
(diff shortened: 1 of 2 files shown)


yusufm (approved):
LGTM. 
Note: even though the data doesn't change frequently, do we want to add an info message somewhere that the data may be delayed by upto 30 mins?

CaseyHillers:
> Note: even though the data doesn't change frequently, do we want to add an info message somewhere that the data may be delayed by upto 30 mins?

Added

christopherfujino:
I saw this via the revert. What static assets are we serving? I can think of fonts off the top of my head...

CaseyHillers:
We serve HTML, javascript, javascript maps, json files, css, css maps, fonts, images.

I think the best path forward for this change is to update StaticRequestHandler to use a cache instead of trying to wrap it in the generic cache request handler.

christopherfujino:
ahh, i see, so all those things are being read from disk. that makes sense. it would be great to just publish them all to cloud storage as part of cocoon builds. we'd just have to figure out how to cache bust on each release.

PR #1118, opened 2021-03-22
[app_dart] Move CocoonConfig datastore model to its own model file

CaseyHillers:
The datastore models should all exist under `src/model/`. Move CocoonConfig to its own file, and simplify the config file to `config.dart` to make it easier to tell them apart.

Removed `InvalidConfigurationException` as it appears to be dead code.


keyonghan (approved):
LGTM

PR #1120, opened 2021-03-22
Reduce cron for /api/refresh-github-commits

CaseyHillers:
Triaging the flutter-dashboard GCP logs shows all commits are being scheduled from the webhook trigger instead of this cron job. Reducing frequency to save some GitHub API quota.


keyonghan (approved):
LGTM

PR #1121, opened 2021-03-23
Revert "[app_dart] Cache static files for 30 minutes"

CaseyHillers:
Reverts flutter/cocoon#1117

The MIME type isn't being saved, so everything is sent as plaintext. This is causing the dashboard to be unusable.


keyonghan (approved):
LGTM

PR #1129, opened 2021-03-24
Revert "Reduce cron for /api/refresh-github-commits"

godofredoc:
Reverts flutter/cocoon#1120


CaseyHillers (changes requested):
No longer necessary

PR #1130, opened 2021-03-25
[scheduler] Add SchedulerConfig and validation tests

CaseyHillers:
Part of go/flutter-build-test-infra

This change is a no-op to Cocoon, and only adds models necessary for future changes.

* Create `SchedulerConfig` and `Target` protos
* Validate scheduler config from a yaml is valid
  * Ensure all dependencies exist
  * No duplicate targets
  * No cycles (the config should only create a tree)
* Add `bin/validate_scheduler_config.dart` for running tests against external configs

# Issues

### DIFF
--- a/app_dart/lib/src/model/proto/internal/scheduler.proto
+++ b/app_dart/lib/src/model/proto/internal/scheduler.proto
@@ -4,6 +4,13 @@
 
 syntax = "proto2";
 
+message SchedulerConfig {
+    // Targets to run from this config.
+    repeated Target targets = 1;
+    // Git branches to run these targets against.
+    repeated string enabled_branches = 2;
+}
+
 message Target {
     // Unique, human readable identifier.
     optional string name = 1;
--- a/app_dart/lib/src/model/proto/internal/scheduler.proto
+++ b/app_dart/lib/src/model/proto/internal/scheduler.proto
@@ -1,0 +1,24 @@
+// Copyright 2019 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+syntax = "proto2";
+
+message Target {
+    // Unique, human readable identifier.
+    optional string name = 1;
+    // Names of other targets required to succeed before triggering this target.
+    repeated string dependencies = 2;
+    // Whether this target is stable and can be used to gate commits.
+    // Defaults to false which blocks builds.
+    optional bool bringup = 3;
+    // Number of minutes this target is allowed to run before being marked as failed.
+    optional int32 timeout = 4;
+    // Name of the testbed this target will run on.
+    // Defaults to a linux vm.
+    optional string testbed = 5;
+    // Properties to configure infrastructure tooling.
+    map<string, string> properties = 6;
+    // Name of the LUCI recipe to trigger.
+    optional string recipe = 7;
+}
--- a/app_dart/lib/src/model/proto/internal/scheduler.proto
+++ b/app_dart/lib/src/model/proto/internal/scheduler.proto
@@ -18,12 +18,12 @@ message Target {
     repeated string dependencies = 2;
     // Whether this target is stable and can be used to gate commits.
     // Defaults to false which blocks builds.
-    optional bool bringup = 3;
+    optional bool bringup = 3 [default = false];
     // Number of minutes this target is allowed to run before being marked as failed.
-    optional int32 timeout = 4;
+    optional int32 timeout = 4 [default = 30];
     // Name of the testbed this target will run on.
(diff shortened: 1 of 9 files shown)


app_dart/lib/src/model/proto/internal/scheduler.proto
```diff
@@ -0,0 +1,31 @@
...
+message SchedulerConfig {
+    // Targets to run from this config.
+    repeated Target targets = 1;
+    // Git branches to run these targets against.
+    repeated string enabled_branches = 2;
+}
+
+message Target {
+    // Unique, human readable identifier.
+    optional string name = 1;
```
keyonghan, line 16: shall the name be required? is it builder name, task name, or any other?

app_dart/lib/src/model/proto/internal/scheduler.proto
```diff
@@ -0,0 +1,31 @@
...
+syntax = "proto2";
+
+message SchedulerConfig {
+    // Targets to run from this config.
+    repeated Target targets = 1;
+    // Git branches to run these targets against.
+    repeated string enabled_branches = 2;
+}
+
+message Target {
```
keyonghan, line 14: Do we need a flaky flag somewhere?

app_dart/lib/src/service/scheduler.dart
```diff
@@ -256,3 +257,57 @@ class Scheduler {
...
+  // Construct [targetGraph]. With a one scan approach, cycles in the graph
+  // cannot exist as it only works forward.
+  for (final Target target in schedulerConfig.targets) {
+    if (targetGraph.containsKey(target.name)) {
+      exceptions.add('ERROR: ${target.name} already exists in graph');
+    } else {
+      targetGraph[target.name] = <Target>[];
+      // Add edges
+      if (target.dependencies.isNotEmpty) {
+        if (target.dependencies.length != 1) {
```
keyonghan, line 290: why is `dependencies` defined as repeated then?

app_dart/lib/src/service/scheduler.dart
```diff
@@ -256,3 +257,57 @@ class Scheduler {
...
+          exceptions
+              .add('ERROR: ${target.name} has multiple dependencies which is not supported. Use only one dependency');
+        } else {
+          if (target.dependencies.first == target.name) {
+            exceptions.add('ERROR: ${target.name} cannot depend on itself');
+          } else if (targetGraph.containsKey(target.dependencies.first)) {
+            targetGraph[target.dependencies.first].add(target);
```
keyonghan, line 297: do we need to maintain the child list of a target?

app_dart/test/service/scheduler_test.dart
```diff
@@ -250,6 +252,135 @@ void main() {
...
+  - master
+targets:
+  - name: A
+    dependencies:
+      - B
+  - name: B
+    dependencies:
+      - A
+      ''') as YamlMap;
+      expect(() => loadSchedulerConfig(configWithCycle), throwsA(isA<FormatException>()));
```
keyonghan, line 343: can we verify the exception summary for different scenarios?

CaseyHillers: Added!

app_dart/lib/src/service/scheduler.dart
CaseyHillers: This graph is specifically for validating configs. For future scheduling, we'll want to use a hash map for tracking what builds should be triggered when. So when `$target` notifies of completion, we call `triggers[$target]` to quickly lookup the next set.
```
trigger['web_engine_linux'] = <Target>['web_engine_linux_chrome', 'web_engine_linux_firefox', ...];
```

app_dart/lib/src/service/scheduler.dart
CaseyHillers: This is future functionality we'd like to support, but it adds a lot of complexity to the initial scheduler. It's easier to support if if this is already a list

app_dart/lib/src/model/proto/internal/scheduler.proto
CaseyHillers: From feedback on go/flutter-build-test-infra, we're renaming it as bringup

app_dart/lib/src/model/proto/internal/scheduler.proto
CaseyHillers: Protos should never use required field as it causes issues in the future if you deprecate the field ([go/protodosdonts](http://go/protodosdonts))

keyonghan (approved):
LGTM

CaseyHillers:
\cc @christopherfujino

PR #1131, opened 2021-03-26
use flutter_test instead of package:test

jonahwilliams:
This makes it easier to version solve since flutter_test has fewer dependencies.

Blocking https://github.com/flutter/flutter/issues/77856

Should be safe as long as `flutter test` is being used to run unit tests, even if they are not flutter tests

### DIFF
--- a/app_flutter/pubspec.yaml
+++ b/app_flutter/pubspec.yaml
@@ -35,7 +35,6 @@ dev_dependencies:
     sdk: flutter
   mockito: ^4.0.0
   path: ^1.6.4
-  test: ^1.6.0
 
 # For information on the generic Dart part of this file, see the
 # following page: https://dart.dev/tools/pub/pubspec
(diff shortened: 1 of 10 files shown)


app_flutter/pubspec.yaml
```diff
@@ -35,7 +35,6 @@ dev_dependencies:
     sdk: flutter
   mockito: ^4.0.0
   path: ^1.6.4
-  test: ^1.6.0
```
CaseyHillers, line 38: cocoon_service (backend) is dependent on package test. Are dev dependencies ignored?

CaseyHillers (approved):
LGTM!

app_flutter/pubspec.yaml
jonahwilliams: Yup!

jonahwilliams: (assuming that is the dev_dependency of cocoon_service)

PR #1132, opened 2021-03-29
Fix device_doctor build script for windows.

yusufm:
The script exists after running the get dependencies command. Precede the
commands with "call" which will execute that command as well as the following build command.


godofredoc (approved):

keyonghan (approved):
LGTM

PR #1134, opened 2021-03-30
[cocoon] Exclude null builds when calling buildbucket API

keyonghan:
When buildbucket API call returns empty builds, it fails `/api/refresh-chromebot-status` which refresh LUCI builds statuses to flutter dashboard.
This is to exclude the empty case.

https://github.com/flutter/flutter/issues/79380

### DIFF
--- a/app_dart/lib/src/service/luci.dart
+++ b/app_dart/lib/src/service/luci.dart
@@ -202,6 +202,7 @@ class LuciService {
     final BatchResponse batchResponse = await buildBucketClient.batch(batchRequest);
     final Iterable<Build> builds = batchResponse.responses
         .map<SearchBuildsResponse>((Response response) => response.searchBuilds)
+        .where((SearchBuildsResponse response) => response.builds != null)
         .expand<Build>((SearchBuildsResponse response) => response.builds);
     return builds;
   }


app_dart/lib/src/service/luci.dart
```diff
@@ -202,6 +202,7 @@ class LuciService {
     final BatchResponse batchResponse = await buildBucketClient.batch(batchRequest);
     final Iterable<Build> builds = batchResponse.responses
         .map<SearchBuildsResponse>((Response response) => response.searchBuilds)
+        .where((SearchBuildsResponse response) => response.builds != null)
```
godofredoc, line 205: This will only hide the problem. Do we know why some builds are returning null?

keyonghan:
This is to open the tree first.
A following up is to find the reason why. Otherwise that empty build will never be updated.

app_dart/lib/src/service/luci.dart
keyonghan: This is to open the tree first.
A following up is to find the reason why. Otherwise that empty build will never be updated.

godofredoc: LGTM if we will be able to find the issue after we land this change

godofredoc (approved):
LGTM if we will be able to root cause the real issue after this mitigation.

app_dart/lib/src/service/luci.dart
keyonghan: Created https://github.com/flutter/flutter/issues/79393 to track.

keyonghan:
It turns out the builder was enabled in prod_builders.json, but missed from builder config https://devops-console-oss.corp.google.com/flutter/infra/+/master:config/framework_config.star

godofredoc:
> It turns out the builder was enabled in prod_builders.json, but missed from builder config https://devops-console-oss.corp.google.com/flutter/infra/+/master:config/framework_config.star

How can we prevent this from happening again?

PR #1135, opened 2021-03-30
[app_dart] Remove refresh github commits cron

CaseyHillers:
Disable the cron job as it is causing issues with the webhook scheduler.

For the next week I'll look at if any commits are being dropped via https://plx.corp.google.com/scripts2/script_60._49462f_0000_2e53_920f_089e0832f92c

This endpoint can still be manually queried if there are outages

# Issues

https://github.com/flutter/flutter/issues/79408

### DIFF
--- a/app_dart/cron.yaml
+++ b/app_dart/cron.yaml
@@ -1,9 +1,6 @@
 # Updates to this file must be pushed with:
 #   gcloud app deploy --project flutter-dashboard cron.yaml
 cron:
-- description: refresh commits from GitHub
-  url: /api/refresh-github-commits
-  schedule: every 15 minutes
 - description: refresh chromebot build status
   url: /api/refresh-chromebot-status
   schedule: every 3 minutes


keyonghan:
Shall we start by deceasing the frequency, rather than removing it completely?

I lost some context about commit event not triggering tasks via webhook pr event. 
/cc @christopherfujino may have more input on this.

CaseyHillers:
https://github.com/flutter/flutter/issues/78351 is I believe the issue. For `dev`, `beta`, `stable` branches they are pushed to. However, Cocoon currently doesn't track those release branches, and shouldn't block this.

keyonghan (approved):
LGTM

christopherfujino (approved):
LGTM

CaseyHillers:
I forgot to mention this, but since we added webhook scheduling Cocoon has not dropped a single commit (Usually there was a ~2% drop rate).

I'm not sure if it's because there was two scheduler systems running, but I'll continue to monitor.

PR #1136, opened 2021-04-01
[scheduler] Rename recipe field to builder

CaseyHillers:
This config isn't being used yet. It's non-trivial to spawn a swarming task to run a recipe in Flutter LUCI infrastructure.

Instead, we'll start by triggering the existing builders from Cocoon.


christopherfujino (approved):
lgtm

PR #1138, opened 2021-04-02
[app_dart] Update engine build status to use config value

CaseyHillers:
Quick fix to unify the framework and engine tree failure status messages (they are identical)

### DIFF
--- a/app_dart/lib/src/request_handlers/push_engine_status_to_github.dart
+++ b/app_dart/lib/src/request_handlers/push_engine_status_to_github.dart
@@ -84,8 +84,7 @@ class PushEngineStatusToGithub extends ApiRequestHandler<Body> {
         request.targetUrl = 'https://ci.chromium.org/p/flutter/g/engine/console';
         request.context = 'luci-engine';
         if (status != GithubBuildStatusUpdate.statusSuccess) {
-          request.description = 'Flutter build is currently broken. Please do not merge this '
-              'PR unless it contains a fix to the broken build.';
+          request.description = config.flutterBuildDescription;
         }
 
         try {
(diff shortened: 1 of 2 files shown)


app_dart/lib/src/request_handlers/push_engine_status_to_github.dart
```diff
@@ -84,8 +84,7 @@ class PushEngineStatusToGithub extends ApiRequestHandler<Body> {
         request.targetUrl = 'https://ci.chromium.org/p/flutter/g/engine/console';
         request.context = 'luci-engine';
         if (status != GithubBuildStatusUpdate.statusSuccess) {
-          request.description = 'Flutter build is currently broken. Please do not merge this '
-              'PR unless it contains a fix to the broken build.';
+          request.description = config.flutterBuildDescription;
```
yusufm, line 87: flutterBuildDescription always going to be populated, should we have a default value?

yusufm (approved):

app_dart/lib/src/request_handlers/push_engine_status_to_github.dart
keyonghan, line 87: Shall we say `Engine build`? `Flutter build` seems confusing as used for engine tree.

CaseyHillers: What about `Tree is currently broken. ... contains a fix for the tree.`? That way it works for all Flutter repos?

CaseyHillers: By default, it will be empty not showing a message. I believe GitHub defaults to "Success" in that case.

keyonghan: SGTM.

keyonghan (approved):
LGTM with nit.

PR #1139, opened 2021-04-03
[app_dart] Create FakeScheduler service + fix empty manifest crash

CaseyHillers:
This is to make it easier to write tests that depend on `Scheduler` but do not care about its implementation.

While implementing this, I noticed an issue where empty devicelab manifests would cause crashes, so I fixed it. Added a regression test.

## Issues
Fixes https://github.com/flutter/flutter/issues/79679

### DIFF
--- a/app_dart/bin/server.dart
+++ b/app_dart/bin/server.dart
@@ -46,10 +46,7 @@ Future<void> main() async {
     );
 
     /// Cocoon scheduler service to manage validating commits in presubmit and postsubmit.
-    final Scheduler scheduler = Scheduler(
-      cache: cache,
-      config: config,
-    );
+    final Scheduler scheduler = Scheduler(config: config);
 
     final Map<String, RequestHandler<dynamic>> handlers = <String, RequestHandler<dynamic>>{
       '/api/append-log': AppendLog(config, authProvider),
--- a/app_dart/bin/server.dart
+++ b/app_dart/bin/server.dart
@@ -70,6 +70,7 @@ Future<void> main() async {
         buildBucketClient: buildBucketClient,
         luciBuildService: luciBuildService,
         githubChecksService: githubChecksService,
+        scheduler: scheduler,
       ),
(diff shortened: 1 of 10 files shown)


CaseyHillers:
Friendly ping

app_dart/lib/src/service/scheduler.dart
```diff
@@ -31,20 +32,27 @@ import 'datastore.dart';
...
-  final DatastoreService datastore;
+  final DatastoreServiceProvider datastoreProvider;
   final HttpClientProvider httpClientProvider;
   final GitHubBackoffCalculator gitHubBackoffCalculator;
+
+  DatastoreService datastore;
   Logging log;
 
+  /// Subcache key to store [SchedulerConfig] for quicker retrievals.
+  static const String schedulerSubcacheName = 'scheduler';
```
keyonghan, line 54: Has this been used anywhere?

app_dart/bin/server.dart
```diff
@@ -45,6 +45,12 @@ Future<void> main() async {
       config,
     );
 
+    /// Cocoon scheduler service to manage validating commits in presubmit and postsubmit.
+    final Scheduler scheduler = Scheduler(
+      cache: cache,
```
keyonghan, line 50: Could we implement cache in a separate PR when needed? It is not clear to me how it helps in this PR. It will help if any issue pops up related to cache (like a revert without affecting other implementations).

app_dart/lib/src/service/scheduler.dart
CaseyHillers: Not in prod - This is for a future PR

app_dart/bin/server.dart
CaseyHillers: Removed. This was left over as this PR is a split from a larger PR.

keyonghan (approved):
LGTM.

PR #1140, opened 2021-04-03
Add additional engine auto-labelling

cbracken:
This adds additional engine auto-labelling for the following labels:
* embedder
* platform-fuchsia
* platform-linux
* platform-macos
* platform-web
* platform-windows

PR #1141, opened 2021-04-03
Extract and test framework auto-labelling

cbracken:
Applies a similar refactoring to the framework auto-labelling for files
as was done in #1140 for the engine. Also adds more test coverage for
framework labelling.

This also fixes a bug where PRs that included files with the word
'accessibility' in their path were missed due to a typo in the path
matching.

Moves the labelling tests into groups for framework and engine to
improve readability.

### DIFF
--- a/app_dart/test/request_handlers/github_webhook_test.dart
+++ b/app_dart/test/request_handlers/github_webhook_test.dart
@@ -272,6 +272,116 @@ void main() {
       ));
     });
 
+    group('getLabelsForFrameworkPath', () {
+      test('Only the team label is applied to pubspec.yaml', () {
+        expect(GithubWebhook.getLabelsForFrameworkPath('packages/flutter_tools/pubspec.yaml'), <String>{'team'});
+      });
+
+      test('Tool label applied', () {
+        expect(GithubWebhook.getLabelsForFrameworkPath('packages/flutter_tools/hot_reload.dart'), contains('tool'));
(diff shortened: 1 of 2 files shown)


app_dart/test/request_handlers/github_webhook_test.dart
```diff
@@ -272,6 +272,116 @@ void main() {
...
+      test('platform-windows applied for windows embedder', () {
+        expect(
+            GithubWebhook.getLabelsForEnginePath('shell/platform/windows/start_menu.cc'), contains('platform-windows'));
+      });
+
+      test('platform-web applied for web paths', () {
+        expect(GithubWebhook.getLabelsForEnginePath('lib/web_ui/shadow_dom.dart'), contains('platform-web'));
+        expect(GithubWebhook.getLabelsForEnginePath('web_sdk/'), contains('platform-web'));
+      });
+    });
```
CaseyHillers, line 383: Up to you, but you could deduplicate some of the boiler plate by doing a table driven test

CaseyHillers (approved):

app_dart/test/request_handlers/github_webhook_test.dart
cbracken: True but table-driven tests aren't too idiomatic in Dart, and are likely to be a little harder to debug, though certainly more concise. If you can think of a way to do it that'll fail with an exactly line-number etc. though, definitely go for it!

PR #1142, opened 2021-04-05
[app_dart] Fix writing GitHub quota rate limit json

CaseyHillers:
The issue was the `resets` was being sent as a `DateTime`, when the table expected a timestamp int. This caused silent failures.

I rewrote the schema of the table to be with `DateTime`, and now seeing results using the `testchillers` version.

# Issues

Fixes https://github.com/flutter/flutter/issues/79466


keyonghan (approved):
LGTM

PR #1143, opened 2021-04-05
Skip reboot for 32 bit phone when recovering device

keyonghan:
Device recovery of device_doctor tool reboots the device. However ios32 phone needs manual intervention when rebooting.
This PR skips reboot for the 32 phone.

Related issue: https://github.com/flutter/flutter/issues/79316

### DIFF
--- a/device_doctor/lib/src/ios_device.dart
+++ b/device_doctor/lib/src/ios_device.dart
@@ -14,6 +14,11 @@ import 'health.dart';
 import 'mac.dart';
 import 'utils.dart';
 
+/// Identifiers for devices that should never be rebooted.
+final Set<String> noRebootForbidList = <String>{
+  '822ef7958bba573829d85eef4df6cbdd86593730', // 32bit iPhone requires manual intervention on reboot.
+};
+
 /// IOS implementation of [DeviceDiscovery].
 ///
 /// Discovers available ios devices and chooses one to work with.
@@ -137,6 +142,9 @@ class IosDevice implements Device {
   Future<bool> restart_device({ProcessManager processManager}) async {
     processManager ??= LocalProcessManager();
     try {
+      if (noRebootForbidList.contains(deviceId)) {
+        return true;
+      }
       await eval('idevicediagnostics', <String>['restart'], processManager: processManager);
     } on BuildFailedError catch (error) {
       logger.severe('device restart fails: $error');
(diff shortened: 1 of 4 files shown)


device_doctor/lib/src/ios_device.dart
```diff
@@ -14,6 +14,11 @@ import 'health.dart';
 import 'mac.dart';
 import 'utils.dart';
 
+/// Identifiers for devices that should never be rebooted.
+final Set<String> noRebootForbidList = <String>{
```
yusufm, line 18: This name sounds like a double negative, maybe noRebootList? or rebootForbidList?

yusufm (approved):

device_doctor/lib/src/ios_device.dart
keyonghan: Renamed to `noRebootList`.

device_doctor/pubspec.lock
```diff
@@ -28,14 +28,14 @@ packages:
       name: async
       url: "https://pub.dartlang.org"
     source: hosted
-    version: "2.5.0"
+    version: "2.5.0-nullsafety.2"
```
jmagman, line 31: These needed to be downgraded?

yusufm (approved):

device_doctor/pubspec.lock
keyonghan: These were auto updated when I upgraded the sdk version to sdk: ">=2.3.0 <3.0.0" from sdk: ">=2.1.0 <3.0.0"
(2.3.0 is needed to support Set literal)
Do we see any concerns here?

jmagman: I wonder why upping the minimum dart sdk requirement would downgrade these packages (`2.5.0-nullsafety.2` is older than `2.5.0`).  You'd think if anything it would make you upgrade packages.

I don't really have any concerns I just don't get why it would happen.

jmagman: (the resolved sdk went from `>=2.12.0-0.0 <3.0.0` down to `>=2.10.0-78 <2.11.0`)

keyonghan: Ah, that's why. Just learned upping the dart sdk version could possibly lowering the real resolved sdk version.

PR #1144, opened 2021-04-06
[scheduler] Add scheduler field to target

CaseyHillers:
This allows for specifying a customer scheduler in the target. This will allow us to consolidate all the existing configs under `.ci.yaml` and make migrations simpler.

When this and https://github.com/flutter/cocoon/pull/1139 land, I'll follow up with a PR to download `.ci.yaml` and add the plumbing to add targets as tasks.

### DIFF
--- a/app_dart/lib/src/model/proto/internal/scheduler.proto
+++ b/app_dart/lib/src/model/proto/internal/scheduler.proto
@@ -31,6 +31,16 @@ message Target {
     // Name of the LUCI builder to trigger.
     optional string builder = 7;
     // Name of the scheduler to trigger this target.
-    // Defaults to being triggered by cocoon. Supports 'cocoon', 'luci'.
-    optional string scheduler = 8 [default = 'cocoon'];
+    // Defaults to being triggered by cocoon.
+    optional SchedulerSystem scheduler = 8 [default = cocoon];
 }
+
+// Schedulers supported in SchedulerConfig.
+// Next ID: 2
+enum SchedulerSystem {
+    // Cocoon will handle all actions for the target (initial trigger, retries).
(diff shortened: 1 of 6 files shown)


app_dart/lib/src/model/proto/internal/scheduler.proto
```diff
@@ -30,4 +30,7 @@ message Target {
     map<string, string> properties = 6;
     // Name of the LUCI builder to trigger.
     optional string builder = 7;
+    // Name of the scheduler to trigger this target.
+    // Defaults to being triggered by cocoon. Supports 'cocoon', 'luci'.
+    optional string scheduler = 8 [default = 'cocoon'];
```
christopherfujino, line 35: I"m guessing it's non-trivial to parse an enum out of a user-maintained yaml file?

CaseyHillers: It is trivial. I didn't think of using an enum. Updated :-)

app_dart/lib/src/model/proto/internal/scheduler.proto
```diff
@@ -30,4 +30,17 @@ message Target {
     map<string, string> properties = 6;
     // Name of the LUCI builder to trigger.
     optional string builder = 7;
+    // Name of the scheduler to trigger this target.
+    // Defaults to being triggered by cocoon.
+    optional SchedulerSystem scheduler = 8 [default = cocoon];
 }
+
+// Schedulers supported in SchedulerConfig.
+// Next ID: 2
```
christopherfujino, line 39: 3?

christopherfujino (approved):
nit about the comment, otherwise LGTM

PR #1145, opened 2021-04-06
wrap vars in double quotes to safeguard strings with spaces

christopherfujino:
for reference: https://tldp.org/LDP/abs/html/quotingvar.html


CaseyHillers (approved):
LGTM! Thank you for fixing this in the other runners as well!

PR #1146, opened 2021-04-07
Allow request API requests from internal accounts.

godofredoc:
There are several APIs that are called manually by gardeners. This
change allows to use gcloud auth to authenticate curl requests.

Bug: https://github.com/flutter/flutter/issues/43123

### DIFF
--- a/app_dart/lib/src/request_handling/authentication.dart
+++ b/app_dart/lib/src/request_handling/authentication.dart
@@ -103,7 +103,6 @@ class AuthenticationProvider {
   /// unauthenticated.
   Future<AuthenticatedContext> authenticate(HttpRequest request) async {
     final String agentId = request.headers.value('Agent-ID');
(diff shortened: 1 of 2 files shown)


app_dart/lib/src/request_handling/authentication.dart
```diff
@@ -191,7 +189,7 @@ class AuthenticationProvider {
 
       final String clientId = await config.oauthClientId;
       assert(clientId != null);
-      if (token.audience != clientId) {
+      if (token.audience != clientId && !token.email.endsWith('@google.com')) {
```
CaseyHillers, line 192: Can you update the dart doc with the gcloud auth info?

CaseyHillers (approved):
LGTM!

app_dart/lib/src/request_handling/authentication.dart
godofredoc: done

PR #1148, opened 2021-04-07
[app_dart] Exclude framework dev/devicelab/bin/tasks and dev/bots from no tests warning

jmagman:
Avoid "It looks like this pull request may not have tests" message for framework PRs that only change dart files in `dev/devicelab/bin/tasks` or `dev/bots`.

Avoid adding the message on PRs like:
https://github.com/flutter/flutter/pull/71953
https://github.com/flutter/flutter/pull/75736
https://github.com/flutter/flutter/pull/76554
etc

### DIFF
--- a/app_dart/lib/src/request_handlers/github_webhook.dart
+++ b/app_dart/lib/src/request_handlers/github_webhook.dart
@@ -217,7 +217,9 @@ class GithubWebhook extends RequestHandler<Body> {
     bool needsTests = false;
 
     await for (PullRequestFile file in files) {
-      if (file.filename.endsWith('.dart')) {
+      if (file.filename.endsWith('.dart') &&
+          !file.filename.startsWith('dev/devicelab/') &&
+          !file.filename.startsWith('dev/bots/')) {
         needsTests = true;
       }
       if (file.filename.endsWith('_test.dart')) {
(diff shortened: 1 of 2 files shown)


app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -217,7 +217,9 @@ class GithubWebhook extends RequestHandler<Body> {
     bool needsTests = false;
 
     await for (PullRequestFile file in files) {
-      if (file.filename.endsWith('.dart')) {
+      if (file.filename.endsWith('.dart') &&
+          !file.filename.startsWith('dev/devicelab/') &&
```
CaseyHillers, line 221: What about `dev/devicelab/bin/` and `dev/devicelab/lib/tasks/`?

There's still framework code where we would tests written

jmagman: How about I just make this `dev/devicelab/bin/tasks` and also remove `dev/bots/`?  Those are the actual integration tests.

CaseyHillers: SGTM. Is the goal to one day have tests for `dev/bots` code?

jmagman: Not a goal of mine 😉  You're saying you're cool with me excluding `dev/bots/`?  I feel like the integration themselves exercise that code.

CaseyHillers: Oh I see what you mean. Yes i'm okay with removing `dev/bots` from asking for tests

CaseyHillers (approved):
LGTM when the devicelab path is updated

christopherfujino (approved):
lgtm

PR #1150, opened 2021-04-08
[app_dart] Differentiate 404s from other HTTP exceptions on GitHub files

CaseyHillers:
* Refactor `remoteFileContent` to `githubFileContent`
  - Only return response content if file was a 200 response
  - Throw NotFoundException on 404s
  - Throw HttpException on remaining exceptions
  - Allows for customized responses where it may be okay if a config file does not exist, otherwise it should be a hard failure
* If DeviceLab `manifest.yaml` 404s, return default response
* Retrieve `.ci.yaml` and store it in cache for future use

## Issues

### DIFF
--- a/app_dart/lib/src/foundation/utils.dart
+++ b/app_dart/lib/src/foundation/utils.dart
@@ -10,8 +10,11 @@ import 'dart:typed_data';
 import 'package:appengine/appengine.dart';
 import 'package:github/github.dart';
 import 'package:googleapis/bigquery/v2.dart';
+import 'package:meta/meta.dart';
+import 'package:retry/retry.dart';
 
 import '../foundation/typedefs.dart';
+import '../request_handling/exceptions.dart';
 import '../service/github_service.dart';
 import '../service/luci.dart';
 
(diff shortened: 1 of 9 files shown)


app_dart/lib/src/foundation/utils.dart
```diff
@@ -29,48 +32,71 @@ Duration twoSecondLinearBackoff(int attempt) {
...
+      return await utf8.decoder.bind(clientResponse).join();
+    } else if (status == HttpStatus.notFound) {
+      throw NotFoundException('HTTP $status: $url');
+    } else {
+      log.warning('HTTP $status: $url');
+      throw HttpException('HTTP $status: $url');
     }
   } finally {
-    client.close(force: true);
+    client.close();
```
christopherfujino, line 78: why not `force: true`?

app_dart/lib/src/foundation/utils.dart
```diff
@@ -29,48 +32,71 @@ Duration twoSecondLinearBackoff(int attempt) {
...
+}) async {
+  String content;
+  try {
+    content = await githubFileContent(
+      '/flutter/cocoon/master/app_dart/dev/branches.txt',
+      httpClientProvider: httpClientProvider,
+      log: log,
+      retryOptions: retryOptions,
+    );
+  } on HttpException {
```
christopherfujino, line 96: In what situation would this happen? And should we trigger some sort of alert if we hit this?

CaseyHillers: This would only happen during a GitHub CDN outage (which we haven't faced issues with). The implementers of this function don't catch this error, so it would surface as an increased error rate in cocoon (we already use error tracking dashboards)

app_dart/lib/src/foundation/utils.dart
CaseyHillers: Typo - Fixed now

app_dart/lib/src/service/scheduler.dart
```diff
@@ -200,20 +208,68 @@ class Scheduler {
...
 
+    final SchedulerConfig schedulerConfig = await getSchedulerConfig(commit);
+    log.debug('Loaded scheduler config $schedulerConfig');
+
     return tasks;
   }
 
+  /// Load in memory the `.ci.yaml`.
+  Future<SchedulerConfig> getSchedulerConfig(Commit commit, {RetryOptions retryOptions}) async {
+    final String ciPath = '${commit.repository}/${commit.sha}/.ci.yaml';
```
keyonghan, line 219: Is a preceding `/`needed?

app_dart/lib/src/service/scheduler.dart
```diff
@@ -200,20 +208,68 @@ class Scheduler {
...
+  ///
   // TODO(chillers): Remove when DeviceLab has migrated to LUCI. https://github.com/flutter/flutter/projects/151
   @visibleForTesting
-  Future<YamlMap> loadDevicelabManifest(Commit commit) async {
-    final String path = '/${commit.repository}/${commit.sha}/dev/devicelab/manifest.yaml';
+  Future<YamlMap> loadDevicelabManifest(
+    Commit commit, {
+    RetryOptions retryOptions,
+  }) async {
+    final String path = '${commit.repository}/${commit.sha}/dev/devicelab/manifest.yaml';
```
keyonghan, line 259: Ditto. a `/` needed?

christopherfujino: this is actually the format that `githubFileContent` expects.

CaseyHillers: I updated the format as if everything passes a leading slash, `githubFileContent` should just add it.

app_dart/lib/src/service/scheduler.dart
```diff
@@ -200,20 +208,68 @@ class Scheduler {
       ));
     });
 
+    final SchedulerConfig schedulerConfig = await getSchedulerConfig(commit);
```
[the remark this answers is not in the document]
CaseyHillers: We retry HttpExceptions. I added back the fails safes for *_builder.json and .ci.yaml to not fail if we can't load those configs with logging to see how common it is.

app_dart/lib/src/foundation/utils.dart
CaseyHillers: If `NotFoundException`, we should create an empty config. This case is expected when we remove the existing configs (devicelab/manifest.yaml, *_builders.json) or when a release branch is created without .ci.yaml.

keyonghan (approved):
LGTM.

CaseyHillers:
This has been at 50% of prod traffic for 4 hours, and I haven't seen any issues (other than unrelated auth warnings https://github.com/flutter/cocoon/pull/1158). The `HttpException` flows require longer soak time, so i'll land this in the current state and follow up if we can remove the logic.

PR #1152, opened 2021-04-09
[app_flutter] Add refresh github commit button

CaseyHillers:
Add a refresh github commits button in the settings. This can help ensure cocoon is up to date without needing to do the cookie hack or launch a gcloud cli

example ui from settings


yusufm (approved):
LGTM

PR #1153, opened 2021-04-09
Add a composite index for flutter repo and creation date on Checklist.

yusufm:
Adding a composite index on Checklist entity for FlutterRepositoryPath and CreationDate, as a precursor to enabling display of different repo data on flutter dashboard.

### DIFF
--- a/app_dart/index.yaml
+++ b/app_dart/index.yaml
@@ -13,6 +13,12 @@ indexes:
   - name: Branch
   - name: CreateTimestamp
     direction: desc
+- kind: Checklist
+  properties:
+  - name: FlutterRepositoryPath
+  - name: Branch
+  - name: CreateTimestamp
+    direction: desc
 - kind: Task
   ancestor: yes
   properties:


app_dart/index.yaml
```diff
@@ -13,6 +13,11 @@ indexes:
   - name: Branch
   - name: CreateTimestamp
     direction: desc
+- kind: Checklist
```
yusufm, line 16: @CaseyHillers based on the prospective queries we will want to make, should we be adding the repo info to Task entities directly, versus using the property in Checklist?

CaseyHillers: It isn't necessary for get-status. It grabs tasks if they have the parent key of the checklist - https://cs.opensource.google/flutter/cocoon/+/869becfbe169ce10c561f507c8984e2233e1be51:app_dart/lib/src/service/datastore.dart;l=124

app_dart/index.yaml
```diff
@@ -13,6 +13,11 @@ indexes:
...
+  properties:
+  - name: FlutterRepositoryPath
+  - name: CreateTimestamp
```
CaseyHillers, line 19: Can we add branch?

yusufm: done

CaseyHillers (approved):
LGTM

PR #1156, opened 2021-04-12
add presubmit, postsubmit to targets

christopherfujino:


app_dart/lib/src/model/proto/internal/scheduler.pbenum.dart
```diff
@@ -1,6 +1,6 @@
 ///
 //  Generated code. Do not modify.
-//  source: lib/src/model/proto/internal/scheduler.proto
+//  source: scheduler.proto
```
christopherfujino, line 3: I'm guessing this was a change to the dart plugin. Tried running the compiler from different working directories, and got the same string.

app_dart/lib/src/model/proto/internal/scheduler.proto
```diff
@@ -33,6 +33,25 @@ message Target {
...
+    optional PhaseConfig presubmit = 9;
+    optional PhaseConfig postsubmit = 10;
+}
+
+// CI Phase (i.e. pre-submit vs post-submit).
+// Next ID: 3
+message PhaseConfig {
+    optional Pool pool = 1;
+    // Allows disabling tests not intended to run in this phase.
+    optional bool enabled = 2;
```
christopherfujino, line 45: in my original design doc, I had both `enabled` and `flaky` here, but it probably makes more sense to have `flaky` at the top level (a target that is flaky post-submit is also flaky pre-submit). I'm assuming we should skip flaky tests pre-submit?

app_dart/lib/src/model/proto/internal/scheduler.proto
```diff
@@ -33,6 +33,25 @@ message Target {
...
+}
+
+enum Pool {
```
CaseyHillers, line 48: We currently have 4 pools for Flutter:
```
luci.flutter.prod
luci.flutter.staging
luci.flutter.try
luci.flutter.tests
```

Should we always mirror from infradata to this proto?

app_dart/lib/src/model/proto/internal/scheduler.proto
CaseyHillers: Yes. `Target` allows setting `bringup` where it'll handle this logic.

Short term, presubmit will override and not schedule tasks marked `bringup=true`. Long term, it would be good to schedule them but not push checks to collect data.

christopherfujino: > Yes. `Target` allows setting `bringup` where it'll handle this logic.
> 
> Short term, presubmit will override and not schedule tasks marked `bringup=true`. Long term, it would be good to schedule them but not push checks to collect data.

SGTM.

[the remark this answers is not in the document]
christopherfujino: I would expect this to be true, I'll set that as the default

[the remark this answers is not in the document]
christopherfujino: done

app_dart/lib/src/model/proto/internal/scheduler.proto
christopherfujino: hmm, what's tests? and by "mirror", do you mean manually?

christopherfujino: also, where is that defined? I'll add a comment to the proto linking to it.

christopherfujino: it looks like luci.flutter.tests is only used for a single fuchsia test? https://cs.opensource.google/search?q=luci.flutter.tests&sq=&ss=flutter

CaseyHillers: @godofredoc do you know what the purpose of `luci.flutter.tests` is?

app_dart/lib/src/model/proto/internal/scheduler.proto
```diff
@@ -33,6 +33,25 @@ message Target {
...
     // Defaults to being triggered by cocoon.
     optional SchedulerSystem scheduler = 8 [default = cocoon];
+    optional PhaseConfig presubmit = 9;
+    optional PhaseConfig postsubmit = 10;
+}
+
+// CI Phase (i.e. pre-submit vs post-submit).
+// Next ID: 3
+message PhaseConfig {
+    optional Pool pool = 1;
```
CaseyHillers, line 43: Is there a use case for pool being different on post submit and presubmit runs?

christopherfujino: pre-submit would be try, and post-submit prod, right?

CaseyHillers: Yes, but what if a target has `bringup=true` we schedule those builds on staging? Then we wouldn't need target to know about the luci pools.

app_dart/lib/src/model/proto/internal/scheduler.proto
godofredoc: tests is a pool of independent devices that can be used either by try or prod tasks.

CaseyHillers: Thanks! Chris and I synced offline, for now we'll hide pool from `.ci.yaml` and leave that as a detail for the scheduler.

CaseyHillers (approved):
LGTM once the pool enum is removed

PR #1157, opened 2021-04-13
[app_dart] Vaccum github commits

CaseyHillers:
* Update RefreshGitHubCommits to work with the webhook scheduler
  * Look at all commits from 3 minutes ago to last 24 hours, and add any missing to Cocoon (previous logic was only recently missed tasks)
  * Add hourly cron job
  * Support any repo the scheduler supports (only flutter/flutter)


app_dart/lib/src/request_handlers/vacuum_github_commits.dart
```diff
@@ -0,0 +1,112 @@
...
+    final DatastoreService datastore = datastoreProvider(config.db);
+
+    for (String repository in Config.schedulerSupportedRepos) {
+      final String owner = repository.split('/').first;
+      final String repo = repository.split('/').last;
+      final RepositorySlug slug = RepositorySlug(owner, repo);
+      await _vacuumRepository(slug, datastore: datastore);
+    }
+
+    return Body.empty;
```
yusufm, line 47: always returning Body.empty?

CaseyHillers: This is a cron utility function where we only care if it passed or not (HTTP 200 versus 500). The information is still kept in the logs if we need to audit a run.

yusufm: If the return value doesn't matter, can we remove it, to simplify?

CaseyHillers: No as `RequestHandler` asserts the body is not null.

yusufm:
LGTM, but someone else should approve.

app_dart/lib/src/request_handlers/vacuum_github_commits.dart
```diff
@@ -0,0 +1,112 @@
...
+  final DatastoreServiceProvider datastoreProvider;
+
+  final Scheduler scheduler;
+
+  @override
+  Future<Body> get() async {
+    final DatastoreService datastore = datastoreProvider(config.db);
+
+    for (String repository in Config.schedulerSupportedRepos) {
+      final String owner = repository.split('/').first;
```
godofredoc, line 41: Should we start saving the owner and repo separately?

CaseyHillers: I like the idea of these being stored as a `RepositorySlug` as then it's simple to pull owner, repo, and fullName. I'll follow up with this refactor in a separate PR - https://github.com/flutter/flutter/issues/80531

app_dart/cron.yaml
```diff
@@ -7,6 +7,9 @@ cron:
 - description: clean up stale datastore records
   url: /api/vacuum-clean
   schedule: every 1 hours
+- description: retrieve missing commits
+  url: /api/vacuum-github-commits
```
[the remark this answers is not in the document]
CaseyHillers: Previously, only a cron job every minute for when we used the cron based method.

app_dart/bin/server.dart
```diff
@@ -78,11 +78,6 @@ Future<void> main() async {
       '/api/push-gold-status-to-github': PushGoldStatusToGithub(config, authProvider),
       '/api/push-engine-build-status-to-github': PushEngineStatusToGithub(config, authProvider, luciBuildService),
       '/api/refresh-chromebot-status': RefreshChromebotStatus(config, authProvider, luciBuildService),
-      '/api/refresh-github-commits': RefreshGithubCommits(
```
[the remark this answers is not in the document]
CaseyHillers: The use case is a bad Cocoon roll out or GitHub is down. Otherwise, the new scheduling logic has a higher consistency rate than the cron method (previously dropped ~1% of commits now 0%).

godofredoc:
Thanks for the explanations.

godofredoc (approved):

PR #1160, opened 2021-04-15
Ensure milo can find builds triggered by cocoon backend.

godofredoc:
This change makes the builds created by cocoon backend discoverable by
the milo queries.


app_dart/lib/src/request_handlers/query_github_graphql.dart
```diff
@@ -0,0 +1,56 @@
...
+class QueryGithubGraphql extends ApiRequestHandler<Body> {
+  const QueryGithubGraphql(
+    Config config,
+    AuthenticationProvider authenticationProvider,
+  ) : super(config: config, authenticationProvider: authenticationProvider);
+
+  @override
+  Future<Body> post() async {
+    final String requestDataString = String.fromCharCodes(requestBody);
+    log.error(requestDataString);
```
yusufm, line 30: should this be a lower log level (info, debug?)

CaseyHillers (commented):
How does this PR surface builds in MILO?

app_dart/lib/src/request_handlers/query_github_graphql.dart
godofredoc: uppps wrong file

> How does this PR surface builds in MILO?



> How does this PR surface builds in MILO?

wrong file, uploaded the correct one.

CaseyHillers (approved):
LGTM - Thanks for the fix!

PR #1162, opened 2021-04-16
[app_dart] Refactor presubmit logic to scheduler

CaseyHillers:
- Start of a refactor to internalize `LuciBuildService` and `BuildBucketClient` in `Scheduler` (i.e. request handlers should call `Scheduler` instead of these services)
  - All references to the presubmit calls were switched to call scheduler directly
- Create fakes for the scheduler internal services (`FakeLuciBuildService` and `FakeBuildBucketClient`)

# Issues
https://github.com/flutter/flutter/issues/77858

# Tests
Since this is a refactor, I moved tests to the area under test. This deduplicated some of the mock calls

# Follow up
- Incorporate `.ci.yaml` targets into presubmit
- Refactor remaining `LuciBuildService` logic under `Scheduler` (postsubmit logic)

### DIFF
--- a/app_dart/bin/server.dart
+++ b/app_dart/bin/server.dart
@@ -49,6 +49,7 @@ Future<void> main() async {
     final Scheduler scheduler = Scheduler(
       cache: cache,
       config: config,
+      luciBuildService: luciBuildService,
     );
 
     final Map<String, RequestHandler<dynamic>> handlers = <String, RequestHandler<dynamic>>{
@@ -65,8 +66,6 @@ Future<void> main() async {
       '/api/get-log': GetLog(config, authProvider),
       '/api/github-webhook-pullrequest': GithubWebhook(
         config,
-        buildBucketClient: buildBucketClient,
-        luciBuildService: luciBuildService,
         githubChecksService: githubChecksService,
         scheduler: scheduler,
       ),
@@ -94,7 +93,7 @@ Future<void> main() async {
       '/api/reset-try-task': ResetTryTask(
         config,
         authProvider,
-        luciBuildService,
+        scheduler,
       ),
       '/api/update-agent-health': UpdateAgentHealth(config, authProvider),
       '/api/update-agent-health-history': UpdateAgentHealthHistory(config, authProvider),
--- a/app_dart/lib/src/request_handlers/github_webhook.dart
+++ b/app_dart/lib/src/request_handlers/github_webhook.dart
@@ -15,9 +15,7 @@ import '../model/github/checks.dart';
 import '../request_handling/body.dart';
 import '../request_handling/exceptions.dart';
 import '../request_handling/request_handler.dart';
-import '../service/buildbucket.dart';
 import '../service/github_checks_service.dart';
-import '../service/luci_build_service.dart';
 import '../service/scheduler.dart';
 
 /// List of repos that require check for labels and tests.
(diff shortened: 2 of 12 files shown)


CaseyHillers:
Friendly ping

app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -76,7 +63,7 @@ class GithubWebhook extends RequestHandler<Body> {
           final CheckRunEvent checkRunEvent = CheckRunEvent.fromJson(
             jsonDecode(stringRequest) as Map<String, dynamic>,
           );
-          await githubChecksService.handleCheckRun(checkRunEvent, luciBuildService);
+          await scheduler.processCheckRun(checkRunEvent);
```
godofredoc, line 66: Thinking about moving forward to an eventual separation of the scheduler from the cocoon backend. Would it make sense to create a new webhook handler just for the scheduler?

app_dart/bin/server.dart
```diff
@@ -49,6 +49,7 @@ Future<void> main() async {
     final Scheduler scheduler = Scheduler(
       cache: cache,
       config: config,
+      luciBuildService: luciBuildService,
```
godofredoc, line 52: Would the use of luciBuildService be encapsulated inside the scheduler?

app_dart/test/src/service/fake_luci_build_service.dart
```diff
@@ -0,0 +1,26 @@
...
+class FakeLuciBuildService extends LuciBuildService {
+  FakeLuciBuildService(
+    Config config, {
+    BuildBucketClient buildbucket,
+    GithubChecksUtil githubChecksUtil,
+  }) : super(
+          config,
+          buildbucket ?? FakeBuildBucketClient(),
+          const ServiceAccountInfo(email: 'test-account'),
+          githubChecksUtil: githubChecksUtil ?? MockGithubChecksUtil(),
```
godofredoc, line 24: The combination of fakes and mock may be a little bit confusing from someone using this class. Should the githubChecksUtil also be a fake?

app_dart/bin/server.dart
CaseyHillers: Yes, the goal is these are internal services that request handlers shouldn't know about. This helps minimize the logic in the handlers

app_dart/lib/src/request_handlers/github_webhook.dart
CaseyHillers: I think the proper fix is `github_webhook` shouldn't execute logic. The more logic we add to webhooks, the more likely a PR is to break all webhook functionality. We should be pushing the events to `TaskQueue` to be processed later (almost immediately, but we can retry if a failure occurs).

app_dart/lib/src/service/github_checks_service.dart
```diff
@@ -75,20 +75,6 @@ class GithubChecksService {
     }
   }
 
-  /// Reschedules a failed build using a [CheckRunEvent]. The CheckRunEvent is
-  /// generated when someone clicks the re-run button from a failed build from
```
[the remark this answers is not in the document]
CaseyHillers: Yes, it was moved into `Scheduler` as `processCheckRun`

app_dart/test/src/service/fake_luci_build_service.dart
CaseyHillers: Ideally, I want to replace all mocks with fakes so `app_dart` can use NNBD. However, that's a large technical debt process. Is it okay to leave this for now? https://github.com/flutter/flutter/issues/80990

app_dart/lib/src/request_handlers/github_webhook.dart
godofredoc: The comment is more about separating logic in the webhook independently of how it is processed. Having logic in the webhook for label management, autosubmit bot, scheduling, etc makes it error prone causing someone adding a label processing logic to potentially break scheduling. Not a requirement for this PR but we should definitely start thinking about independent services for very focused functionality e.g a scheduler service just doing scheduling and nothing else.

app_dart/test/src/service/fake_luci_build_service.dart
godofredoc: sounds good, thanks!

godofredoc (approved):
LGTM

app_dart/lib/src/request_handlers/github_webhook.dart
CaseyHillers: That makes sense, and I believe our concerns go hand in hand. With a microservice approach, we would still want some way to retry failed webhooks. The microservices would reduce the impact a bad change has to other parts of the system.

PR #1165, opened 2021-04-22
Prevent retries on tasks already scheduled or running.

godofredoc:
When a task has failed and the dashboard latency is high multiple people
can retry the same build multiple times wasting resources and preventing
tasks with lower priority from running.

Bug: https://github.com/flutter/flutter/issues/80918

### DIFF
--- a/app_dart/lib/src/request_handlers/reset_prod_task.dart
+++ b/app_dart/lib/src/request_handlers/reset_prod_task.dart
@@ -8,10 +8,12 @@ import 'package:appengine/appengine.dart';
 import 'package:cocoon_service/cocoon_service.dart';
 import 'package:cocoon_service/src/model/appengine/commit.dart';
 import 'package:cocoon_service/src/model/appengine/task.dart';
+import 'package:cocoon_service/src/model/luci/buildbucket.dart';
 import 'package:cocoon_service/src/request_handling/exceptions.dart';
 import 'package:cocoon_service/src/service/luci.dart';
 import 'package:cocoon_service/src/service/luci_build_service.dart';
 import 'package:gcloud/db.dart';
+import 'package:github/github.dart';
 import 'package:meta/meta.dart';
 
 import '../datastore/config.dart';
@@ -45,6 +47,8 @@ class ResetProdTask extends ApiRequestHandler<Body> {
     final String encodedKey = requestData[taskKeyParam] as String;
     final ClientContext clientContext = authContext.clientContext;
     final KeyHelper keyHelper = KeyHelper(applicationContext: clientContext.applicationContext);
+    final String owner = request.uri.queryParameters['owner'] ?? 'flutter';
+    final String repo = request.uri.queryParameters['repo'] ?? '';
     final Key<int> key = keyHelper.decode(encodedKey) as Key<int>;
     log.info('Rescheduling task with Key: ${key.id}');
     final Task task = (await datastore.lookupByKey<Task>(<Key<int>>[key])).single;
@@ -65,6 +69,12 @@ class ResetProdTask extends ApiRequestHandler<Body> {
           .where((LuciBuilder builder) => builder.taskName == task.name)
           .map((LuciBuilder builder) => builder.name)
           .single;
+    }
+    final RepositorySlug slug = RepositorySlug(owner, repo);
+    final Build currentBuild = await luciBuildService.getBuild(slug, commit.sha, builder, 'prod');
+    final List<Status> noReschedule = <Status>[Status.started, Status.scheduled, Status.success];
+    if (currentBuild != null && noReschedule.contains(currentBuild.status)) {
+      throw const ConflictException();
     }
     await luciBuildService.rescheduleProdBuild(
       commitSha: commit.sha,
--- a/app_dart/lib/src/request_handlers/reset_prod_task.dart
+++ b/app_dart/lib/src/request_handlers/reset_prod_task.dart
@@ -39,6 +39,8 @@ class ResetProdTask extends ApiRequestHandler<Body> {
   final LuciBuildService luciBuildService;
 
   static const String taskKeyParam = 'Key';
+  static const String ownerParam = 'Owner';
+  static const String repoParam = 'Repo';
 
   @override
   Future<Body> post() async {
@@ -47,8 +49,8 @@ class ResetProdTask extends ApiRequestHandler<Body> {
     final String encodedKey = requestData[taskKeyParam] as String;
     final ClientContext clientContext = authContext.clientContext;
     final KeyHelper keyHelper = KeyHelper(applicationContext: clientContext.applicationContext);
-    final String owner = request.uri.queryParameters['owner'] ?? 'flutter';
-    final String repo = request.uri.queryParameters['repo'] ?? '';
+    final String owner = requestData[ownerParam] as String ?? 'flutter';
+    final String repo = requestData[repoParam] as String ?? '';
     final Key<int> key = keyHelper.decode(encodedKey) as Key<int>;
     log.info('Rescheduling task with Key: ${key.id}');
     final Task task = (await datastore.lookupByKey<Task>(<Key<int>>[key])).single;
@@ -72,6 +74,7 @@ class ResetProdTask extends ApiRequestHandler<Body> {
     }
     final RepositorySlug slug = RepositorySlug(owner, repo);
     final Build currentBuild = await luciBuildService.getBuild(slug, commit.sha, builder, 'prod');
+    log.info('Owner: $owner, Repo: $repo, Builder: $builder');
     final List<Status> noReschedule = <Status>[Status.started, Status.scheduled, Status.success];
     if (currentBuild != null && noReschedule.contains(currentBuild.status)) {
       throw const ConflictException();
--- a/app_dart/lib/src/request_handlers/reset_prod_task.dart
+++ b/app_dart/lib/src/request_handlers/reset_prod_task.dart
@@ -73,10 +73,15 @@ class ResetProdTask extends ApiRequestHandler<Body> {
           .single;
     }
     final RepositorySlug slug = RepositorySlug(owner, repo);
-    final Build currentBuild = await luciBuildService.getBuild(slug, commit.sha, builder, 'prod');
-    log.info('Owner: $owner, Repo: $repo, Builder: $builder');
+    final Iterable<Build> currentBuilds = await luciBuildService.getProdBuilds(slug, commit.sha, builder, repo);
     final List<Status> noReschedule = <Status>[Status.started, Status.scheduled, Status.success];
-    if (currentBuild != null && noReschedule.contains(currentBuild.status)) {
+    final Build build = currentBuilds.firstWhere(
+      (Build element) => noReschedule.contains(element.status),
+      orElse: () => null,
+    );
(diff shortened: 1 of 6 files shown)


app_dart/lib/src/request_handlers/reset_prod_task.dart
```diff
@@ -66,6 +72,22 @@ class ResetProdTask extends ApiRequestHandler<Body> {
...
+    final List<Status> noReschedule = <Status>[Status.started, Status.scheduled, Status.success];
+    final Build build = currentBuilds.firstWhere(
+      (Build element) {
+        log.info('Found build status: ${element.status} inNoReschedule ${noReschedule.contains(element.status)}');
+        return noReschedule.contains(element.status);
+      },
+      orElse: () => null,
+    );
+    log.info(
+        'Owner: $owner, Repo: $repo, Builder: $builder, CommitSha: ${commit.sha}, CurrentBuild: $build, CurrentBuild: $currentBuilds');
```
keyonghan, line 86: Nit: Build: $build, CurrentBuilds: $currentBuilds

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -45,35 +45,62 @@ class LuciBuildService {
...
     String builderName,
+  ) async {
+    final Map<String, List<String>> tags = <String, List<String>>{
+      'buildset': <String>['sha/git/$commitSha'],
+      'user_agent': const <String>['flutter-cocoon'],
+    };
+    return getBuilds(slug, commitSha, builderName, 'try', tags);
+  }
+
+  /// Returns an Iterable of try BuildBucket build for a given Github [slug], [commitSha],
```
keyonghan, line 62: Nit: doc update.

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -45,35 +45,62 @@ class LuciBuildService {
...
+    String builderName,
+    String repo,
+  ) async {
+    final Map<String, List<String>> tags = <String, List<String>>{
+      'buildset': <String>['commit/gitiles/chromium.googlesource.com/external/github.com/flutter/$repo/+/$commitSha'],
+    };
+    return getBuilds(slug, commitSha, builderName, 'prod', tags);
+  }
+
+  /// Returns a BuildBucket build for a given Github [slug], [commitSha],
```
keyonghan, line 76: build list?

app_dart/lib/src/request_handlers/reset_prod_task.dart
```diff
@@ -45,6 +49,8 @@ class ResetProdTask extends ApiRequestHandler<Body> {
     final String encodedKey = requestData[taskKeyParam] as String;
     final ClientContext clientContext = authContext.clientContext;
     final KeyHelper keyHelper = KeyHelper(applicationContext: clientContext.applicationContext);
+    final String owner = requestData[ownerParam] as String ?? 'flutter';
```
CaseyHillers, line 52: Have you thought about moving these to query params? If this was a GET request. we could use the CacheRequestHandler to rate limit this handler with specific args

app_dart/lib/src/request_handlers/reset_prod_task.dart
```diff
@@ -45,6 +49,8 @@ class ResetProdTask extends ApiRequestHandler<Body> {
...
+    final String repo = requestData[repoParam] as String ?? '';
     final Key<int> key = keyHelper.decode(encodedKey) as Key<int>;
     log.info('Rescheduling task with Key: ${key.id}');
     final Task task = (await datastore.lookupByKey<Task>(<Key<int>>[key])).single;
```
CaseyHillers, line 56: Instead of passing `Owner` and `Repo`, you can look up the parent Commit of task to get these values.

app_dart/test/request_handlers/reset_prod_task_test.dart
```diff
@@ -130,6 +169,75 @@ void main() {
...
+          builderName: 'Windows');
+      config.db.values[task.key] = task;
+      config.db.values[commit.key] = commit;
+      when(mockLuciBuildService.getProdBuilds(any, any, any, any)).thenAnswer((_) async {
+        return <Build>[startedBuild];
+      });
+      expect(() => tester.post(handler), throwsA(isA<ConflictException>()));
+    });
+
+    test('Fails if task already succeded', () async {
```
keyonghan, line 202: Nit: succeeded.

app_dart/lib/src/request_handlers/reset_prod_task.dart
```diff
@@ -66,6 +72,22 @@ class ResetProdTask extends ApiRequestHandler<Body> {
           .map((LuciBuilder builder) => builder.name)
           .single;
     }
+    final RepositorySlug slug = RepositorySlug(owner, repo);
+    final Iterable<Build> currentBuilds = await luciBuildService.getProdBuilds(slug, commit.sha, builder, repo);
+    final List<Status> noReschedule = <Status>[Status.started, Status.scheduled, Status.success];
+    final Build build = currentBuilds.firstWhere(
```
keyonghan, line 78: Do we need to check every build in `currentBuilds`, or checking the first is enough (similar to try case)?

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -45,35 +45,62 @@ class LuciBuildService {
...
+  /// [builderName] and [bucket].
```
CaseyHillers, line 77: nit: and [tags]

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -45,35 +45,62 @@ class LuciBuildService {
...
+  /// Returns an Iterable of try BuildBucket build for a given Github [slug], [commitSha],
   /// [builderName].
-  Future<Build> getTryBuild(
+  Future<Iterable<Build>> getTryBuilds(
     github.RepositorySlug slug,
     String commitSha,
     String builderName,
+  ) async {
+    final Map<String, List<String>> tags = <String, List<String>>{
+      'buildset': <String>['sha/git/$commitSha'],
```
CaseyHillers, line 56: Should we explicitly add a tag for differentiate our try and prod builds?

app_dart/test/request_handlers/reset_prod_task_test.dart
```diff
@@ -49,6 +53,33 @@ void main() {
         ),
         sha: '7d03371610c07953a5def50d500045941de516b8',
       );
+      startedBuild = const Build(
```
CaseyHillers, line 56: If these are const, can they be instantiated as static const outside of setup?

CaseyHillers (commented):
Can you rebase with master? This may have merge conflicts in the test code with https://github.com/flutter/cocoon/pull/1162

app_dart/lib/src/request_handlers/reset_prod_task.dart
godofredoc: Updated and removed currentBuilds as it is just a list and its elements are being logged above.

app_dart/lib/src/service/luci_build_service.dart
godofredoc: done

app_dart/lib/src/service/luci_build_service.dart
godofredoc: updated.

app_dart/test/request_handlers/reset_prod_task_test.dart
godofredoc: updated

app_dart/lib/src/request_handlers/reset_prod_task.dart
godofredoc: The first one is enough. Ideally if there is a started, succeeded and or queued task we should not trigger a new one.

godofredoc:
> #1162

it was rebased yesterday night.

app_dart/lib/src/request_handlers/reset_prod_task.dart
godofredoc: Not possible, one of the reasons for this change is to allow engine retries which do not have entries in the data store.

CaseyHillers: This is pulling builderName from task, but we currently do not store engine tasks. Do we need to add `taskKey || builderName` as the param logic?

app_dart/lib/src/request_handlers/reset_prod_task.dart
godofredoc: The reason it was added as post is because per convention rest apis modifying state are usually post and get requests do not modify state.

app_dart/lib/src/service/luci_build_service.dart
godofredoc: done.

app_dart/lib/src/service/luci_build_service.dart
godofredoc: we usually use the bucket for that pupose.

app_dart/test/request_handlers/reset_prod_task_test.dart
godofredoc: done

app_dart/lib/src/request_handlers/reset_prod_task.dart
godofredoc: yes, the support for engine will be completed in a separate PR

CaseyHillers (approved):
LGTM

app_flutter/lib/service/appengine_cocoon.dart
```diff
@@ -164,6 +164,9 @@ class AppEngineCocoonService implements CocoonService {
         },
         body: jsonEncode(<String, String>{
           'Key': task.key.child.name,
+          // This prepares this api to support different repos.
+          'Owner': 'flutter',
```
CaseyHillers, line 168: nit: I'm hesitant to hard code repo logic. Can we pass Commit to this function and pull the values from it?

godofredoc: This is temporarily only,  

> nit: I'm hesitant to hard code repo logic. Can we pass Commit to this function and pull the values from it?
We may not be able to get the commit if we try to retry a task that was skipped by the scheduler. Is there any other way to get the repo from the commit?

godofredoc: > 
> > nit: I'm hesitant to hard code repo logic. Can we pass Commit to this function and pull the values from it?
> > 
We may not be able to get the commit if we try to retry a task that was skipped by the scheduler. Is there any other way to get the repo from the commit?

CaseyHillers: In the flutter app, we're guaranteed to always have a commit, but not a task. 

1. If the scheduler is skipping tasks, I would assume we would not want to allow retry/manual executions of those tasks
2. The frontend could backfill missing tasks with the information from task grid to ensure all tasks are non-null  - https://cs.opensource.google/flutter/cocoon/+/master:app_flutter/lib/widgets/task_grid.dart;l=78

godofredoc: 1) We do, the use case is: a task fails consistently and we need manually bisect n tasks that were skipped by the scheduler.
2) Sounds overkilling for not hardcoding the repo temporarily. Once the dashboard supports multiple repos we can use the selected repo instead of the hardcoded value.

keyonghan (approved):
LGTM

PR #1166, opened 2021-04-23
Update server.dart

yusufm:
Moving some internal documentation into code.

### DIFF
--- a/app_dart/bin/server.dart
+++ b/app_dart/bin/server.dart
@@ -53,9 +53,42 @@ Future<void> main() async {
     );
 
     final Map<String, RequestHandler<dynamic>> handlers = <String, RequestHandler<dynamic>>{
+      /// /api/append-log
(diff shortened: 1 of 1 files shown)


CaseyHillers (changes requested):
See the dartdocs style guide at https://dart.dev/guides/language/effective-dart/documentation
 These should be class level dartdocs

yusufm:
> See the dartdocs style guide at https://dart.dev/guides/language/effective-dart/documentation
> These should be class level dartdocs

I was thinking about that... in this case I was trying to keep the docs as close to each api as possible. Since there are many of them in that function. I can move it up to a big block, but I'm not sure thats going to be better for ongoing maintenance.

yusufm:
Abandoning this, since a lot has been cleaned up. Will open a new PR for the docs.

PR #1170, opened 2021-04-24
Remove agent health update cron job.

yusufm:
Agents have been deprecated and this job is generating errors, disable the cron job.

### DIFF
--- a/app_dart/cron.yaml
+++ b/app_dart/cron.yaml
@@ -4,9 +4,6 @@ cron:
 - description: refresh chromebot build status
   url: /api/refresh-chromebot-status
   schedule: every 3 minutes
-- description: clean up stale datastore records
-  url: /api/vacuum-clean
-  schedule: every 1 hours
 - description: retrieve missing commits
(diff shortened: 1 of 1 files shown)


app_dart/cron.yaml
```diff
@@ -22,9 +22,6 @@ cron:
 - description: check for mergeable commits waiting for the tree to go green
   url: /api/check-waiting-pull-requests
   schedule: every 5 minutes
-- description: insert agent health status to bigquery
```
CaseyHillers, line 25: We should remove the request handler as well

yusufm: Will do that in following PR that will target app_dart code.

CaseyHillers, line 25: vacuum clean is another agent cron to remove

CaseyHillers (approved):
LGTM!

Are you able to push these cron changes?

yusufm:
Thanks, yes, I was able to push the cron changes.

PR #1171, opened 2021-04-26
Add functionality to retry engine builds using curl commands.

godofredoc:
This is to allow the engine gardeners to re-trigger failed builds even
though they are not Flutter LUCI admins.

Bug: https://github.com/flutter/flutter/issues/81102

### DIFF
--- a/app_dart/lib/src/request_handlers/reset_prod_task.dart
+++ b/app_dart/lib/src/request_handlers/reset_prod_task.dart
@@ -56,6 +56,15 @@ class ResetProdTask extends ApiRequestHandler<Body> {
     RepositorySlug slug;
     String builder = requestData[builderParam] as String ?? '';
     Task task;
+    Commit commit;
+
+    if (encodedKey.isNotEmpty) {
+      // Check params required for dashboard.
+      checkRequiredParameters(<String>[taskKeyParam]);
+    } else {
+      // Checks params required when this API is called with curl.
+      checkRequiredParameters(<String>[commitShaParam, builderParam, repoParam]);
+    }
 
     if (encodedKey.isNotEmpty) {
       // Request coming from the dashboard.
@@ -65,7 +74,7 @@ class ResetProdTask extends ApiRequestHandler<Body> {
       if (task.status == 'Succeeded') {
         return Body.empty;
       }
-      final Commit commit = await datastore.db.lookupValue<Commit>(task.commitKey, orElse: () {
+      commit = await datastore.db.lookupValue<Commit>(task.commitKey, orElse: () {
         throw BadRequestException('No such commit: ${task.commitKey}');
       });
       slug = commit.slug;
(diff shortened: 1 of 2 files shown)


app_dart/lib/src/request_handlers/reset_prod_task.dart
```diff
@@ -41,39 +41,54 @@ class ResetProdTask extends ApiRequestHandler<Body> {
...
+        builder = builders
+            .where((LuciBuilder builder) => builder.taskName == task.name)
+            .map((LuciBuilder builder) => builder.name)
+            .single;
+      }
+    } else {
+      // Request not coming from dashboard means we need to create everything from
+      // parameters.
+      if (commitSha.isEmpty || builder.isEmpty || repo.isEmpty) {
+        throw const BadRequestException(
```
CaseyHillers, line 85: `checkRequiredParameters` makes this easier to read (see https://cs.opensource.google/flutter/cocoon/+/master:app_dart/lib/src/request_handlers/update_task_status.dart)

app_dart/test/request_handlers/reset_prod_task_test.dart
```diff
@@ -237,10 +261,6 @@ void main() {
       expect(task.attempts, equals(1));
     });
 
-    test('Fails if task does not exist', () async {
```
CaseyHillers, line 240: Isn't this still true when the extra params are not passed?

app_dart/lib/src/request_handlers/reset_prod_task.dart
```diff
@@ -41,39 +41,54 @@ class ResetProdTask extends ApiRequestHandler<Body> {
...
+      if (builder == null) {
+        final List<LuciBuilder> builders = await config.luciBuilders('prod', slug);
+        builder = builders
+            .where((LuciBuilder builder) => builder.taskName == task.name)
+            .map((LuciBuilder builder) => builder.name)
+            .single;
+      }
+    } else {
+      // Request not coming from dashboard means we need to create everything from
+      // parameters.
```
CaseyHillers, line 83: We could create an in memory engine commit to minimize the changes needed below:
`final Commit commit = Commit(repository: slug.fullName, commitSha, sha);`

app_dart/lib/src/request_handlers/reset_prod_task.dart
```diff
@@ -82,20 +97,24 @@ class ResetProdTask extends ApiRequestHandler<Body> {
...
+      commitSha: commitSha,
       builderName: builder,
+      repo: repo,
     );
-    task
-      ..status = Task.statusNew
-      ..startTimestamp = 0
-      ..attempts += 1;
-    await datastore.insert(<Task>[task]);
+    if (task != null) {
```
keyonghan, line 110: The logic allows retry framework build via curl as well, for which case the task is null and will not be updated in the datastore. Shall we enforce engine only, and restricting the framework retry to dashboard only?

app_dart/lib/src/request_handlers/reset_prod_task.dart
godofredoc: Done

app_dart/lib/src/request_handlers/reset_prod_task.dart
godofredoc: Forced an exception if we try to use curl with flutter repo.

app_dart/lib/src/request_handlers/reset_prod_task.dart
godofredoc: done

app_dart/test/request_handlers/reset_prod_task_test.dart
godofredoc: That use case is covered by "Re-schedule without all the parameters raises exception"

keyonghan (approved):
LGTM

PR #1178, opened 2021-04-28
Adds a graphql query service.

godofredoc:
This is required to validate github graphql queries that require
authentication.

### DIFF
--- a/app_dart/bin/server.dart
+++ b/app_dart/bin/server.dart
@@ -70,6 +70,9 @@ Future<void> main() async {
         githubChecksService: githubChecksService,
         scheduler: scheduler,
       ),
+
+      /// API to run authenticated graphql queries. It requires to pass the graphql query as the body
+      /// of a POST request.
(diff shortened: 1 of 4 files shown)


app_dart/bin/server.dart
```diff
@@ -69,6 +70,7 @@ Future<void> main() async {
         githubChecksService: githubChecksService,
         scheduler: scheduler,
       ),
+      '/api/query-github-graphql': QueryGithubGraphql(config, authProvider),
```
yusufm, line 73: Can you add api user doc here. I'm trying to add more documentation, and closest to where the api is defined seems like a good place.

Implementation doc can be added at the handler class.

yusufm: Note, that dart doc style says to add doc near the class def, but for api's I think it makes it more readable to have them intertwined here. But your call on where to add it.

yusufm (approved):
Approving, with doc request.

yusufm (approved):

PR #1179, opened 2021-04-28
[app_dart] Update readiness check to match Dart

CaseyHillers:
This should reduce issues in the daily promotion

### DIFF
--- a/app_dart/app.yaml
+++ b/app_dart/app.yaml
@@ -4,7 +4,12 @@
 
 runtime: custom
 env: flex
-service: default
+readiness_check:
+  check_interval_sec: 20
+  timeout_sec: 10
+  failure_threshold: 3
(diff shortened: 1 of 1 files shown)


app_dart/app.yaml
```diff
@@ -4,7 +4,12 @@
 
 runtime: custom
 env: flex
-service: default
+readiness_check:
```
keyonghan, line 7: Could you please share how Dart does this?

CaseyHillers: https://dart.googlesource.com/dart_ci/+/refs/heads/master/appengine/app.yaml

keyonghan: Thanks.

keyonghan (approved):
LGTM

PR #1181, opened 2021-04-28
Replaced accentColor with colorScheme.secondary

HansMuller:
Replaced Theme accentColor dependency with colorScheme.secondary because the accentColor theme properties are going to be deprecated. See https://github.com/flutter/flutter/pull/81336


CaseyHillers (approved):
LGTM!

Looks like you need to rebase with tip of tree to fix presubmit. Afterwards, you need to add `waiting for tree to go green` label to submit the PR (Cocoon is experimenting with bot only submissions)

PR #1185, opened 2021-04-29
[app_dart] Remove devicelab manifest scheduling

CaseyHillers:
Since there are no more cocoon agents, this scheduling can not run. Removing the dead code.


app_dart/test/request_handlers/vacuum_github_commits_test.dart
```diff
@@ -181,10 +169,9 @@ void main() {
           throw StateError('Commit failed');
         }
       };
-      scheduler.devicelabManifest = singleTaskManifestYaml;
       final Body body = await tester.get<Body>(handler);
       expect(db.values.values.whereType<Commit>().length, 3);
-      expect(db.values.values.whereType<Task>().length, 10);
+      expect(db.values.values.whereType<Task>().length, 8);
```
yusufm, line 174: For my information only: where do these tasks get populated?

yusufm (approved):
LGTM

app_dart/test/request_handlers/vacuum_github_commits_test.dart
CaseyHillers: They should all be created in `service/scheduler.dart` (after my recent refactors). Updates still need to be refactored into that class, and are spread in various parts of the codebase

PR #1186, opened 2021-04-29
Fix duplicates tasknames get status list

yusufm:
If there are failures for a task in the consecutive latest commits, we will duplicate the task name in the list of failed tasks being returned.

Ensure that we only return a task name once in the list of failing tasks.

Bug:https://github.com/flutter/flutter/issues/81402
Test:app_dart/dart test

### DIFF
--- a/app_dart/test/service/build_status_provider_test.dart
+++ b/app_dart/test/service/build_status_provider_test.dart
@@ -113,6 +113,14 @@ void main() {
         expect(status, BuildStatus.failure(const <String>['task2']));
       });
 
+      test('ensure failed task do not have duplicates when last consecutive commits contains red tasks', () async {
+        db.addOnQuery<Commit>((Iterable<Commit> results) => twoCommits);
+        db.addOnQuery<Task>((Iterable<Task> results) => middleTaskFailed);
+
+        final BuildStatus status = await buildStatusService.calculateCumulativeStatus();
+        expect(status, BuildStatus.failure(const <String>['task2']));
+      });
+
       test('ignores failures on flaky commits', () async {
         db.addOnQuery<Commit>((Iterable<Commit> results) => oneCommit);
         db.addOnQuery<Task>((Iterable<Task> results) => middleTaskFlakyFailed);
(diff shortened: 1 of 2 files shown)


app_dart/test/service/build_status_provider_test.dart
```diff
@@ -113,6 +113,14 @@ void main() {
         expect(status, BuildStatus.failure(const <String>['task2']));
       });
 
+      test('ensure failed task do not have duplicates when last consecutive commits contains red tasks', () async {
+        db.addOnQuery<Commit>((Iterable<Commit> results) => twoCommits);
+        db.addOnQuery<Task>((Iterable<Task> results) => middleTaskFailed);
```
godofredoc, line 118: Does this mean that if the last task passed but one in the middle failed we will still print it as a failure?

yusufm: In this case, any task should register as a failure, correctly. It just happens to be the middle task in the list of tasks that we provided.

The duplication on happens when starting with the latests commits, there are consecutive failures for a task, then that task is reported multiple times..

godofredoc: In this case G-R-R or G-R-G should be marked as a green. If the last task succeeded then it should not be printed as a failure.

The use case is a task is run and fails, later on someone retries the build and passes opening the tree.

yusufm: I'm confused, the test is based on:
|                | T1 | T2 | T3 |
|commit-1| G | R | G |
|commit-2| G | R | G |

So the top runs for T2 are fails, which means that our current system lists T2 twice. This change does not affect the behavior of the algorithm at all, it only dedupes the list of task names.

CaseyHillers: It would be clearer this is a bug fix if we fixed the algorithm as changing the datastructure hides this issue. 

We should mark the task as scanned once we have found it be in the failed state so we don't look at it again.

```
            } else if (_isFailed(task) || _isRerunning(task)) {
              failedTasks.add(task.name);
+              tasksInProgress[task.name] = false;
            }
```

yusufm: Done.

CaseyHillers (approved):
LGTM, thanks for the fix!

PR #1190, opened 2021-04-30
Remove stage

CaseyHillers:
- Remove stage from backend
  - Simplify `/api/public/get-status` endpoint
 - Remove stage from frontend
  - Simplify `TaskGrid` to be a 2D matrix (previous was multiple 2D matrices)
  - Remove some devicelab agent-isms in the code


app_flutter/lib/widgets/task_grid.dart
```diff
@@ -198,45 +198,43 @@ class _TaskGridState extends State<TaskGrid> {
     int commitCount = 0;
     for (final CommitStatus status in filteredStatuses) {
       commitCount += 1;
-      for (final Stage stage in status.stages) {
```
CaseyHillers, line 201: Here's an example of the general pay off we get from removing stages. We don't need 3 loops to scan the grid

app_flutter/lib/service/dev_cocoon.dart
```diff
@@ -101,36 +101,6 @@ class DevelopmentCocoonService implements CocoonService {
...
   }
 
-  static const List<String> _stages = <String>[
-    'cirrus',
-    'chromebot',
-    'devicelab',
-    'devicelab_win',
-    'devicelab_ios',
-  ];
-  static const List<int> _stageCount = <int>[
```
CaseyHillers, line 111: @Hixie I changed the cocoon fake data to be 120 LUCI tasks (we no longer serve these other tasks), SG?

yusufm (approved):
👍🏼

PR #1191, opened 2021-04-30
[app_dart] Log errors for failed scheduled builds

CaseyHillers:
Add `error` to `Response` to match [https://source.chromium.org/chromium/infra/infra/+/main:recipes-py/recipe_proto/go.chromium.org/luci/buildbucket/proto/builds_service.proto;l=159?q=BatchRequest&ss=chromium%2Finfra%2Finfra](upstream)

If this identifies try builds that are failing to schedule, I will follow up with implementing retry logic in the buildbucket service.

https://github.com/flutter/flutter/issues/80376

### DIFF
--- a/app_dart/lib/src/model/luci/buildbucket.dart
+++ b/app_dart/lib/src/model/luci/buildbucket.dart
@@ -7,6 +7,7 @@ import 'package:meta/meta.dart';
 
 import '../../request_handling/body.dart';
 import '../common/json_converters.dart';
+import '../google/grpc.dart' as grpc;
 
 part 'buildbucket.g.dart';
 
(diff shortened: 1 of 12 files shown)


app_dart/lib/src/model/luci/buildbucket.dart
```diff
@@ -126,6 +127,9 @@ class Response extends JsonBody {
   /// The [Build] response corresponding to a cancelBuild request.
   final Build cancelBuild;
 
+  /// Error code of the unsuccessful request.
+  final int error;
```
keyonghan, line 131: From the referenced link: `// Error code and details of the unsuccessful RPC.`. Is it the type of error `int`?

keyonghan: https://source.corp.google.com/piper///depot/google3/google/rpc/status.proto?q=google.rpc.Status

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -237,12 +237,18 @@ class LuciBuildService {
       maxAttempts: 3,
       delayFactor: Duration(seconds: 2),
     );
+    BatchResponse batchResponse;
     await r.retry(
       () async {
-        await buildBucketClient.batch(BatchRequest(requests: requests));
+        batchResponse = await buildBucketClient.batch(BatchRequest(requests: requests));
```
godofredoc, line 243: I took a quick look at this and it seems like the problem is that buildBucketClient.batch sometimes takes too long for a regular request(usually 60s timeout). We may need to move the call to this method in something that can run reliably for more than 60s.

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -237,12 +237,18 @@ class LuciBuildService {
...
       },
       retryIf: (Exception e) => e is BuildBucketException,
     );
+    for (Response response in batchResponse.responses) {
```
godofredoc, line 247: You may want to add some logs before/after running buildBucketClient.batch  to also catch request timeouts.

app_dart/lib/src/model/luci/buildbucket.dart
```diff
@@ -126,10 +127,43 @@ class Response extends JsonBody {
...
+  final ResponseStatus error;
+
   @override
   Map<String, dynamic> toJson() => _$ResponseToJson(this);
 }
 
+/// [Status] defines a logical error model that is suitable for
+/// different programming environments, including REST APIs and RPC APIs. It is
+/// used by [gRPC](https://github.com/grpc). Each [Status] message contains
+/// three pieces of data: error code, error message, and error details.
```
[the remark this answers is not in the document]
CaseyHillers: Since this is gRPC specific, I moved it into its own gRPC file

app_dart/lib/src/service/luci_build_service.dart
CaseyHillers: My expectation is that is already handled, is that correct?

```
// https://cs.opensource.google/flutter/cocoon/+/master:app_dart/lib/src/service/buildbucket.dart;l=83-86
// Null response or statusCode will move to the throw
if (response.statusCode < 300) {
      return responseFromJson(json.decode(rawResponse.substring(kRpcResponseGarbage.length)) as Map<String, dynamic>);
    }
    throw BuildBucketException(response.statusCode, rawResponse);
```

godofredoc (approved):

app_dart/lib/src/service/luci_build_service.dart
CaseyHillers: Ohh, in that case we need to push the work to a TaskQueue as GitHub sets the 10 second timeout on their request. I created a tracking issue in https://github.com/flutter/flutter/issues/81579

PR #1194, opened 2021-05-03
Remove agent relate stackdriver logging and taskattempt code from app_dart

yusufm:
Remove stackdriver logging and task attempt summary code which was used by now defunct agent code.

TEST=app_flutter/flutter test
TEST=app_dart/dart test


app_flutter/lib/widgets/task_overlay.dart
```diff
@@ -291,7 +290,6 @@ class TaskOverlayContents extends StatelessWidget {
                               'Task was run outside of devicelab',
                               style: Theme.of(context).textTheme.bodyText2,
                             ),
-                          if (QualifiedTask.fromTask(task).isDevicelab) TaskAttemptSummary(task: task),
                           if (QualifiedTask.fromTask(task).isLuci) LuciTaskAttemptSummary(task: task),
```
CaseyHillers, line 293: Can we rename `LuciTaskAttemptSummary` to `TaskAttemptSummary` now?

yusufm: I'm leaning to leaving it as is for simplicities sake, unless there is a compelling reason to rename.

CaseyHillers: With the code cleanup from last week all tasks are `LuciTask`, so I think it adds TD to leave this naming here (how does `LuciTask` differ from `Task`). I'll defer to @godofredoc for what we should do.

yusufm: With your changes to remove the stage concept, is QualifiedTask going away as well?

CaseyHillers: I removed `Stage` from the backend (+ whatever was needed in the frontend). Feel free to remove it if it no longer makes sense.

yusufm: .isLuci is based on the stage thats set when creating QualifiedTask. That still seems to be around after your change?

CaseyHillers: Yes, that's a frontend relic. My PR should have been labeled "remove stage from backend"

CaseyHillers (approved):
LGTM - Talked offline, Yusuf to follow up with separate PR with remaining cleanup

yusufm:
Updated goldens for flutter test.

PR #1196, opened 2021-05-03
[app_dart] move datastore/config to service/config

CaseyHillers:
`datastore` is an obsolete directory, and this class is just a service now.


yusufm (approved):
👍🏼

PR #1198, opened 2021-05-04
[app_dart] Set logger in vacuum github commits

CaseyHillers:
This is causing some errors when we do want to add a commit from the cron method as the scheduler logger is null


keyonghan (approved):
LGTM

PR #1199, opened 2021-05-05
[scheduler] Add ci yaml postsubmit targets to datastore

CaseyHillers:
- Add `enabled_branches` to target (for knowing which targets should be enabled
- Pass `ref` for presubmit targets to set presubmit commit branch
- Add postsubmit tasks to datastore
  - There are no existing postsubmit tasks in flutter/flutter/.ci.yaml, so this will not block the tree

# Issues

# Future work
- Update /api/refresh-chromebot-status to poll these builders as well (only necessary for those scheduled by luci during migration)


app_dart/lib/src/service/scheduler.dart
```diff
@@ -196,6 +198,60 @@ class Scheduler {
...
+    // 1. Add targets with local definition
+    final Iterable<Target> overrideBranchTargets = targets.where((Target target) => target.enabledBranches.isNotEmpty);
+    final Iterable<Target> enabledTargets =
+        overrideBranchTargets.where((Target target) => target.enabledBranches.contains(commit.branch));
+    filteredTargets.addAll(enabledTargets);
+
+    // 2. Add targets with global definition
+    if (config.enabledBranches.contains(commit.branch)) {
+      final Iterable<Target> defaultBranchTargets = targets.where((Target target) => target.enabledBranches.isEmpty);
+      filteredTargets.addAll(defaultBranchTargets);
```
christopherfujino, line 249: If both the target and the global config specify to enable the current branch, won't this lead to it the target showing up in `filteredTargets` twice?

CaseyHillers: No, but I added a test for it

app_dart/lib/src/service/scheduler.dart
```diff
@@ -196,6 +198,60 @@ class Scheduler {
...
+
+    // 1. Add targets with local definition
+    final Iterable<Target> overrideBranchTargets = targets.where((Target target) => target.enabledBranches.isNotEmpty);
+    final Iterable<Target> enabledTargets =
+        overrideBranchTargets.where((Target target) => target.enabledBranches.contains(commit.branch));
+    filteredTargets.addAll(enabledTargets);
+
+    // 2. Add targets with global definition
+    if (config.enabledBranches.contains(commit.branch)) {
+      final Iterable<Target> defaultBranchTargets = targets.where((Target target) => target.enabledBranches.isEmpty);
```
christopherfujino, line 248: Ah, I see, it's because you're first filtering for only targets that don't locally define it.

christopherfujino (approved):
LGTM

PR #1201, opened 2021-05-05
Adding some api docstrings

yusufm:
Adding some docstrings for dahsboard apis.


app_dart/bin/server.dart
```diff
@@ -46,6 +46,17 @@ Future<void> main() async {
...
+      /// Check if any github pull requests are marked with label
+      /// "waiting for tree to go green" and merge up to the number
+      /// specified within the handler.
+      ///
+      /// This api is called via cron job.
+      ///
+      /// GET: /api/check-waiting-pull-requests
+      ///
+      /// Response: Status 200 OK
       '/api/check-waiting-pull-requests': CheckForWaitingPullRequests(config, authProvider),
```
CaseyHillers, line 60: Can we move these docs to class level dart docs? IDEs show class level docs, which makes it easy to surface this info here and in the tests.

app_dart/bin/server.dart
```diff
@@ -82,29 +93,135 @@ Future<void> main() async {
...
       ),
+
+      /// Updates cocoon task status.
+      ///
+      /// This API updates task status in datastore.
+      ///
+      /// POST: /api-update-status
+      ///
+      /// Parameters:
+      ///   CommitBranch: (string in body). Branch of commit.
```
CaseyHillers, line 104: In general, we should have defined RPC protos which would enforce documentation of requests and responses (this doesn't have to be done this PR). WDYT?

app_dart/bin/server.dart
```diff
@@ -82,29 +93,135 @@ Future<void> main() async {
         authProvider,
         scheduler,
       ),
+
+      /// Updates cocoon task status.
+      ///
+      /// This API updates task status in datastore.
```
CaseyHillers, line 99: It also pushes performance metrics to skia-perf

app_dart/bin/server.dart
```diff
@@ -82,29 +93,135 @@ Future<void> main() async {
         authProvider,
         scheduler,
       ),
+
+      /// Updates cocoon task status.
```
CaseyHillers, line 97: I would call this a holistic task update as it also updates performance data

app_dart/bin/server.dart
```diff
@@ -82,29 +93,135 @@ Future<void> main() async {
...
+      ///
+      /// Requires authentication: Status: 401 Unauthorized
       '/api/update-task-status': UpdateTaskStatus(config, swarmingAuthProvider),
       '/api/vacuum-github-commits': VacuumGithubCommits(
         config,
         authProvider,
         scheduler: scheduler,
       ),
+
+      /// Returns status of the tree.
```
CaseyHillers, line 121:
```suggestion
      /// Returns status of the framework tree.
```

app_dart/bin/server.dart
```diff
@@ -82,29 +93,135 @@ Future<void> main() async {
...
+      ///
+      /// Returns serialized proto with enum representing the
+      /// status of the tree and list of offending tasks.
+      ///
+      /// GET: /api/public/build-status
+      ///
+      /// Parameters:
+      ///   branch: (string in query) default: 'master'. Name of the repo branch.
+      ///
+      /// Response: Status 200 OK
```
CaseyHillers, line 131: Can you link to the proto instead? For example, 2 doesn't make sense, but with context it maps to a logical enum

app_dart/bin/server.dart
```diff
@@ -82,29 +93,135 @@ Future<void> main() async {
...
+      ///        ]},
+      ///       },
+      ///     }
       '/api/public/get-status': CacheRequestHandler<Body>(
         cache: cache,
         config: config,
         delegate: GetStatus(config),
       ),
+
+      /// Return supported branches.
```
CaseyHillers, line 194:
```suggestion
Get supported branches to run infrastructure on the framework repo.
```

app_dart/bin/server.dart
```diff
@@ -82,29 +93,135 @@ Future<void> main() async {
...
+      ///   CommitBranch: (string in body). Branch of commit.
+      ///   CommitSha: (string in body). Sha of commit.
+      ///   BuilderName: (string in body). Name of the luci builder.
+      ///   NewStatus: (string in body) required. Status of the task.
+      ///   ResultData: (string in body) optional. Benchmark data.
+      ///   BenchmarkScoreKeys: (string in body) optional. Benchmark data.
+      ///
+      /// Response: Status 200 OK
+      ///
+      /// Requires authentication: Status: 401 Unauthorized
```
CaseyHillers, line 113: We should separate what requires authentication from the response as well (for example, update-task-status only allows the swarming bots to update tasks)

app_dart/bin/server.dart
yusufm: changed

app_dart/bin/server.dart
yusufm: Agreed, I am all for making the RPCs more defined.

app_dart/bin/server.dart
yusufm: done

app_dart/bin/server.dart
yusufm: Note sure what you mean here? Are you saying to add a note that the update-task-status is only authenticated by the bots? Can you show an example comment you want here?

app_dart/bin/server.dart
yusufm: It would be nice for the docs to show up in the IDEs and stuff. I'm torn between having each section next to the call itself, which might make it easier to navigate, versus a massive block of text for the class.

Would having a class level doc saying, "go look at inline comments" be effective?

app_dart/bin/server.dart
```diff
@@ -193,10 +193,10 @@ Future<void> main() async {
         delegate: GetStatus(config),
       ),
 
-      /// Return supported branches.
+      /// Get supported branches for the framework repo.
       ///
-      /// Returns branches that are supported by the dashboard
-      /// for the default repo.
+      /// Get list of supported branches to run infrastructure
```
yusufm, line 198: I wasn't sure what " run infrastructure" meant. Is it correct to say "run infrastructure tasks"?

See change to this comment and if it still makes sense.

CaseyHillers: sgtm

app_dart/bin/server.dart
yusufm: ok, will move the whole block to class level after all feedback is incorporated.

app_dart/bin/server.dart
yusufm: These are intended more as api user docs rather than implementation, I'm following what seems to be general api doc principles (https://docs.github.com/en/rest/reference/actions) where potential responses are stated for user.

Maybe I should add this under a "Responses" heading to make it more clear?

app_dart/bin/server.dart
yusufm: Is this what you meant?

CaseyHillers: Breadcrumbs are dart types in brackets

For example:
```dart
/// Returns [BuildStatusResponse]
```

app_dart/bin/server.dart
CaseyHillers: FWIW reviewing this PR would be easier if I could load the handlers inline

app_dart/bin/server.dart
CaseyHillers: Since this is a Flutter owned project, the Flutter and Dart style guides will take precedence. An example of how we can balance the two styles:

```
/// Returns [HttpStatus.ok] if a corresponding task exists, can be updated, and metrics center can push performance payload to skia-perf.
/// If [authProvider] cannot authenticate the current request, [HttpStatus.forbidden] will be returned.
```

As a side note, I'm hesitant to include auth info as it's dependent on the auth provider (some return both 401's and 403's)

CaseyHillers:
@googlebot I consent.

CaseyHillers (commented):
The latest git push appears to have added unrelated commits to this PR. Can you rebase your change on ToT?

yusufm:
@CaseyHillers rebased, try it now.

CaseyHillers:
Rebase with https://github.com/flutter/cocoon/pull/1212 to fix the latest Dart compiler issues

app_dart/bin/server.dart
godofredoc: Is great to have docs for all the different handlers. If refactoring is taking too much effort let's land this PR to add the docs and improve the docs in a different PR.

godofredoc (approved):

PR #1203, opened 2021-05-06
Make api/get-status repository aware

yusufm:
We now filter on an (optional) repo query parameter when returning list of commits from datastore.

Due to constraints in datatstore querying, application of filtering by repo will now reject any values that do not match the expected repo name, so if data rows contain any empty/null values for the repository, they will also be filtered out.

TEST=app_dart/dart test

### DIFF
--- a/app_dart/bin/server.dart
+++ b/app_dart/bin/server.dart
@@ -106,6 +106,22 @@ Future<void> main() async {
         authProvider,
         scheduler: scheduler,
       ),
+
+      /// Returns status of the tree.
+      ///
+      /// Returns serialized proto with enum representing the
+      /// status of the tree and list of offending tasks.
+      ///
+      /// GET: /api/public/build-status
+      ///
+      /// Parameters:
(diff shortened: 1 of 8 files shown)


app_dart/bin/server.dart
```diff
@@ -82,23 +82,119 @@ Future<void> main() async {
...
+
+      /// Returns task results for commits.
+      ///
+      /// Returns result details about each task in each checklist for every commit.
+      ///
+      /// GET: /api/public/get-status
+      ///
+      /// Parameters:
+      ///   branch: (string in query) default: 'master'. Name of the branch to filter on.
+      ///   repo: (string in query) default: 'flutter/flutter'. Name of repo to filter on.
```
yusufm, line 140: Diff in this particular file are a carry over from another PR. The only change specific to this PR is addition of this "repo:" line to the doc.

yusufm:
Closing this for another PR.

PR #1207, opened 2021-05-11
[scheduler] Set checkrun details page to LUCI build url when initially scheduled

CaseyHillers:
This is done on pubsub notifications from the build, but during the scheduled -> started phase redirects to the build dashboard. We can minimize this to show users the queued time.

Update the `GrpcStatus` check to look for non-0 exit codes and log them

# Issues

https://github.com/flutter/flutter/issues/80376

### DIFF
--- a/app_dart/lib/src/service/luci_build_service.dart
+++ b/app_dart/lib/src/service/luci_build_service.dart
@@ -147,7 +147,7 @@ class LuciBuildService {
     ]));
     final Iterable<Build> builds = batch.responses
         .map((Response response) => response.searchBuilds)
-        .expand((SearchBuildsResponse response) => response.builds ?? <Build>[]);
+        .expand((SearchBuildsResponse response) => response?.builds ?? <Build>[]);
     return Map<String, Build>.fromIterable(builds,
         key: (dynamic b) => b.builderId.builder as String, value: (dynamic b) => b as Build);
   }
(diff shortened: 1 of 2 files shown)


app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -249,8 +252,25 @@ class LuciBuildService {
...
+
+      final Build scheduleBuild = response.scheduleBuild;
+      // Tags are List<String> so we need to decode to a single int
+      final List<String> checkrunIdStrings = scheduleBuild.tags['github_checkrun'];
+      final int checkRunId = checkrunIdStrings.map((String id) => int.parse(id)).single;
+      final String buildUrl = 'https://ci.chromium.org/ui/b/${scheduleBuild.id}';
+      // Not all scheduled builds have check runs
+      if (checkRunId != null) {
+        final github.CheckRun checkRun = await githubChecksUtil.getCheckRun(config, slug, checkRunId);
+        await githubChecksUtil.updateCheckRun(config, slug, checkRun, detailsUrl: buildUrl);
```
godofredoc, line 273: Are these calls fast? They may increase the execution time of this request considerably.

godofredoc (approved):
LGTM let's give it a try and monitor the execution time for this handler.

app_dart/lib/src/service/luci_build_service.dart
CaseyHillers: I'm not sure. We already run one set of these to create the checks in this handler (ideally this would be a non-issue when we have https://github.com/flutter/flutter/issues/81579). Worst case it'll time out on this nice-to-have functionality

PR #1208, opened 2021-05-11
Use a patched version of uuid_enhanced that does not subclass typed_data classes

jason-simmons:
The typed_data classes are sealed in the latest Dart SDK
(see https://github.com/dart-lang/sdk/commit/3e9cdc56440bded76d8078a2567910236ad6f7b9)


jason-simmons:
That was not intentional.  Regenerated the PR.

app_dart/uuid_enhanced/.cirrus.yml
```diff
@@ -0,0 +1,13 @@
+container:
```
CaseyHillers, line 1: Cocoon has `tests.yaml` where we could add this as a standard Cocoon dart project (would be added to the existing LUCI builder)

CaseyHillers (approved):
Overall, this LGTM.

I'm hesitant to fork the package if https://github.com/flutter/cocoon/pull/1209 can unblock the Dart roll

CaseyHillers:
@jason-simmons can we close this, or is the roll still blocked?

PR #1202, opened 2021-05-06
[scheduler] ci migration script

CaseyHillers:
- Migrate cocoon as an example

# Issues

https://github.com/flutter/flutter/issues/76140

PR #1213, opened 2021-05-13
[scheduler] ci migration script 

CaseyHillers:
Originally https://github.com/flutter/cocoon/pull/1202 - but I had a git hiccup

- Migrate cocoon as an example

# Issues


scheduler/lib/models/scheduler.proto
```diff
@@ -33,10 +33,10 @@ message Target {
     // Name of the scheduler to trigger this target.
     // Defaults to being triggered by cocoon.
     optional SchedulerSystem scheduler = 8 [default = cocoon];
-    // Whether target should run pre-submit.
-    optional bool presubmit = 9 [default = true];
-    // Whether target should run post-submit.
-    optional bool postsubmit = 10 [default = true];
+    // Whether target should run pre-submit. Defaults to false, will not run presubmit.
+    optional bool presubmit = 9 [default = false];
```
christopherfujino, line 37: Won't this lead to an overall more verbose `.ci.yaml` file, once most of the targets are running post and pre?

CaseyHillers: Good point, I was looking at this from the cocoon perspective. I swapped backed the default=true to better accomodate the framework

christopherfujino: This LGTM.

As an aside, once we have the whole flutter world using this proto for configuration, I foresee that there will be a lot of things we'll want to change but can't easily do without breaking everything. Down the line, we may want to think about versioning the `.ci.yaml` files, so we can rev the protobuf with breaking changes.

christopherfujino (approved):
LGTM

PR #1210, opened 2021-05-12
[scheduler] Update prod tasks listed in ci.yaml

CaseyHillers:
- Explicitly pass postsubmit builders in `LuciService.getRecentTasks`
  - Update `/api/refresh-chromebot-status` to get postsubmit builders from `Scheduler`
  - Update engine tree status to get postsubmit builders from `Scheduler`

# Tests
Add test to refresh-chromebot-status to check builders in ci.yaml are used to update status

# Issues
https://github.com/flutter/flutter/issues/76140

PR #1215, opened 2021-05-13
[scheduler] Update prod tasks listed in ci.yaml

CaseyHillers:
Recreation of https://github.com/flutter/cocoon/pull/1210 due to some git hiccups

- Explicitly pass postsubmit builders in `LuciService.getRecentTasks`
  - Update `/api/refresh-chromebot-status` to get postsubmit builders from `Scheduler`
  - Update engine tree status to get postsubmit builders from `Scheduler`

# Tests
Add test to refresh-chromebot-status to check builders in ci.yaml are used to update status

# Issues

### DIFF
--- a/app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
+++ b/app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
@@ -58,24 +58,7 @@ class RefreshChromebotStatus extends ApiRequestHandler<Body> {
   Future<Body> get() async {
     final LuciService luciService = luciServiceProvider(this);
     final DatastoreService datastore = datastoreProvider(config.db);
-    for (String branch in await config.flutterBranches) {
-      await _updateBranch(branch, luciService: luciService, datastore: datastore);
-    }
-    return Body.empty;
-  }
-
-  /// Update LUCI tasks in [datastore] for [branch] on the framework repo.
-  Future<void> _updateBranch(
-    String branch, {
-    LuciService luciService,
-    DatastoreService datastore,
-  }) async {
-    final List<Commit> commits = await datastore.queryRecentCommits(limit: 1, branch: branch).toList();
-    if (commits.isEmpty) {
-      log.debug('Branch $branch does not have any commits');
-      return;
-    }
-    final Commit latestCommit = commits.first;
+    final Commit latestCommit = await datastore.queryRecentCommits(limit: 1).single;
     final SchedulerConfig schedulerConfig = await scheduler.getSchedulerConfig(latestCommit);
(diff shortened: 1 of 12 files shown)


app_dart/lib/src/request_handlers/refresh_chromebot_status.dart
```diff
@@ -53,8 +58,13 @@ class RefreshChromebotStatus extends ApiRequestHandler<Body> {
   Future<Body> get() async {
     final LuciService luciService = luciServiceProvider(this);
     final DatastoreService datastore = datastoreProvider(config.db);
+    final Commit latestCommit = await datastore.queryRecentCommits(limit: 1).single;
+    final SchedulerConfig schedulerConfig = await scheduler.getSchedulerConfig(latestCommit);
+    final List<Target> postsubmitTargets = scheduler.getPostSubmitTargets(latestCommit, schedulerConfig);
+    final List<LuciBuilder> postsubmitBuilders =
```
keyonghan, line 64: I may miss some context. Does `postsubmitBuilders` here include those in builders.json as well, or only those from .ci.yaml? Or we will wait for targets migrated to .ci.yaml before landing this?

CaseyHillers: Thanks, added `scheduler.getPostSubmitBuilders` to aggregate these. This will support both configs during the migration.

app_dart/test/src/datastore/fake_datastore.dart
```diff
@@ -107,7 +107,10 @@ class FakeDatastoreDB implements DatastoreDB {
 
   @override
   FakeQuery<T> query<T extends Model<dynamic>>({Partition partition, Key<dynamic> ancestorKey}) {
-    final List<T> results = values.values.whereType<T>().toList();
+    List<T> results = values.values.whereType<T>().toList();
+    if (ancestorKey != null) {
```
CaseyHillers, line 111: I noticed this while debugging we only return by type, so I extended it to filter the ancestor key (none of the tests take advantage of this yet)

keyonghan (approved):
This LGTM with tests fix.

CaseyHillers:
Pushing this as a traffic test via version testchillers

CaseyHillers:
Reverted traffic test. There's an issue where we don't expect a builder to be specified in both configs. Quick fix in https://github.com/flutter/flutter/pull/82860

CaseyHillers:
Fixed bug saw in prod due to duplicate targets in prod_builders and ci.yaml. As a temporary fix, it'll prioritize the prod_builders.json for updates in datastore, and after the migration tomorrow this will be a non-issue.

PR #1220, opened 2021-05-15
[app_dart] Create service account authentication provider

CaseyHillers:
Refactor `luci-status` to use the existing auth methods

This is necessary to create `KeyHelper` in LuciStatus so we can update prod tasks in datastore via pubsub


app_dart/test/request_handling/authentication_test.dart
```diff
@@ -60,67 +62,6 @@ void main() {
         );
       });
 
-      test('auth succeeds with authenticated cookie but unauthenticated header', () async {
```
CaseyHillers, line 63: These tests should have been removed in the agent cleanup. They have equivalents else where

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -456,6 +456,7 @@ class LuciBuildService {
     @required String builderName,
     String branch = 'master',
     String repo = 'flutter',
+    String encodedTaskKey,
```
keyonghan, line 459: Nit: doc for `encodedTaskKey`.

keyonghan (commented):
LGTM with nit.
As before, can we verify the functionality with some traffic before landing?

CaseyHillers:
Starting traffic test now with testchillers

app_dart/lib/src/service/luci_build_service.dart
CaseyHillers: Done

CaseyHillers:
Traffic test showed authentication is still working as intended

keyonghan (approved):

PR #1229, opened 2021-05-20
[scheduler] ci.yaml check run

CaseyHillers:
This adds a check to all Cocoon scheduled presubmits that validates the ci.yaml + validates all builds were scheduled.

If `.ci.yaml` does not exist, it will pass as it uses an empty `SchedulerConfig`.

## Issues

https://github.com/flutter/flutter/issues/82303

### DIFF
--- a/app_dart/lib/src/service/scheduler.dart
+++ b/app_dart/lib/src/service/scheduler.dart
@@ -359,6 +359,7 @@ class Scheduler {
         conclusion: github.CheckRunConclusion.success,
       );
     } else {
+      log.info('Marking PR #$prNumber ci.yaml validation as failed');
       // Failure when validating ci.yaml
       await githubChecksService.githubChecksUtil.updateCheckRun(
         config,
--- a/app_dart/lib/src/service/scheduler.dart
+++ b/app_dart/lib/src/service/scheduler.dart
@@ -345,6 +345,7 @@ class Scheduler {
         commitSha: commitSha,
       );
     } catch (e) {
+      log.info(e.toString());
       exception = e;
     }
 
--- a/app_dart/lib/src/service/scheduler.dart
+++ b/app_dart/lib/src/service/scheduler.dart
@@ -333,21 +333,23 @@ class Scheduler {
     );
     dynamic exception;
     final Commit presubmitCommit = Commit(branch: branch, repository: slug.fullName, sha: commitSha);
+    List<LuciBuilder> presubmitBuilders = <LuciBuilder>[];
     try {
-      final List<LuciBuilder> presubmitBuilders = await getPresubmitBuilders(
+      presubmitBuilders = await getPresubmitBuilders(
         commit: presubmitCommit,
         prNumber: prNumber,
       );
-      await luciBuildService.scheduleTryBuilds(
-        builders: presubmitBuilders,
-        slug: slug,
-        prNumber: prNumber,
-        commitSha: commitSha,
-      );
-    } catch (e) {
+    } on FormatException catch (e) {
       log.info(e.toString());
       exception = e;
     }
+
+    await luciBuildService.scheduleTryBuilds(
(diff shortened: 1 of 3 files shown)


app_dart/test/request_handlers/github_webhook_test.dart
```diff
@@ -904,35 +904,6 @@ void main() {
         request.headers.set('X-GitHub-Event', 'pull_request');
       });
 
-      test('Exception is raised when no builders available', () async {
```
CaseyHillers, line 907: This is no longer true as there's always at least one check we generate on presubmit

app_dart/lib/src/service/scheduler.dart
```diff
@@ -322,17 +325,54 @@ class Scheduler {
...
+        prNumber: prNumber,
+      );
+      await luciBuildService.scheduleTryBuilds(
+        builders: presubmitBuilders,
+        slug: slug,
+        prNumber: prNumber,
+        commitSha: commitSha,
+      );
+    } catch (e) {
+      exception = e;
```
godofredoc, line 348: Should we log this exception?

app_dart/lib/src/service/scheduler.dart
```diff
@@ -322,17 +325,54 @@ class Scheduler {
...
+      await githubChecksService.githubChecksUtil.updateCheckRun(
+        config,
+        slug,
+        ciValidationCheckRun,
+        status: github.CheckRunStatus.completed,
+        conclusion: github.CheckRunConclusion.success,
+      );
+    } else {
+      // Failure when validating ci.yaml
+      await githubChecksService.githubChecksUtil.updateCheckRun(
```
godofredoc, line 363: we can move this part to the catch section and remove the if - then - else

app_dart/lib/src/service/scheduler.dart
```diff
@@ -322,17 +325,54 @@ class Scheduler {
...
+        commitSha: commitSha,
+      );
+    } catch (e) {
+      exception = e;
+    }
+
+    // Update validate ci.yaml check
+    if (exception == null) {
+      // Success in validating ci.yaml
+      await githubChecksService.githubChecksUtil.updateCheckRun(
```
godofredoc, line 354: We can add this to the end of the try block and remove the if - then else.

app_dart/lib/src/service/scheduler.dart
christopherfujino: we should also only catch a more specific type of Exception, rather than everything.

christopherfujino: https://dart.dev/guides/language/effective-dart/usage#avoid-catches-without-on-clauses

CaseyHillers: I added a log.

We could catch `FormatException` and `HttpException` for ci.yaml related issues, but I figured if there's an issue in scheduling the LUCI builds (anywhere), we should mark it as failed. That way people will retry the presubmit instead of having a forever pending checkrun. WDYT?

app_dart/lib/src/service/scheduler.dart
CaseyHillers: This is separate to minimize the code the try block deals with (only scheduler config validation + schedule builds), otherwise we would silently fail on quota issues

app_dart/lib/src/service/scheduler.dart
CaseyHillers: Moved to catch on FormatException. If another exception fails, this will leave the ci.yaml validation as pending.

christopherfujino: Will the exception bubbling up lead to a stronger signal that monitoring will detect? If not, then I suppose there's no advantage to disambiguating `FormatException`s.

godofredoc (approved):

PR #1230, opened 2021-05-21
Reland "[app_dart] Create service account authentication provider" #1220

CaseyHillers:
https://github.com/flutter/cocoon/pull/1220

Diff between this PR and the original is in last commit. I removed the expiration check as it was new functionality that does not work with service accounts.


keyonghan (approved):
LGTM.

The ci.yaml validation test doesn't seem to point to the correct build.
Screen Shot 2021-05-21 at 11 06 58 AM
And it is not showing up https://ci.chromium.org/p/flutter/g/cocoon-try/builders.

CaseyHillers:
`ci.yaml validation isn't a builder`.  It only validates Cocoon did not have issues when triggering the presubmit builds (such as malformed ci.yaml or in the future failed buildbucket ScheduleBuild RPCs).

If it fails, it gives the detailed error of what happened (such as specifically the issue with the ci.yaml)

PR #1235, opened 2021-06-02
Add support for properties in the reset_prod_task.

godofredoc:
This is allow retry builds that override properties.

Bug: https://github.com/flutter/flutter/issues/81102

### DIFF
--- a/app_dart/lib/src/request_handlers/reset_prod_task.dart
+++ b/app_dart/lib/src/request_handlers/reset_prod_task.dart
@@ -43,6 +43,7 @@ class ResetProdTask extends ApiRequestHandler<Body> {
   static const String repoParam = 'Repo';
   static const String commitShaParam = 'Commit';
   static const String builderParam = 'Builder';
+  static const String propertiesParam = 'Properties';
 
   @override
   Future<Body> post() async {
@@ -53,6 +54,8 @@ class ResetProdTask extends ApiRequestHandler<Body> {
     final String owner = requestData[ownerParam] as String ?? 'flutter';
     final String repo = requestData[repoParam] as String ?? 'flutter';
     String commitSha = requestData[commitShaParam] as String ?? '';
+    final Map<String, String> properties = requestData[propertiesParam] as Map<String, String> ?? <String, String>{};
+
     RepositorySlug slug;
     String builder = requestData[builderParam] as String ?? '';
     Task task;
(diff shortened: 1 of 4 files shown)


app_dart/lib/src/request_handlers/reset_prod_task.dart
```diff
@@ -53,6 +54,9 @@ class ResetProdTask extends ApiRequestHandler<Body> {
     final String owner = requestData[ownerParam] as String ?? 'flutter';
     final String repo = requestData[repoParam] as String ?? 'flutter';
     String commitSha = requestData[commitShaParam] as String ?? '';
+    final Map<String, dynamic> properties =
```
CaseyHillers, line 57: What is the use case for configuring this? Shouldn't we be reusing the properties from the previously scheduled build?

godofredoc: The engine post-submit tasks sometimes need to override the generated artifacts on retries. Rather than always overriding the artifacts by default the builders raise an exception if the artifacts and we are re-running the build and only succeed if we we manually pass a property telling the builder to override the artifacts.

A second use case is to run builds deleting the caches first. This is useful for the cases when a gclient sync pollutes the source code cache and we need to force a build with a fresh source code checkout.

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -476,9 +479,7 @@ class LuciBuildService {
         ],
         'user_agent': const <String>['luci-scheduler'],
       },
-      properties: <String, String>{
-        'git_ref': commitSha,
-      },
+      properties: localProperties,
```
CaseyHillers, line 482: Will this override the existing input properties? For example, are you required to copy all properties into the POST request, then add the cache clobber or clear build artifacts property?

keyonghan (approved):
LGTM

app_dart/lib/src/service/luci_build_service.dart
keyonghan: it will be merged and replace any existing ones: https://github.com/flutter/cocoon/blob/master/app_dart/lib/src/model/luci/buildbucket.dart#L387

godofredoc: Thanks for checking Keyong!

PR #1238, opened 2021-06-03
Update the details url as soon as we retrigger a task.

godofredoc:
This is to be able to link to the LUCI task as soon as the build is
scheduled.

Bug: https://github.com/flutter/flutter/issues/82417

### DIFF
--- a/app_dart/lib/src/service/luci_build_service.dart
+++ b/app_dart/lib/src/service/luci_build_service.dart
@@ -377,7 +377,7 @@ class LuciBuildService {
     userData['repo_owner'] = slug.owner;
     userData['repo_name'] = slug.name;
     userData['user_agent'] = 'flutter-cocoon';
-    await buildBucketClient.scheduleBuild(ScheduleBuildRequest(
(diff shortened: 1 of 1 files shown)


keyonghan (commented):
LGTM

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -397,6 +397,8 @@ class LuciBuildService {
         userData: json.encode(userData),
       ),
     ));
+    final String buildUrl = 'https://ci.chromium.org/ui/b/${scheduleBuild.id}';
+    await githubChecksUtil.updateCheckRun(config, slug, githubCheckRun, detailsUrl: buildUrl);
     return true;
```
CaseyHillers, line 402: Not necessary for this PR, but we should start returning the builds instead of true so we can pass this information around

CaseyHillers (approved):

app_dart/lib/src/service/luci_build_service.dart
godofredoc: Added https://github.com/flutter/flutter/issues/83945 to not forget updating the methods to return a Build object.

PR #1242, opened 2021-06-07
[app_dart] Check bash script exit codes in local deploy script

CaseyHillers:
Fixes https://github.com/flutter/flutter/issues/84139

Issue was the bash script exit code were being ignored by the local Dart script. This only impacted local uses, and not production uses.


godofredoc (approved):
Thanks for the quick fix!

PR #1250, opened 2021-06-17
Adds auto retry for framework TOT failures

keyonghan:
This adds auto retry for framework TOT failures. 

Issues: https://github.com/flutter/flutter/issues/84004, https://github.com/flutter/flutter/issues/84818

### DIFF
--- a/app_dart/lib/src/request_handlers/push_engine_status_to_github.dart
+++ b/app_dart/lib/src/request_handlers/push_engine_status_to_github.dart
@@ -124,9 +124,6 @@ class PushEngineStatusToGithub extends ApiRequestHandler<Body> {
         case Task.statusFailed:
         case Task.statusInfraFailure:
           log.debug('Using ${task.status} from commit ${task.commitSha} ref ${task.ref} builder ${task.builderName}');
-          final int retries = tasks.where((LuciTask element) => element.commitSha == task.commitSha).length - 1;
-          log.debug('markdown: ${task.summaryMarkdown}, builderName: ${task.builderName}, retries: $retries');
-          await luciBuildService.checkRerunBuilder(commit: Commit(), luciTask: task, retries: retries, repo: 'engine');
           return GithubBuildStatusUpdate.statusFailure;
         case Task.statusSucceeded:
           log.debug('Using ${task.status} from commit ${task.commitSha} ref ${task.ref} builder ${task.builderName}');
--- a/app_dart/lib/src/request_handlers/push_engine_status_to_github.dart
+++ b/app_dart/lib/src/request_handlers/push_engine_status_to_github.dart
@@ -126,8 +126,7 @@ class PushEngineStatusToGithub extends ApiRequestHandler<Body> {
           log.debug('Using ${task.status} from commit ${task.commitSha} ref ${task.ref} builder ${task.builderName}');
           final int retries = tasks.where((LuciTask element) => element.commitSha == task.commitSha).length - 1;
           log.debug('markdown: ${task.summaryMarkdown}, builderName: ${task.builderName}, retries: $retries');
-          await luciBuildService.checkRerunBuilder(
-              commitSha: task.commitSha, luciTask: task, retries: retries, repo: 'engine');
+          await luciBuildService.checkRerunBuilder(commit: Commit(), luciTask: task, retries: retries, repo: 'engine');
           return GithubBuildStatusUpdate.statusFailure;
         case Task.statusSucceeded:
           log.debug('Using ${task.status} from commit ${task.commitSha} ref ${task.ref} builder ${task.builderName}');
--- a/app_dart/lib/src/service/luci_build_service.dart
+++ b/app_dart/lib/src/service/luci_build_service.dart
@@ -22,6 +22,8 @@ import '../service/datastore.dart';
 import 'buildbucket.dart';
 import 'luci.dart';
 
+const Set<String> taskFailStatusSet = <String>{Task.statusInfraFailure, Task.statusFailed};
+
 /// Class to interact with LUCI buildbucket to get, trigger
 /// and cancel builds for github repos. It uses [config.luciTryBuilders] to
 /// get the list of available builders.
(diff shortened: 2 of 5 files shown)


jmagman (approved):
Nice, you can get rid of that -9 workaround.
Logic LGTM

godofredoc (commented):
Isn't this the functionality still being debated in the auto-retries design doc?

app_dart/lib/src/request_handlers/push_engine_status_to_github.dart
```diff
@@ -126,8 +126,7 @@ class PushEngineStatusToGithub extends ApiRequestHandler<Body> {
           log.debug('Using ${task.status} from commit ${task.commitSha} ref ${task.ref} builder ${task.builderName}');
           final int retries = tasks.where((LuciTask element) => element.commitSha == task.commitSha).length - 1;
           log.debug('markdown: ${task.summaryMarkdown}, builderName: ${task.builderName}, retries: $retries');
-          await luciBuildService.checkRerunBuilder(
-              commitSha: task.commitSha, luciTask: task, retries: retries, repo: 'engine');
+          await luciBuildService.checkRerunBuilder(commit: Commit(), luciTask: task, retries: retries, repo: 'engine');
```
godofredoc, line 129: Do we need to create the commit passing the CommitSha?

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -510,19 +508,14 @@ class LuciBuildService {
     return false;
   }
 
-  bool _shouldRerunBuilder(LuciTask luciTask, String repo) {
-    if (luciTask.summaryMarkdown == null || !luciTask.builderName.contains('Mac')) {
+  /// Only blocking failures on framework TOT will be triggered.
```
godofredoc, line 511: Should this get updated to match what the method is actually doing?

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -510,19 +508,14 @@ class LuciBuildService {
...
+  Future<bool> _shouldRerunBuilder(LuciTask luciTask, int retries, Commit commit, DatastoreService datastore) async {
+    if ((luciTask.status != Task.statusFailed && luciTask.status != Task.statusInfraFailure) ||
```
godofredoc, line 513: Creating a set of the expected status and comparing the current one with the 'in' operator will make this logic simpler.

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -510,19 +508,14 @@ class LuciBuildService {
...
-      case 'flutter':
-        return luciTask.summaryMarkdown.contains('retcode: -9') ||
-            (kMacBuildersWithShards.contains(luciTask.builderName) &&
-                luciTask.summaryMarkdown ==
-                    'recipe infra failure: Infra Failure: Step(\'display builds.build(s) failed\') (retcode: 1)');
-      case 'engine':
-        return luciTask.summaryMarkdown.contains('retcode: -9');
-    }
-    return false;
+    final Commit latestCommit = await datastore.queryRecentCommits(limit: 1).single;
```
godofredoc, line 518: How do we prevent this from running on almost every failed task of every commit?

app_dart/test/service/luci_build_service_test.dart
```diff
@@ -485,140 +487,138 @@ void main() {
...
         luciTask: luciTask,
         retries: 0,
         repo: 'flutter',
+        datastore: datastore,
       );
       expect(rerunFlag, false);
     });
 
-    test('Do not rerun a flutter Mac builder with non -9 retcode', () async {
+    test('Do not rerun a non-supported repo builder', () async {
```
godofredoc, line 565: To which use case does this test applies?

app_dart/test/service/luci_build_service_test.dart
```diff
@@ -485,140 +487,138 @@ void main() {
...
-        retries: 0,
+        retries: 1,
         repo: 'flutter',
+        datastore: datastore,
       );
       expect(rerunFlag, false);
     });
 
-    test('Do not rerun a flutter Mac builder exceeding attempt limit', () async {
+    test('Do not rerun a flutter builder when not blocking the tree', () async {
```
godofredoc, line 604: Do we need at least 3 commits and more than two tasks to make this test relevant? e.g. oldest fails, the one in the middle fails with a different task, etc?

app_dart/lib/src/request_handlers/push_engine_status_to_github.dart
keyonghan: This is legacy code for `-9` retry. The auto retry here will be a no-op for engine repo. Removed to clean up.

app_dart/lib/src/service/luci_build_service.dart
keyonghan: Updated.

app_dart/lib/src/service/luci_build_service.dart
keyonghan: Done.

app_dart/lib/src/service/luci_build_service.dart
keyonghan: If the commit is TOT, it is expected that all failed tasks will be rerun.
If the commit is not TOT, then failed tasks will not be triggered.

app_dart/test/service/luci_build_service_test.dart
keyonghan: Since the logic is for framework only, removing the test to avoid confusion.

app_dart/test/service/luci_build_service_test.dart
keyonghan: Not necessary. We cares only the TOT commit: if the current failed task comes from the TOT commit, then it is blocking the tree and we need to rerun; if the task is not from the TOT commit, then we would expect there is a newer commit on top of this one and the same task of the newer commit would potentially help open the tree, and we will not rerun this task.
To make it more clear, updated to add both the current commit and a TOT commit to the fake datastore.

app_dart/lib/src/service/luci_build_service.dart
godofredoc: Let's keep an eye on the behavior once it lands, given that the cron job is not in sync with the scheduler it may cause unexpected race conditions potentially leading to retry almost all failures.

godofredoc (approved):
LGTM

app_dart/lib/src/service/luci_build_service.dart
keyonghan: We have deprecated the cron job and solely rely on the scheduler now, so we should be okay in this area.
But I will closely monitor the dashboard on Monday.

PR #1251, opened 2021-06-21
[scheduler] Add recipe field and generate_jspb script

CaseyHillers:

generate_jspb will take a given yaml, and make it into a format that luci's starlark files can read.

### DIFF
--- a/scheduler/bin/generate_jspb.dart
+++ b/scheduler/bin/generate_jspb.dart
@@ -10,7 +10,7 @@ import 'package:yaml/yaml.dart';
 
 Future<void> main(List<String> args) async {
   if (args.length != 2) {
-    print('generate_builder_config.dart \$repo \$sha');
+    print('generate_jspb.dart \$repo \$sha');
     exit(1);
   }
   final String repo = args.first;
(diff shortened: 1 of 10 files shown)


scheduler/bin/generate_jspb.dart
```diff
@@ -0,0 +1,34 @@
...
+    throw HttpException('HTTP ${clientResponse.statusCode}: $ciYamlUrl');
+  }
+  final String configContent = await utf8.decoder.bind(clientResponse).join();
+  client.close(force: true);
+  final YamlMap configYaml = loadYaml(configContent) as YamlMap;
+  // There's an assumption that we're only generating builder configs from commits that
+  // have already landed with validation. Otherwise, this will fail.
+  final SchedulerConfig schedulerConfig = schedulerConfigFromYaml(configYaml);
+  const JsonEncoder encoder = JsonEncoder.withIndent('  ');
+  final String prettyJson = encoder.convert(schedulerConfig.writeToJsonMap());
```
christopherfujino, line 32: if you use `schedulerConfig.toProto3Json()` it will preserve the plaintext key names (rather than the proto integers), which is (I think) what we want.

CaseyHillers: Thanks! Done

christopherfujino (approved):
lgtm

PR #1256, opened 2021-06-24
Update flutter status name.

godofredoc:
The status name was changed to make it consistent with the name used in
the engine and the the "waiting for tree to go green" was failing to
submit the PR instead of waiting for the tree to become green.

Bug: https://github.com/flutter/flutter/issues/84835


jonahwilliams (approved):
LGTM

PR #1258, opened 2021-06-25
[app_dart] Add google_internal target support

CaseyHillers:
Adds initial plumbing to ingest target from `.ci.yaml`

```yaml
# .ci.yaml
- name: Google Internal Roll
   # Breakages in the Google roll usually require patches, not reverts. Including as FYI.
   bringup: true
   presubmit: false
   scheduler: google_internal
```

Using `/api/update-task-status` this target can be updated via FRoB

# Issues

https://github.com/flutter/flutter/issues/79658


renyou (approved):
Thank you so much for adding this!

PR #1265, opened 2021-06-29
[app_flutter] Temporarily use builder name as column

CaseyHillers:
To make https://github.com/flutter/flutter/issues/84998 a softer transition, we'll temporarily move to builder names as the index value.

Once the framework .ci.yaml has been fixed and soaked for 20 commits, this will be reverted.

The golden diffs are due to outdated changes from the task icons being changed (<5% overall diff)


keyonghan (approved):
LGTM

PR #1267, opened 2021-06-30
Add link(s) to failure page when a PR has failing checks/statuses

dnfield:
Fixes https://github.com/flutter/flutter/issues/85657

### DIFF
--- a/app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
+++ b/app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
@@ -221,7 +221,7 @@ class CheckForWaitingPullRequests extends ApiRequestHandler<Body> {
             (commit['checkSuites']['nodes']?.first['checkRuns']['nodes'] as List<dynamic>).cast<Map<String, dynamic>>();
       }
       checkRuns ??= <Map<String, dynamic>>[];
-      final Set<String> failures = <String>{};
+      final Set<_FailureDetail> failures = <_FailureDetail>{};
       final bool ciSuccessful = await _checkStatuses(
         sha,
         failures,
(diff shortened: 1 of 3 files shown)


app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
```diff
@@ -284,7 +284,7 @@ class CheckForWaitingPullRequests extends ApiRequestHandler<Body> {
         allSuccess = false;
```
CaseyHillers, line 284: `CheckRun.fromJson(checkRun)` would be clearer than using the `Map<String, dynamic>`

app_dart/test/request_handlers/check_for_waiting_pull_requests_test.dart
```diff
@@ -393,7 +393,7 @@ This pull request is not suitable for automatic merging in its current state.
             'sBody': '''
 This pull request is not suitable for automatic merging in its current state.
 
-- The status or check suite Linux has failed. Please fix the issues identified (or deflake) before re-applying this label.
+- The status or check suite [Linux](https://Linux) has failed. Please fix the issues identified (or deflake) before re-applying this label.
```
CaseyHillers, line 396: nit: This test seems misleading, can we pipe a realistic luci detailsUrl?

CaseyHillers (approved):
Overall, LGTM - Just some small, non-blocking comments

keyonghan (approved):
LGTM

app_dart/test/request_handlers/check_for_waiting_pull_requests_test.dart
dnfield: I did this to make it a little easier to construct the test - we can but the important part here is just that it gets the link in there.

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
dnfield: I'd rather we change that in a separate commit in case it breaks anything.

dnfield:
I apparently do not have permission to merge in this repo.

keyonghan:
> I apparently do not have permission to merge in this repo.

Cocoon is now disabled for manual merge. We need to label `waiting for tree green label` for auto merge.

PR #1268, opened 2021-07-01
update Tags format

chunhtai:


chunhtai:
For the incoming ci format change https://github.com/flutter/flutter/pull/85565/files

app_dart/lib/src/model/proto/internal/scheduler.proto
```diff
@@ -57,7 +57,22 @@ message Target {
     // Name of the LUCI recipe to use for the builder.
     optional string recipe = 13;
     // Additional tag for the target.
-    repeated string tags = 14;
+    optional Tags tags = 14;
```
CaseyHillers, line 60: It seems simpler to do

```
repeated Tags tags = 14;
```

And have tags be an enum

chunhtai: The plan is tags can be extended to have complex value
for example,

   tags:
        shard: true
        platform: ios
        duration: 1000

CaseyHillers: Is there an issue or a doc I can look at? It sounds like we're working on similar functionality (https://github.com/flutter/flutter/issues/84998, go/flutter-automatic-builders)

chunhtai: This is a part of
https://docs.google.com/document/d/1DRw_J3zjuVpiVRN3iRRCjeI0RGoLrB6zOOTtj4h-hHI/edit#heading=h.4qsx9pjitde1

the problem i am try to solve here is that we need to find testowner and file pr to mark test flaky. Depends on the type of the test, the way to find test owner and file pr will be different. That is why I want to add different tag to those tests, so the script can pick it up. While I was making the change, @keyonghan suggest we make the tags to be a complex object so people extend it in the future.

CaseyHillers: Thanks for the context. I believe we can pull this information without needing to modify `.ci.yaml`.

`platform` can be retrieved by the first word of the target.name: `Linux_android`, `Linux`, etc
`shard` is determined by if there is `_#` or `_last` at the end of the target name
`duration` is `target.timeout` (I'm working on a PR to mass enable these)
The test types are harder to translate, but if we could simplify to on-device tests + vm tests we could reuse platform.

The word `tag` is misleading as this file is intended to be used by `LUCI` where there's a different connotation for tag.

chunhtai: > shard is determined by if there is _# or _last at the end of the target name

I think some do not follow that, e.g. framework_tests_libraries, framework_tests_misc. I also need to know whether it is framework test, devicelab test.. etc

platform and duration are just examples for future extension that people can create complex key value pair

> The word tag is misleading as this file is intended to be used by LUCI where there's a different connotation for tag.

If so, we can consider a different name

app_dart/lib/src/model/proto/internal/scheduler.proto
```diff
@@ -57,7 +57,22 @@ message Target {
...
+}
+
+// Tags for a specific infrastructure task
+// Next ID: 6
+message Tags {
```
CaseyHillers, line 65: Based on [builds_service.proto](https://source.chromium.org/chromium/infra/infra/+/main:go/src/go.chromium.org/luci/buildbucket/proto/builds_service.proto;l=280;drc=adbd1e588d722858df732b1ecb23e09c00ab1b54), can we make this a StringPair as well?

chunhtai: sounds good

chunhtai: it looks like StringPair is a custom structure defined in https://source.chromium.org/chromium/infra/infra/+/main:luci/appengine/swarming/proto/api/swarming.proto?q=StringPair&ss=chromium%2Finfra%2Finfra

I will need to duplicate the code I assume?

CaseyHillers: Yeah, easier to copy it here since it's relatively simple

chunhtai:
Hi Casey, I updated the format PTAL :)

app_dart/lib/src/model/proto/internal/scheduler.proto
```diff
@@ -57,7 +57,17 @@ message Target {
     // Name of the LUCI recipe to use for the builder.
     optional string recipe = 13;
     // Additional tag for the target.
-    repeated string tags = 14;
+    repeated StringPair tags = 14;
```
CaseyHillers, line 60: doc?

app_dart/lib/src/model/proto/internal/scheduler.proto
```diff
@@ -57,7 +57,17 @@ message Target {
...
+}
+
+// A string key-value pair. Typically used for tagging.
+//
+// This structure need to be in sync with the StringPair in LUCI infra.
+// See the `go/src/go.chromium.org/luci/buildbucket/proto/common.proto` in the
+// https://source.chromium.org
+message StringPair {
```
CaseyHillers, line 68: nit: next id?

app_dart/lib/src/model/proto/internal/scheduler.proto
```diff
@@ -57,7 +57,17 @@ message Target {
     // Name of the LUCI recipe to use for the builder.
     optional string recipe = 13;
     // Additional tag for the target.
-    repeated string tags = 14;
```
CaseyHillers, line 60: Can we link to the LUCI docs on what a tag is?

CaseyHillers (approved):
LGTM once the docs comments are addressed

app_dart/lib/src/model/proto/internal/scheduler.proto
chunhtai: This structure should not be changed, do we still need to document next id?

CaseyHillers: Good point, SGTM

app_dart/lib/src/model/proto/internal/scheduler.proto
chunhtai: I am hesitate to do that now seems these tags are not feed into the luci yet. Should we wait for that once we actually feed this tags into luci?

PR #1272, opened 2021-07-12
[ci.yaml] Add platform_properties, add_recipe_cq

CaseyHillers:

Related infra CL: https://flutter-review.googlesource.com/c/infra/+/15360

### DIFF
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -1,17 +1,40 @@
+# Describes the targets run in continuous integration environment.
+#
+# Flutter infra uses this file to generate a checklist of tasks to be performed
+# for every commit.
+#
+# More information at:
+#  * https://github.com/flutter/cocoon/blob/master/CI_YAML.md
 enabled_branches:
   - master
+
+platform_properties:
+  linux:
+    properties:
+      caches: dart_pub_cache:.pub_cache
+      os: Linux
+  mac:
+    properties:
+      os: Mac-10.15
+  windows:
+    properties:
+      os: Windows
 
 targets:
   - name: Linux Cocoon
     builder: Linux Cocoon
     postsubmit: false
     recipe: cocoon/cocoon
+    properties:
+      add_recipes_cq: "true"
     scheduler: luci
 
   - name: Linux device_doctor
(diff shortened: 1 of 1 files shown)


.ci.yaml
```diff
@@ -1,17 +1,40 @@
...
+# for every commit.
+#
+# More information at:
+#  * https://github.com/flutter/cocoon/blob/master/CI_YAML.md
 enabled_branches:
   - master
 
+platform_properties:
+  linux:
+    properties:
```
christopherfujino, line 13: do we need a `properties` key here? are there gonna be other keys at this same level?

christopherfujino: nevermind, I see this is already defined in the proto

CaseyHillers: Unfortunately, yes. I can't find a way around it as protos do not make it possible to nest values (in this case we map->properties->map)

We could hardcode the platforms in the proto (linux_properties->map), but i'd prefer to make it easy to add new platforms (like web, linux_android)

christopherfujino: agreed. this is annoying, but fine.

christopherfujino:
Looks like a parse error:

```
00:00 [32m+2[0m[31m -1[0m: /b/s/w/ir/k/cocoon/app_dart/integration_test/validate_all_ci_configs_test.dart: validate config file of flutter/flutter/master/.ci.yaml [1m[31m[E][0m[0m
  Protobuf JSON decoding failed at: root["targets"]["0"]["tags"]["0"]. Expected String value
  package:test_api                                         fail
  integration_test/validate_all_ci_configs_test.dart 37:9  main.<fn>
```

.ci.yaml
```diff
@@ -1,17 +1,40 @@
...
+      caches: "[{\"name\": \"dart_pub_cache\", \"path\": \".pub_cache\"}]"
```
CaseyHillers, line 14: @christopherfujino I updated this to be json encoded. For repos with more caches, we can use a yaml multiline block to make this prettier

christopherfujino (approved):
LGTM

PR #1276, opened 2021-07-13
[ci.yaml] LSC boiler plate script

CaseyHillers:
This is a boiler plate script that allows you to easily make LSC to ci.yaml. It includes logic on how to write the file in yaml.

The useful part of this is the `writeYaml`. This functionality could probably be generalized and upstreamed into `package:yaml`, but  there's a lot of edge cases it doesn't handle.


CaseyHillers:
> Is this script to make name to be equal to builder? It looks good for a boiler plate.

Yes, this specifically does name==builderName and imports tags into properties.

PR #1277, opened 2021-07-14
[ci.yaml] Deprecate tags field

CaseyHillers:

### DIFF
--- a/app_dart/lib/src/model/proto/internal/scheduler.proto
+++ b/app_dart/lib/src/model/proto/internal/scheduler.proto
@@ -56,19 +56,7 @@ message Target {
     repeated string enabled_branches = 12;
     // Name of the LUCI recipe to use for the builder.
     optional string recipe = 13;
-    // Additional tags for the target.
-    // The tags is list of key-value pairs that can be used to categorize different targets.
-    repeated StringPair tags = 14;
-}
-
-// A string key-value pair. Typically used for tagging.
-//
-// This structure must not be changed and need to be in sync with the StringPair in LUCI infra.
-// See the `go/src/go.chromium.org/luci/buildbucket/proto/common.proto` in the
-// https://source.chromium.org
-message StringPair {
-  required string key = 1;
-  required string value = 2;
+    reserved 14;
 }
 
 // Schedulers supported in SchedulerConfig.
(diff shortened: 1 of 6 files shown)


app_dart/lib/src/model/proto/internal/scheduler.proto
```diff
@@ -58,7 +58,7 @@ message Target {
     optional string recipe = 13;
     // Additional tags for the target.
     // The tags is list of key-value pairs that can be used to categorize different targets.
-    repeated StringPair tags = 14;
```
chunhtai, line 61: Do we need to keep and mark deprecated consider how short live it is?

chunhtai (approved):
LGTM

app_dart/lib/src/model/proto/internal/scheduler.proto
CaseyHillers: My concern is there may be a dev release with tags, and we still need to support it. /cc @christopherfujino

chunhtai: oh, do you mean the cocoon may point to a dev flutter branch? I thought it always grab ci.yaml from the master branch. For the cocoon itself, the only code that relies on the tags should be the request handler i added, and it is not active yet since the periodic http request is not set up yet

CaseyHillers: Yes. Cocoon's tip of tree needs to support all ci.yamls from all supported release branches. Removing tags entirely would cause Cocoon to fail to schedule presubmit jobs, in this case `tags` would be a non-existent field. Instead, we deprecate for a few weeks then remove.

CaseyHillers: Once a new dev release is available, we should be able to remove the entry

christopherfujino: unfortunately [this dev release](https://github.com/flutter/flutter/blob/dev/.ci.yaml) is going to be a beta release. I can try to cherry-pick the migration

CaseyHillers:
When https://github.com/flutter/flutter/pull/86445 lands, we can remove the tags. I updated the PR to reserve the field to prevent data corruption, and cleaned up the helper StringPair message.

chunhtai (approved):
LGTM

chunhtai (changes requested):
Actually, can you also update cocoon/CI_YAML.md

christopherfujino (approved):
LGTM

chunhtai (approved):
LGTM

PR #1278, opened 2021-07-14
Update the script to match the ci.yaml

chunhtai:
ci.yaml format has changed, This pr updates the script to be in sync.

This pr also change the way we detect existing issue and pr. It now write a hidden meta tags to the issue/pr body and use them to detect which builders they belong to

### DIFF
--- a/app_dart/lib/src/request_handlers/check_flaky_tests_and_update_github.dart
+++ b/app_dart/lib/src/request_handlers/check_flaky_tests_and_update_github.dart
@@ -205,9 +205,9 @@ class CheckForFlakyTestAndUpdateGithub extends ApiRequestHandler<Body> {
     }
     bool hasFrameworkTag = false;
     bool hasHostOnlyTag = false;
-    // If tags contains 'shard', it must be a shard test.
-    // If tags contains 'devicelab', it must be a devicelab test.
-    // Otherwise, it is framework host only test if its tags contains both
+    // If tags contain 'shard', it must be a shard test.
+    // If tags contain 'devicelab', it must be a devicelab test.
+    // Otherwise, it is framework host only test if its tags contain both
     // 'framework' and 'hostonly'.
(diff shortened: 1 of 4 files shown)


app_dart/lib/src/request_handlers/check_flaky_tests_and_update_github.dart
```diff
@@ -185,7 +188,7 @@ class CheckForFlakyTestAndUpdateGithub extends ApiRequestHandler<Body> {
     return target != null && target[_ciYamlTargetIsFlakyKey] == true;
   }
 
-  YamlList _getTags(String builderName, YamlMap ci) {
+  List<dynamic> _getTags(String builderName, YamlMap ci) {
```
keyonghan, line 190: Shall we use String explicitly? do we have other type other than string

chunhtai:
@keyonghan I am not sure where may be a good place to put the readme since the folder contains multiple handler. I think I will just document that in the wiki, but I think it should be updated after the script is deployed and running?

I will prepare an announcement email before It is deployed

keyonghan (approved):
LGTM with nit.

keyonghan:
> @keyonghan I am not sure where may be a good place to put the readme since the folder contains multiple handlers. I think I will just document that in the wiki, but I think it should be updated after the script is deployed and running?
>
> I will prepare an announcement email before It is deployed

A wiki page sounds just fine. We can prepare it first, and release it once the script in in prod.

app_dart/lib/src/request_handlers/check_flaky_tests_and_update_github.dart
chunhtai: I don't know whether there is plan to extend to usage of tags. The problem with forcing it to be string is that we may break it without knowing it if we change it later since the test is using mock data not the real ci.yaml in master branch.

PR #1280, opened 2021-07-14
Add TestRunFlaky property for task

keyonghan:
This is to mark test, which has rerun from test runner side, as flaky if any.

Existing logic counts a test as flaky if multiple runs exist for the whole build. With https://github.com/flutter/flutter/pull/86394, a test can be flaky even if there is only one build run but there are multiple reruns from the test runner. This PR adds a task property to hold that flaky flag.

One TODO is update this property when task finishes if any flake exists from test runner side.

Related: https://github.com/flutter/flutter/issues/86324

### DIFF
--- a/app_dart/lib/src/model/appengine/task.dart
+++ b/app_dart/lib/src/model/appengine/task.dart
@@ -30,7 +30,7 @@ class Task extends Model<int> {
     this.name,
     this.attempts = 0,
     this.isFlaky = false,
-    this.isTestRunFlaky = false,
+    this.isTestFlaky = false,
     this.timeoutInMinutes,
     this.reason = '',
     this.requiredCapabilities,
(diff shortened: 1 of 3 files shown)


app_dart/lib/src/model/appengine/task.dart
```diff
@@ -178,6 +179,13 @@ class Task extends Model<int> {
   @JsonKey(name: 'Flaky')
   bool isFlaky;
 
+  /// Whether the test run of this task is flaky.
+  ///
+  /// Test runner supports rerun, and this flag tracks if a flake happens.
+  @BoolProperty(propertyName: 'TestRunFlaky')
+  @JsonKey(name: 'TestRunFlaky')
+  bool isTestRunFlaky;
```
CaseyHillers, line 187: `isTestFlaky` is clearer to me (run was confusing me)

app_dart/lib/src/model/appengine/task.dart
```diff
@@ -178,6 +179,13 @@ class Task extends Model<int> {
   @JsonKey(name: 'Flaky')
   bool isFlaky;
```
CaseyHillers, line 180: Are you able to update this doc to help differentiate it from the new property?

app_dart/lib/src/model/appengine/task.dart
keyonghan: Updated.

app_dart/lib/src/model/appengine/task.dart
keyonghan: Updated the doc to reflect latest .ci.yaml, also to distinguish from the new property.

CaseyHillers (approved):

PR #1286, opened 2021-07-16
Switch to use github bot for check flaky handler

chunhtai:
Also fixes an issue where github may use a different line break


keyonghan (approved):
LGTM!

PR #1289, opened 2021-07-20
Reflect test flakiness in dashboard when test step flakes from test runner

keyonghan:
This is part of https://github.com/flutter/flutter/issues/86324.

For flakes from test runner with one build, existing dashboard simply shows green without exclamation icon.

This PR adds the exclamation icon for the above case, so that the flaky issues are easily identifiable.


app_flutter/lib/model/task.proto
```diff
@@ -25,4 +25,5 @@ message Task {
     optional string buildNumberList = 16;
     optional string builderName = 17;
     optional string luciBucket = 18;
+    optional bool is_test_flaky = 19;
```
CaseyHillers, line 28: Would it make sense to add `bool bringup` to match what's being done in the ci.yaml, and reuse the is_flaky field?

CaseyHillers (approved):
LGTM w/ comment

app_flutter/lib/model/task.proto
keyonghan: It would be good to keep consistent between frontend and backend, but I would prefer these names refreshed in a separate PR to make things easier.
https://github.com/flutter/flutter/issues/86721

PR #1292, opened 2021-07-21
[ci.yaml] Enable CI and add ci.yaml roller

CaseyHillers:

### DIFF
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -26,7 +26,6 @@ platform_properties:
 targets:
   - name: Linux Cocoon
     builder: Linux Cocoon
-    postsubmit: false
     recipe: cocoon/cocoon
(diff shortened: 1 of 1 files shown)


.ci.yaml
```diff
@@ -26,7 +26,6 @@ platform_properties:
 targets:
   - name: Linux Cocoon
     builder: Linux Cocoon
-    postsubmit: false
```
keyonghan, line 29: Do we want to run cocoon for post submit?

CaseyHillers: Yes, it would be helpful to verify changes do not break the tree, and it's not very expensive to add a new linux vm test.

keyonghan (approved):
LGTM

PR #1293, opened 2021-07-21
Load TestFlaky property for task app_flutter

keyonghan:
This is a follow up of https://github.com/flutter/cocoon/pull/1289.

We need to manually load `TestFlaky` to `task`, so that this value can be reflected on dashboard.


app_flutter/lib/service/appengine_cocoon.dart
```diff
@@ -253,7 +253,8 @@ class AppEngineCocoonService implements CocoonService {
       ..requiredCapabilities.add(objectRequiredCapabilities.toString())
       ..reservedForAgentId = taskData['ReservedForAgentID']
       ..stageName = taskData['StageName']
-      ..status = taskData['Status'];
+      ..status = taskData['Status']
+      ..isTestFlaky = taskData['TestFlaky'];
```
CaseyHillers, line 257:
```suggestion
      ..isTestFlaky = taskData['TestFlaky'] ?? false;
```

To handle release branches

CaseyHillers (approved):

app_flutter/lib/service/appengine_cocoon.dart
keyonghan: Release tasks should be the same as master ones, with default value `false`.
But added to incorporate old tasks with `null` value.

PR #1301, opened 2021-07-28
Support test ownership validation

keyonghan:
This PR uses cocoon's existing functionality to obtain ownership data based on builder configs defined in .ci.yaml, and creates a support to validate test ownership.

This PR also adds support for `Firebaselab tests`.

A companion recipe change: https://flutter-review.googlesource.com/c/recipes/+/16300.

This is part of https://github.com/flutter/flutter/issues/87144


chunhtai:
One thing I am worry about this approach is that is may be a pain to migrate TESTOWNERS format, also one can break the presubmit by changing the script without knowing.

Otherwise it looks good

keyonghan:
> One thing I am worry about this approach is that is may be a pain to migrate TESTOWNERS format, also one can break the presubmit by changing the script without knowing.

That's true, but I don't think we have a short plan to finish ownership info for shards tests or any big format migration. On the other hand, we have seen people adding tests without adding ownership over time, which causes pain to the infra team.

The main reason I put it under cocoon is we can extend the functionality to support other repos in the future. One way to avoid any unknown breakage is put it under `flutter/<repo>`, but we need separate supports for different repos and need to maintain multiple set of codes.

For now, I added more unit tests for other types to cover all scenarios to try avoiding breakage change from cocoon side without knowing.

We need to revisit the logic when we have ownership for shard tests, or when we expand the logic to other repos.

app_dart/bin/validate_task_ownership.dart
```diff
@@ -0,0 +1,92 @@
...
+import 'package:cocoon_service/src/request_handlers/flaky_handler_utils.dart';
+import 'package:yaml/yaml.dart';
+
+/// Remote check based on flutter `repo` and the commit `ref`.
+///
+/// This currently supports `flutter/flutter` only.
+Future<List<String>> remoteCheck(String repo, String ref) async {
+  final HttpClient client = HttpClient();
+  final Uri ciYamlUrl = Uri.https('raw.githubusercontent.com', 'flutter/$repo/$ref/.ci.yaml');
+  final HttpClientRequest ciYamlRequest = await client.getUrl(ciYamlUrl);
```
CaseyHillers, line 17: Some of the other bin scripts get ciYaml from a network request. Should we refactor this into a utility so they can all share that logic?

app_dart/bin/validate_task_ownership.dart
```diff
@@ -0,0 +1,92 @@
...
+  return noOwnerBuilders;
+}
+
+/// Validate test ownership defined in `testOwnersContenct` for tests configured in `ciYamlContent`.
+List<String> validateOwnership(String ciYamlContent, String testOwnersContenct) {
+  final List<String> noOwnerBuilders = <String>[];
+  final YamlMap ciYaml = loadYaml(ciYamlContent) as YamlMap;
+  final YamlList yamlList = ciYaml['targets'] as YamlList;
+  final List<YamlMap> targets = yamlList.map<YamlMap>((dynamic target) => target as YamlMap).toList();
+  for (final YamlMap target in targets) {
```
CaseyHillers, line 55: Prefer using `loadSchedulerConfig` and the `SchedulerConfig` APIs instead of pulling from yaml (pulling from yaml isn't supported)

CaseyHillers (commented):
+1 to Chun-Heng's comment about testability. You should add an integration test that ensures this is testing the tip of tree to prevent breakages landing here

app_dart/bin/validate_task_ownership.dart
```diff
@@ -0,0 +1,92 @@
...
+
+/// Remote check based on flutter `repo` and the commit `ref`.
+///
+/// This currently supports `flutter/flutter` only.
+Future<List<String>> remoteCheck(String repo, String ref) async {
+  final HttpClient client = HttpClient();
+  final Uri ciYamlUrl = Uri.https('raw.githubusercontent.com', 'flutter/$repo/$ref/.ci.yaml');
+  final HttpClientRequest ciYamlRequest = await client.getUrl(ciYamlUrl);
+  final HttpClientResponse ciYamlResponse = await ciYamlRequest.close();
+  final Uri testOwnersUrl = Uri.https('raw.githubusercontent.com', 'flutter/$repo/$ref/TESTOWNERS');
```
chunhtai, line 19: I have defined kTestOwnerPath in the flaky_handler_utils

app_dart/bin/validate_task_ownership.dart
```diff
@@ -0,0 +1,92 @@
...
+
+import 'package:cocoon_service/src/request_handlers/flaky_handler_utils.dart';
+import 'package:yaml/yaml.dart';
+
+/// Remote check based on flutter `repo` and the commit `ref`.
+///
+/// This currently supports `flutter/flutter` only.
+Future<List<String>> remoteCheck(String repo, String ref) async {
+  final HttpClient client = HttpClient();
+  final Uri ciYamlUrl = Uri.https('raw.githubusercontent.com', 'flutter/$repo/$ref/.ci.yaml');
```
chunhtai, line 16: you can use kCiYamlPath

app_dart/bin/validate_task_ownership.dart
```diff
@@ -0,0 +1,92 @@
+// Copyright 2021 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:cocoon_service/src/request_handlers/flaky_handler_utils.dart';
```
chunhtai, line 8: Should we move some the getTestowner code out to a more generic util file?

app_dart/bin/validate_task_ownership.dart
```diff
@@ -0,0 +1,92 @@
...
+    }
+  }
+  return noOwnerBuilders;
+}
+
+/// Validates task ownership.
+///
+/// It expects two parameters for remote validation: the flutter `repo` and the `commit`.
+///
+/// It expects three parameters for remote validation: `local` arg, the full path to the config
```
chunhtai, line 70:
```suggestion
/// It expects three parameters for local validation: `local` arg, the full path to the config
```

chunhtai: What is the use case for local check?

app_dart/bin/validate_task_ownership.dart
```diff
@@ -0,0 +1,92 @@
...
+/// Validate test ownership defined in `testOwnersContenct` for tests configured in `ciYamlContent`.
+List<String> validateOwnership(String ciYamlContent, String testOwnersContenct) {
+  final List<String> noOwnerBuilders = <String>[];
+  final YamlMap ciYaml = loadYaml(ciYamlContent) as YamlMap;
+  final YamlList yamlList = ciYaml['targets'] as YamlList;
+  final List<YamlMap> targets = yamlList.map<YamlMap>((dynamic target) => target as YamlMap).toList();
+  for (final YamlMap target in targets) {
+    final String builder = target['builder'] as String;
+    final String owner = getTestOwner(builder, getTypeForBuilder(builder, ciYaml), testOwnersContenct);
+    print('$builder: $owner');
```
chunhtai, line 58: print statement

app_dart/bin/validate_task_ownership.dart
keyonghan: Done!

app_dart/bin/validate_task_ownership.dart
keyonghan: Used existing utils function instead. Also updated `bing/generate_jspb.dart`.

> +1 to Chun-Heng's comment about testability. You should add an integration test that ensures this is testing the tip of tree to prevent breakages landing here

https://flutter-review.googlesource.com/c/recipes/+/16300 is ready to accept trigger from cocoon. Will create a new builder once the recipe cl is in.

app_dart/bin/validate_task_ownership.dart
keyonghan: It would be useful/handy to validate changes locally when in development.

app_dart/bin/validate_task_ownership.dart
keyonghan: I would prefer to do that in a separate PR.

app_dart/bin/validate_task_ownership.dart
keyonghan: Done.

app_dart/bin/validate_task_ownership.dart
keyonghan: changed to refer there. Also the kCiYamlPath.

app_dart/bin/validate_task_ownership.dart
keyonghan: Can you elaborate on here?

chunhtai: was this print statement intentional? I thought that was something you forgot to remove.

keyonghan: Yeah, this is intentionally to print summary message to the stdout.

app_dart/bin/validate_task_ownership.dart
chunhtai: the original suggestion is not addressed yet

app_dart/lib/src/foundation/utils.dart
```diff
@@ -12,10 +12,17 @@ import 'package:github/github.dart';
 import 'package:googleapis/bigquery/v2.dart';
 import 'package:meta/meta.dart';
 import 'package:retry/retry.dart';
+import 'package:yaml/yaml.dart';
 
+import '../../protos.dart';
 import '../foundation/typedefs.dart';
+import '../request_handlers/flaky_handler_utils.dart';
```
chunhtai, line 19: This is weird dependency, I will probably move the methods here

app_dart/lib/src/request_handlers/flaky_handler_utils.dart
```diff
@@ -292,6 +296,30 @@ String getTestOwner(String builderName, BuilderType type, String testOwnersConte
         }
         break;
       }
+    case BuilderType.firebaselab:
```
chunhtai, line 299: not related to this pr, Do we want to file mark flaky pr for firebaselab test?

app_dart/lib/src/request_handlers/flaky_handler_utils.dart
```diff
@@ -312,7 +340,9 @@ BuilderType getTypeForBuilder(String builderName, YamlMap ci) {
   // Otherwise, it is framework host only test if its tags contain both
```
chunhtai, line 340: can you also update the comment here

keyonghan: Done.

app_dart/lib/src/request_handlers/flaky_handler_utils.dart
keyonghan: I am not sure about that, but i guess we still want to have ownership for those tests.

app_dart/lib/src/foundation/utils.dart
keyonghan: Yeah, I know, but don't wanna to make this PR too complex.
Created https://github.com/flutter/flutter/issues/87287 to refactor.

app_dart/bin/validate_task_ownership.dart
keyonghan: Sorry missed here. Updated.

app_dart/bin/validate_task_ownership.dart
keyonghan: Didn't know we have integration tests here. Just created one instead.

app_dart/lib/src/request_handlers/flaky_handler_utils.dart
```diff
@@ -292,6 +296,30 @@ String getTestOwner(String builderName, BuilderType type, String testOwnersConte
...
+        if (match != null && match.namedGroup(kOwnerGroupName) != null) {
+          final List<String> lines = match
+              .namedGroup(kOwnerGroupName)
+              .split('\n')
+              .where((String line) => line.isNotEmpty && !line.startsWith('#'))
+              .toList();
+
+          for (final String line in lines) {
+            final List<String> words = line.trim().split(' ');
+            // e.g. words = ['/xxx/xxx/xxx_test.dart', '@stuartmorgan' '@flutter/tool']
```
chunhtai, line 313: nit: This does not seems like a good example

chunhtai (approved):
LGTM

app_dart/lib/src/request_handlers/flaky_handler_utils.dart
keyonghan: Removed as this is not necessary here.

CaseyHillers (approved):

PR #1305, opened 2021-07-30
[ci.yaml] Update lsc script

CaseyHillers:
Moving this to be a local only as in practice, remote ci.yaml's are impractical to update (instead checkout a local copy and iterate on it).

Added some of the fields to the yaml now that they're being used

### DIFF
--- a/app_dart/bin/yaml_converter.dart
+++ b/app_dart/bin/yaml_converter.dart
@@ -11,6 +11,7 @@ import 'package:yaml/yaml.dart';
 const int kDefaultTimeout = 30;
 
 void writeYaml(SchedulerConfig config) {
+  // Header
   final List<String> configYaml = <String>[
     '# Describes the targets run in continuous integration environment.',
     '#',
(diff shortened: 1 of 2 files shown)


app_dart/bin/yaml_converter.dart
```diff
@@ -63,27 +81,21 @@ void writeYaml(SchedulerConfig config) {
...
-  // There's an assumption that we're only generating builder configs from commits that
-  // have already landed with validation. Otherwise, this will fail.
+  final File ciYamlFile = File(args[0]);
+  final YamlMap configYaml = loadYaml(ciYamlFile.readAsStringSync()) as YamlMap;
   final SchedulerConfig schedulerConfig = schedulerConfigFromYaml(configYaml);
-  for (Target target in schedulerConfig.targets) {
-    target.name = target.builder;
-  }
+
+  //////////////////////////////////////////////////////////////////////////////
```
christopherfujino, line 92: nit. rather than the big comments, maybe a largeScaleChange helper function?

app_dart/bin/yaml_converter.dart
```diff
@@ -30,6 +29,15 @@ void writeYaml(SchedulerConfig config) {
   for (Target target in config.targets) {
     configYaml.add('  - name: ${target.name}');
     configYaml.add('    builder: ${target.builder}');
+    if (target.enabledBranches.isNotEmpty) {
```
christopherfujino, line 32: how about adding a single test for this file that verifies the output? That way in code review we can easily preview the diff a change here would make to the generated output.

CaseyHillers: Good point, i'll add an integration test to validate an empty LSC returns the same Cocoon ci.yaml. I added a boiler plate, but it still needs some work

PR #1308, opened 2021-07-30
[ci.yaml] Deprecate builder

CaseyHillers:
This is obsolete, and all instances (including releases) should have been mirroring this field.

Once we update all the ci.yamls to remove this field, we can move this to reserved so it is no longer accessible.


christopherfujino (approved):
LGTM

PR #1309, opened 2021-07-30
Lock the first row/col (labels) of the build dashboard

flar:
Note I have only tested this on the default TextDirection.ltr - not sure how to run the app with a different text direction.

And I think the popups that appear seem to match what I expect them to be, but it's not clear how to prove that other than "the task seems to have an icon that matches its cell color". Is there a more authoritative way to make sure that the data for the 12th cell on the 4th row remains consistent as we scroll around?


CaseyHillers:
>  Is there a more authoritative way to make sure that the data for the 12th cell on the 4th row remains consistent as we scroll around?

If the expectation is the task icons will be in a static place when vertically scrolling, we could validate that clicking any y level below a task icon triggers an overlay to pop up with the same task name.

I'm not sure it's feasible to horizontally do this, other than to know what LUCI build corresponds to each commit (and check the log link)

flar:
> > Is there a more authoritative way to make sure that the data for the 12th cell on the 4th row remains consistent as we scroll around?
> 
> If the expectation is the task icons will be in a static place when vertically scrolling, we could validate that clicking any y level below a task icon triggers an overlay to pop up with the same task name.
> 
> I'm not sure it's feasible to horizontally do this, other than to know what LUCI build corresponds to each commit (and check the log link)

It looks like the tests include constructing the grid with a fixed random seed so they would conceivably cover this. I can visually see that the cells launch the same notice after I scroll and re-click them, but it seems like it is something the tests might cover.

The tests are failing when run here, but they pass for me locally on my Mac so these may be golden failures that only trigger on Linux. I'll try cloning to my Linux machine and running there.

flar:
The tests are failing as a result of this recent change to scrollbars in the framework: https://github.com/flutter/flutter/pull/82687

I'll investigate and see if I can submit a separate PR to fix the tests and then rebase this one.

flar:
After rebasing on top of the fixed tests and (checking and) updating the goldens, this change is ready to go...

CaseyHillers (approved):
LGTM! Goldens make sense as the test

flar:
The one thing I thought of after I made all the changes was that maybe column 0 of the LOADING... line could have a "loading" icon, but I didn't have any ideas of what would be appropriate to put there. If someone thinks of something...

PR #1310, opened 2021-07-31
Fix golden tests after change in scrolling

flar:
A recent change in scrolling has led to an assertion failure in our scrolling golden tests. Apparently you need to use pumpAndSettle() to avoid the assertion.

I've verified the above solution, but I'm starting this PR off with an empty commit to verify that the current tree cannot pass its own tests. When it fails, I'll push a second commit with the fixes to verify that it passes.

~Note that this points out another issue with the tests here - the tests that failed were attempting to see the "LOADING..." banner after a scroll (as per the comments on the tests themselves), and I'm guessing that is why they used pump() rather than pumpAndSettle() because they needed to get the screen capture before the new data came in. As far as I can tell, that never happened and the golden files for those tests never had the "LOADING..." banner in the first place. I believe that something will need to be done with the data generator for the tests so that you can pause it, scroll, pumpAndSettle, see the banner, then resume the data source...~
So, it looks like the single pump() used to be testing the LOADING... banner by getting a redraw before new data came in, but it was hard to see because everything was white on white. Switching to pumpAndSettle() meant that the banner row was showing up as valid data now. So, I implemented explicit pausing of the testing cocoon data source so that the "LOADING" banner would still show up.

### DIFF
--- a/app_flutter/lib/service/dev_cocoon.dart
+++ b/app_flutter/lib/service/dev_cocoon.dart
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'dart:async';
 import 'dart:math' as math;
 
 import 'package:fixnum/fixnum.dart';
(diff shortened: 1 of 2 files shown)


flar:
It looks  like the tests use pump() almost everywhere instead of pumpAndSettle(). Those uses should be double-checked.

CaseyHillers (approved):
LGTM!

PR #1311, opened 2021-08-02
[cocoon device doctor] Kill android process before running tasks

keyonghan:
This PR adds a support for a new action `prepare`, which prepares the phone before running a task. One feature added here for android device is: clean up the top activity in android devices.

Related: https://github.com/flutter/flutter/issues/83938

### DIFF
--- a/device_doctor/lib/src/android_device.dart
+++ b/device_doctor/lib/src/android_device.dart
@@ -245,8 +245,8 @@ class AndroidDevice implements Device {
     // Example result:
     //
     // Proc # 0: fore  T/A/T  trm: 0 4544:com.google.android.googlequicksearchbox/u0a66 (top-activity)
-    final RegExp processRegExp = RegExp(' 0 [0-9]+:(?<process>.+)/');
-    final List<String> processes = results.map((e) => processRegExp.firstMatch(e).namedGroup('process')).toList();
+    final List<String> processes =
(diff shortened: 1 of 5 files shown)


device_doctor/lib/src/android_device.dart
```diff
@@ -215,4 +222,40 @@ class AndroidDevice implements Device {
...
+    // Skip uninstalling process when no device is available or no application exists.
+    if (result == 'adb: no devices/emulators found' || result == null || result.isEmpty) {
+      stdout.write('no process is running');
+      return true;
+    }
+    final List<String> results = result.trim().split('\n');
+    // Example result:
+    //
+    // Proc # 0: fore  T/A/T  trm: 0 4544:com.google.android.googlequicksearchbox/u0a66 (top-activity)
+    final RegExp processRegExp = RegExp(' 0 [0-9]+:(?<process>.+)/');
```
gaaclarke, line 248: Will that `0` always be there?  I'm not sure what it represents.  `trm: \d+ \d+:(.+)/` would be more robust.

gaaclarke (commented):
LGTM, nit about a fragile regex

device_doctor/lib/src/android_device.dart
keyonghan: Was following the [first answer](https://stackoverflow.com/questions/31079853/how-do-i-kill-all-active-tasks-apps-using-adb/31079988). 
Now switched to use index instead of regexp to extract the process more accurate.

godofredoc (approved):

PR #1312, opened 2021-08-02
Add more logs when pushing build status to github

keyonghan:
Related: https://github.com/flutter/flutter/issues/87512

### DIFF
--- a/app_dart/lib/src/request_handlers/push_build_status_to_github.dart
+++ b/app_dart/lib/src/request_handlers/push_build_status_to_github.dart
@@ -73,7 +73,11 @@ class PushBuildStatusToGithub extends ApiRequestHandler<Body> {
     final Map<LuciBuilder, List<LuciTask>> luciTasks = await luciService.getRecentTasks(builders: postsubmitBuilders);
 
     String status = GithubBuildStatusUpdate.statusSuccess;
-    for (List<LuciTask> tasks in luciTasks.values) {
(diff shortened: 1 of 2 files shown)


app_dart/lib/src/request_handlers/push_build_status_to_github.dart
```diff
@@ -73,7 +73,11 @@ class PushBuildStatusToGithub extends ApiRequestHandler<Body> {
     final Map<LuciBuilder, List<LuciTask>> luciTasks = await luciService.getRecentTasks(builders: postsubmitBuilders);
 
     String status = GithubBuildStatusUpdate.statusSuccess;
-    for (List<LuciTask> tasks in luciTasks.values) {
+    for (LuciBuilder luciBuilder in luciTasks.keys) {
+      final List<LuciTask> tasks = luciTasks[luciBuilder];
```
jmagman, line 77: Avoid the value lookup, iterate over the keys and values at the same time.
```suggestion
    luciTasks.forEach((LuciBuilder luciBuilder, List<LuciTask> tasks) {
```

jmagman (approved):
LGTM with iteration nit.

app_dart/lib/src/request_handlers/push_build_status_to_github.dart
keyonghan: Thanks. Updated.

jmagman (approved):
LGTM!

PR #1313, opened 2021-08-02
Change to use name instead of builder defined in .ci.yaml

keyonghan:
To support https://github.com/flutter/flutter/pull/87522

### DIFF
--- a/app_dart/lib/src/request_handlers/flaky_handler_utils.dart
+++ b/app_dart/lib/src/request_handlers/flaky_handler_utils.dart
@@ -23,7 +23,7 @@ const String kP5Label = 'P5';
 const String kP6Label = 'P6';
 const String kBigQueryProjectId = 'flutter-dashboard';
 const String kCiYamlTargetsKey = 'targets';
-const String kCiYamlTargetBuilderKey = 'builder';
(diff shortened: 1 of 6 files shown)


app_dart/lib/src/request_handlers/flaky_handler_utils.dart
```diff
@@ -360,7 +360,7 @@ BuilderType getTypeForBuilder(String builderName, YamlMap ci) {
 List<dynamic> _getTags(String builderName, YamlMap ci) {
   final YamlList targets = ci[kCiYamlTargetsKey] as YamlList;
   final YamlMap target = targets.firstWhere(
-    (dynamic element) => element[kCiYamlTargetBuilderKey] == builderName,
+    (dynamic element) => element[kCiYamlTargetNameKey] == builderName,
```
CaseyHillers, line 363: This would've been caught in https://github.com/flutter/cocoon/pull/1308 if it used the proto. This isn't using the dedicated API, and is likely to be broken again. I filed https://github.com/flutter/flutter/issues/87529

CaseyHillers (approved):
Thanks for the quick fix!

chunhtai (approved):
LGTM

PR #1316, opened 2021-08-03
Add scheduler_job_id tag when reset a prod task from build dashboard

keyonghan:
Issue: https://github.com/flutter/flutter/issues/87599

### DIFF
--- a/app_dart/lib/src/service/luci_build_service.dart
+++ b/app_dart/lib/src/service/luci_build_service.dart
@@ -470,7 +470,7 @@ class LuciBuildService {
     ];
     tags['user_agent'] = <String>['luci-scheduler'];
     // Tag `scheduler_job_id` is needed when calling buildbucket search build API.
-    tags['scheduler_job_id'] = <String>[builderName];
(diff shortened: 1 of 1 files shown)


jmagman (approved):
RSLGTM

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -469,6 +469,8 @@ class LuciBuildService {
       'commit/gitiles/chromium.googlesource.com/external/github.com/flutter/$repo/+/$commitSha',
     ];
     tags['user_agent'] = <String>['luci-scheduler'];
+    // Tag `scheduler_job_id` is needed when calling buildbucket search build API.
+    tags['scheduler_job_id'] = <String>[builderName];
```
godofredoc, line 473: I believe we also need the repository. E.g 'flutter/Linux Android AOT Engine'

keyonghan: Good catch! Added.

godofredoc (approved):

PR #1318, opened 2021-08-05
Skip cocoon backend cloud build if no new commits

keyonghan:
This is for https://github.com/flutter/flutter/issues/87749.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,21 +1,73 @@
 # Provide instructions for google Cloud Build to auto-build flutter
 # dashboard to flutter-dashboard project. Auto-build will be triggered
-# by new commits to `master` branch.
+# by daily schedule on `master` branch.
+#
+# The auto-build will be skipped if no new commits since last deployment.
+#
+# For Monday, check if any new commits in last three days
+# For Tuesday through Friday, check if any new commits in last day.
+# 
+# When comparing `commit_time` with `last_deploy_time`, there is a 7-hour difference:
+# `commit_time` is PST whereas the `last_deploy_time` is UTC.
 
 steps:
   # Build app_flutter.
   - name: gcr.io/$PROJECT_ID/flutter
     entrypoint: '/bin/bash'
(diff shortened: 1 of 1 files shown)


cloudbuild.yaml
```diff
@@ -1,21 +1,61 @@
...
+        else
+          bash cloud_build/repo_dash_build.sh
+        fi
 
   # Deploy a new version to google cloud.
   - name: 'gcr.io/cloud-builders/gcloud'
     entrypoint: '/bin/bash'
-    args: ['cloud_build/deploy.sh', '$PROJECT_ID', '$SHORT_SHA', '$_GAE_PROMOTE']
+    args:  
+      - '-c'
```
CaseyHillers, line 52: Couldn't we just add the check in the deploy step? e.g. build everything, but only deploy if it's a new version?

keyonghan: I don't see the value to build everything we have already built/deployed before.

CaseyHillers: Cocoon changes aren't the only thing that break our deploys, and it's helpful to know when those start to occur

cloudbuild.yaml
```diff
@@ -1,21 +1,61 @@
...
+        echo "latest_version: $latest_version"
+        echo "current_version: $current_version"
+        if [ "$latest_version" = "$current_version" ]; then
+          echo "No updates since last deployment."
+        else
+          bash cloud_build/app_flutter_build.sh
+        fi
 
-    # Build repo_dashboard.
+  #   # Build repo_dashboard.
```
CaseyHillers, line 35: nit: extra #

cloudbuild.yaml
keyonghan: Ah, missed that part. Updated to always build.

cloudbuild.yaml
keyonghan: Done.

CaseyHillers (approved):

PR #1319, opened 2021-08-06
Fixes flaky script

chunhtai:
fixes https://github.com/flutter/flutter/issues/87500

script will now add team label
Skip filing the pr if no entry in ci.yaml
updated ignored list for deflake script

### DIFF
--- a/app_dart/lib/src/request_handlers/flaky_handler_utils.dart
+++ b/app_dart/lib/src/request_handlers/flaky_handler_utils.dart
@@ -15,12 +15,17 @@ import '../service/github_service.dart';
 // String constants.
 const String kTeamFlakeLabel = 'team: flakes';
 const String kSevereFlakeLabel = 'severe: flake';
+const String kFrameworkLabel = 'framework';
+const String kToolLabel = 'tool';
+const String kEngineLabel = 'engine';
+const String kWebLabel = 'platform-web';
 const String kP1Label = 'P1';
 const String kP2Label = 'P2';
 const String kP3Label = 'P3';
 const String kP4Label = 'P4';
 const String kP5Label = 'P5';
 const String kP6Label = 'P6';
+
 const String kBigQueryProjectId = 'flutter-dashboard';
 const String kCiYamlTargetsKey = 'targets';
 const String kCiYamlTargetNameKey = 'name';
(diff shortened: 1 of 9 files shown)


app_dart/lib/src/request_handlers/flaky_handler_utils.dart
```diff
@@ -74,11 +85,16 @@ Please follow https://github.com/flutter/flutter/wiki/Reducing-Test-Flakiness#fi
...
 
   List<String> get issueLabels {
-    return <String>[
+    final List<String> labels = <String>[
       kTeamFlakeLabel,
       kSevereFlakeLabel,
       kP1Label,
     ];
+    final String teamLabel = _getTeamLabelFromTeam(ownership.team);
+    if (teamLabel?.isNotEmpty == true) {
```
keyonghan, line 94: Nit: skip `== true`.

app_dart/lib/src/request_handlers/flaky_handler_utils.dart
```diff
@@ -425,10 +454,56 @@ String _issueBuildLink({String builder, String build}) {
...
+      return kFrameworkLabel;
+    case Team.engine:
+      return kEngineLabel;
+    case Team.tool:
+      return kToolLabel;
+    case Team.web:
+      return kWebLabel;
+    case Team.unknown:
+      return null;
+  }
```
keyonghan, line 483: A return is needed out of the switch block.

keyonghan (commented):
Seems CI is failing.

app_dart/lib/src/request_handlers/flaky_handler_utils.dart
chunhtai: is
```dart
if (teamLabel != null && teamLabel!. isNotEmpty)
```
prefer over this?

In flutter/flutter we usually do this instead

keyonghan: TIL. Let's keep it with flutter/flutter then.

app_dart/lib/src/request_handlers/flaky_handler_utils.dart
chunhtai: fixed

keyonghan (approved):
LGTM with tests pass.

PR #1320, opened 2021-08-06
Redirect traffic only after the deployment is successful.

godofredoc:
This is to address the problems with the backend not being available
during the deployment process.

Bug: https://github.com/flutter/flutter/issues/87222


keyonghan (approved):
LGTM

PR #1321, opened 2021-08-06
Document upgrading dependencies via ci.yaml

CaseyHillers:
This is to support easier bringup of new dependencies like https://github.com/flutter/flutter/issues/87649


CaseyHillers:
@jmagman you may find this useful when supporting new versions of xcode as well (feel free to give an feedback on how we can improve this)

blasten (approved):

PR #1322, opened 2021-08-06
add a progress indicator in column 0 of LOADING...

flar:
Replace the odd space in column 0 of the loading row with a progress indicator and make sure at least one full instance of "LOADING..." is visible.

This PR also adds synthetic loading times to the DevCocoon source (not used for tests) for better visual inspection of the loading indicators.


flar:
> This PR also adds synthetic loading times to the DevCocoon source (not used for tests) for better visual inspection of the loading indicators.

To be clear - that source is used to drive the tests as well, but the synthetic loading delays are not enabled except for the UI. The same is true of the animated progress indicator, it is used only for the full UI, not for the tests/goldens.

CaseyHillers (approved):
LGTM! Thanks for cleaning up the loading spam

PR #1325, opened 2021-08-10
[ci.yaml] Generate Linux ci_yaml roller in presubmit

CaseyHillers:
This won't run in presubmit, but will generate the try builder in LUCI config. Once this has propagated, we can remove the bringup: true tag.

This will ensure breakages can't land in presubmit


CaseyHillers:
It did add the presubmit check, let me go and fix this.

keyonghan (approved):
LGTM.

PR #1326, opened 2021-08-10
[scheduler] Don't run bringup presubmit targets

CaseyHillers:
This is necessary for adding new presubmit targets in the .ci.yaml

The expectation is developers add a target that is in bringup, and wait for it to soak. Once prod is green, they can remove bringup to enable the presubmit.

Currently, this doesn't work as the ci.yaml change can't be landed as the builder doesn't exist yet.

### DIFF
--- a/app_dart/lib/src/service/scheduler.dart
+++ b/app_dart/lib/src/service/scheduler.dart
@@ -241,7 +241,8 @@ class Scheduler {
   /// and that do not have any dependencies.
   List<Target> getPreSubmitTargets(Commit commit, SchedulerConfig config) {
     // Filter targets to only those run in presubmit.
-    final Iterable<Target> presubmitTargets = config.targets.where((Target target) => target.presubmit);
+    final Iterable<Target> presubmitTargets =
+        config.targets.where((Target target) => target.presubmit && !target.bringup);
 
     return _filterEnabledTargets(commit, config, presubmitTargets.toList());
   }
--- a/app_dart/test/service/scheduler_test.dart
+++ b/app_dart/test/service/scheduler_test.dart
@@ -310,6 +310,10 @@ targets:
     enabled_branches:
       - master
     presubmit: true
+  - name: Linux D
+    scheduler: luci
+    bringup: true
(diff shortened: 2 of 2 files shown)


app_dart/lib/src/service/scheduler.dart
```diff
@@ -241,7 +241,8 @@ class Scheduler {
   /// and that do not have any dependencies.
   List<Target> getPreSubmitTargets(Commit commit, SchedulerConfig config) {
     // Filter targets to only those run in presubmit.
-    final Iterable<Target> presubmitTargets = config.targets.where((Target target) => target.presubmit);
+    final Iterable<Target> presubmitTargets =
+        config.targets.where((Target target) => target.presubmit && !target.bringup);
```
keyonghan, line 245: I have been thinking about whether we need to distinguish pre submit and post submit regarding `bringup`.
With this change here, if a test is `bringup: true`, it will not run in presubmit, but run in post submit without affecting the tree.
When it comes to flakiness in presubmit, I would be interested in running presubmit tests as well, but without affecting the pre-submit tree. A separate control sounds fit here. Does that make sense?
This is based on the assumption that sometimes for some reason, a test maybe flaky in presubmit, but the same test may not be flaky in post submit. I guess that's why we also have the [pre-submit flaky label](https://github.com/flutter/flutter/issues?q=is%3Aopen+is%3Aissue+label%3A%22team%3A+presubmit+flakes%22).
If the flakiness status of the same test is same for both pre-submit and post-submit, then we may not need to separately address the flaky issues for presubmit. I still need to look into the data to see if this is necessary.

CaseyHillers: This is necessary from the LUCI config side, otherwise it's not possible to add new presubmit targets.

We could enable running bringup presubmit builds, but not updating the check. However, we need to validate this doesn't impact SLOs.

My thought is if presubmit and postsubmit infra are the same, we shouldn't spend extra capacity on the presubmit runs.

keyonghan (approved):
LGTM

PR #1327, opened 2021-08-11
Fix list formatting in ci.yaml docs

stuartmorgan:
Fixes indentation of inline blocks and sublists, and replaces unsupported use of letters as list markers, so that the markdown will display correctly on GitHub.


CaseyHillers:
FYI Cocoon does not allow human submissions, you'll need to use the waiting for tree to go green label here.

PR #1329, opened 2021-08-13
[ci.yaml] Add section on adding new targets

CaseyHillers:


gspencergoog (approved):
32384589-a60f0e74-c078-11e7-9bc1-e5b5287aea9d

CI_YAML.md
```diff
@@ -54,6 +54,19 @@ targets:
...
+
+All new targets should be added as `bringup: true` to ensure they do not block the tree.
+
+Targets based on the LUCI or Cocoon schedulers will first need to be mirrored to flutter/infra
+before they will be run. This propagation takes about 30 minutes, and will only run as non-blocking
+in postsubmit.
+
+The target will show runs in https://ci.chromium.org/p/flutter (under the repo). See
+https://github.com/flutter/flutter/wiki/Adding-a-new-Test-Shard for up to date information
+on the steps to promote your target to blocking. In general, this involves 10 consecutive green
```
keyonghan, line 67: For framework, 50 consecutive green runs are needed as of now (based on SLO 2%).

CI_YAML.md
```diff
@@ -54,6 +54,19 @@ targets:
        ["framework", "hostonly"]
 ```
 
+## Adding new targets
+
+All new targets should be added as `bringup: true` to ensure they do not block the tree.
```
keyonghan, line 59: We want to enforce adding the target in staging pool first before moving it to the prod, even with `bringup: true`. Could you add a comment here? 
BTW: do we have plans to support staging targets in the .ci.yaml?

CaseyHillers: My understanding of staging is that we want it for new testbeds, not necessarily new tests. Is there a doc with more details?

I can add support for `.staging_ci.yaml` with all the plumbing to generate these staging builders, but this doc is intended for developers.

keyonghan: It will be for any new test, not only a new testbed.
Thanks. I will /cc you when that section is ready.

CI_YAML.md
keyonghan: Nit: the auto deflake process is running only on framework repo and the 50 is also only applicable to the framework.

keyonghan (approved):
LGTM with nit.

CI_YAML.md
CaseyHillers: Would the plan be to deprecate the `bringup` field then? For example, what about re-sharding tests. Are those new tests?

keyonghan: No. The bringup will still be helpful for existing tests that are flaky. Only those new tests or old ones that have not been running for a while need to go through the staging pool.
Since the shard tests have been running, the re-sharding probably not need to follow.

PR #1333, opened 2021-08-16
Upload task metrics for staging tests

keyonghan:
The `update-task-status` API has been updating task metrics for prod tests only.

As requested in https://github.com/flutter/flutter/issues/88296, we are adding support for staging tests. Here the task metrics will only be uploaded to `metrics_center`, rather than datastore (which will show up in flutter build dashboard).

### DIFF
--- a/app_dart/lib/src/request_handlers/update_task_status.dart
+++ b/app_dart/lib/src/request_handlers/update_task_status.dart
@@ -55,25 +55,31 @@ class UpdateTaskStatus extends ApiRequestHandler<UpdateTaskStatusResponse> {
     final Map<String, dynamic> resultData =
         requestData[resultsParam] as Map<String, dynamic> ?? const <String, dynamic>{};
     final List<String> scoreKeys = (requestData[scoreKeysParam] as List<dynamic>)?.cast<String>() ?? const <String>[];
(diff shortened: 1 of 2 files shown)


CaseyHillers (commented):
Have you looked into porting the skia perf upload logic into the test runner? We already do that for the engine, and it might not make sense to do it in Cocoon now that the migration is complete.

app_dart/lib/src/request_handlers/update_task_status.dart
```diff
@@ -55,25 +55,31 @@ class UpdateTaskStatus extends ApiRequestHandler<UpdateTaskStatusResponse> {
     final Map<String, dynamic> resultData =
         requestData[resultsParam] as Map<String, dynamic> ?? const <String, dynamic>{};
     final List<String> scoreKeys = (requestData[scoreKeysParam] as List<dynamic>)?.cast<String>() ?? const <String>[];
+    final String builderName = requestData[builderNameParam] as String;
```
CaseyHillers, line 58: Should we trim here as well?

CaseyHillers (commented):
Is this still needed, or should we close the pull request?

PR #1336, opened 2021-08-24
Updates status calculation to pass if any success.

godofredoc:
Retry logic in flutter dashboard does not allow to retry if there was at
least a successful execution and the logic to paint the task-commit in
the grid was using the signal of the last task. This PR updates the
logic to calculate the grid status to make it consistent with the retry
logic.

Bug: https://github.com/flutter/flutter/issues/88197

### DIFF
--- a/app_dart/lib/src/service/build_status_provider.dart
+++ b/app_dart/lib/src/service/build_status_provider.dart
@@ -65,7 +65,7 @@ class BuildStatusService {
       return BuildStatus.failure();
     }
 
-    final List<String> failedTasks = <String>[];
+    final Set<String> failedTasks = <String>{};
     for (CommitStatus status in statuses) {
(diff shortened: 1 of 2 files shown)


CaseyHillers (changes requested):
The proposed change does not align with the linked issue. Is there some missing context?

Datastore only stores the last run of a build in `Task`. Your proposal would let the tree be green as long any of the last 20 commits had a passing run of this task.

godofredoc:
> The proposed change does not align with the linked issue. Is there some missing context?
> 
> Datastore only stores the last run of a build in `Task`. Your proposal would let the tree be green as long any of the last 20 commits had a passing run of this task.

Thanks for the information. I'm closing this PR as implementing the functionality will require a full refactoring of the datastore.

PR #1342, opened 2021-08-31
Prepare to remove devicelab service account.

godofredoc:
As the pubsub messages are coming from the same project as the appengine
instance there is no reason to use an additional service account. This
PR removes the dependency on the devicelab service account.

### DIFF
--- a/app_dart/lib/src/request_handlers/luci_status.dart
+++ b/app_dart/lib/src/request_handlers/luci_status.dart
@@ -117,7 +117,10 @@ class LuciStatusHandler extends RequestHandler<Body> {
     if (info.expiresIn == null || info.expiresIn < 1) {
       return false;
     }
-    final ServiceAccountInfo devicelabServiceAccount = await config.deviceLabServiceAccount;
-    return info.email == devicelabServiceAccount.email;
+    final Set<String> allowedServiceAccounts = <String>{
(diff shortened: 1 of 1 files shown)


app_dart/lib/src/request_handlers/luci_status.dart
```diff
@@ -117,7 +116,10 @@ class LuciStatusHandler extends RequestHandler<Body> {
     if (info.expiresIn == null || info.expiresIn < 1) {
       return false;
     }
-    final ServiceAccountInfo devicelabServiceAccount = await config.deviceLabServiceAccount;
-    return info.email == devicelabServiceAccount.email;
+    final Set<String> allowedServiceAccounts = <String>{
+      'flutter-devicelab@flutter-dashboard.iam.gserviceaccount.com',
+      'flutter-dashboard@appspot.gserviceaccount.com'
```
CaseyHillers, line 121: Can we add this in `Config`? That's where I've been adding service accounts so it's easy to reference what's being used

godofredoc: This is a transitional step, I'll update the configs with the default account once we have migrated all the use cases.

CaseyHillers: Ack

CaseyHillers (approved):

PR #1343, opened 2021-09-01
Remove the use of devicelab service account.

godofredoc:
This is to use the application default account for everything in the
backend.


CaseyHillers (approved):

keyonghan (approved):
LGTM

PR #1344, opened 2021-09-02
Prepare search requests considering realm when calling builderbucket API

keyonghan:
The `staging` suffix was removed in https://flutter-review.googlesource.com/c/infra/+/17400 for staging builders, and same builder name exists in both staging and prod pools. This results in `realm` appended to `scheduler_job_id`:
Screen Shot 2021-09-02 at 11 19 43 AM

Cocoon has been querying without specifying the `realm`, and miss those builds.

This PR adds the missing `realm` case.

### DIFF
--- a/app_dart/lib/src/service/luci.dart
+++ b/app_dart/lib/src/service/luci.dart
@@ -180,7 +180,7 @@ class LuciService {
     // Log error of any response if existing.
     // Any response code other than 200 suggests there is an error.
     // https://cloud.google.com/apis/design/errors#handling_errors
-    if (batchResponse.responses.any((Response response) => response.error.code != 200)) {
+    if (batchResponse.responses.any((Response response) => response.error != null && response.error.code != 200)) {
       final String responseError = batchResponse.responses
           .map<String>((Response response) {
             if (response.error.code == null) {
--- a/app_dart/lib/src/service/luci.dart
+++ b/app_dart/lib/src/service/luci.dart
@@ -47,9 +47,13 @@ class LuciService {
     @required this.buildBucketClient,
     @required this.config,
     @required this.clientContext,
+    @required this.log,
   })  : assert(buildBucketClient != null),
         assert(config != null),
-        assert(clientContext != null);
+        assert(clientContext != null),
+        assert(log != null);
+
+  final Logging log;
 
   /// Client for making buildbucket requests to.
   final BuildBucketClient buildBucketClient;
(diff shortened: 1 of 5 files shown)


app_dart/lib/src/service/luci.dart
```diff
@@ -181,26 +196,26 @@ class LuciService {
...
-    return builds;
+
+    searchRequests.addAll(builders.where(includeBuilder).map<Request>((LuciBuilder builder) {
+      return _getRequest(<String>['flutter/prod-${builder.name}']);
+    }).toList());
+
+    return searchRequests;
+  }
+
+  Request _getRequest(List<String> schedulerJobId) {
```
godofredoc, line 209: docs?

app_dart/lib/src/service/luci.dart
```diff
@@ -170,6 +170,21 @@ class LuciService {
...
+        .map<SearchBuildsResponse>((Response response) => response.searchBuilds)
+        .where((SearchBuildsResponse response) => response.builds != null)
+        .expand<Build>((SearchBuildsResponse response) => response.builds);
+    return builds;
+  }
+
+  /// Prepare search requests via tags predicate.
+  ///
+  /// For builders only enabled in prod, the scheduler_job_id follows: `flutter/<builder.name>`
+  /// For builders enabled in both prod and staging, the scheduler_job_id follows: `flutter/<realm>-<builder.name>`.
```
godofredoc, line 186: Why we need new tags if we can use the bucket? ideally the predicate just adds the bucket to the query.

app_dart/lib/src/service/luci.dart
```diff
@@ -170,6 +170,21 @@ class LuciService {
   ///
   /// Latest builds of each builder will be returned from new to old.
   Future<Iterable<Build>> getBuilds(String repo, bool requireTaskName, List<LuciBuilder> builders) async {
+    final List<Request> searchRequests = prepareSearchRequests(repo, requireTaskName, builders);
+    final BatchRequest batchRequest = BatchRequest(requests: searchRequests);
+    final BatchResponse batchResponse = await buildBucketClient.batch(batchRequest);
```
godofredoc, line 175: Can we add logging for any failed response?

app_dart/lib/src/service/luci.dart
keyonghan: Added.

app_dart/lib/src/service/luci.dart
keyonghan: That would be easy, but the Buildbucket API doesn't support bucket search: https://chromium.googlesource.com/infra/luci/luci-go/+/refs/heads/main/buildbucket/proto/builds_service.proto#381

app_dart/lib/src/service/luci.dart
keyonghan: Added.

app_dart/lib/src/service/luci.dart
godofredoc: It supports BuilderId https://chromium.googlesource.com/infra/luci/luci-go/+/refs/heads/main/buildbucket/proto/builder.proto#15 which is a combination of project, bucket and builder name.

keyonghan: I recalled why we switched from BuildId to Tags: https://github.com/flutter/cocoon/pull/1314

Dart HHH is sharing the same builder as us, and their builds miss commit sha info which is needed from our backend (https://github.com/flutter/flutter/issues/87512)

godofredoc (approved):

PR #1345, opened 2021-09-02
[ci.yaml] Add documentation about properties

CaseyHillers:


christopherfujino (approved):
LGTM

CI_YAML.md
```diff
@@ -76,6 +76,49 @@ promote a test that has been passing for the past 50 runs.
 To prevent tests from rotting, all targets are required to have a clear owner. Add an
 owner in [TESTOWNERS](https://github.com/flutter/flutter/blob/master/TESTOWNERS)
 
+## Properties
+
+Targets support specifying properties that can be passed throughout infrastructure. The
+following are a list of keys that are reserved for special use. Note, properties is a
```
christopherfujino, line 82: nit. maybe make this bold or something, because it's probably the trickiest part of this whole setup and easy to miss when skimming this doc.

CI_YAML.md
```diff
@@ -76,6 +76,49 @@ promote a test that has been passing for the past 50 runs.
...
+Map<String, String> and any special values can be JSON encoded.
+
+**add_recipes_cq**: String boolean whether to add this target to flutter/recipes CQ. This ensures
+changes to flutter/recipes pass on this target before landing.
+
+**caches**: JSON representation of swarming caches the bot running the target should have.
+Name is what flutter/recipes will refer to it as, and path is where it will be stored on the bot.
+Path should be versioned.
```
stuartmorgan, line 90: What does this last bit mean? I don't see anything that looks like a version in the example, which seems very confusing.

CI_YAML.md
```diff
@@ -76,6 +76,49 @@ promote a test that has been passing for the past 50 runs.
...
+
+Example
+```yaml
+caches: >-
+  [
+    {"name":"gradle","path":"gradle"}
+  ]
+```
+
+**dependencies**: JSON list of objects with "dependency" and optionally "version".
```
stuartmorgan, line 100: Maybe include a version in one of the examples?

CI_YAML.md
CaseyHillers: I added some more color, to the doc.

We've now had LUCI long enough that we are starting to need to support versioning dependencies. All the original deps and caches are not versioned, so we upgrade deps they will become versioned (see below an example for openjdk 11)

```
      caches: >-
        [
          {"name": "openjdk", "path": "java11"}
        ]
```

PR #1346, opened 2021-09-08
Fix vacuum handler.

godofredoc:
The vacuum handler was raising an exception trying to call a method in a
null.

Bug: https://github.com/flutter/flutter/issues/89622

### DIFF
--- a/app_dart/lib/src/request_handlers/vacuum_github_commits.dart
+++ b/app_dart/lib/src/request_handlers/vacuum_github_commits.dart
@@ -64,8 +64,11 @@ class VacuumGithubCommits extends ApiRequestHandler<Body> {
     final DateTime queryAfter = DateTime.now().subtract(const Duration(days: 1));
     final DateTime queryBefore = DateTime.now().subtract(const Duration(minutes: 3));
     try {
+      log!.debug(
+          'Listing commit for slug: $slug branch: $branch and msSinceEpoch: ${queryAfter.millisecondsSinceEpoch}');
       commits = await githubService.listCommits(slug, branch, queryAfter.millisecondsSinceEpoch);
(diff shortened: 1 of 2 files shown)


app_dart/lib/src/request_handlers/vacuum_github_commits.dart
```diff
@@ -65,7 +65,10 @@ class VacuumGithubCommits extends ApiRequestHandler<Body> {
     final DateTime queryAfter = DateTime.now().subtract(const Duration(days: 1));
     final DateTime queryBefore = DateTime.now().subtract(const Duration(minutes: 3));
     try {
+      log.debug(
+          'Listing commit for slug: $slug branch: $branch and msSinceEpoch: ${queryAfter.millisecondsSinceEpoch}');
       commits = await githubService.listCommits(slug, branch, queryAfter.millisecondsSinceEpoch);
+      commits = commits ?? <RepositoryCommit>[];
```
yusufm, line 71: Is this ok to be empty, will it cause issues below?

CaseyHillers (approved):

app_dart/lib/src/request_handlers/vacuum_github_commits.dart
godofredoc: As long as it is not null it will prevent errors below. We still need to fix the root case which is failures in the github requests.

yusufm (approved):

CaseyHillers:
Cocoon is now accepting pull requests again. To make your PR compatible,
  1. Please update the base to point to main (edit in the GitHub UI)
  2. Please rebase on tip of tree. There is now null safety support and you'll have to update your PR to be null safe. There may be some rough edges around the initial migration, so feel free to add fixes as you see fit.

PR #1348, opened 2021-09-15
Create checks and batch build requests asynchronously.

godofredoc:
This is to prevent the task timing out before the builds are scheduled.

### DIFF
--- a/app_dart/lib/src/service/luci_build_service.dart
+++ b/app_dart/lib/src/service/luci_build_service.dart
@@ -161,7 +161,7 @@ class LuciBuildService {
   /// Creates BuildBucket [Request] using [checkSuiteEvent], [slug], [prNumber], [commitSha], [builder], [builderId]
   /// and [userData].
   Future<Request> _createBuildRequest(CheckSuiteEvent? checkSuiteEvent, github.RepositorySlug slug, int prNumber,
-      String commitSha, String builder, BuilderId builderId, Map<String, dynamic> userData) async {
+      String commitSha, String? builder, BuilderId builderId, Map<String, dynamic> userData) async {
     int? checkRunId;
     if (checkSuiteEvent != null || config.githubPresubmitSupportedRepo(slug)) {
       final github.CheckRun checkRun = await githubChecksUtil.createCheckRun(
@@ -197,16 +197,16 @@ class LuciBuildService {
   /// Schedules BuildBucket builds for a given [prNumber], [commitSha]
   /// and Github [slug].
   Future<void> scheduleTryBuilds({
-    @required List<LuciBuilder> builders,
-    @required int prNumber,
-    @required String commitSha,
-    @required github.RepositorySlug slug,
-    CheckSuiteEvent checkSuiteEvent,
+    required List<LuciBuilder> builders,
+    required int prNumber,
+    required String commitSha,
+    required github.RepositorySlug slug,
+    CheckSuiteEvent? checkSuiteEvent,
   }) async {
     if (!config.githubPresubmitSupportedRepo(slug)) {
       throw BadRequestException('Repository ${slug.name} is not supported by this service.');
     }
-    List<String> builderNames = await _builderNamesForRepo(
+    List<String?> builderNames = await _builderNamesForRepo(
       builders,
       prNumber,
       commitSha,
@@ -234,7 +234,7 @@ class LuciBuildService {
     String commitSha,
     github.RepositorySlug slug,
   ) async {
-    final Map<String, Build> builds = await tryBuildsForRepositoryAndPr(
+    final Map<String?, Build?> builds = await tryBuildsForRepositoryAndPr(
       slug,
       prNumber,
       commitSha,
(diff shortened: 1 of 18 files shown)


godofredoc:
This is ready for review.

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -204,56 +249,24 @@ class LuciBuildService {
...
-      retryIf: (Exception e) => e is BuildBucketException,
-    );
-    for (Response response in batchResponse.responses) {
-      if (response.error?.code != 0) {
-        log.warning('BatchResponse error: $response');
-        continue;
+    final List<Request> requests = await Future.wait(requestFutures);
+    List<BatchResponse> batchResponseList;
+    final List<Future<BatchResponse>> futureBatchResponses = <Future<BatchResponse>>[];
+    final Iterable<List<Request>> requestPartitions = await shard(requests, 1);
```
keyonghan, line 257: Is this intentional to have batch size 1? This will incur too many buildbucket API calls, have we checked this comply with QPS requirement from buildbucket side?

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -204,56 +249,24 @@ class LuciBuildService {
...
+    for (List<Request> requestPartition in requestPartitions) {
+      futureBatchResponses.add(buildBucketClient.batch(BatchRequest(requests: requestPartition)));
```
keyonghan, line 259: If we only want to query based on size 1, then no batch request is needed.

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -262,17 +275,35 @@ class LuciBuildService {
       }
 
       final Build scheduleBuild = response.scheduleBuild;
+      if (scheduleBuild.tags == null) {
```
keyonghan, line 278: what does `tags==null` mean?

app_dart/lib/src/service/scheduler.dart
```diff
@@ -430,25 +434,37 @@ class Scheduler {
...
     final Iterable<LuciBuilder> ciYamlBuilders =
         presubmitLuciTargets.map((Target target) => LuciBuilder.fromTarget(target, commit.slug));
     builders.addAll(ciYamlBuilders);
     // Filter builders based on the PR diff
     final GithubService githubService = await config.createGithubService(commit.slug);
+    log.debug('About to get list of files');
+    githubService.setLogger(log);
     final List<String> files = await githubService.listFiles(commit.slug, prNumber);
-    return await getFilteredBuilders(builders, files);
+    log.debug('About to filter filter builds');
```
keyonghan, line 461: Nit: duplicated `filter`.

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -44,6 +45,15 @@ class LuciBuildService {
     this.log = log;
   }
 
+  /// Shards [rows] into several sublists of size [maxEntityGroups].
```
keyonghan, line 48: Shall we define the `maxEntityGroups` as a const variable?

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -44,6 +45,15 @@ class LuciBuildService {
...
+  Future<List<List<Request>>> shard(List<Request> requests, int max) async {
```
keyonghan, line 49: Curious what caused the index issue you hit? Some tests on this function should help.

app_dart/lib/src/service/luci_build_service.dart
godofredoc: I made this a configuration value with a value of 5. We don't have a way to measure the flutter QPS but this is internal traffic and minimal compared to other clients like fuchsia.

app_dart/lib/src/service/luci_build_service.dart
godofredoc: This was updated to use a batch of 5.

app_dart/lib/src/service/luci_build_service.dart
godofredoc: There are two types of errors:

* An error that has tags information that we can use to update the check as a failed task
* An error with no tags that we can not provide any feedback, tags == null is this case.

app_dart/lib/src/service/scheduler.dart
godofredoc: Removed the additional logging. This was here only to identify where we were getting the error.

app_dart/lib/src/service/luci_build_service.dart
godofredoc: Not in here, but the calling function is already using a configuration value.

app_dart/lib/src/service/luci_build_service.dart
godofredoc: it was a problem with trying to schedule a build that didn't exist. Turns out that modifying presubmit tags requires to also provide the bringup property.

app_dart/lib/src/service/scheduler.dart
keyonghan: Oh, I meant there are two duplicated words `filter` in the log message.

keyonghan (commented):
LGTM to nit.

keyonghan (approved):

CaseyHillers:
Cocoon is now accepting pull requests again. To make your PR compatible,

1. Please update the base to point to main (edit in the GitHub UI)
2. Please rebase on tip of tree. There is now null safety support and you'll have to update your PR to be null safe. There may be some rough edges around the initial migration, so feel free to add fixes as you see fit.

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -151,9 +161,45 @@ class LuciBuildService {
...
+          'github_link': <String>['https://github.com/${slug.fullName}/pull/$prNumber'],
+          if (checkRunId != null) 'github_checkrun': <String>[checkRunId.toString()],
+        },
+        properties: <String, String>{
+          'git_url': 'https://github.com/${slug.owner}/${slug.name}',
+          'git_ref': 'refs/pull/$prNumber/head',
+        },
+        notify: NotificationConfig(
+          pubsubTopic: 'projects/flutter-dashboard/topics/luci-builds',
+          userData: json.encode(userData),
```
CaseyHillers, line 194: nit: base64encode

PR #1355, opened 2021-09-21
Grant explicit write permissions.

godofredoc:
Flutter repositories allow readonly workflows by default. Workflows that
require write permissions need to explicitly request them in the
configuration file.

### DIFF
--- a/.github/workflows/mirror.yml
+++ b/.github/workflows/mirror.yml
@@ -1,5 +1,4 @@
 # Mirror master to main branches in the cocoon repository.
-
 on:
   push:
     branches:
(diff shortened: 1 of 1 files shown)


.github/workflows/mirror.yml
```diff
@@ -1,19 +1,19 @@
...
     runs-on: ubuntu-latest
     name: Mirror master branch to main branch
     steps:
-    - name: Mirror action step
-      id: mirror
-      uses: google/mirror-branch-action@c6b07e441a7ffc5ae15860c1d0a8107a3a151db8
-      with:
-        github-token: ${{ secrets.GITHUB_TOKEN }}
-        source: 'master'
-        dest: 'main'
```
keyonghan, line 19: Is `dest` no longer needed here?

godofredoc: Nice catch! it was removed as part of the conflict resolution. Added it back.

keyonghan (approved):
LGTM

PR #1358, opened 2021-09-22
Update master references to main

CaseyHillers:


app_dart/lib/src/service/buildbucket.dart
```diff
@@ -70,6 +70,7 @@ class BuildBucketClient {
     }
 
     httpRequest.write(json.encode(request.toJson()));
+    print(json.encode(request.toJson());
```
yusufm, line 73: Print a description of whats being output here?

keyonghan (approved):
LGTM

app_dart/lib/src/service/buildbucket.dart
godofredoc: Would print show in the remote logs?

CaseyHillers: Removed, this was from a mock checkout

Prints do show in stackdriver, but they're not properly correlated (it's very useful if you don't want to plumb logging instances to somewhere like here)

godofredoc (approved):

CaseyHillers:
Submitting as the checks are expected to fail, and this is blocking the null safety migration

app_dart/lib/src/service/buildbucket.dart
godofredoc: Good to know, thanks!

PR #1359, opened 2021-09-23
[app_dart] Truncate luci build push messages before sending as check

CaseyHillers:
Fixes https://github.com/flutter/flutter/issues/87507

### DIFF
--- a/app_dart/test/service/github_checks_service_test.dart
+++ b/app_dart/test/service/github_checks_service_test.dart
@@ -133,7 +133,7 @@ void main() {
 
     test('really large summaryMarkdown', () async {
       String summaryMarkdown = '';
-      for (int i = 0; i < 65535; i++) {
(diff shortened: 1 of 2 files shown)


app_dart/test/service/github_checks_service_test.dart
```diff
@@ -130,6 +130,15 @@ void main() {
       const String expectedSummary = '${kGithubSummary}Empty summaryMarkdown';
       expect(githubChecksService.getGithubSummary(null), expectedSummary);
     });
+
+    test('really large summaryMarkdown', () async {
+      String summaryMarkdown = '';
+      for (int i = 0; i < 65535; i++) {
```
keyonghan, line 136: Nit: a smaller number like 20000 is enough.

keyonghan (approved):
LGTM with nit.

PR #1362, opened 2021-09-24
[app_dart] Switch to logging singleton

CaseyHillers:
This removes the need for passing around a logging instance. Instead import 'service/logging.dart' and just use `log` from it.

This fixes issues with the late initialization of logs, and will make it easier to add more logs in the future.

All that was needed was adding to bin/server.dart `useLoggingPackageAdaptor()`

https://github.com/flutter/flutter/issues/90641

## Test

I deployed to a test version of appengine to validate the logs are coming through on stackdriver


godofredoc (approved):
Seems like you'll need a rebase.

CaseyHillers:
@godofredoc it seems like there's some issue with the scheduling (I tried a few commits, one got one build scheduled, this one got 0)

CaseyHillers:
I'm working on fixing some test failures. For some reason, these aren't appearing in my local environment so it's hard to debug. I have a few fixes I've added to CI

1. Ensure flutter upgrade to keep flutter at tip of tree (one debug run showed it was using a version from several weeks ago)
2. Fix some cocoon_service imports (should use the specific file, not the package export)

CaseyHillers:
@godofredoc do you mind taking another look? I migrated the license check to null safety so CI could run on flutter master

godofredoc (approved):

PR #1363, opened 2021-09-27
Fix scheduling problems.

godofredoc:
Scheduling is working correctly but the github checks were not updated
as the request objects were not returning the tags.


CaseyHillers (approved):
LGTM

PR #1366, opened 2021-09-29
Remove autogenerated files and add them to gitignore.

godofredoc:
This is to prevent checking in autogenerated files.


keyonghan (approved):
LGTM

PR #1367, opened 2021-09-29
Fixes build re-run failing with a 400 status code.

godofredoc:
This was caused by pubsub notification topic not being encoded as
base64.

Bug: https://github.com/flutter/flutter/issues/90914


keyonghan (approved):
LGTM

PR #1368, opened 2021-09-29
Do not try to submit if PR is not mergeable.

godofredoc:
This will remove several 500 status code in the cocoon backend and it
will also improve the bot's throughput to submit PRs.

Bug: https://github.com/flutter/flutter/issues/90851


CaseyHillers (approved):
Good catch! It would be nice if this could rely on `package:github` as those models surface the mergeable field

PR #1369, opened 2021-09-29
[app_dart] Log all writes

CaseyHillers:
https://github.com/flutter/flutter/issues/90979 - Fix logs so they show on non-error logs

Follow up from https://github.com/flutter/cocoon/pull/1362


Piinks (approved):
LGTM

PR #1370, opened 2021-09-29
Add screen on check for android devices

keyonghan:
This is to ensure devicelab android devices are running with screen on all the time.

https://github.com/flutter/flutter/issues/89260


CaseyHillers (approved):
LGTM

PR #1372, opened 2021-09-30
Revert "Do not try to submit if PR is not mergeable."

christopherfujino:
Reverts flutter/cocoon#1368

Fixes https://github.com/flutter/flutter/issues/91035

We need the bot to leave the "waiting for tree to go green" label on PRs that are green EXCEPT for the `luci-flutter` check


christopherfujino:
Further discussion is ongoing in https://github.com/flutter/flutter/issues/91035

godofredoc:
Should we land the logging part first? we may not be able to find the issue if we revert now.

CaseyHillers (approved):

godofredoc (approved):
We may also need to revert the logging PR

christopherfujino:
> We may also need to revert the logging PR

I resolved the conflict. We can revert the logging change independently if desired after.

PR #1374, opened 2021-09-30
[app_dart] Only show bringup outside of prod environments

CaseyHillers:
This is a demo of using environment toggles for different functionality in Cocoon

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/app_dart/lib/src/service/scheduler.dart
+++ b/app_dart/lib/src/service/scheduler.dart
@@ -12,6 +12,7 @@ import 'package:retry/retry.dart';
 import 'package:truncate/truncate.dart';
 import 'package:yaml/yaml.dart';
 
+import '../foundation/environment.dart';
 import '../foundation/providers.dart';
 import '../foundation/typedefs.dart';
 import '../foundation/utils.dart';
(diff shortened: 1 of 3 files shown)


app_dart/lib/src/service/scheduler.dart
```diff
@@ -213,8 +214,10 @@ class Scheduler {
   /// Filter [SchedulerConfig] to only the targets expected to run for the branch,
   /// and that do not have any dependencies.
   List<Target> getPostSubmitTargets(Commit commit, SchedulerConfig config) {
-    // Filter targets to only those run in postsubmit.
-    final List<Target> postsubmitTargets = config.targets.where((Target target) => target.postsubmit).toList();
+    // Filter targets to only those run in this postsubmit environment.
+    final List<Target> postsubmitTargets =
+        config.targets.where((Target target) => target.postsubmit && target.bringup != isProduction).toList();
```
yusuf-goog, line 219: Would it be possible to avoid this "is this a specific environment" check in this case?

It would be better if the system has the same components/services in each of our environments. So rather than doing something specific/different in "prod", it would do the same thing with the correct instances in the particular environment. 

Rather than something like:

if env == prod: push to prod_DB, else, skip

do:

push to 'env'_DB

CaseyHillers: Thanks, I updated it so it just filters `SchedulerConfig` when we first get it

app_dart/lib/src/service/scheduler.dart
```diff
@@ -195,7 +196,13 @@ class Scheduler {
       ),
       ttl: const Duration(hours: 1),
     ))!;
-    return SchedulerConfig.fromBuffer(configBytes);
+    final SchedulerConfig cachedConfig = SchedulerConfig.fromBuffer(configBytes);
+    // Filter targets so non-bringup is shown in production, otherwise only show bringup.
+    return SchedulerConfig(
+      enabledBranches: cachedConfig.enabledBranches,
+      targets: cachedConfig.targets.where((Target target) => target.bringup != isProduction),
```
keyonghan, line 203: Clean up nit: https://github.com/flutter/cocoon/blob/main/app_dart/lib/src/request_handlers/push_build_status_to_github.dart#L70

keyonghan (approved):
LGTM with nit.

PR #1380, opened 2021-10-01
Test task queues.

godofredoc:


app_dart/lib/src/request_handlers/reset_try_task.dart
```diff
@@ -25,7 +30,31 @@ class ResetTryTask extends ApiRequestHandler<Body> {
 
   @override
   Future<Body> get() async {
-    final String owner = request!.uri.queryParameters['owner'] ?? 'flutter';
+    // move to cloud tasks
+    final Client client = authenticationProvider.httpClientProvider();
```
keyonghan, line 34: Check [`createAccessClient`](https://github.com/flutter/cocoon/blob/main/app_dart/lib/src/service/access_client_provider.dart#L10) which returns an OAuth 2.0 authenticated client. Bigquery API is authenticated this way.

godofredoc:
This is going to be implemented in a different way.

PR #1385, opened 2021-10-08
Always put PR title as commit headline

nt4f04uNd:
Fixes https://github.com/flutter/flutter/issues/88615

## Pre-launch Checklist

- [x] I signed the [CLA].


app_dart/test/src/service/fake_graphql_client.dart
```diff
@@ -38,20 +38,20 @@ class FakeGraphQLClient implements GraphQLClient {
...
     }
   }
 
   void verifyMutations(List<MutationOptions> expected) {
     expect(mutations.length, expected.length);
     for (int i = 0; i < mutations.length; i++) {
-      /// [BaseOptions.toKey] serializes all of the relevant parts of the query
-      /// or mutation for us, except the fetch policy.
-      expect(mutations[i].toString(), expected[i].toString());
-      expect(mutations[i].fetchPolicy, expected[i].fetchPolicy);
```
nt4f04uNd, line 54: This is not working with the current version of graphql library. The toString now returns `Instance of <MutationOptions>`. The list of `properties` includes `policies`, so we can leave only one `expect`

app_dart/test/src/service/fake_graphql_client.dart
```diff
@@ -38,20 +38,20 @@ class FakeGraphQLClient implements GraphQLClient {
...
+      expect(
+        mutations[i].properties,
+        equals(expected[i].properties),
+      );
```
nt4f04uNd, line 54: This is not working with the current version of graphql library. The toString now returns `Instance of <MutationOptions>`. The list of properties includes `policies`, so we can leave only one `expect`

christopherfujino (commented):
I'd like @CaseyHillers to double-check the graphql code as he has more context, but this looks like the right change to make. Thanks for making this PR!

PR #1391, opened 2021-10-13
Detect benchmarks as tests in engine repo

dnfield:
Engine uses things like `shell_benchmarks.cc` for benchmark related tests.

@xster fyi


keyonghan (commented):
Cocoon has been switched to `main` a while ago.

dnfield:
Hm. I tried switching the base, if that doesn't work for tests/CI I'll create a new PR.

PR #1392, opened 2021-10-13
Detect benchmarks as tests in engine repo

dnfield:
Engine uses things like shell_benchmarks.cc for benchmark related tests.

@xster fyi

Another attempt here, this time starting witht he right base.


dnfield:
My particular change may need a different test, but it seems fair to leave it to reviewers rather than the bot whether a particular change needs a benchmark or unit test

PR #1411, opened 2021-10-21
[app_dart] Fix presubmit builds setting incorrect repo_owner

CaseyHillers:
There was an outage from yesterday's github model removal PR. It incorrectly set the repo_owner in the buildbucket rpc, which cannot be parsed (it passes `pullRequest.base.repo` instead of ` pullRequest.base.repo.owner`)

Fixes https://github.com/flutter/flutter/issues/92270

### DIFF
--- a/app_dart/lib/src/request_handlers/luci_status.dart
+++ b/app_dart/lib/src/request_handlers/luci_status.dart
@@ -81,7 +81,7 @@ class LuciStatusHandler extends RequestHandler<Body> {
       // create the slug from the data in the message and send the check status
       // update.
       slug = RepositorySlug(
-        'flutter', // Hard code as there was an outage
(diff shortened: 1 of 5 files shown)


app_dart/lib/src/request_handlers/luci_status.dart
```diff
@@ -81,7 +81,7 @@ class LuciStatusHandler extends RequestHandler<Body> {
       // create the slug from the data in the message and send the check status
       // update.
       slug = RepositorySlug(
-        userData['repo_owner'] as String,
+        'flutter', // Hard code as there was an outage
```
keyonghan, line 84: shall we restore this?

keyonghan (approved):
LGTM

PR #1415, opened 2021-10-26
[release_dashboard] cli-start connected

Yugue:
*Release initialization step of the desktop App is connected with the conductor CLI now. Upon clicking on a button, the desktop would initialize a `StartContext` class with the parameters captures by the input forms and drop-downs and executes `StartContext.run()`*

Migrated from this [PR](https://github.com/flutter/flutter/pull/92225), because the release dashboard was moved from flutter/flutter to flutter/cocoon.

Main Issue:
- [x] https://github.com/flutter/flutter/issues/93823

Side Issues:
- [x] https://github.com/flutter/flutter/issues/93823
- [x] https://github.com/flutter/flutter/issues/93824
- [x] https://github.com/flutter/flutter/issues/93825

PRs that needed to be landed before for this PR:
- [x] https://github.com/flutter/cocoon/pull/1427
- [x] https://github.com/flutter/cocoon/pull/1438
- [x] https://github.com/flutter/flutter/pull/92200
- [x] https://github.com/flutter/flutter/pull/93580
- [x] https://github.com/flutter/flutter/pull/93514
- [x] https://github.com/flutter/flutter/pull/92594

When the initialization is being executed, a loading icon will appear. After the initialization has been completed, a release state file is created, and it will be displayed at the `conductor status` widget section. Currently, the user has to manually refresh the app for the status to appear. It will be changed such that `conductor status` will be automatically displayed after the release has been completed in this [issue](https://github.com/flutter/flutter/issues/93156).

Here is a video recording of the workflow:
https://user-images.githubusercontent.com/20194490/138746587-746ba866-b9a4-4846-8ba9-b56289a3659d.mov

It asks for the passphrase in a window prompt for the SSH key if it hasn't been provided or saved into an agent:
image

If the user doesn't provide the correct SSH passphrase, this error message will be displayed:

image

## Pre-launch Checklist

- [x] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [x] I signed the [CLA].

[Features we expect every widget to implement]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#features-we-expect-every-widget-to-implement


Yugue:
Currently using mocks from Mockito. I can make another PR to migrate to fakes to reproduce this [PR](https://github.com/flutter/flutter/pull/92453), since this PR is already very large on its own.

CaseyHillers (commented):
This PR will be easier to review when we can rebase it on top of https://github.com/flutter/cocoon/pull/1427

Yugue:
Rewrote the tests by using Fakes instead of mocks by following @christopherfujino's [draft](https://github.com/flutter/flutter/pull/92453).

release_dashboard/lib/logic/helper_functions.dart
```diff
@@ -0,0 +1,9 @@
+// Copyright 2014 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+// Function that converts the cherrypick captured by input forms from string to
+// a string array
+List<String> cherrypickStringtoArray(String? cherrypickString) {
```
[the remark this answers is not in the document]
Yugue: ok, I can factor it to a class. Should `cherrypickStringtoArray` be a method in the class? And I pass the whole cherrypick class as an argument to the widget that is using the helper function `CreateReleaseSubsteps`?

release_dashboard/lib/widgets/create_release_substeps.dart
```diff
@@ -2,20 +2,42 @@
...
+import 'package:file/file.dart';
+import 'package:file/local.dart';
 import 'package:flutter/material.dart';
+import 'package:platform/platform.dart';
+import 'package:process/process.dart';
 
+import '../logic/cherrypicks.dart';
 import 'common/tooltip.dart';
 
+const FileSystem fileSystem = LocalFileSystem();
```
christopherfujino, line 18: Do we still need all of these if we have `ConductorService`?

release_dashboard/lib/widgets/create_release_substeps.dart
```diff
@@ -33,10 +55,36 @@ class CreateReleaseSubsteps extends StatefulWidget {
...
+
+  /// Executes the [run] function to initialize a release.
+  ///
+  /// If there is a [startContext] provided as a parameter, e.g. the fake [startContext] in the test,
+  /// executes the [run] method of that [startContext] instead, otherwise use the [conductor]'s.
+  Future<void> runCreateRelease() {
+    if (widget.startContext != null) {
+      return widget.startContext!.run();
+    }
+    // data captured by the input forms and dropdowns are transformed to conform the formats of StartContext
```
christopherfujino, line 72: why do we have two branches here?

release_dashboard/lib/widgets/create_release_substeps.dart
```diff
@@ -88,23 +162,50 @@ class CreateReleaseSubstepsState extends State<CreateReleaseSubsteps> {
...
           ),
+        Row(
+          mainAxisAlignment: MainAxisAlignment.center,
+          children: <Widget>[
+            ElevatedButton(
+              key: const Key('step1continue'),
+              onPressed: _isLoading
+                  ? null // if the release initialization is loading, disable this button
+                  : () async {
+                      setError(null);
```
christopherfujino, line 187: why do we clear out the error here?

release_dashboard/test/fakes/fake_start_context.dart
```diff
@@ -0,0 +1,150 @@
...
+import 'package:file/file.dart';
+import 'package:file/memory.dart';
+import 'package:platform/platform.dart';
+
+import 'fake_process_manager.dart';
+
+const String kFlutterRoot = '/flutter';
+const String kCheckoutsParentDirectory = '$kFlutterRoot/dev/tools/';
+
+class TestStdio extends Stdio {
```
christopherfujino, line 16: This probably shouldn't be in a file named `fake_start_context.dart`. same for the constants above.

release_dashboard/lib/logic/cherrypicks.dart
```diff
@@ -0,0 +1,9 @@
+// Copyright 2014 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+// Function that converts the cherrypick captured by input forms from string to
+// a string array
+List<String> cherrypickStringtoArray(String? cherrypickString) {
+  return cherrypickString == '' || cherrypickString == null ? <String>[] : cherrypickString.split(',');
```
christopherfujino, line 8: should we validate this input with our regex?

release_dashboard/lib/services/conductor.dart
```diff
@@ -14,15 +14,16 @@ abstract class ConductorService {
   /// Returns null when there is no active release, such as in the case when first initialized.
   pb.ConductorState? get state;
 
+  File get stateFile;
```
christopherfujino, line 17: Why was this added?

release_dashboard/lib/widgets/create_release_substeps.dart
Yugue: Oh my bad, I forgot to remove them.

release_dashboard/lib/widgets/create_release_substeps.dart
Yugue: Yeah, so in `create_release_substeps_test.dart`, a fake `startContext` is passed as a parameter of the `CreateReleaseSubsteps` widget. [Here](https://github.com/Yugue/cocoon/blob/5494852519fe5fc811d05665781f0c9f503d9d5b/release_dashboard/test/widgets/create_release_substeps_test.dart#L164) is an example. This makes sure that when there is a fake `startContext` passed, that one gets executed instead. 
Currently, only the test uses that parameter. Should I rename the parameter `startContext` to `fakeStartContext`?

release_dashboard/lib/widgets/create_release_substeps.dart
Yugue: yea, this makes sure there is no error present during the `loading` state. For example, a user uses the conductor, and gets an error. After, the user corrects the inputs and hits `continue` again. `continue` becomes disabled, and a loading animation appears. Without setting the error object to null, the user would still see the error they got previously during the loading phase.

release_dashboard/test/fakes/fake_start_context.dart
Yugue: Moved `TestStdio`, is it ok to keep `const String kFlutterRoot = '/flutter';
const String kCheckoutsParentDirectory = '$kFlutterRoot/dev/tools/';` here though? since they are being referenced directly in this file.

release_dashboard/lib/logic/cherrypicks.dart
Yugue: Yes, it is being validated in this [PR](https://github.com/flutter/cocoon/pull/1423).

release_dashboard/lib/services/conductor.dart
Yugue: Oh yeah, I should have added a comment about this. Because I am using the [`stateFile` getter](https://github.com/Yugue/cocoon/blob/5494852519fe5fc811d05665781f0c9f503d9d5b/release_dashboard/lib/services/local_conductor.dart#L31) of `local_conductor.dart` outside of the class in [here](https://github.com/Yugue/cocoon/blob/5494852519fe5fc811d05665781f0c9f503d9d5b/release_dashboard/lib/widgets/create_release_substeps.dart#L84). The getter wouldn't be accessible outside of its class without it being defined in the abstract class.

release_dashboard/lib/logic/cherrypicks.dart
christopherfujino: I mean in this function

christopherfujino: maybe at least leave a TODO here and then implement an additional validation here when you land https://github.com/flutter/cocoon/pull/1423 (assuming you land this one first and rebase)

release_dashboard/lib/widgets/create_release_substeps.dart
christopherfujino: What you should do is instead provide a `FakeConductorService`. Overall, you should try to avoid functions/constructors having special test-only override parameters (this makes the code more difficult to reason about). This function, which has essentially two code paths, one only for tests the other for real use, is a code smell. It means that the tests aren't really testing your code. I recommend you remove the `StartContext` from the widget, and always call `conductor.createRelease()`

release_dashboard/lib/widgets/create_release_substeps.dart
christopherfujino: ahh, i see. that makes sense.

christopherfujino: as a nit, maybe in the `setError()` method add a check that the value we're trying to set this to isn't already the value of `_error`. if it is, then no need to call `setState` and no need to re-rebuild the widget tree.

release_dashboard/test/fakes/fake_start_context.dart
christopherfujino: hmm, I guess this was test-related globals that are unused in this test suite. i guess it's fine to leave it here.

release_dashboard/lib/services/conductor.dart
christopherfujino: Ahh, I see. I don't think that should be public or required as a dependency of calling `createRelease`. This is sort of a circular dependency situation, where you need to get a value from the ConductorService in order to give it back to the ConductorService. My recommendations:

1. For the `LocalConductorService` implementation, remove the `stateFile` parameter from `createRelease` (alternatively make it optional, and if it's not provided fallback to calling the getter)
2. Make the `stateFile` getter private, so that only the `LocalConductorService` itself can call it.

release_dashboard/lib/widgets/create_release_substeps.dart
Yugue: Ok yea, that sounds a really good idea. Changed!

release_dashboard/lib/logic/cherrypicks.dart
Yugue: There is one [here](https://github.com/Yugue/cocoon/blob/80846d7d8c51aea9eb6207acda0fd2efebccef26/release_dashboard/lib/widgets/create_release_substeps.dart#L99). The validations in #1423 will make sure the strings are valid before hitting this `cherrypickStringtoArray` function

release_dashboard/lib/services/conductor.dart
Yugue: Oh, thanks for pointing that out. That's a really good idea. Changed!!

release_dashboard/lib/widgets/create_release_substeps.dart
```diff
@@ -13,9 +16,13 @@ class CreateReleaseSubsteps extends StatefulWidget {
   const CreateReleaseSubsteps({
     Key? key,
     required this.nextStep,
+    this.startContext,
+    required this.conductor,
   }) : super(key: key);
 
   final VoidCallback nextStep;
+  final StartContext? startContext;
```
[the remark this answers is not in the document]
Yugue: Oh right, I forgot about it. Removed!

release_dashboard/test/fakes/services/fake_conductor.dart
```diff
@@ -0,0 +1,46 @@
...
+    required String frameworkMirror,
+    required String incrementLetter,
+    required String releaseChannel,
+  }) async {
+    if (fakeStartContextProvided != null) {
+      return fakeStartContextProvided!.run();
+    } else {
+      /// If there is no [fakeStartContextProvided], initialize a fakeStartContext
+      /// with no exception thrown when [run] is called.
+      final FakeStartContext fakeStartContext = FakeStartContext();
```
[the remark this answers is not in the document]
Yugue: yes, the argument `fakeStartContextProvided` is optional. If no `fakeStartContextProvided` is provided, then the class initializes a `fakeStartContext` that does not throw any error by default. I think that most of the time when we use `FakeConductor`, we do not want `createRelease` method to throw an error. With this logic, we don't have to pass a `fakeStartContext` every time when we want to create a `createRelease` that doesn't throw any error.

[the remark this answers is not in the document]
Yugue: I added a dart doc to the class to explain this, it might be a bit confusing.

release_dashboard/test/fakes/services/fake_conductor.dart
```diff
@@ -0,0 +1,46 @@
...
+      return fakeStartContext.run();
+    }
+  }
+
+  @override
+  ConductorState? get state {
```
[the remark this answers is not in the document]
Yugue: Ok, I changed it to
```
@override
ConductorState? state = ConductorState(
  conductorVersion: 'abcdef',
  releaseChannel: 'dev',
  releaseVersion: '2.7.0-4.0.pre',
);
```
How do we change it in tests in the future?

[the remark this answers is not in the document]
christopherfujino:
```
// This callback function can be attached to the handler of a button or some other trigger that will happen in your test
void Function() callback = () => fakeConductor.state = updatedState;

tester.tap(someButtonThatCallsCallback);
// assert that the state was really updated
expect(fakeConductor.state.interestingField == updatedState.interestingField, true);
```

release_dashboard/test/fakes/services/fake_conductor.dart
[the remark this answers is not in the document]
christopherfujino: My point is, if there is a branch where we actually create a `FakeStartContext` but ignore the `createRelease()` arguments, this is surprising behavior. If we already have the arguments, we should use them. Note: these parameters are already required.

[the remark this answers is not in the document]
christopherfujino: Also, please don't mark conversations resolved just because you disagree. It's ok to have a different opinion, but adding a dartdoc did not resolve my concern.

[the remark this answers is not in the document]
Yugue: Sorry, I resolved this by accident.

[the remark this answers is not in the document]
Yugue: Oh I see what you mean, my bad I misunderstood you at the beginning. Added the parameters.

release_dashboard/test/fakes/fake_start_context.dart
```diff
@@ -0,0 +1,127 @@
...
+import 'fake_stdio.dart';
+
+const String kFlutterRoot = '/flutter';
+const String kCheckoutsParentDirectory = '$kFlutterRoot/dev/tools/';
+
+/// Initializes a fake [StartContext] in a fake local environment.
+class FakeStartContext extends StartContext {
+  factory FakeStartContext({
+    String candidateBranch = 'flutter-1.2-candidate.3',
+    Checkouts? checkouts,
```
christopherfujino, line 21: this doesn't need to be a parameter, right? we're just overriding it on line 46.

release_dashboard/test/fakes/services/fake_conductor.dart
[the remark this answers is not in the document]
Yugue: Adding this right now...

release_dashboard/test/fakes/fake_start_context.dart
Yugue: Added null-aware operators `??=`

release_dashboard/test/fakes/services/fake_conductor.dart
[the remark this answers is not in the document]
Yugue: Regarding this issue, could we do something like [this](https://github.com/Yugue/cocoon/blob/cdaeeca77c8907701e3970a54b5427d1cf666277/release_dashboard/test/fakes/services/fake_conductor.dart#L54) for `fake_conductor.dart`? And have `testState` as a parameter in `FakeConductor` [here](https://github.com/Yugue/cocoon/blob/cdaeeca77c8907701e3970a54b5427d1cf666277/release_dashboard/test/fakes/services/fake_conductor.dart#L17)? 

In that way, it might be easier to pass a testState to `FakeConductor` by doing [this](https://github.com/Yugue/cocoon/blob/cdaeeca77c8907701e3970a54b5427d1cf666277/release_dashboard/test/widgets/conductor_status_test.dart#L111). Otherwise, three lines is required in order to pass a testState by doing this:
```
    FakeConductor fakeConductor = FakeConductor();
    fakeConductor.state = state;
    await tester.pumpWidget(MyApp(fakeConductor));
    ```

**Comment by christopherfujino** [comment]
@Yugue nice job listing the dependency PRs. As far as I know, this should be unblocked. Can you roll the `conductor_core` revision in `pubspec.yaml` to the latest and see if the tests pass?

File: release_dashboard/lib/services/release_dashboard_start_context.dart [hunk 12]
```diff
@@ -0,0 +1,49 @@
...
+          frameworkMirror: frameworkMirror,
+          frameworkUpstream: frameworkUpstream,
+          incrementLetter: incrementLetter,
+          processManager: processManager,
+          releaseChannel: releaseChannel,
+          stateFile: stateFile,
+          force: force,
+        );
+
+// TODO(Yugue)]: Add prompt UI to confirm tag creation at startContext, https://github.com/flutter/flutter/issues/94072.
```
christopherfujino, line 44:
```suggestion
  // TODO(Yugue)]: Add prompt UI to confirm tag creation at startContext, https://github.com/flutter/flutter/issues/94072.
```

release_dashboard/lib/services/dev_local_conductor.dart
```diff
@@ -31,10 +31,20 @@ class DevLocalConductorService extends LocalConductorService {
...
     final EngineRepository localEngineUpstream =
-        EngineRepository(checkouts, localUpstream: true, additionalRequiredLocalBranches: [candidateBranch]);
+        EngineRepository(
+          checkouts,
+          localUpstream: true,
+          name: 'engine-upstream',
+          additionalRequiredLocalBranches: [candidateBranch],
+        );
 
     // TODO: turn force to false
```
CaseyHillers, line 49: This should follow the flutter todo style (TODO(yugue): explaination, issue url)

Yugue: Done, that was a temporary placeholder.

christopherfujino: Yeah, I told Alex that I add these invalid TODO comments to WIP PRs so that I know I have to fix them before merging.

christopherfujino (approved):
LGTM

Yugue:
> LGTM

This is a big moment in my life 😭

PR #1428, opened 2021-10-29
[release_dashboard] add fake_process_manager

christopherfujino:
The upstream FakeProcessManager is defined in test files, and thus not importable. Thus, this was copy pasted from https://github.com/flutter/flutter/blob/master/packages/flutter_tools/test/src/fake_process_manager.dart with some changes to satisfy the analyzer over null checks on non-nullable fields. I added an explicit test exercising it.

### DIFF
--- a/release_dashboard/test/fake_process_manager.dart
+++ b/release_dashboard/test/fake_process_manager.dart
@@ -132,14 +132,13 @@ class _FakeProcess implements io.Process {
     this._stdout,
     this._completer,
     bool outputFollowsExit,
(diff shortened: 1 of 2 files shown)


Yugue (approved):

Yugue:
In the case the `fakeprocessmanager` gets updated in `flutter/flutter`, we have to these files manually?

christopherfujino:
> In the case the `fakeprocessmanager` gets updated in `flutter/flutter`, we have to these files manually?

no, this code can live independently.

PR #1437, opened 2021-11-03
Use staging data when auto deflake tests

keyonghan:
This solves: https://github.com/flutter/flutter/issues/92999

changes:
1) switch query to staging runs instead of prod (with support of cl/407390665)
2) fix corner case when no rows are returned.


CaseyHillers (approved):
flutter-lgtm

PR #1440, opened 2021-11-04
Initial draft version of results_pipleine microservice.

yusuf-goog:
Showcases a possible layout for microservice in cocoon:

Sets up a new microservice (results-pipeline) that can be deployed to appengine via:

   gcloud app deploy --no-promote app.yaml

Creating an endpoint that can receive push messages from a pubsub topic at:

  /api/process_build_update

Currently logs out the received messaged.
Returns a 401 Unauthorized if message is not sent via a specified service account.
Returns a 404 for any other page request.

Contains common libs, useful to most microservices (auth, logging, response), but
not visible outside the microservice directory. Will need a mechanism to share
dart libs.

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/app_dart/micro_services/results_pipeline/analysis_options.yaml
+++ b/app_dart/micro_services/results_pipeline/analysis_options.yaml
@@ -1,0 +1,1 @@
+include: package:lints/recommended.yaml
--- a/app_dart/micro_services/results_pipeline/bin/server.dart
+++ b/app_dart/micro_services/results_pipeline/bin/server.dart
@@ -1,0 +1,62 @@
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:appengine/appengine.dart';
+import 'package:results_pipeline/common/auth_utils.dart';
+import 'package:results_pipeline/common/logging.dart';
+import 'package:results_pipeline/common/response.dart';
+
+/// Default handler, returns a 404
+void _defaultHandler(HttpRequest request) {
+  respond(request, status: HttpStatus.notFound);
+}
+
+/// Check authorization of incoming request.
+Future<void> _checkAuthorization(HttpRequest request) async {
+  if (!await authenticateRequest(request.headers)) {
+    await respond(request, status: HttpStatus.unauthorized);
+    throw HttpException('Unauthorized request: ${request.headers}');
+  }
+}
+
+/// Processes LUCI build status updates messages.
+///
+/// Endpoint for push messages from the pubsub topic that
+/// LUCI send build update status messages to.
+/// Tasks:
+///  1) Verify authorization on incoming message.
+///  2) Decodes build update message.
+///  3) Saves update to datatstore.
+///  4) Triggers retry for failing build.
+///
+/// POST: /api/process_build_request
+///
+/// Response: Status 200 OK
+///
+/// Response: Status 401 Unauthorized
+///
+Future<void> _processUpdate(HttpRequest request) async {
+  final String requestString = await utf8.decodeStream(request);
+  log.info(requestString);
+  await respond(request);
+}
+
+/// Federate incoming requests to the appropriate handlers.
+Future<void> _requestHandler(HttpRequest request) async {
+  if (request.uri.path == '/api/process_build_update') {
+    await _checkAuthorization(request);
+    await _processUpdate(request);
+  } else {
+    _defaultHandler(request);
+  }
+}
+
+Future<void> main() async {
+  useLoggingPackageAdaptor();
+
+  await runAppEngine(_requestHandler,
+      onAcceptingConnections: (InternetAddress address, int port) {
+    final String host = address.isLoopback ? 'localhost' : address.host;
+    print('Serving requests at http://$host:$port/');
+  });
+}
--- a/app_dart/micro_services/results_pipeline/lib/common/auth_utils.dart
+++ b/app_dart/micro_services/results_pipeline/lib/common/auth_utils.dart
@@ -1,0 +1,39 @@
+// Copyright 2021 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:io';
+
+import 'package:googleapis/oauth2/v2.dart';
+import 'package:http/http.dart' as http;
+
+/// Authenticate and authorize incoming requests.
+///
+/// Ensure that the incoming requests are from specified
+/// set of service accounts.
+Future<bool> authenticateRequest(HttpHeaders headers) async {
+  final http.Client client = http.Client();
+  final Oauth2Api oauth2api = Oauth2Api(client);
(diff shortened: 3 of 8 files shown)


app_dart/micro_services/results_pipeline/lib/common/logging.dart
```diff
@@ -0,0 +1,7 @@
+// Copyright 2021 The Flutter Authors. All rights reserved.
```
godofredoc, line 1: Seems like this file should go to common libraries.

app_dart/micro_services/results_pipeline/lib/common/auth_utils.dart
```diff
@@ -0,0 +1,39 @@
+// Copyright 2021 The Flutter Authors. All rights reserved.
```
godofredoc, line 1: Seems like this file should go to common libraries.

app_dart/micro_services/results_pipeline/bin/server.dart
```diff
@@ -0,0 +1,63 @@
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:appengine/appengine.dart';
+import 'package:results_pipeline/common/auth_utils.dart';
+import 'package:results_pipeline/common/logging.dart';
+import 'package:results_pipeline/common/response.dart';
+
+/// Default handler, returns a 404
```
godofredoc, line 9: You may to keep server as the integration point for one or more handlers. I'd recommend to move most of this logic to a separate file.

app_dart/micro_services/results_pipeline/lib/common/response.dart
```diff
@@ -0,0 +1,88 @@
+// Copyright 2021 The Flutter Authors. All rights reserved.
```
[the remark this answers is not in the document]
yusuf-goog: Yes, we need to figure out the mechanism for the common libraries across microservices.

I would be much happier if we could have a versioned dart package that can be imported by each microservice rather than just copying code over. As our number of services increase, and each could have very varied deployment and test cycles, breakages due to incompatible changes could start becoming a real problem.

app_dart/micro_services/results_pipeline/lib/common/logging.dart
yusuf-goog: ditto as above

app_dart/micro_services/results_pipeline/lib/common/auth_utils.dart
yusuf-goog: ditto comment as above

app_dart/micro_services/results_pipeline/bin/server.dart
yusuf-goog: I had originally written it up as a separate file. I was trying this flat layout to see if most of our microservices will be super small and would they really need the added break out. 

I'm fine with just standardizing on having the handlers broken out even if they are super small. Any thoughts?

app_dart/micro_services/results_pipeline/bin/server.dart
```diff
@@ -0,0 +1,63 @@
...
+///  4) Triggers retry for failing build.
+///
+/// POST: /api/process_build_request
+///
+/// Response: Status 200 OK
+///
+/// Response: Status 401 Unauthorized
+///
+Future<void> _processUpdate(HttpRequest request) async {
+  await _checkAuthorization(request);
```
yusuf-goog, line 38: @godofredoc also, I was thinking of moving the auth checks out to _requestHandler(), rather than make it the responsibility of each handler. It might help avoid times where someone forgets to add necessary auth to a handler, and cosolidates the checks at the top level. Thougts?

app_dart/micro_services/results_pipeline/analysis_options.yaml
```diff
@@ -0,0 +1,196 @@
+# Specify analysis options.
```
CaseyHillers, line 1: nit: we should move to using pub.dev/packages/lints

app_dart/micro_services/results_pipeline/bin/server.dart
CaseyHillers: I think there should be a generic server they all extend from (which would handle this generic logic), and then the microservices can extend it with the RPCs they support

app_dart/micro_services/results_pipeline/bin/server.dart
CaseyHillers: The reason for auth per request handler is it enables ACLs. Otherwise, we have one generic auth for the entire backend

app_dart/micro_services/results_pipeline/lib/common/response.dart
```diff
@@ -0,0 +1,88 @@
...
+
+import 'dart:convert';
+import 'dart:io';
+import 'dart:typed_data';
+
+import 'package:meta/meta.dart';
+
+/// Class that represents an HTTP response body before it has been serialized.
+@immutable
+abstract class Body {
```
CaseyHillers, line 13: Going forward, I'm against reusing `Body`. I'd prefer we use something like `package:http` instead of writing our own classes on top of the `dart:io` calls. This class has slowed my ability to develop features due to it's low level functionality.

app_dart/micro_services/results_pipeline/lib/common/auth_utils.dart
CaseyHillers: One issue I faced while making the scheduler microservice is how appengine handles dart packages. You can't rely on the local references or git, and you'll need to land your package change, publish it, then land your microservice change. I'd like to see if we can find a way around it and get the local paths working here.

app_dart/micro_services/results_pipeline/lib/common/response.dart
yusuf-goog: Very happy to hear this! :) will remove.

app_dart/micro_services/results_pipeline/bin/server.dart
yusuf-goog: I didn't mean have one for every handler, but have the calls made just above the handler call each time. E.g.:

Future<void> _requestHandler(HttpRequest request) async {
  if (request.uri.path == '/api/process_build_update') {
    checkAUTH here;
    await _processUpdate(request);
  } else {
   checkAUTH here;
    _defaultHandler(request);
  }
}

This separates responsibility from the handler to do one additional function, simplifying it.

app_dart/micro_services/results_pipeline/analysis_options.yaml
yusuf-goog: done

godofredoc:
Should we close this PR?

Hixie:
Since nobody objected to closing the PR after @godofredoc asked, I'm going to close it. Please feel free to reopen if it's still useful!

PR #1444, opened 2021-11-08
Don't request tests for some commment-only changes

stuartmorgan:
Adds the ability to recognize PRs that only change single-line comments,
and exempts them from test requests.

- Currently only handles a specific set of languages that use `//`.
- Makes no attempt to handle multi-line (`/* */`) blocks, so tests will
  still sometimes be requested for comment-only changes (e.g., to
  Javadoc comments).
- Currently enabled only for flutter/plugins and flutter/packages in
  order to evaluate it before rolling it out to other repositories.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/app_dart/lib/src/request_handlers/github_webhook.dart
+++ b/app_dart/lib/src/request_handlers/github_webhook.dart
@@ -398,7 +398,7 @@ class GithubWebhook extends RequestHandler<Body> {
           !filename.contains('.ci/') &&
           !filename.contains('.github/') &&
           !filename.endsWith('.md')) {
-        needsTests = true;
+        needsTests = !_allChangesAreCodeComments(file);
       }
 
       // See https://github.com/flutter/flutter/wiki/Plugin-Tests for discussion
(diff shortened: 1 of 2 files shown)


app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -515,4 +515,50 @@ class GithubWebhook extends RequestHandler<Body> {
...
+      'dart',
+      'java',
+      'kt',
+      'm',
+      'mm',
+      'swift',
+    };
+    final String filename = file.filename!;
+    final String? extension = filename.contains('.') ?
+      filename.split('.').last.toLowerCase() : null;
```
Hixie, line 546: do we need the toLowerCase?

app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -515,4 +515,50 @@ class GithubWebhook extends RequestHandler<Body> {
...
+    if (extension == null || !codeExtensions.contains(extension)) {
+      return false;
+    }
+
+    // Only handle single-line comments; identifying multi-line comments
+    // would require the full file and non-trivial parsing.
```
Hixie, line 552: This also doesn't notice changes that only affect the commend side of a line with code and a trailing comment (which is perfectly fine, just saying we might want to mention that here in case someone is confused about it).

Hixie:
> * Currently enabled only for flutter/plugins and flutter/packages in
>   order to evaluate it before rolling it out to other repositories.

Is this true? I couldn't see where in the PR that was implemented.

Hixie (approved):
This is great!

LGTM

Hixie:
fyi @Piinks since you worked in this area in the past and may have opinions

app_dart/lib/src/request_handlers/github_webhook.dart
stuartmorgan: Good point, added that to the comment.

stuartmorgan:
> > * Currently enabled only for flutter/plugins and flutter/packages in
> >   order to evaluate it before rolling it out to other repositories.
> 
> Is this true? I couldn't see where in the PR that was implemented.

Line 401 is in a function that is specific to those repos. Since the existing pattern was that flutter/engine and flutter/flutter had completely separate functions, I did the same for packages+plugins when I enabled it. I think there's definitely room to refactor to combine things though.

Do you want me to enable this for flutter/flutter? I could add that here, I just figured it might be good to try it out in a smaller environment first.

Hixie:
Totally fine with starting small to test it. This will be hugely valuable, I think. Especially on flutter/flutter, there's a lot of contributions that are purely docs.

PR #1447, opened 2021-11-10
Kill adb server before running any device healthcheck

keyonghan:
This fixes: https://github.com/flutter/flutter/issues/93075

### DIFF
--- a/device_doctor/lib/src/android_device.dart
+++ b/device_doctor/lib/src/android_device.dart
@@ -93,9 +93,11 @@ class AndroidDeviceDiscovery implements DeviceDiscovery {
   @override
   Future<Map<String, List<HealthCheckResult>>> checkDevices({ProcessManager processManager}) async {
     processManager ??= LocalProcessManager();
+    final List<HealthCheckResult> defaultChecks = <HealthCheckResult>[];
+    defaultChecks.add(await killAdbServerCheck(processManager: processManager));
     final Map<String, List<HealthCheckResult>> results = <String, List<HealthCheckResult>>{};
     for (AndroidDevice device in await discoverDevices(processManager: processManager)) {
-      final List<HealthCheckResult> checks = <HealthCheckResult>[];
+      final List<HealthCheckResult> checks = defaultChecks;
       checks.add(HealthCheckResult.success(kDeviceAccessCheckKey));
       checks.add(await adbPowerServiceCheck(processManager: processManager));
       checks.add(await developerModeCheck(processManager: processManager));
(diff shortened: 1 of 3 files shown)


device_doctor/test/src/android_device_test.dart
```diff
@@ -326,4 +326,36 @@ void main() {
...
+      process = FakeProcess(0);
+      when(processManager.start(<dynamic>['adb', 'kill-server'], workingDirectory: anyNamed('workingDirectory')))
+          .thenAnswer((_) => Future.value(process));
+
+      HealthCheckResult healthCheckResult = await deviceDiscovery.killAdbServerCheck(processManager: processManager);
+      expect(healthCheckResult.succeeded, true);
+      expect(healthCheckResult.name, kKillAdbServerCheckKey);
+    });
+
+    test('returns failure when adb returns none 0 code', () async {
```
yusuf-goog, line 350: none -> non

yusuf-goog (approved):
LGTM, with minor comment

device_doctor/test/src/android_device_test.dart
keyonghan: Fixed.

device_doctor/lib/src/android_device.dart
```diff
@@ -93,9 +93,11 @@ class AndroidDeviceDiscovery implements DeviceDiscovery {
   @override
   Future<Map<String, List<HealthCheckResult>>> checkDevices({ProcessManager processManager}) async {
     processManager ??= LocalProcessManager();
+    final List<HealthCheckResult> defaultChecks = <HealthCheckResult>[];
```
godofredoc, line 96: do we need a new variable? seems like adding check.add( .. killAdb? will be simpler.

keyonghan: We want to run killAdb before running any `adb` command, including the ones in `discoverDevices`.

godofredoc: Assuming all the checks are running sequentially we only need to run kill-server once at the beginning, no need to run it on every step as the adb daemon will be already running on the same user.

godofredoc: That seems to contradict your first response. Anyway the comment was only cosmetic to remove a few unnecessary lines.

keyonghan: Oh, I see the confusing part.
But here we need the `killAdb` before the `device for loop`, because the `discoverDevices` also calls `adb`.

PR #1454, opened 2021-11-10
[app_dart] Allow waiting for green to accept empty statuses

CaseyHillers:
Google's CLA status was moved to a GitHub check. Instead, we should be checking iff the checks are empty.

Most repos still have a status (like WIP), but Cocoon only had the CLA bot. Now it's blocked on having PRs submitted by the bot.

Fixes https://github.com/flutter/flutter/issues/93418

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
+++ b/app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
@@ -247,7 +247,7 @@ class CheckForWaitingPullRequests extends ApiRequestHandler<Body> {
           title: title,
           sha: sha,
           labelId: labelId!,
-          emptyValidations: checkRuns.isEmpty || statuses.isEmpty,
+          emptyValidations: checkRuns.isEmpty && statuses.isEmpty,
           isConflicting: isConflicting,
           unknownMergeableState: unknownMergeableState,
           labels: labels));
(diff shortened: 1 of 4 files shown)


app_dart/test/request_handlers/check_for_waiting_pull_requests_test.dart
```diff
@@ -292,20 +292,7 @@ void main() {
       flutterRepoPRs.add(prInProgress);
       await tester.get(handler);
       _verifyQueries();
-      githubGraphQLClient.verifyMutations(<MutationOptions>[
```
CaseyHillers, line 295: These tests are actually expecting the wrong thing. They didn't set a status, and are reproducing the bug Cocoon has now. Instead, we expect no more graphql mutations as it shouldn't leave a comment.

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
```diff
@@ -247,7 +247,7 @@ class CheckForWaitingPullRequests extends ApiRequestHandler<Body> {
           title: title,
           sha: sha,
           labelId: labelId!,
-          emptyValidations: checkRuns.isEmpty || statuses.isEmpty,
+          emptyChecks: checkRuns.isEmpty,
```
godofredoc, line 250: This needs to be handled in a per repo basis. We don't want to allow empty statuses in all the repos. E.g. luci-status has to be present on engine and flutter

CaseyHillers: Updated. I added this check explicitly in the _checkStatuses to ensure repos with tree status ensure the status is set

godofredoc (approved):

CaseyHillers:
@godofredoc can you submit this?

PR #1455, opened 2021-11-11
Use a fixed number of commits when calculating flaky percentage

keyonghan:
This fixes part of https://github.com/flutter/flutter/issues/92502.

Instead of querying build stats in past two weeks, this PR updates to query recent 100 (default value) builds for each builder.

### DIFF
--- a/app_dart/lib/src/service/bigquery.dart
+++ b/app_dart/lib/src/service/bigquery.dart
@@ -26,6 +26,8 @@ import 'access_client_provider.dart';
 /// success_count	    INTEGER
 /// failure_count	    INTEGER
 /// is_flaky	        INTEGER
+///
+/// This returns latest [LIMIT] number of build stats for each builder.
 const String getBuilderStatisticQuery = r'''
 select builder_name,
        sum(is_flaky) as flaky_number,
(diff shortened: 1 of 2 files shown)


app_dart/lib/src/service/bigquery.dart
```diff
@@ -35,14 +37,15 @@ select builder_name,
...
-where date>=date_sub(current_date(), interval 14 day) and
-      date<=current_date() and
+from (select *, row_number() over (partition by builder_name order by time desc) as rank from `flutter-dashboard.datasite.luci_prod_build_status`)
+where date>=date_sub(current_date(), interval 30 day) and
       builder_name not like '%Drone' and
       repo='flutter' and
       branch='master' and
       pool = 'luci.flutter.prod' and
-      builder_name not like '%Beta%'
+      builder_name not like '%Beta%' and
```
CaseyHillers, line 46: What about stable?

keyonghan: Stable and dev do not run on `master` branch (we do not have any data under both `master` and `stable/dev`).

CaseyHillers: Out of curiosity, why does beta run?

keyonghan: @godofredoc may have more context.
BTW: here is the beta executions under `master`: go/flutter_beta_runs.

godofredoc: Flakiness should be calculated only on master/main. Flakiness on release branches is not actionable as only approved cherry picks are allowed to land on those branches.

CaseyHillers: This sounds like a bug in how we're saving the data (branch isn't being set correctly)

CaseyHillers (approved):
Everything else LGTM

app_dart/lib/src/service/bigquery.dart
keyonghan: Created: https://github.com/flutter/flutter/issues/93557

PR #1464, opened 2021-11-15
Use correct bucket stats when file/update flaky bugs

keyonghan:
This solves: https://github.com/flutter/flutter/issues/93678

This PR does
1) Update flaky bug body to reflect recent changes (https://github.com/flutter/flutter/issues/92502)
2) Update flaky bugs using staging stats for builders marked as `bringup: true`, which are expected to run in `staging` pool.
3) Update flaky bugs using prod stats for builders (such as shard builders) without `bringup: true`.

### DIFF
--- a/app_dart/lib/src/request_handlers/flaky_handler_utils.dart
+++ b/app_dart/lib/src/request_handlers/flaky_handler_utils.dart
@@ -140,9 +140,6 @@ class IssueUpdateBuilder {
   }
 
   String get issueUpdateComment {
-    if (statistic.flakyBuilds!.length + statistic.succeededBuilds!.length < kFlayRatioBuildNumberList) {
-      return 'Current flaky ratio is not available (< $kFlayRatioBuildNumberList commits).\n';
-    }
     String result = 'Current flaky ratio for the past (up to) 100 commits is ${_formatRate(statistic.flakyRate)}%.\n';
     if (statistic.flakyRate > 0.0) {
       result = result +
--- a/app_dart/lib/src/request_handlers/flaky_handler_utils.dart
+++ b/app_dart/lib/src/request_handlers/flaky_handler_utils.dart
@@ -120,7 +120,7 @@ class IssueUpdateBuilder {
   final BuilderStatistic statistic;
   final double threshold;
   final Issue existingIssue;
-  final String bucket;
+  final Bucket bucket;
 
   bool get isBelow => statistic.flakyRate < threshold;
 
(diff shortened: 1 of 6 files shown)


app_dart/lib/src/request_handlers/flaky_handler_utils.dart
```diff
@@ -113,11 +114,13 @@ class IssueUpdateBuilder {
     required this.statistic,
     required this.threshold,
     required this.existingIssue,
+    required this.bucket,
   });
 
   final BuilderStatistic statistic;
   final double threshold;
   final Issue existingIssue;
+  final String bucket;
```
chunhtai, line 123: Would be cleaner to use enum instead

chunhtai: or just a boolean if it is not expected to increase

app_dart/lib/src/request_handlers/update_existing_flaky_issues.dart
```diff
@@ -79,4 +90,48 @@ class UpdateExistingFlakyIssue extends ApiRequestHandler<Body> {
...
+    pb.SchedulerConfig schedulerConfig, {
+    required List<BuilderStatistic> prodBuilderStatisticList,
+    required List<BuilderStatistic> stagingBuilderStatisticList,
+    required Map<String?, Issue> nameToExistingIssue,
+  }) async {
+    final Map<String, bool> builderFlakyMap = <String, bool>{};
+    for (pb.Target target in schedulerConfig.targets) {
+      builderFlakyMap[target.name] = target.bringup;
+    }
+    // For prod builder stats, updates an existing flaky bug when the builder is not marked as flaky.
```
chunhtai, line 107: do we need to worry about cases where a test is recently move to a staging, there are recent builds both in prod and stage

app_dart/lib/src/request_handlers/flaky_handler_utils.dart
keyonghan: We may support `try` in the future. Updated with enum.

app_dart/lib/src/request_handlers/update_existing_flaky_issues.dart
chunhtai: If that is the case, would that be noisy? for example, a test recently moved from `prod` to `stage` and there is only 1 `stage` while other existing builds are in prod. wouldn't we want to grab at least 100 recent builds for the stats?

keyonghan: That's a good point. I added a threshold (10 commits) to control if we want to update the flaky ratio. Does that sound good to you?

app_dart/lib/src/request_handlers/flaky_handler_utils.dart
```diff
@@ -139,6 +140,9 @@ class IssueUpdateBuilder {
   }
 
   String get issueUpdateComment {
+    if (statistic.flakyBuilds!.length + statistic.succeededBuilds!.length < kFlayRatioBuildNumberList) {
```
chunhtai, line 143: Maybe we just skip if this is below threshold? on one hand i know printing this comment let us know script is still working, but this comment is not really useful for triaging. I am neutral on this.

app_dart/lib/src/request_handlers/update_existing_flaky_issues.dart
chunhtai: sounds good

chunhtai (approved):
LGTM

app_dart/lib/src/request_handlers/flaky_handler_utils.dart
keyonghan: It makes sense to me to skip the comment. Updated.

PR #1467, opened 2021-11-16
Skip adb retry to avoid adb kill server

keyonghan:
Tests sometimes fail due to adb logcat dying: https://github.com/flutter/flutter/issues/93661

This PR removes `adb kill server` when getting device dimensions from device doctor. This logic is being called from chromium config side, and we do not want adb interruption during test execution.

### DIFF
--- a/device_doctor/lib/src/android_device.dart
+++ b/device_doctor/lib/src/android_device.dart
@@ -8,11 +8,9 @@ import 'dart:io';
 
 import 'package:meta/meta.dart';
 import 'package:process/process.dart';
-import 'package:retry/retry.dart';
 
 import 'device.dart';
 import 'health.dart';
-import 'host_utils.dart';
 import 'mac.dart';
 import 'utils.dart';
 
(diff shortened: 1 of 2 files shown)


device_doctor/lib/src/android_device.dart
```diff
@@ -45,26 +43,8 @@ class AndroidDeviceDiscovery implements DeviceDiscovery {
...
-    );
-  }
-
-  void _killAdbServer({ProcessManager processManager}) async {
-    if (Platform.isWindows) {
-      await killAllRunningProcessesOnWindows('adb', processManager: processManager);
-    } else {
-      await eval('adb', <String>['kill-server'], canFail: false, processManager: processManager);
-    }
+    String result = await _deviceListOutput(deviceOutputTimeout, processManager: processManager);
```
godofredoc, line 46: What are we trying to accomplish with this change? removing the retries will probably make things more flaky

dnfield: Can we keep the retry but remove the kill server?

Can we somehow detect if a test is running on the device during this?

Can we move this code from infra to a setup/teardown step on the test(s)?

godofredoc: I believe this runs only when the bot is idle, are we running it somewhere else?

keyonghan: Intended to remove the `adb kill-server` to avoid interrupting test execution.

However, after revisiting the chromium log: `get_dimensions` or `get_state` only runs when no test is running. So this should not be the culprit of issue: https://github.com/flutter/flutter/issues/93661.

Closing.

keyonghan: > I believe this runs only when the bot is idle, are we running it somewhere else?

(didn't see this comment before posting)

Currently we have only two parts running `adb kill-server`: one in `get_dimension`, the other in `get_state`. Doesn't sound caused by device doctor. Wondering if anything else may cause `adb logcat` dying.

godofredoc: If this is happening only on adb logcat then my guess is that the device is crashing. Is there a particular test where this is happening?

PR #1478, opened 2021-11-22
[scheduler] Retry ci.yaml checks

CaseyHillers:
Fixes https://github.com/flutter/flutter/issues/93923
Fixes https://github.com/flutter/flutter/issues/87965

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/app_dart/lib/src/service/scheduler.dart
+++ b/app_dart/lib/src/service/scheduler.dart
@@ -244,13 +244,8 @@ class Scheduler {
   Future<void> triggerPresubmitTargets({
     required github.PullRequest pullRequest,
     String reason = 'Newer commit available',
+    RetryOptions retryOptions = const RetryOptions(maxAttempts: 3),
   }) async {
-    // Always cancel running builds so we don't ever schedule duplicates.
-    log.fine('about to cancel presubmit targets');
-    await cancelPreSubmitTargets(
-      pullRequest: pullRequest,
-      reason: reason,
-    );
     final github.CheckRun ciValidationCheckRun = await githubChecksService.githubChecksUtil.createCheckRun(
       config,
       pullRequest.base!.repo!.slug(),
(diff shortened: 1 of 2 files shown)


app_dart/lib/src/service/scheduler.dart
```diff
@@ -244,13 +244,8 @@ class Scheduler {
   Future<void> triggerPresubmitTargets({
     required github.PullRequest pullRequest,
     String reason = 'Newer commit available',
+    RetryOptions retryOptions = const RetryOptions(maxAttempts: 3),
```
godofredoc, line 247: Should this PR's description get update to retry ci.yaml script independently of the scheduling workflow.

app_dart/test/service/scheduler_test.dart
```diff
@@ -43,7 +44,6 @@ targets:
       - stable
     scheduler: luci
   - name: Google Internal Roll
-    postsubmit: true
```
godofredoc, line 46: Any reason to remove the postsubmit|presubmit configs? it seems like that logic was not changed by this PR

CaseyHillers: These are already the default values (if not present or not false, it will run in that environment). This is cleanup to align with how we use the ci.yaml today.

app_dart/lib/src/service/scheduler.dart
CaseyHillers: There's a few issues I've noticed that lead me to choose here:
1. https://github.com/flutter/flutter/issues/93923 - Failure in filtering the presubmit targets as the GitHub token failed to be auth
2. https://github.com/flutter/flutter/issues/87965 - Failed to find ci.yaml as force pushes require a bit of extra time
3. Network flakes where the ci.yaml contents were off by a few bytes
Currently we don't throw errors if the a build fails to schedule on LUCI, so this should be quick to retry. In the worst case, I expect it to only add 10 seconds to the runtime.

godofredoc (approved):

PR #1481, opened 2021-11-24
Clean up github token for test flaky bot

keyonghan:
The `GitHubFlakyBotToken` was created for testing purpose, this PR cleans up related codes as we have switched to use `fluttergithubbot`.


chunhtai (approved):
LGTM, we may also want to delete it from the cocoon db

keyonghan:
> LGTM, we may also want to delete it from the cocoon db

Already removed:-)

PR #1485, opened 2021-11-27
fix: platform script path works on windows

Letiste:
Fix https://github.com/flutter/flutter/issues/94285

Here is the explanation: https://github.com/dart-lang/sdk/issues/31610

### DIFF
--- a/dashboard/test/utils/task_icons.dart
+++ b/dashboard/test/utils/task_icons.dart
@@ -13,7 +13,7 @@ import 'package:path/path.dart' as path;
 Future<void> precacheTaskIcons(WidgetTester tester) async {
   // Depending on how we're invoked, Platform.script.path will have extra parts
   // after app_flutter. Just trim them off.
-  final List<String> pathParts = path.split(Platform.script.path);
+  final List<String> pathParts = path.split(Platform.script.toFilePath());
   while (pathParts.last != 'dashboard') {
     pathParts.removeLast();
   }


dashboard/test/utils/task_icons.dart
```diff
@@ -13,7 +13,7 @@ import 'package:path/path.dart' as path;
 Future<void> precacheTaskIcons(WidgetTester tester) async {
   // Depending on how we're invoked, Platform.script.path will have extra parts
   // after app_flutter. Just trim them off.
-  final List<String> pathParts = path.split(Platform.script.path);
+  final List<String> pathParts = path.split(Platform.script.toFilePath());
```
CaseyHillers, line 16: Will this work as expected?

Based on the the linked issue, it'll add file:// which is not what we want here. For context, this app is primarily deployed to the web, but we do try and support the other platforms when possible

Letiste: This is only used in the tests, before pumping any widgets. We need to get the correct file path to read the images (or at least that's my understanding, sorry if I'm mistaken)

CaseyHillers: It's using `ExactAssetImage` which does the resolution for the platform the app is running on. I'll close this PR as obsolete

CaseyHillers:
Thank you for the PR! I'm closing this as obsolete, however, please reopen if I missed anything.

PR #1487, opened 2021-11-29
[release_dashboard] Push release UI

Yugue:
UI for `Push the release` step:

Main Issue:
- [x] https://github.com/flutter/flutter/issues/94272*

UI:

image

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/release_dashboard/lib/widgets/progression.dart
+++ b/release_dashboard/lib/widgets/progression.dart
@@ -10,8 +10,8 @@ import 'codesign_engine_substeps.dart';
 import 'conductor_status.dart';
 import 'create_release_substeps.dart';
 import 'publish_tag_substeps.dart';
+import 'push_release_substep.dart';
 import 'release_completed.dart';
-import 'substeps.dart';
 
 /// Displays the progression and each step of the release from the conductor.
 ///
@@ -147,7 +147,7 @@ class MainProgressionState extends State<MainProgression> {
                 ),
                 Step(
                   title: Text(MainProgression._stepTitles[6]),
-                  content: ConductorSubsteps(nextStep: nextStep),
+                  content: PushReleaseSubsteps(nextStep: nextStep),
                   isActive: true,
                   state: handleStepState(6),
                 ),
--- a/release_dashboard/lib/widgets/progression.dart
+++ b/release_dashboard/lib/widgets/progression.dart
@@ -87,6 +87,8 @@ class MainProgressionState extends State<MainProgression> {
           children: <Widget>[
             const ConductorStatus(),
             const SizedBox(height: 20.0),
+            // TODO(Yugue):  render stepper content widget only if the release is at that step,
+            // https://github.com/flutter/flutter/issues/94755.
             Stepper(
               controlsBuilder: (BuildContext context, ControlsDetails details) => Row(),
               physics: const ScrollPhysics(),
--- a/release_dashboard/lib/widgets/push_release_substep.dart
+++ b/release_dashboard/lib/widgets/push_release_substep.dart
@@ -1,0 +1,113 @@
+// Copyright 2014 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:conductor_core/conductor_core.dart';
+import 'package:flutter/material.dart';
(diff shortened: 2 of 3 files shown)


release_dashboard/lib/widgets/progression.dart
```diff
@@ -31,7 +32,10 @@ class MainProgression extends StatefulWidget {
     'Apply Engine Cherrypicks',
     'Codesign Engine Binaries',
     'Publish the Release',
-    'Release is Successfully published'
+    'Release is Successfully published',
```
christopherfujino, line 35: This seems redundant with `'Release is Completed'` and also out of order

release_dashboard/lib/widgets/push_release_substep.dart
```diff
@@ -0,0 +1,113 @@
...
+      'packaging',
+    );
+
+    return Column(
+      crossAxisAlignment: CrossAxisAlignment.center,
+      children: <Widget>[
+        CheckboxAsSubstep(
+          substepName: PushReleaseSubsteps.substepTitles[PushReleaseSubstep.pushRelease]!,
+          subtitle: SelectableText('${PushReleaseSubsteps.substepSubtitles[PushReleaseSubstep.pushRelease]!} '
+              '${releaseStatus?[ConductorStatusEntry.releaseChannel]} branch directly. '
```
christopherfujino, line 78: can this be null at this point?

christopherfujino: > No, but when the whole app gets loaded in tests, it cases a null check failure. They are only here for the tests to pass.

Can we provide a `releaseStatus` in the tests?

Yugue: In `conductor_status.dart`, there are 2 tests where the app doesn't have any `releaseStatus` (to simulate the situation where a release has not been initialized yet). And these tests would fail exactly right here indicating that a null check has been performed on a null value.
Here is the test: https://github.com/Yugue/cocoon/blob/0f9fd2fdd379bd5ec6b487d6248b46e354fa531e/release_dashboard/test/widgets/conductor_status_test.dart#L22

christopherfujino: Why would that test render `PushReleaseSubsteps`?

Yugue: It doesn't directly. But I think `Stepper` paints all the widgets. It displays `PushReleaseSubsteps` although it is not at that step yet. Even if the tests in `conductor_status_test.dart` were changed, the test in `main_test.dart` still fails because it all the entire stepper with a null state [here](https://github.com/Yugue/cocoon/blob/d7623bedb388aafa1535b2c21ec0b09fb673264e/release_dashboard/test/main_test.dart#L16). A a null check error would throw exactly at this line if it were changed to `!`. Therefore, `?` was used to not crash the tests when the entire stepper was painted with no state (which could happen in a scenario where a release hasn't been initialized yet).

christopherfujino: > But it seems that Stepper paints all the widgets in each step. It renders PushReleaseSubsteps although it is not at that step yet.
Ahh, I see.

I think the solution here would be to conditionally render the `MainProgression` widget only when there is a existing state, rather than coding the `MainProgression` and its subwidgets to handle a null state when that doesn't really make sense.

Yugue: I don't think I can conditionally render `MainProgression`. Since when the state is null (a scenario of no release started yet), the user should still be able to access the first step of `MainProgression`, which is to create a new release.

I can see if I can implement a logic to only paint the current step's widget when we are at the current step.

christopherfujino: Oh, good point

Yugue: I added a todo [here](https://github.com/Yugue/cocoon/blob/1aa19c0752d3f7bd34a37548591ea6a0f2d5edf5/release_dashboard/lib/widgets/progression.dart#L90). 

It will be handled in this [PR](https://github.com/flutter/cocoon/pull/1500).

christopherfujino (approved):
LGTM

PR #1484, opened 2021-11-27
[release_dashboard] push tag step UI

Yugue:
UI for the last two steps:

Main Issue:
- [x] https://github.com/flutter/flutter/issues/94271

UI:

image

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read the [Flutter Style Guide] _recently_, and have followed its advice.
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[test-exempt]: https://github.com/flutter/flutter/wiki/Tree-hygiene#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat

PR #1483, opened 2021-11-26
[release_dashboard] Add framework merge PR UI

Yugue:
Refactored previously `codesign_engine_substeps.dart` to `merge_pr_substeps.dart` => a common widget for both `Codesign Engine Binaries` and `Merge Framework PR` steps. 

Two more steps'UI (shorter ones) to go.

Main Issues:
- [x] https://github.com/flutter/flutter/issues/93938
- [x] https://github.com/flutter/flutter/issues/94268

PR that needed to be landed before:
- [x] https://github.com/flutter/cocoon/pull/1466
- [x] https://github.com/flutter/cocoon/pull/1477 (needs to be rebased from this)

4 UI scenarios:
1. When an engine PR is needed:

image

2. When an engine PR is not needed:

image

3. When a framework PR is needed:

image

4. When a framework PR is not needed:

image

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read the [Flutter Style Guide] _recently_, and have followed its advice.
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[test-exempt]: https://github.com/flutter/flutter/wiki/Tree-hygiene#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat

PR #1482, opened 2021-11-25
[release_dashboard] nextContext Connect

Yugue:
*Added `conductor next` implementation for `Apply engine cherrypicks` step*, and for `Apply framework cherrypicks` step.

Main Issue:
- [x] https://github.com/flutter/flutter/issues/94627

PR that needed to be landed before this:
- [ ] https://github.com/flutter/cocoon/pull/1494
- [x] https://github.com/flutter/flutter/pull/93896
- [x] https://github.com/flutter/flutter/pull/91768

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read the [Flutter Style Guide] _recently_, and have followed its advice.
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[test-exempt]: https://github.com/flutter/flutter/wiki/Tree-hygiene#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat

PR #1494, opened 2021-12-02
[release_dashboard] continue button common widget

Yugue:
Made `continueButton` of each step a common widget so it could be used across any steps.

Main Issue:
- [x] https://github.com/flutter/flutter/issues/94582

PR needs to be rebased and updated after this has been merged:
https://github.com/flutter/cocoon/pull/1482
https://github.com/flutter/cocoon/pull/1483
https://github.com/flutter/cocoon/pull/1484
https://github.com/flutter/cocoon/pull/1487

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/release_dashboard/lib/widgets/cherrypicks_substeps.dart
+++ b/release_dashboard/lib/widgets/cherrypicks_substeps.dart
@@ -152,6 +152,7 @@ class CherrypicksSubstepsState extends State<CherrypicksSubsteps> {
             substepPressed(CherrypicksSubstep.applyCherrypicks);
           },
         ),
+        const SizedBox(height: 20.0),
         ContinueButton(
             elevatedButtonKey: Key('apply${repositoryName(widget.repository, true)}CherrypicksContinue'),
             enabled: !_isEachSubstepChecked.containsValue(false),
--- a/release_dashboard/test/widgets/common/continue_button_test.dart
+++ b/release_dashboard/test/widgets/common/continue_button_test.dart
@@ -1,0 +1,113 @@
+// Copyright 2014 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:conductor_ui/widgets/common/continue_button.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+void main() {
+  const Key buttonKey = Key('testButton');
+  const String errorMsg = 'There is an error';
+
+  group('Continue button tests', () {
+    testWidgets('Renders elements correctly', (WidgetTester tester) async {
+      await tester.pumpWidget(
+        MaterialApp(
+          home: Material(
+            child: ContinueButton(
+              elevatedButtonKey: buttonKey,
+              enabled: true,
+              onPressedCallback: () async {},
+              isLoading: false,
+            ),
+          ),
+        ),
+      );
+
+      expect(find.byKey(buttonKey), findsOneWidget);
+      expect(find.byType(ElevatedButton), findsOneWidget);
+      expect(find.byType(SelectableText), findsNothing);
+      expect(tester.widget<ElevatedButton>(find.byKey(buttonKey)).enabled, equals(true));
+      expect(find.byType(CircularProgressIndicator), findsNothing);
+    });
+
+    testWidgets('Display the error message', (WidgetTester tester) async {
+      await tester.pumpWidget(
+        MaterialApp(
+          home: Material(
+            child: ContinueButton(
+              elevatedButtonKey: buttonKey,
+              error: errorMsg,
+              enabled: true,
+              onPressedCallback: () async {},
+              isLoading: false,
+            ),
+          ),
+        ),
+      );
+
+      expect(find.text(errorMsg), findsOneWidget);
+      expect(find.byType(SelectableText), findsOneWidget);
+    });
+
+    testWidgets('Disable the button', (WidgetTester tester) async {
+      await tester.pumpWidget(
+        MaterialApp(
+          home: Material(
+            child: ContinueButton(
+              elevatedButtonKey: buttonKey,
+              enabled: false,
+              onPressedCallback: () async {},
+              isLoading: false,
+            ),
+          ),
+        ),
+      );
+
+      expect(find.byType(ElevatedButton), findsOneWidget);
+      expect(tester.widget<ElevatedButton>(find.byKey(buttonKey)).enabled, equals(false));
+    });
+
+    testWidgets('Display the loading widget', (WidgetTester tester) async {
+      await tester.pumpWidget(
+        MaterialApp(
+          home: Material(
+            child: ContinueButton(
+              elevatedButtonKey: buttonKey,
+              enabled: true,
+              onPressedCallback: () async {},
+              isLoading: true,
+            ),
+          ),
+        ),
+      );
+
+      expect(find.byType(CircularProgressIndicator), findsOneWidget);
+    });
+
+    testWidgets('Execute onPressedCallback if button is enabled', (WidgetTester tester) async {
+      bool callbackExecuted = false;
+      Future<void> onPressedCallback() async {
+        callbackExecuted = true;
+      }
+
+      await tester.pumpWidget(
+        MaterialApp(
+          home: Material(
+            child: ContinueButton(
+              elevatedButtonKey: buttonKey,
+              enabled: true,
+              onPressedCallback: onPressedCallback,
+              isLoading: false,
+            ),
+          ),
+        ),
+      );
+
+      await tester.tap(find.byKey(buttonKey));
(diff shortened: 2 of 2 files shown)


release_dashboard/lib/widgets/common/continue_button.dart
```diff
@@ -0,0 +1,72 @@
...
+              error!,
+              style: Theme.of(context).textTheme.subtitle1!.copyWith(color: Colors.red),
+            ),
+          ),
+        Row(
+          mainAxisAlignment: MainAxisAlignment.center,
+          children: <Widget>[
+            ElevatedButton(
+              key: elevatedButtonKey,
+              onPressed: enabled == true
```
CaseyHillers, line 55: When `isLoading` is true, what marks it back to false? Usually you would confirm the `DynamicFuture` is completed.

Yugue: The parent can change it back. Because it is passed as the parent's state. For example here: 
https://github.com/Yugue/cocoon/blob/eaed2969a2accc1cbf772d1eb1cb73bccda84599/release_dashboard/lib/widgets/create_release_substeps.dart#L221

CaseyHillers: The parent shouldn't have control over this, or need to code it. You can use `Future.whenComplete` to mark it as finished.

https://api.dart.dev/stable/2.14.4/dart-async/Future-class.html

release_dashboard/lib/widgets/common/continue_button.dart
```diff
@@ -0,0 +1,72 @@
...
+                  ? () async {
+                      await onPressedCallback();
```
CaseyHillers, line 57: This should be

```
onPressedCallback().whenComplete(() => isLoading = false);
```

In this case, the UI doesn't need to wait hang for this action to be complete, and it will update itself once the future is complete.

Yugue: Ok, is it also possible to set `isLoading` to true when there is `onPressedCallback()` is started or `onPressedCallback()` encounters an error?

CaseyHillers:
```
isLoading = true;
onPressedCallback().whenComplete(() => isLoading = false);
```

You can implement [catchError](https://api.flutter.dev/flutter/dart-async/Future/catchError.html), but IIRC `whenComplete` acts like `finally` so it is guaranteed to run last.

Yugue: I realized I can't offset the `isLoading` and `error` to `ContinueButton`, they will have to be passed by the parents. Since the parent widget is using `isLoading` to disable the inputs and dropdowns, and needs to modify `error` to display other errors not thrown by the callbacks of `ContinueButton`.

Yugue: I changed the code to use `whenComplete` and `catchError`, please take another look.

CaseyHillers: This widget is missing the logic (it's still in the parent `CreateReleaseSubsteps`). Leaving it in the parent means we'll need to duplicate it everywhere else we use ContinueButton

Yugue: But the parents needs `isLoading` to disable the dropdowns and inputforms. And the parent widget might also updates `_error`. If I offload the logic to the child, I will have to pass two additional parameters `setError` and `setIsLoading`.

Would it be ok to leave the logic as a callback to be passed to the child? In this way we could customize the `onPressedCallback`, and make the widget more flexible to accommodate for future changes. Because every step of the release will be using the same `ContinueButton` widget. If I bring the logic of `CreateReleaseSubsteps` inside the `ContinueButton`, the other steps are not guaranteed to have the same logic as `CreateReleaseSubsteps`. For example, another step could have a `dialoguePromp` appear as `onPressedCallback`, if `Yes` is clicked on the prompt, then executes the rest of the callback, otherwise do not execute.

CaseyHillers: Adding the additional callbacks sgtm. This is a good example of a well defined api that can be used everywhere

release_dashboard/lib/widgets/common/continue_button.dart
Yugue: The parent needs to control the enabling behaviour of the continue button. For `CreateReleaesSubsteps`, when all the inputs and dropdowns of the parent are validated, the continue button becomes enabled. For other steps, the continue button enables when all the parent's checkboxes are checked.

CaseyHillers (commented):
https://github.com/flutter/cocoon/pull/1494#discussion_r763370154 needs to be addressed

release_dashboard/lib/widgets/common/continue_button.dart
Yugue: By adding the additional callbacks, do you mean by passing `setError` and `setIsLoading` as additional parameters of  `ContinueButton`? This wouldn't be necessary if we keep the logic in the parent, since `setError` and `setIsLoading` are parts of the logic of `onPressedCallback` which is defined in the parent, and passed to the child.

The main concern is having the logic declared in the parent and pass as the parameter `onPressedCallback` to the child. This way, `onPressedCallback` is flexible, since the parent can define it however it wants. If we offload it to the child `ContinueButton`, we are bound to use every instance of `ContinueButton` (used by every step) according to the logic `ContinueButton` defined, which will be difficult to accommodate for different behaviours of all continue buttons of each step, for example the one mentioned above.

CaseyHillers: The goal with any API is to make it as intuitive as possible for users to pick up. The assumption you're baking in makes the `ContinueButton`'s onPressed to be very complicated. It's easier to use if you explicitly enable `onError` and `onLoading` as optional parameters in `ContinueButton`.

In general, it's easier to have an API like
``` dart
// Better API design where its explicit
ContinueButton({
  Future<void> onPressed,
  onError?,
  onLoading?,
})
```

versus 

``` dart
/// It's easy to invoke this widget, but it's requiring everywhere else to implement the future logic.
///
/// [onPressed] is a future that should have error handles specified before being passed.
///
/// An example for [onPressed] with error handling and loading:
///
/// ```
///.   myOnPressedFuture.catchError((error, stacktrace) {
///       setError(errorToString(error, stacktrace));
///     }).whenComplete(() {
///       setIsLoading(false);
///       if (_error == null) {
///         myNextFuture();
///       }
///     });
/// ```
///
/// This Continue button doesn't have a call in for `onNext` either, just to emphasize the issues of relying
/// on users to setup everything themselves before calling the widget.
ContinueButton(onPressed)
```

Yugue: The only way I was able to offset the logic to the `ContinueButton` is to pass 3 additional parameters, and making `onPressedCallback` optional:
```
  final Key elevatedButtonKey;
  final String? error;
  final bool enabled;
  final DynamicFuture? onPressedCallback;
  final bool isLoading;
  final SetIsLoading? setIsLoading;
  final SetError? setError;
  final VoidCallback? nextStep;

The `ElevatedButton`'s `onPressed` becomes:
```
              onPressed: enabled == true
                  ? () async {
                      if (setError != null) setError!(null);
                      if (setIsLoading != null) setIsLoading!(true);
                      try {
                        if (onPressedCallback != null) await onPressedCallback!();
                      } catch (error, stacktrace) {
                        if (setError != null) {
                          setError!(errorToString(error, stacktrace));
                        }
                      } finally {
                        if (setIsLoading != null) setIsLoading!(false);
                      }
                      if (error == null && nextStep != null) nextStep!();
                    }
                  : null,
                  ```
As you can see, the logic in `onPressed` becomes much more complicated. Because I have to accommodate for all the steps that are using the `ContinueButton`. Some steps do not have a `onPressedCallback`, and they simply executes `nextStep`. If they simply call `nextStep`, it is unnecessary for them to have `loading` or `error` (since `nextStep` does not load, and it can't throw any error), thus, there is no need to have `setError` or `setLoading` for these steps. That's why I have to do all these null checks.

The main problem is `isLoading`, `error` of `ContinueButton` must come from the parents, so does `setError`, and `nextStep`. Because parents needs to consume these states for other widgets. If the parent doesn't need these, then these states could simply be declared in `ContinueButton`, and there won't be any problem offloading all the logic to `ContinueButton`. 

It would be nicer to have the logic in `ContinueButton`, but the latter is used in many steps, and it should accommodate for all different behaviours. It doesn't appear to be beneficial to offload the logic to `ContinueButton`, unless you have a better approach. 

What do you think if we leave the logic in the parent as it is, and when all the continue buttons are made, I can revisit to see if there is an efficient way to offload all the logic used by each step to `ContinueButton`. It will be simpler to do when the logic for each step is made.

CaseyHillers (approved):

PR #1498, opened 2021-12-06
Merge a clean revert without waiting for checks

keyonghan:
https://github.com/flutter/flutter/issues/94592

### DIFF
--- a/app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
+++ b/app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
@@ -22,6 +22,9 @@ import 'refresh_cirrus_status.dart';
 /// goes green.
 const int _kMergeCountPerCycle = 2;
 
+/// The regular expression string of the default revert commit message.
+const String _regExpCommitMessage = 'This reverts commit (?<revertCommitSha>.+).';
+
 /// Injected latency per repository. Engine and Flutter use an injected latency of 1h meaning
 /// that the bot skips any commits younger than 1h. However 1h is too long for some repositories
 /// whose builds are faster. Use this constant to override the default 1h latency for a given repository.
@@ -67,7 +70,14 @@ class CheckForWaitingPullRequests extends ApiRequestHandler<Body> {
     );
     final List<_AutoMergeQueryResult> queryResults = await _parseQueryData(data, slug.name);
     for (_AutoMergeQueryResult queryResult in queryResults) {
-      if (mergeCount < _kMergeCountPerCycle && queryResult.shouldMerge) {
+      final RegExp commitRegExp = RegExp(_regExpCommitMessage);
+      final RegExpMatch? match = commitRegExp.firstMatch(queryResult.message);
+      String? revertCommitSha;
+      if (match != null) {
+        revertCommitSha = match.namedGroup('revertCommitSha');
+      }
+      // If the PR is a revert of the tot commit, merge without waiting for checks passing.
+      if (await isShaFromTOT(revertCommitSha, slug) || mergeCount < _kMergeCountPerCycle && queryResult.shouldMerge) {
         final bool merged = await _mergePullRequest(
           queryResult.graphQLId,
           queryResult.sha,
@@ -88,6 +98,16 @@ class CheckForWaitingPullRequests extends ApiRequestHandler<Body> {
         );
       }
     }
+  }
+
+  /// Check if the `commitSha` is from TOT.
+  Future<bool> isShaFromTOT(
+    String? commitSha,
+    RepositorySlug slug,
+  ) async {
+    final GitHub github = await config.createGitHubClient(slug: slug);
+    final RepositoryCommit totCommit = await github.repositories.getCommit(slug, 'HEAD');
+    return commitSha != null && totCommit.sha == commitSha;
   }
 
   Future<Map<String, dynamic>> _queryGraphQL(
@@ -217,6 +237,9 @@ class CheckForWaitingPullRequests extends ApiRequestHandler<Body> {
           );
 
       final String sha = commit['oid'] as String;
+      // The commit message will be used to identify is a revert happens, and further to check if a
+      // TOT commit is being reverted. If yes, merge the PR without waiting for all tests passing.
+      final String message = commit['message'] as String;
       List<Map<String, dynamic>>? statuses;
       if (commit['status'] != null &&
           commit['status']['contexts'] != null &&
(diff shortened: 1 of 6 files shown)


app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
```diff
@@ -67,7 +70,14 @@ class CheckForWaitingPullRequests extends ApiRequestHandler<Body> {
     );
     final List<_AutoMergeQueryResult> queryResults = await _parseQueryData(data, slug.name);
     for (_AutoMergeQueryResult queryResult in queryResults) {
-      if (mergeCount < _kMergeCountPerCycle && queryResult.shouldMerge) {
+      final RegExp commitRegExp = RegExp(_regExpCommitMessage);
+      final RegExpMatch? match = commitRegExp.firstMatch(queryResult.message);
+      String? revertCommitSha;
+      if (match != null) {
+        revertCommitSha = match.namedGroup('revertCommitSha');
+      }
```
keyonghan, line 78: Github doesn't treat revert PR/commit separately, and there is not a simple mapping between the revert and the to-be-reverted commit.

Checked recent reverts, and all are with the default commit message which includes the to-be-reverted commit sha info. (this is what is proposed here).

Another more stable way is use git diff between the current revert commit and the second TOT commit: if diff is zero, then we can conclude the current revert is  based on the TOT commit. Filed https://github.com/SpinlockLabs/github.dart/issues/283 to request support on git diff comparison.

CaseyHillers: With https://github.com/flutter/flutter/commit/86cb2598a3b683ebb7bdb9ed4ba6bb0b57424d05 as an example,

Cocoon could grab
1. The proposed revert PR diff: https://github.com/flutter/flutter/commit/86cb2598a3b683ebb7bdb9ed4ba6bb0b57424d05.patch
2. The tip of tree commit diff: https://github.com/flutter/flutter/commit/9412edf1ca69121fe733160e22bc48131e7a3fcd.patch

It can then split the lines of each, and for each line that starts with `+` or `-` add it to a map. When going through the second diff, it can reverse the operator and remove it from the map. If the map has any keys remaining, it's not a pure revert and should be skipped.

keyonghan: https://docs.github.com/en/rest/reference/commits#compare-two-commits does exactly the same thing. It sounds a re-implementation of this API. Submitted https://github.com/SpinlockLabs/github.dart/pull/285 to support the `files` field so that we can use this API.

This is the diff result of a revert commit: https://github.com/flutter/flutter/pull/94611/commits/9d2bf9140499984aff7d78145029cfc2a39a762d and the second tot commit: 86d229dc399136aa7a0b3935f723ffa1289217f8
https://api.github.com/repos/flutter/flutter/compare/86d229dc399136aa7a0b3935f723ffa1289217f8...9d2bf9140499984aff7d78145029cfc2a39a762d, where the empty `files` field validates this is a clean revert.

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
```diff
@@ -67,7 +67,8 @@ class CheckForWaitingPullRequests extends ApiRequestHandler<Body> {
     );
     final List<_AutoMergeQueryResult> queryResults = await _parseQueryData(data, slug.name);
     for (_AutoMergeQueryResult queryResult in queryResults) {
-      if (mergeCount < _kMergeCountPerCycle && queryResult.shouldMerge) {
+      // If the PR is a revert of the tot commit, merge without waiting for checks passing.
+      if (mergeCount < _kMergeCountPerCycle && queryResult.shouldMerge || await isTOTRevert(queryResult.sha, slug)) {
```
CaseyHillers, line 71: Could we simplify this? Maybe exit early by checking `_kMergeCountPerCycle < mergeCount` and add a helper function

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
```diff
@@ -90,6 +91,23 @@ class CheckForWaitingPullRequests extends ApiRequestHandler<Body> {
...
+  /// `GitHubComparison` validates a clean revert of TOT commit.
+  Future<bool> isTOTRevert(
+    String commitSha,
+    RepositorySlug slug,
+  ) async {
+    final GitHub github = await config.createGitHubClient(slug: slug);
+    final RepositoryCommit secondTotCommit = await github.repositories.getCommit(slug, 'HEAD~');
+    log.info('Current commit is: $commitSha');
+    log.info('Second TOT commit is: ${secondTotCommit.sha}');
+    final GitHubComparison gitHubComparison =
```
CaseyHillers, line 106: nit: gitHub -> github

app_dart/pubspec.yaml
```diff
@@ -19,7 +19,7 @@ dependencies:
   file: ^6.1.2
   fixnum: ^1.0.0
   gcloud: ^0.8.4
-  github: ^8.2.4
+  github: ^8.3.0
```
CaseyHillers, line 22: Nice! Thanks for upstreaming this logic

app_dart/test/request_handlers/check_for_waiting_pull_requests_test.dart
```diff
@@ -362,6 +387,34 @@ void main() {
       ]);
     });
 
+    test('Merge a clean revert PR with in progress tests', () async {
```
CaseyHillers, line 390: Can we add a test for a non-clean revert?

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
keyonghan: Added a helper function to make it clear.

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
```diff
@@ -90,6 +91,23 @@ class CheckForWaitingPullRequests extends ApiRequestHandler<Body> {
     }
   }
 
+  /// Check if the `commitSha` is a clean revert of TOT commit.
```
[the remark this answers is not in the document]
keyonghan: Done

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
keyonghan: Done

app_dart/test/request_handlers/check_for_waiting_pull_requests_test.dart
keyonghan: Please see `githubComparison = GitHubComparison('abc', 'def', 0, 0, 0, <CommitFile>[CommitFile(name: 'test')]);` in other tests, which is mocking the non-clean revert cases.

CaseyHillers (approved):
flutter-lgtm

PR #1493, opened 2021-12-02
[app_dart] Pass repo for Task and Commit

CaseyHillers:
https://github.com/flutter/flutter/issues/94522

## Test Env

http://testchillers.flutter-dashboard.appspot.com/api/public/get-status?flushCache=true&repo=engine

## Future work

1. Add other repos to supported scheduler repos (to start writing to datastore)
  A. This should get a soak time of 30 commits for each repo to populate enough data for dashboards and tree status
2. Add refresh chromebot cron jobs for each repo (temporary measure until cocoon scheduler is available)
3. Revert tree status back to the datastore algorithm
3. Update frontend to allow for repo switching

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read the [Flutter Style Guide] _recently_, and have followed its advice.
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

PR #1503, opened 2021-12-06
[scheduler] Write every Flutter repo to datastore

CaseyHillers:

## Order of PRs

[X] - https://github.com/flutter/cocoon/pull/1493 - Use repo when reading all data
[ ] - **Current**: Write data from other repos
[ ] - Next: Update front end to load other repos


app_dart/lib/src/service/config.dart
```diff
@@ -47,11 +47,6 @@ class Config {
     RepositorySlug('flutter', 'plugins'),
   };
 
-  /// List of GitHub repositories that are supported by [Scheduler].
-  static Set<RepositorySlug> schedulerSupportedRepos = <RepositorySlug>{
```
CaseyHillers, line 51: This is the key logic here. Removing it enables the scheduler to start writing other repos.

app_dart/lib/src/service/scheduler.dart
```diff
@@ -117,7 +117,7 @@ class Scheduler {
   }
 
   Future<void> _addCommit(Commit commit) async {
-    if (!Config.schedulerSupportedRepos.contains(commit.slug)) {
+    if (!Config.supportedRepos.contains(commit.slug)) {
```
CaseyHillers, line 120: This is the key logic. Moving to supportedRepos will support flutter, engine, plugins, packages, cocoon

keyonghan: This will write all tasks to datastore as well. How about the task status update? I guess we will wait for the result processing as the current refresh API seems already overloaded just considering framework repo.

CaseyHillers: update-task-status is hard coded to only work for the framework - https://cs.opensource.google/flutter/cocoon/+/main:app_dart/lib/src/request_handlers/update_task_status.dart;l=111;drc=fddb3f5f9887e9621b1650bf36a345398847f724

The plan is to deprecate it in favor of the results processing microservice

CaseyHillers: For the results being overloaded, I plan to add cron jobs for each separate repo (which shouldn't be an issue as flutter/flutter has by far the most tasks). Early Q1 that will no longer be necessary as I'll have everything wired to update via pubsub

keyonghan (approved):
LGTM

PR #1504, opened 2021-12-07
Fix auto flake comment on open flaky bug for a recently deflaked builder

keyonghan:
https://github.com/flutter/cocoon/pull/1464/files updated the bot to comment on open flaky bugs based on different pool stats. But it missed a corner case when a builder was marked as unflaky but a flaky bug is still open. For this case, the bot logic updates the bug twice.

This PR skip the update from staging side, as the builder has been marked as unflaky.

Context: https://github.com/flutter/flutter/issues/93804#


chunhtai (approved):
LGTM

PR #1505, opened 2021-12-07
[web] Add the flutter-gold check in web engine PRs

mdebbar:
Add a cron entry to send gold status to flutter/engine PRs that run web engine tests.


app_dart/cron.yaml
```diff
@@ -6,30 +6,43 @@ cron:
...
   url: /api/push-build-status-to-github?repo=flutter/engine
   schedule: every 2 minutes
+
+- description: sends pr-specific gold status to GitHub to annotate flutter PRs and commits
+  url: /api/push-gold-status-to-github?repo=flutter/flutter
+  schedule: every 5 minutes
+
+- description: sends pr-specific gold status to GitHub to annotate engine PRs and commits
+  url: /api/push-gold-status-to-github?repo=flutter/engine
+  schedule: every 5 minutes
```
Piinks, line 28: Having separate cron jobs are probably less ideal.
Gold already has a very tight budget for the requests we make to it, so if these were running simultaneously, we might go over.
We should do this more like `check_for_waiting_pull_requests.dart`, with one cron job running through the supported repos.

Here is where that is happening:
https://github.com/flutter/cocoon/blob/main/app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart#L44

Piinks (commented):
This will need to also update the URLs/messages in `push_gold_status_to_github.dart` and `config.dart`, currently they point to the framework instance of Gold, I think you will want them to point to the engine instance.

The `pullRequestChecksQuery` is also specific to flutter/flutter. I think this needs a bit more refactoring.

Tests should be easy, you can probably just copy-paste all of the framework version of these tests, and change the repo to `flutter/engine` to make sure everything is working: https://github.com/flutter/cocoon/blob/main/app_dart/test/request_handlers/push_gold_status_to_github_test.dart

app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
```diff
@@ -46,7 +48,8 @@ class PushGoldStatusToGithub extends ApiRequestHandler<Body> {
       return Body.empty;
     }
 
-    final RepositorySlug slug = Config.flutterSlug;
+    final String repo = request!.uri.queryParameters[kRepoParam] ?? 'flutter/flutter';
```
CaseyHillers, line 51: Generally, Cocoon refers to `RepositorySlug.full` in string form as `repository`. `repo` refers to the RepositorySlug.name, defaulting the owner to 'flutter'

app_dart/cron.yaml
mdebbar: We send a separate request to Gold for each PR, right? If that's the case, then the number of cron jobs doesn't matter. If there are 10 PRs in flutter and 10 PRs in the engine, we'll be sending 20 requests to Gold regardless of whether we have a single cron job or separate ones. I might be missing something here, so please correct me if I'm wrong :)

mdebbar:
> Tests should be easy, you can probably just copy-paste all of the framework version of these tests, and change the repo to `flutter/engine` to make sure everything is working: https://github.com/flutter/cocoon/blob/main/app_dart/test/request_handlers/push_gold_status_to_github_test.dart

I wanted to first send the PR to make sure I'm on the right path before I add/update tests.

app_dart/cron.yaml
Piinks: > We send a separate request to Gold for each PR, right?

We do, but one at a time.

There is a strict time threshold on those Gold endpoints. So the quantity does not matter 10 framework + 10 engine PRs = 20, but if they are running in parallel like this, they could max out that rate limit.

mdebbar:
@Piinks there's a check for `master` here: https://github.com/flutter/cocoon/blob/c9a5c6d022f2067cd9d8bce52372e729d2901cbb/app_dart/lib/src/request_handlers/push_gold_status_to_github.dart#L68 

I'm gonna change it to:
```dart
if (pr.base!.ref != 'master' || pr.base!.ref != 'main') {
```
Does that sound good to you?

Piinks:
> I wanted to first send the PR to make sure I'm on the right path before I add/update tests.

For sure! I was hoping to save you some time. 😜

Piinks:
> ```dart
> if (pr.base!.ref != 'master' || pr.base!.ref != 'main') {
> ```
> 
> Does that sound good to you?

SGTM! I think the framework will make that change to `main` eventually. I hope. :) Thanks!

app_dart/cron.yaml
mdebbar: Looking at the code, I see the requests being made sequentially anyway (unless I'm misunderstanding that code): https://github.com/flutter/cocoon/blob/c9a5c6d022f2067cd9d8bce52372e729d2901cbb/app_dart/lib/src/request_handlers/push_gold_status_to_github.dart#L53-L55

I'm happy to unify the cron jobs into one. Just trying to fully understand the reason behind it :)

Piinks: Yes, they are done sequentially. But if you have two cron jobs, you will have two running simultaneously.

It is the requests we make to the Gold service itself (not github or the datastore) that are so restricted. 
https://github.com/flutter/cocoon/blob/c9a5c6d022f2067cd9d8bce52372e729d2901cbb/app_dart/lib/src/request_handlers/push_gold_status_to_github.dart#L203

Since it is a separate Gold instance, it may not be a problem, but it causes a lot of issues if we hit that rate limit.
I never noticed the build status were separate cron jobs, so if it fits the pattern and does not put us over the limit, I am fine with keeping them separate. :)

mdebbar: Done.

Makes perfect sense now, thanks for explaining!

app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
mdebbar: Done.

app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
```diff
@@ -65,8 +76,9 @@ class PushGoldStatusToGithub extends ApiRequestHandler<Body> {
         continue;
       }
 
-      if (pr.base!.ref != 'master') {
-        log.fine('This change is not staged to land on master, skipping.');
+      final String? baseRef = pr.base!.ref;
+      if (baseRef != 'master' && baseRef != 'main') {
```
CaseyHillers, line 80:
```suggestion
      if (baseRef != Config.defaultBranch(pr.base!.repo!.slug())) {
```

mdebbar: Nice! I'm assuming you'll update the `defaultBranch` method when flutter/flutter switches to the `main` branch?

CaseyHillers: Yep! Infra is maintaining this API for the default branch instead of hard coding it everywhere

app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
```diff
@@ -319,13 +346,14 @@ Future<Map<String, dynamic>?> _queryGraphQL(GraphQLClient client, int? prNumber)
 
 const String pullRequestChecksQuery = r'''
 query ChecksForPullRequest($sPullRequest: Int!) {
-  repository(owner: "flutter", name: "flutter") {
+  repository(owner: $sRepoOwner, name: $sRepoName) {
     pullRequest(number: $sPullRequest) {
       commits(last: 1) {
         nodes {
           commit {
             # (appId: 64368) == flutter-dashboard. We only care about
```
mdebbar, line 354: @Piinks Do we need to change the `appId` for the engine?

mdebbar (commented):
@Piinks I had to adjust some things in the tests to be able to test for multiple repos. Could you please take a look to make sure I didn't mess things up?

app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
```diff
@@ -65,8 +74,9 @@ class PushGoldStatusToGithub extends ApiRequestHandler<Body> {
         continue;
       }
 
-      if (pr.base!.ref != 'master') {
-        log.fine('This change is not staged to land on master, skipping.');
+      final String defaultBranch = Config.defaultBranch(slug);
```
Piinks, line 77: Nice!

app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
```diff
@@ -245,8 +256,16 @@ class PushGoldStatusToGithub extends ApiRequestHandler<Body> {
     }
   }
 
-  String _getTriageUrl(int? number) {
-    return 'https://flutter-gold.skia.org/cl/github/$number';
+  String _getTriageUrl(RepositorySlug slug, int? number) {
```
Piinks, line 259: We should probably change this to be non nullable. It would be invalid if this got a null PR number.

app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
```diff
@@ -291,21 +310,27 @@ class PushGoldStatusToGithub extends ApiRequestHandler<Body> {
...
     }
     return true;
   }
 }
 
-Future<Map<String, dynamic>?> _queryGraphQL(GraphQLClient client, int? prNumber) async {
+Future<Map<String, dynamic>?> _queryGraphQL(
+  GraphQLClient client,
+  RepositorySlug slug,
+  int? prNumber,
```
Piinks, line 324: Same here. I do not recall why these would ever be null.

app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
Piinks: I do not think so, it is the same in `check_for_waiting_pull_requests_queries.dart`, which is used by the merge bot across multiple repositories. @godofredoc would know for sure though.

app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
mdebbar: Probably because the `PullRequest` class has the `number` property as nullable.

Piinks: Yeah I was thinking that may be so. In that case, we should probably assert somewhere that we have gotten a PR number to work with, otherwise this all breaks.

Piinks (commented):
LGTM with the null safety nits.
This repo is frozen until 1/4/22, so we should be able to land this then.

mdebbar:
@Piinks ping 🙂

Piinks (commented):
Can you add the assertions that the PR numbers aren't null from the last review?

mdebbar:
> Can you add the assertions that the PR numbers aren't null from the last review?

Ah, thanks for the reminder. Done.

Piinks (approved):
Flutter_LGTM

mdebbar:
@Piinks I'm unable to merge this PR because only "authorized users" are allowed to. Is this something you can do, or do I have to add my name somewhere?

Piinks:
I am not authorized either. @CaseyHillers can you merge this PR?

CaseyHillers:
> I am not authorized either. @CaseyHillers can you merge this PR?

As FYI, you'll need to use the `waiting for tree to green` label. That has the bot handle merges, which will be the defacto way across all repos in a week.

PR #1510, opened 2021-12-09
[scheduler] Pass git_branch

CaseyHillers:
Quick fix. Cocoon's scheduler wasn't passing anything so it was using the starlark defaults (which for presubmit can be incorrect on release branches)

### DIFF
--- a/app_dart/lib/src/service/luci_build_service.dart
+++ b/app_dart/lib/src/service/luci_build_service.dart
@@ -176,6 +176,7 @@ class LuciBuildService {
     }
     properties ??= <String, dynamic>{};
     properties.addAll(<String, String>{
+      'git_branch': pullRequest.base!.ref!.replaceAll('refs/heads/', ''),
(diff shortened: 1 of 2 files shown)


godofredoc (approved):

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -176,6 +176,7 @@ class LuciBuildService {
     }
     properties ??= <String, dynamic>{};
     properties.addAll(<String, String>{
+      'git_branch': pullRequest.base!.ref!.replaceAll('refs/heads/', ''),
```
godofredoc, line 179: Do we have a property to identify this is presubmit?

CaseyHillers: No, we don't. You could infer that if `git_ref.contains('refs/pull')` it's presubmit, but I wouldn't recommend it.

In the recipe, you can check the buildbucket bucket == try, which is what the devicelab recipe does.

PR #1513, opened 2021-12-10
[app_dart] Remove repo logic from luci build service

CaseyHillers:
This fixes the chromebot statuses on other repos as (1) it'll get the correct default branch and (2) fixes incorrectly filtering builders

### DIFF
--- a/app_dart/lib/src/service/luci.dart
+++ b/app_dart/lib/src/service/luci.dart
@@ -109,7 +109,6 @@ class LuciService {
   /// retries for [repo] including the task name or not.
   Future<List<Build>> getBuildsForBuilderList(
     List<LuciBuilder> builders, {
-    String? repo,
     bool requireTaskName = false,
   }) async {
     final List<Build> builds = <Build>[];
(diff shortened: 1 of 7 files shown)


app_dart/lib/src/service/luci.dart
```diff
@@ -165,11 +164,8 @@ class LuciService {
   /// predefined in cocoon config.
   ///
   /// Latest builds of each builder will be returned from new to old.
-  Future<Iterable<Build>> getBuilds(String? repo, bool requireTaskName, List<LuciBuilder> builders) async {
+  Future<Iterable<Build>> getBuilds(bool requireTaskName, List<LuciBuilder> builders) async {
     bool includeBuilder(LuciBuilder builder) {
-      if (repo != null && builder.repo != repo) {
```
CaseyHillers, line 170: This logic isn't necessary. Flutter's LUCI project has all builders globally available to every repo.

Thus, builders are repos specific because of the ci.yaml, and the passed `builders` here has already filtered it.

keyonghan (approved):
LGTM

PR #1525, opened 2021-12-14
Fix vacuum github test

CaseyHillers:
Fixes https://github.com/flutter/flutter/issues/95212

The issue was the test was not clearing the datastore between runs, so previous data filled the datastore

### DIFF
--- a/app_dart/test/request_handlers/vacuum_github_commits_test.dart
+++ b/app_dart/test/request_handlers/vacuum_github_commits_test.dart
@@ -180,7 +180,7 @@ void main() {
         }
       };
       final Body body = await tester.get<Body>(handler);
-      expect(db.values.values.whereType<Commit>().length, 3 + Config.supportedRepos.length);
+      expect(db.values.values.whereType<Commit>().length, githubCommits.length + Config.supportedRepos.length);
       expect(db.values.values.whereType<Commit>().map<String>(toSha), containsAll(<String>['1', '2', '4']));
       expect(db.values.values.whereType<Commit>().map<int>(toTimestamp), containsAll(<int>[1, 2, 4]));
       expect(await body.serialize().toList(), isEmpty);
--- a/app_dart/test/request_handlers/vacuum_github_commits_test.dart
+++ b/app_dart/test/request_handlers/vacuum_github_commits_test.dart
@@ -3,7 +3,6 @@
 // found in the LICENSE file.
 
 import 'package:cocoon_service/src/model/appengine/commit.dart';
-import 'package:cocoon_service/src/model/appengine/task.dart';
 import 'package:cocoon_service/src/request_handlers/vacuum_github_commits.dart';
 import 'package:cocoon_service/src/request_handling/body.dart';
 import 'package:cocoon_service/src/service/config.dart';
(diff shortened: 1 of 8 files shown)


app_dart/test/request_handlers/vacuum_github_commits_test.dart
```diff
@@ -177,8 +180,7 @@ void main() {
         }
       };
       final Body body = await tester.get<Body>(handler);
-      expect(db.values.values.whereType<Commit>().length, 4 + Config.supportedRepos.length);
-      expect(db.values.values.whereType<Task>().length, 2 + (2 * 3 * Config.supportedRepos.length));
```
CaseyHillers, line 181: I'm removing this assertion as it's not relevant to the test (it's testing the scheduler implementation which changes from time to time)

app_dart/test/request_handlers/vacuum_github_commits_test.dart
```diff
@@ -164,11 +167,11 @@ void main() {
 
     test('skips commits for which transaction commit fails', () async {
       githubCommits = <String>['2', '3', '4'];
-      config.flutterBranchesValue = <String>['master'];
+      config.flutterBranchesValue = <String>['main'];
 
       /// This test is simulating an existing branch, which must already
       /// have at least one commit in the datastore.
-      final Commit commit = shaToCommit('1', 'master', Config.flutterSlug);
+      final Commit commit = shaToCommit('1', 'main', Config.engineSlug);
```
stuartmorgan, line 174: Is this just to avoid this test being disrupted later when flutter/flutter changes default branches?

app_dart/test/request_handlers/vacuum_github_commits_test.dart
```diff
@@ -98,6 +97,10 @@ void main() {
       when(githubService.github.repositories).thenReturn(repositories);
     });
 
+    tearDown(() {
+      db.values.clear();
```
stuartmorgan, line 101: `db` is being set to a new `FakeDatastoreDB` in `setUp`; why is this necessary?

app_dart/test/request_handlers/vacuum_github_commits_test.dart
CaseyHillers: Yep!

app_dart/test/request_handlers/vacuum_github_commits_test.dart
CaseyHillers: I'm not sure. I've noticed similar weirdness with other async tests before where we need to add tearDown logic.

The core issue was commits from other tests were in this test's `db`

stuartmorgan: How is that possible?

If somehow tests are running in parallel with cross-contaminated `db` instances then this doesn't seem like a safe fix, because you could just as easily be clearing the commits from a different test while it was executing.

stuartmorgan: I was curious, so I tried running just this one test locally, and it still fails exactly the same way, so I don't see how it can be caused by other tests.

stuartmorgan: Just replacing the magic number `4` with `githubCommits.length` (which is 3) as you did below fixes the test, without this change. If the core issue is that `db` was wrong, why are the expectations changing?

CaseyHillers: I've removed the tearDown. I'm on my linux workstation today and I'm not able to reproduce.

stuartmorgan (approved):
LGTM

(Looks like there's another completely unrelated failure at the moment.)

PR #1252, opened 2021-06-22
Merge the handlers to update engine and flutter status.

godofredoc:
This is in preparation to enable status updates for the flutter/plugin repository.

Bug: https://github.com/flutter/flutter/issues/84835

PR #1527, opened 2021-12-15
[app_dart] Move tree status check to be based on datastore algorithm

CaseyHillers:
Fixes https://github.com/flutter/flutter/issues/90569
https://github.com/flutter/flutter/issues/94522 - Enabling this to unify more of the logic in Cocoon
https://github.com/flutter/flutter/issues/94483 - More state in Cocoon makes it easier to audit why issues happen in the tree state

Reverts https://github.com/flutter/cocoon/pull/1252

### DIFF
--- a/app_dart/lib/src/request_handlers/push_build_status_to_github.dart
+++ b/app_dart/lib/src/request_handlers/push_build_status_to_github.dart
@@ -18,12 +18,10 @@ import '../service/logging.dart';
 class PushBuildStatusToGithub extends ApiRequestHandler<Body> {
   const PushBuildStatusToGithub(
     Config config,
-    AuthenticationProvider authenticationProvider,
-     {
+    AuthenticationProvider authenticationProvider, {
     @visibleForTesting DatastoreServiceProvider? datastoreProvider,
     @visibleForTesting BuildStatusServiceProvider? buildStatusServiceProvider,
-  })  : 
-        datastoreProvider = datastoreProvider ?? DatastoreService.defaultProvider,
+  })  : datastoreProvider = datastoreProvider ?? DatastoreService.defaultProvider,
         buildStatusServiceProvider = buildStatusServiceProvider ?? BuildStatusService.defaultProvider,
         super(config: config, authenticationProvider: authenticationProvider);
 
--- a/app_dart/lib/src/request_handlers/push_build_status_to_github.dart
+++ b/app_dart/lib/src/request_handlers/push_build_status_to_github.dart
@@ -8,10 +8,7 @@ import 'package:github/github.dart';
 import 'package:meta/meta.dart';
 
 import '../../cocoon_service.dart';
-import '../../src/service/luci.dart';
-import '../model/appengine/commit.dart';
 import '../model/appengine/github_build_status_update.dart';
-import '../model/ci_yaml/ci_yaml.dart';
 import '../request_handling/api_request_handler.dart';
 import '../service/build_status_provider.dart';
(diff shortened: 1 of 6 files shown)


app_dart/lib/src/request_handlers/push_build_status_to_github.dart
```diff
@@ -55,37 +37,15 @@ class PushBuildStatusToGithub extends ApiRequestHandler<Body> {
...
-      }
-      final String latestStatus = await buildStatusService.latestLUCIStatus(tasks);
-      if (status == GithubBuildStatusUpdate.statusSuccess && latestStatus == GithubBuildStatusUpdate.statusFailure) {
-        status = GithubBuildStatusUpdate.statusFailure;
-      }
-    });
-    await _insertBigquery(slug, status, Config.defaultBranch(slug), config);
-    await _updatePRs(slug, status, datastore, Config.defaultBranch(slug));
-    log.fine('All the PRs for $repo have been updated with $status');
+    final BuildStatus status = (await buildStatusService.calculateCumulativeStatus(slug))!;
```
keyonghan, line 45: I guess the main concern here is the latency. This depends on refresh-chromebot-status API, which sometimes causes ~10 min delay.
As we are implementing pubsub support, I would recommend we hold this change until the pubsub is ready.

CaseyHillers: https://github.com/flutter/flutter/issues/90569 is a P1 because the tree status isn't deterministic.

For flutter/flutter, the latency is <5 minutes (3 minute cron + 2 minute refresh chromebot status), whereas flutter/engine is <3 minutes (3 minute cron + <15 second refresh crhomebot status)

keyonghan: Yeah, we have the cron job frequency set up for short, but it usually takes longer.
If you take a look at the logs, the refresh api runs every ~6 min for flutter, and ~3 min for engine, whereas the push-build-status-to-github API runs every ~3min for flutter, and 2min for engine.

I am wondering when we can have the pubsub set up. If it is expected soon, then we can wait for that as the bug is open for a couple of months anyway.

@godofredoc may have some thoughts here.

CaseyHillers:
Friendly ping @godofredoc . I think this is still something we want.

app_dart/lib/src/service/build_status_provider.dart
```diff
@@ -139,31 +137,6 @@ class BuildStatusService {
     }
   }
 
-  /// This function gets called with the last 40 builds fo a given builder ordered
```
godofredoc, line 142: In the medium term we would like to get the luci-status as an independent microservice relying only in buildbucket, this functionality will be useful there although most likely in a different class.

app_dart/lib/src/request_handlers/push_build_status_to_github.dart
godofredoc: Making the current logic deterministic requires small fixes in the way we sort the results from the buildbucket responses. Something like:

* Get the last 30 commits, sort them by commit and associate builds to commits
* Iterate over the set of tasks starting from the most recent commit all the way to the oldest
* If a task fails or if a task does not contain any results close the tree otherwise keep the tree green

However we can wait for this logic to be implemented when we separate the bot to its own microservice.

godofredoc (approved):

CaseyHillers:
Cocoon deploys are lagging behind. I'll land this tomorrow if everything is green in the latest deployment.

PR #1528, opened 2021-12-15
[ci.yaml] Make generate_jspb compilable

CaseyHillers:
https://github.com/flutter/flutter/issues/92121

This enables us to do `dart compile exe app_dart/bin/generate_jspb.dart -o build/generate_jsppb` and upload it to cipd. This will enable the ci.yaml roller recipe to not need a check out Cocoon. Which then addresses the engine's concern checking out an unpinned version of Cocoon.


christopherfujino (approved):
nifty, lgtm

PR #1529, opened 2021-12-15
[dashboard] Process CommitStatus in isolate

CaseyHillers:
This is processing ~200 KB of data every 30 seconds. It would speed up the app a bit to use an isolate for processing this data instead.


keyonghan (approved):
LGTM

PR #1531, opened 2021-12-16
[scheduler] Fix xcode not being forwarded to recipe module

CaseyHillers:
https://github.com/flutter/flutter/issues/95429

This was an issue with platform specific xcode not being applied to the properties

### DIFF
--- a/app_dart/lib/src/model/ci_yaml/target.dart
+++ b/app_dart/lib/src/model/ci_yaml/target.dart
@@ -70,9 +70,9 @@ class Target {
     mergedProperties['dependencies'] = mergedDependencies.values.map((Dependency dep) => dep.toJson()).toList();
 
     // xcode is a special property as there's different download policies if its in the devicelab.
-    if (properties.containsKey('xcode')) {
+    if (mergedProperties.containsKey('xcode')) {
       final Object xcodeVersion = <String, Object>{
-        'sdk_version': properties['xcode']!,
+        'sdk_version': mergedProperties['xcode']!,
       };
 
       if (iosPlatforms.contains(getPlatform())) {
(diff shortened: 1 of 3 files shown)


app_dart/lib/src/model/ci_yaml/target.dart
```diff
@@ -70,9 +70,9 @@ class Target {
     mergedProperties['dependencies'] = mergedDependencies.values.map((Dependency dep) => dep.toJson()).toList();
 
     // xcode is a special property as there's different download policies if its in the devicelab.
-    if (properties.containsKey('xcode')) {
+    if (mergedProperties.containsKey('xcode')) {
       final Object xcodeVersion = <String, Object>{
-        'sdk_version': properties['xcode']!,
+        'sdk_version': mergedProperties['xcode']!,
```
CaseyHillers, line 75: TODO(chillers): Write a regression test

CaseyHillers:
\cc @jmagman

jmagman (approved):
Code LGTM, @keyonghan should review as well.

app_dart/test/model/ci_yaml/target_test.dart
```diff
@@ -0,0 +1,56 @@
...
+        'bringup': false,
+        'dependencies': <String>[],
+        '\$flutter/devicelab_osx_sdk': <String, Object>{
+          'sdk_version': '12abc',
+        },
+        'xcode': '12abc',
+      });
+    });
+
+    test('platform properties with xcode', () {
```
keyonghan, line 38: We need to have a test when both platform and target properties exist, which is the main purpose of this PR.

CaseyHillers: Added!

keyonghan (approved):
LGTM

PR #1532, opened 2021-12-21
Add screen rotation health check for android devices

keyonghan:
The screen rotation is expected to be disabled for a healthy device.

https://github.com/flutter/flutter/issues/95145

### DIFF
--- a/device_doctor/test/src/android_device_test.dart
+++ b/device_doctor/test/src/android_device_test.dart
@@ -6,6 +6,7 @@ import 'dart:async';
 import 'dart:convert';
 import 'dart:io';
 
+import 'package:file/src/backends/memory/memory_file_system.dart';
 import 'package:mockito/mockito.dart';
 import 'package:test/test.dart';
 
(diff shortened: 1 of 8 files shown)


device_doctor/test/src/android_device_test.dart
```diff
@@ -266,6 +266,43 @@ void main() {
...
   });
 
+  group('AndroidScreenRotationCheck', () {
+    AndroidDeviceDiscovery deviceDiscovery;
+    MockProcessManager processManager;
+    Process process;
+    List<List<int>> output;
+
+    setUp(() {
+      deviceDiscovery = AndroidDeviceDiscovery('/tmp/output');
```
godofredoc, line 276: Would this try to write to that location? if it will I'd recommend to use an in memory file instead.

keyonghan: Thanks for the comment, used memory file instead.

godofredoc (approved):

PR #1537, opened 2022-01-06
CommitFilter highlight mode

flar:
This PR adds a new "highlight or filter" option to querying for commits. By default it starts in "highlight" mode as that is the least disruptive mode in that it won't continually ping the server for more commit statuses when a very tight query is made.

One thing I haven't done (yet?) is to automatically scroll to the next highlighted commit as that would take a much larger deep dive into how scrolling controllers work to get right. For now I'm adding the highlight mode and relying on the user to scroll down and look for matches.

Tests and goldens have been updated to match.


flar:
The new UI showing highlighted commits:
Screen Shot 2022-01-06 at 2 05 51 PM

dashboard/lib/logic/task_grid_filter.dart
```diff
@@ -7,9 +7,40 @@ import 'dart:collection';
...
+//
+//   @override
+//   bool get isDefault => _value == defaultValue;
+//
+//   @override
+//   void reset() => _value = defaultValue;
+//
+//   @override
+//   CommitFilterMode get value => _value;
+// }
```
keyonghan, line 42: do we need a cleanup?

dashboard/lib/logic/task_grid_filter.dart
```diff
@@ -7,9 +7,40 @@ import 'dart:collection';
...
 import '../logic/qualified_task.dart';
-import '../model/commit_status.pb.dart';
+import '../model/commit.pb.dart';
 import '../widgets/filter_property_sheet.dart';
 
+enum CommitFilterMode {
+  /// Show only those commits that match the relevant properties in the filter.
+  filter,
+
+  /// Highlight the commits that match the relevant properties (and scroll to the
```
keyonghan, line 17: will it automatically scroll to the first matching commit? What if the first match doesn't show up in the default page of rows?

flar: Actually I don't do that so I should remove the comment.

dashboard/lib/logic/task_grid_filter.dart
flar: done

dashboard/lib/logic/task_grid_filter.dart
flar: removed...

keyonghan: I am not sure if there is an easy way to remind users to keep scrolling to find matched `highlighted` rows if no highlight in the default page. Otherwise, people may simply conclude no match exists? Based on my personal experience, I at most scroll down twice.. But I believe this is still useful especially for recent commits query.

keyonghan (approved):
LGTM with test fix (format seems needed).

dashboard/lib/logic/task_grid_filter.dart
flar: I'm still thinking on that...

- How to collect the data needed to know when scrolling is needed so as not to fight their own manual scrolling
- How to indicate that we are loading more data to find a match
- Scroll to next/prev match capabilities
- Scroll to ? match when they modify the parameters (leave the position if there is match on screen? scroll to next or prev when there is none? etc)
- What is "on screen" is only known deep in the LatticeBody, but the matching is done in TaskGridState
- A count of how many matches there are perhaps?

The last item, a count of matches might help them know if they failed or if they just need to scroll. But, how to make it clear that it is only "count of matches among commits currently loaded"?

flar: So, what is the gold standard for how this works?
And what is the minimal standard for merging it?

Is just "they'll be highlighted if you hunt around and hopefully there was one visible while you are typing" above the minimum standard?

keyonghan: Existing logic on the filter dialog is filtering only matched commits, and the page keeps loading to find ones that haven't shown up.

I would suggest use the `filter` option (proposed in this PR) as the default, so that this will not affect anyone using the dialog as it would be the same as before.

If people want to see how neighborhood commits around the target commit perform, they can switch to the `highlight` option. I think we can still allow the page to keep auto-loading and then highlight the rows on the fly. This way, people will know if the target exists (with the `filter` option beforehand), and if it does and they want to check its neighborhood, they surely want to scroll down to find it (where the auto loading already starts the loading process).

Does it make sense to you?

flar: Changing the default would hide the lack of auto-scrolling where only users that choose the highlight option would experience it. That makes sense.

With respect to auto-loading new data - that's already handled for the filter option via the standard scrolling mechanism right now. For the highlight mode I think it would make sense to make it contingent on whether there are any matches? So it would keep loading until it found at least one match and then only load more if they scroll past that.

keyonghan: > For the highlight mode I think it would make sense to make it contingent on whether there are any matches? So it would keep loading until it found at least one match and then only load more if they scroll past that.

The contingent idea SGTM.
If the second match is weeks/months ago, it may load a while before a match is found. For this case, we may wait until the follow up feature - `filter with neighborhood`. 
If the second match is not far away, 1-2 manual scrolls should be enough.

Hixie:
@flar Looks like you have an LGTM, but let me know if you would still like me to review.

flar:
> @flar Looks like you have an LGTM, but let me know if you would still like me to review.

I feel that this isn't a good feature until I can figure out how to make the list auto-scroll to highlighted items.

Hixie:
@flag sounds good. Do you want the PR still open or is a local branch enough?

flar:
I'm fine with the local branch.

PR #1539, opened 2022-01-07
Fix deflake query and reference builds link

keyonghan:
This fixes: https://github.com/flutter/flutter/issues/96303 to use only recent staging builds when deflaking a test.

This PR also updates the reference link to check the flake status of a test: instead of pointing to the prod builds, the staging builds should be used and are recently available. ([example](https://data.corp.google.com/sites/flutter_infra_metrics_datasite/flutter_check_test_flakiness_status_dashboard/?p=BUILDER_NAME:%22Mac%20module_test_ios%22))


chunhtai (approved):
LGTM

PR #1543, opened 2022-01-11
Fix PR merge bot when checking if a clean revert happens

keyonghan:
This fixes https://github.com/flutter/flutter/issues/96478.

Switching the order of base and head commit (ref: https://docs.github.com/en/rest/reference/commits#compare-two-commits)

### DIFF
--- a/app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
+++ b/app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
@@ -107,7 +107,7 @@ class CheckForWaitingPullRequests extends ApiRequestHandler<Body> {
   ///
   /// By comparing the current commit with second TOT commit, an empty `files` in
   /// `GitHubComparison` validates a clean revert of TOT commit.
-  /// 
+  ///
   /// Note: [compareCommits] expects base commit first, and then head commit.
   Future<bool> isTOTRevert(
     String headSha,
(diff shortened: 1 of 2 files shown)


app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
```diff
@@ -107,17 +107,23 @@ class CheckForWaitingPullRequests extends ApiRequestHandler<Body> {
...
+  /// Note: [compareCommits] expects base commit first, and then head commit.
   Future<bool> isTOTRevert(
-    String commitSha,
+    String headSha,
     RepositorySlug slug,
   ) async {
     final GitHub github = await config.createGitHubClient(slug: slug);
     final RepositoryCommit secondTotCommit = await github.repositories.getCommit(slug, 'HEAD~');
-    log.info('Current commit is: $commitSha');
+    log.info('Current commit is: $headSha');
```
CaseyHillers, line 118: I missed this in the last PR, but can we assure that a PR has a single commit before ensuring this is a ToT revert?

For example, someone could land a PR with
Commit A - Random changes
Commit B - Revert of the the ToT commit
And this would be able to land

keyonghan: This should not be a problem. The `check for waiting pull request` API uses the latest commit in the PR when trying merging. In the given example: it will not land commit A, but will land commit B if it is the revert.

CaseyHillers: PRs can have multiple commits [docs](https://docs.github.com/en/rest/reference/pulls#:~:text=Pull%20Request%27s%20repository.-,commits,The%20API%20location%20of%20this%20Pull%20Request%27s%20commits.,-statuses)

You could upload a commit that deletes all of Cocoon, but then have the latest commit be a pure revert of ToT to get approvals on it.

For presubmit, the GitHub API doesn't store what potentially squashing all the commits into one would look like other than using the pull request refspec.

keyonghan: I am not sure if I understand correctly.
This is our graph query: https://github.com/flutter/cocoon/blob/main/app_dart/lib/src/request_handlers/check_for_waiting_pull_requests_queries.dart#L27. For each PR, we query only the last commit, which is the one we use to do comparison.

Back to the example, when we have two commits A and B, the graph query will return commit B. It will then merge the PR as it is the revert of the ToT commit.

CaseyHillers: I put up [this pr](https://github.com/flutter/cocoon/pull/1544) as an example.

Since we're only checking the last commit, anything else can sneak in.

CaseyHillers: I missed the `..` operator. This looks good to me.

CaseyHillers (approved):

PR #1548, opened 2022-01-13
[app_dart] Speed up autosubmit bot graphql queries

CaseyHillers:
## Summary

- GitHub limits GraphQL queries to <10 seconds
- flutter/flutter has grown the number of contributors it receives, and the endpoint is timing out very frequently
- Simplify some of the operations in the GraphQL code to the bare minimum of what Cocoon needs
  - Remove check runs and instead rely on the single checksuite conclusion (fixing a bug where we only checked the first 100 check runs)

## Issues

https://github.com/flutter/flutter/issues/96503
Fixes https://github.com/flutter/flutter/issues/96540

## Future work

- Remove cirrus graphql and rely on pulling all check suite conclusions

### DIFF
--- a/app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
+++ b/app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
@@ -132,6 +132,8 @@ class CheckForWaitingPullRequests extends ApiRequestHandler<Body> {
   ) async {
     final String labelName = config.waitingForTreeToGoGreenLabelName;
 
+    log.fine('Querying GitHub\'s GraphQL API...');
+    // GitHub's GraphQL has a timeout of 10 seconds.
     final QueryResult result = await client.query(
       QueryOptions(
         document: labeledPullRequestsWithReviewsQuery,
@@ -143,6 +145,7 @@ class CheckForWaitingPullRequests extends ApiRequestHandler<Body> {
         },
       ),
     );
+    log.fine('Queried GitHub\'s GraphQL API.');
 
     if (result.hasException) {
       log.severe(result.exception.toString());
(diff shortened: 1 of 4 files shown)


app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
```diff
@@ -132,6 +132,8 @@ class CheckForWaitingPullRequests extends ApiRequestHandler<Body> {
   ) async {
     final String labelName = config.waitingForTreeToGoGreenLabelName;
 
+    log.fine('Querying GitHub\'s GraphQL API...');
```
dnfield, line 135: nit: put the query in here and the result in the log below

app_dart/test/request_handlers/check_for_waiting_pull_requests_test.dart
```diff
@@ -307,67 +307,20 @@ void main() {
...
-        ],
+        checkSuiteConclusion: CheckRunConclusion.neutral,
       );
       flutterRepoPRs.add(prInProgress);
       await tester.get(handler);
       _verifyQueries();
       githubGraphQLClient.verifyMutations(<MutationOptions>[]);
     });
 
-    test('Does not merge PR with queued checks', () async {
```
dnfield, line 323: Why are we deleting these tests?

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
CaseyHillers: Thanks! I added a log for the result.data.

I'm skipping before as `QueryOptions` from package:graphql doesn't offer an override for toString() or toJson() so only the type would be printed. Additionally, the graphql query is mostly const (except for the variables inserted)

app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
```diff
@@ -304,7 +305,7 @@ class CheckForWaitingPullRequests extends ApiRequestHandler<Body> {
     String sha,
     Set<_FailureDetail> failures,
```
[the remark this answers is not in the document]
CaseyHillers: This is necessary for the statuses (such as missing tree status)

app_dart/test/request_handlers/check_for_waiting_pull_requests_test.dart
CaseyHillers: Most of these are duplicates as conclusions are simpler than scanning all the checks ourselves. I added an explicit test to ensure null checksuite conclusion does not submit the pr.

https://docs.github.com/en/rest/guides/getting-started-with-the-checks-api#about-check-suites

dnfield (approved):

keyonghan (approved):
LGTM

PR #1551, opened 2022-01-14
[app_dart] Extend backoff for github file content

CaseyHillers:
## Summary

- GitHub is taking longer to propagate to the CDN. Instead of doing 200ms, 400ms delays, we should do 3s, 6s to give the system more time to propagate
- Additionally, they're no longer reporting these issues as 500s, but as 404s

## Issues
Fixes https://github.com/flutter/flutter/issues/96503


dnfield (approved):

keyonghan (approved):
This should help reduce the .ci.yml failures. LGTM.

PR #1555, opened 2022-01-19
[app_dart] Shrink auto submission graphql query pagination

CaseyHillers:
## Summary

These are more realistic values, and are quick fixes to the timeouts.

According to GitHub, reducing page size from 100 to 20 on the PRs would result in a speed up of a factor of 5 (even when there isn't extra data).

## Issues

https://github.com/flutter/flutter/issues/96836

## Future work

If this doesn't fix timeout issues, we'll need to split this query into smaller parts.


godofredoc (approved):

CaseyHillers:
https://github.com/flutter/cocoon/pull/1555/commits/304e7618f9d9e7d81d81de88cef2865e57b1535d is from #1548 and was accidentally reverted. Relanding here as I experienced the flaky test.

PR #1556, opened 2022-01-19
Skip filing a new issue when the builder is bringup: true

keyonghan:
The a builder is already `bringup: true`, it means there is a tracking issue (either open or recently closed). For either case, there is no need to create a new bug.

A deflake PR has been skipped for this case, and this PR skips issue creation as well.

https://github.com/flutter/flutter/issues/96799

### DIFF
--- a/app_dart/lib/src/request_handlers/file_flaky_issue_and_pr.dart
+++ b/app_dart/lib/src/request_handlers/file_flaky_issue_and_pr.dart
@@ -113,7 +113,9 @@ class FileFlakyIssueAndPR extends ApiRequestHandler<Body> {
   }
 
   bool _shouldNotFileIssueAndPR(_BuilderDetail builderDetail, Issue? issue) {
-    // Don't create a new issue or deflake PR if the builder has been marked as flaky.
+    // Don't create a new issue or deflake PR using prod builds statuses if the builder has been marked as flaky.
+    // If the builder is `bringup: true`, but still hit flakes, a new bug will be filed in `/api/check_flaky_builders`
+    // based on staging builds statuses.
     if (builderDetail.isMarkedFlaky) {
       return true;
     }
--- a/app_dart/lib/src/request_handlers/file_flaky_issue_and_pr.dart
+++ b/app_dart/lib/src/request_handlers/file_flaky_issue_and_pr.dart
@@ -72,12 +72,7 @@ class FileFlakyIssueAndPR extends ApiRequestHandler<Body> {
   }) async {
     Issue? issue = builderDetail.existingIssue;
 
-    // Don't create a new issue or a new deflake PR if there is an open issue or a recent closed
-    // issue within kGracePeriodForClosedFlake days. It takes time for the flaky ratio to go
-    // down after the fix is merged.
-    if (issue != null &&
-        (issue.state != 'closed' ||
-            DateTime.now().difference(issue.closedAt!) <= const Duration(days: kGracePeriodForClosedFlake))) {
+    if (_shouldNotFileIssueAndPR(builderDetail, issue)) {
       return;
     }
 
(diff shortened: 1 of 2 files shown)


chunhtai:
>it means there is a tracking issue (either open or recently closed). For either case, there is no need to create a new bug.

shouldn't the script pick up this case and only skip if the issue either open or recently closed? I looked at the change, it looks like the code skips it unconditionally. If a test is marked bringup and there is no tracking issue, how does the owner track the flaky test?

keyonghan:
> > it means there is a tracking issue (either open or recently closed). For either case, there is no need to create a new bug.
> 
> shouldn't the script pick up this case and only skip if the issue either open or recently closed? I looked at the change, it looks like the code skips it unconditionally. If a test is marked bringup and there is no tracking issue, how does the owner track the flaky test?

Have been monitoring the process, and it is common that test owners assume a presumed fix and close the bug without validation, waiting for the bot to deflake and remove the flaky issue from .ci.yml. So the tracking bug remains in the .ci.yml, and the test remains as flaky.

On the other hand, `file flaky bug` API uses `prod` data to file bugs and PR to mark test as flaky. This is not correct for the `bringup: true` tests, which are running in staging.

chunhtai:
> Have been monitoring the process, and it is common that test owners assume a presumed fix and close the bug without validation, waiting for the bot to deflake and remove the flaky issue from .ci.yml. So the tracking bug remains in the .ci.yml, and the test remains as flaky.

yes in that case, shouldn't the bot file a new issue to notify the owner it has not been fixed?

> On the other hand, file flaky bug API uses prod data to file bugs and PR to mark test as flaky. This is not correct for the bringup: true tests, which are running in staging.

It sounds like the bot should know which db to query data from when the bringup is true?

keyonghan:
> > Have been monitoring the process, and it is common that test owners assume a presumed fix and close the bug without validation, waiting for the bot to deflake and remove the flaky issue from .ci.yml. So the tracking bug remains in the .ci.yml, and the test remains as flaky.
> 
> yes in that case, shouldn't the bot file a new issue to notify the owner it has not been fixed?
> 
> > On the other hand, file flaky bug API uses prod data to file bugs and PR to mark test as flaky. This is not correct for the bringup: true tests, which are running in staging.
> 
> It sounds like the bot should know which db to query data from when the bringup is true?

Thanks for the comments. Focused on avoiding bothering test owners.
I agree that we should re-purpose the change to use staging data when checking flake status of `bringup:true` tests. Closing the PR and will track in https://github.com/flutter/flutter/issues/96799.

keyonghan:
Reopen here:
1) New bug will be handled in https://github.com/flutter/cocoon/pull/1558
2) We need to stop creating bug from the prod runs.

app_dart/lib/src/request_handlers/file_flaky_issue_and_pr.dart
```diff
@@ -118,6 +112,24 @@ class FileFlakyIssueAndPR extends ApiRequestHandler<Body> {
     await gitHub.assignReviewer(slug, reviewer: prBuilder.pullRequestReviewer, pullRequestNumber: pullRequest.number);
   }
 
+  bool _shouldNotFileIssueAndPR(_BuilderDetail builderDetail, Issue? issue) {
+    // Don't create a new issue or deflake PR if the builder has been marked as flaky.
```
chunhtai, line 116: This comment seems inaccurate. Maybe mentions this is handled in other request_handler.

chunhtai (approved):
LGTM, except some nit

app_dart/lib/src/request_handlers/file_flaky_issue_and_pr.dart
keyonghan: Good catch! Updated.

PR #1558, opened 2022-01-21
File flaky bug for `bringup:true` builders if no open tracking bug

keyonghan:
Existing workflow is like:
1) flake bot file a flaky bug & deflake the builder
2) bug is closed due to a fix or decreased flake number
3) flake bot will file a new flaky bug still based on prod runs of this builder (INCORRECT)

This PR does:
1) file a new flaky bug based on staging runs (if there exists any flake)
2) create sharable function `fileFlakyIssue` in utils so different APIs can access
3) rename `deflake_flaky_builders` to `check_flaky_builders`.

https://github.com/flutter/flutter/issues/96799

### DIFF
--- a/app_dart/bin/server.dart
+++ b/app_dart/bin/server.dart
@@ -59,7 +59,7 @@ Future<void> main() async {
       ///
       /// Response: Status 200 OK
       '/api/check-waiting-pull-requests': CheckForWaitingPullRequests(config, authProvider),
-      '/api/deflake_flaky_builders': DeflakeFlakyBuilders(config, authProvider),
+      '/api/deflake_flaky_builders': CheckFlakyBuilders(config, authProvider),
       '/api/file_flaky_issue_and_pr': FileFlakyIssueAndPR(config, authProvider),
       '/api/flush-cache': FlushCache(
         config,
--- a/app_dart/bin/server.dart
+++ b/app_dart/bin/server.dart
@@ -59,7 +59,7 @@ Future<void> main() async {
       ///
       /// Response: Status 200 OK
       '/api/check-waiting-pull-requests': CheckForWaitingPullRequests(config, authProvider),
-      '/api/deflake_flaky_builders': CheckFlakyBuilders(config, authProvider),
+      '/api/check_flaky_builders': CheckFlakyBuilders(config, authProvider),
       '/api/file_flaky_issue_and_pr': FileFlakyIssueAndPR(config, authProvider),
       '/api/flush-cache': FlushCache(
         config,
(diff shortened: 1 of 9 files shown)


app_dart/lib/src/request_handlers/check_flaky_builders.dart
```diff
@@ -52,20 +52,44 @@ class DeflakeFlakyBuilders extends ApiRequestHandler<Body> {
...
       'Status': 'success',
     });
   }
 
+  /// A new issue should be filed for staging builders if
+  ///   1) there is any flake in recent runs
+  ///   2) there is no open flaky bug tracking the flake
+  bool _shouldFileIssue(List<BuilderRecord> builderRecords, _BuilderInfo info) {
+    final bool noExistingOpenIssue =
+        info.existingIssue == null || info.existingIssue != null && info.existingIssue!.isClosed;
```
chunhtai, line 89: we may want to also check for grace period?

app_dart/lib/src/request_handlers/check_flaky_builders.dart
```diff
@@ -52,20 +52,44 @@ class DeflakeFlakyBuilders extends ApiRequestHandler<Body> {
...
+    return builderRecords.any((BuilderRecord record) => record.isFlaky) && noExistingOpenIssue;
```
chunhtai, line 90: maybe flip the order to short-circuit the buildRecords check?

app_dart/lib/src/request_handlers/check_flaky_builders.dart
keyonghan: Yeah, we should check that. Added.

app_dart/lib/src/request_handlers/check_flaky_builders.dart
keyonghan: Done.

app_dart/bin/server.dart
```diff
@@ -59,7 +59,7 @@ Future<void> main() async {
       ///
       /// Response: Status 200 OK
       '/api/check-waiting-pull-requests': CheckForWaitingPullRequests(config, authProvider),
-      '/api/deflake_flaky_builders': DeflakeFlakyBuilders(config, authProvider),
+      '/api/check_flaky_builders': CheckFlakyBuilders(config, authProvider),
```
chunhtai, line 62: we need to update the cron job too

chunhtai (approved):
LGTM

app_dart/bin/server.dart
keyonghan: Will update when this is in.

PR #1559, opened 2022-01-21
[scheduler] Idempotent subscriptions + scheduler request subscription

CaseyHillers:
## Summary

- Only process pubsub messages once (based on topic and message id)
- Add subscription for scheduling buildbucket requests with `BatchRequest`
  - This gives flexibility to pass RPCs beyond ScheduleBuild as there's no definitive way of knowing the passed model
 

## Issues

https://github.com/flutter/flutter/issues/86948
https://github.com/flutter/flutter/issues/84450

## Future work

- Publish messages to scheduler-requests pubsub


CaseyHillers:
Friendly ping

app_dart/lib/src/request_handlers/scheduler/request.dart
```diff
@@ -0,0 +1,63 @@
+// Copyright 2019 The Flutter Authors. All rights reserved.
```
godofredoc, line 1: Should we call this file scheduler_request_subscription? request is too generic and it does not say much about the real purpose.

app_dart/lib/src/request_handling/subscription_handler.dart
```diff
@@ -19,23 +25,124 @@ import 'body.dart';
...
+    }
+
+    if (envelope == null) {
+      throw const BadRequestException('Failed to get message');
+    }
+    log.finer(envelope.toJson());
+
+    final String messageId = envelope.message!.messageId!;
+
+    final Uint8List? messageLock = await cache.getOrCreate(topicName, messageId);
```
godofredoc, line 109: Do we need an implementation that allows to specify a TTL? ideally the TTL should match the configuration in the pub/sub subscription

app_dart/test/request_handlers/scheduler/request_test.dart
```diff
@@ -0,0 +1,51 @@
...
+import 'package:test/test.dart';
+
+import '../../src/datastore/fake_config.dart';
+import '../../src/request_handling/fake_authentication.dart';
+import '../../src/request_handling/fake_http.dart';
+import '../../src/request_handling/subscription_tester.dart';
+import '../../src/utilities/mocks.dart';
+
+void main() {
+  late SchedulerRequestSubscription handler;
```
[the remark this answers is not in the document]
CaseyHillers: Done!

app_dart/lib/src/request_handling/subscription_handler.dart
godofredoc: No, I'm referring for the TTL for the item in the cache. One week seems like too for messages that will expire in 24h with the default pubsub configurations. It would be great if we can add those optimizations now that we have the context fresh.

app_dart/bin/server.dart
```diff
@@ -111,6 +113,12 @@ Future<void> main() async {
         authProvider,
         scheduler,
       ),
+      '/api/scheduler/requests': SchedulerRequestSubscription(
```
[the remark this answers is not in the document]
godofredoc: request-subscription is still very generic, it can apply to any request handler dealing with pubsub. Can we use something the implies what the handler is actually doing? e.g build-request-subscription if you envision multiple subscription handlers for the scheduler.

app_dart/lib/src/request_handling/subscription_handler.dart
CaseyHillers: I'll decrease this to 1 day, but a message id is guaranteed to be unique over the lifetime of a topic. By default that allows for 64^19 messages, and PubSub will dynamically increase the range to guarantee this.

app_dart/bin/server.dart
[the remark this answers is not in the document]
CaseyHillers: Updated to `/api/scheduler/batch-request-subscription` as this schedules `BatchRequest` to BuildBucket

godofredoc (approved):

PR #1564, opened 2022-01-25
[scheduler] Gracefully fail runIf

CaseyHillers:
Instead of failing, this should run all targets.

Fixes https://github.com/flutter/flutter/issues/97152


app_dart/lib/src/service/scheduler.dart
```diff
@@ -359,7 +359,15 @@ class Scheduler {
 
     // Filter builders based on the PR diff
```
yusuf-goog, line 360: Add to function docstring to what happens based on files being returned from githubService?

yusuf-goog (approved):
LGTM, with request for expanded docstring.

PR #1572, opened 2022-01-28
[app_dart] Generalize pubsub authentication

CaseyHillers:
This is a quick fix to make writing subscriptions easier. I fell into this when I used the wrong auth for testing the scheduler batch request subscription.

## Summary
- Default `SubscriptionHandler` to use the `PubsubAuthenticationProvider`
- Rename `LuciPubsubAuthenticationProvider` to `PubsubAuthenticationProvider` and fix related names and docs

https://github.com/flutter/flutter/issues/86948 - Enabling pubsub in Cocoon

### DIFF
--- a/app_dart/lib/cocoon_service.dart
+++ b/app_dart/lib/cocoon_service.dart
@@ -31,7 +31,7 @@ export 'src/request_handlers/vacuum_github_commits.dart';
 export 'src/request_handling/authentication.dart';
 export 'src/request_handling/body.dart';
 export 'src/request_handling/cache_request_handler.dart';
-export 'src/request_handling/luci_pubsub_authentication.dart';
+export 'src/request_handling/pubsub_authentication.dart';
 export 'src/request_handling/request_handler.dart';
 export 'src/request_handling/static_file_handler.dart';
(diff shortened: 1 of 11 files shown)


app_dart/lib/cocoon_service.dart
```diff
@@ -31,7 +31,7 @@ export 'src/request_handlers/vacuum_github_commits.dart';
 export 'src/request_handling/authentication.dart';
 export 'src/request_handling/body.dart';
 export 'src/request_handling/cache_request_handler.dart';
-export 'src/request_handling/luci_pubsub_authentication.dart';
+export 'src/request_handling/pubsub_authentication.dart';
```
yusuf-goog, line 34: Do we want to make this name even more generic (e.g. service_account auth) since its not specifically tied to pubsub service?

CaseyHillers: No, because in the future I want to rename https://github.com/flutter/cocoon/blob/main/app_dart/lib/src/request_handling/swarming_authentication.dart to that

yusuf-goog (approved):

PR #1578, opened 2022-02-02
[scheduler] Trigger cocoon targets on postsubmit

CaseyHillers:
## Summary
- This enables targets with `scheduler: cocoon` to be scheduled via Cocoon (instead of LUCI)
  - This is a no-op PR as it requires updating the ci.yaml to enable

## Issues

https://github.com/flutter/flutter/issues/95037


app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -527,6 +533,70 @@ class LuciBuildService {
...
+
+  /// Creates a [ScheduleBuildRequest] for [target] and [task] against [commit].
+  ScheduleBuildRequest _createPostsubmitScheduleBuild({
+    required Commit commit,
+    required Target target,
+    required Task task,
+  }) {
+    final Map<String, List<String>> tags = <String, List<String>>{
+      'buildset': <String>[
+        'commit/git/${commit.sha}',
```
yusuf-goog, line 565: should we have some helper functions to generate these different git related strings?

yusuf-goog (approved):
LGTM with one comment

app_dart/lib/src/service/luci_build_service.dart
CaseyHillers: I want to overall simplify `LuciBuildService` (there's a lot of redundancy within it that I want to move into Scheduler). I filed https://github.com/flutter/flutter/issues/97687 for tracking this.

My assumption is it will be more obvious to me in the coming weeks as I work on this as to what the right way is to de-duplicate this service.

PR #1587, opened 2022-02-04
[dashboard] Merge columns for luci and cocoon tasks

CaseyHillers:
Tasks with `scheduler:cocoon` and `scheduler:luci` are effectively the same. They are managed by the ci.yaml tooling, but change their triggering policy and results processing.

After starting the migration, there's a few issues that shouldn't show:

https://flutter-dashboard.appspot.com/#/build?repo=cocoon&branch=master

Instead, these duplicate columns shouldn't exist.

## Issues

https://github.com/flutter/flutter/issues/92300 - Ensure a smooth migration (no disruption in the dashboard)

### DIFF
--- a/dashboard/lib/logic/qualified_task.dart
+++ b/dashboard/lib/logic/qualified_task.dart
@@ -59,11 +59,23 @@ class QualifiedTask {
     if (other.runtimeType != runtimeType) {
       return false;
     }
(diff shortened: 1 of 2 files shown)


keyonghan (approved):
LGTM

PR #1588, opened 2022-02-04
[dashboard] randomness to queue time shown in dev dashboard

XilaiZhang:
This change aims to add randomness to the queue times displayed in dev dashboard. Previously the display of queue times always show as 0 minutes in dev dashboard.


dashboard/lib/service/dev_cocoon.dart
```diff
@@ -274,10 +274,11 @@ class DevelopmentCocoonService implements CocoonService {
     final int minAttempts = _minAttempts[status];
     final int maxAttempts = _maxAttempts[status];
     final int attempts = minAttempts + random.nextInt(maxAttempts - minAttempts + 1);
+    final int randomQueueTime = random.nextInt(1000 * 60 * 15);
```
CaseyHillers, line 277: Generally, avoid random in fakes or tests. The reason we use it here is for the weights which we intentionally don't want to be deterministic.

dashboard/lib/service/dev_cocoon.dart
```diff
@@ -274,10 +274,11 @@ class DevelopmentCocoonService implements CocoonService {
...
     final Task task = Task()
       ..createTimestamp = Int64(commitTimestamp + index)
-      ..startTimestamp = Int64(commitTimestamp + index + 10000)
-      ..endTimestamp = Int64(commitTimestamp + index + 10000 + random.nextInt(1000 * 60 * 15))
+      ..startTimestamp = Int64(commitTimestamp + index + randomQueueTime)
+      ..endTimestamp = Int64(commitTimestamp + index + randomQueueTime + random.nextInt(1000 * 60 * 15))
```
CaseyHillers, line 281: Another approach would be to do 

```
      ..startTimestamp = Int64(commitTimestamp + (index * 1000 * 60))
```

XilaiZhang: Thanks for the reply! Would it be better to change random.nextInt(1000 * 60 * 15) in endTimestamp as well? like would having `..endTimestamp = Int64(commitTimestamp + random.nextInt(1000 * 60 * 15) + random.nextInt(1000 * 60 * 15))` be better, or is endTimestamp encouraged to have more randomness encoded?

CaseyHillers: Yes, let's remove random from endTimestamp as well. The more deterministic fake data is, the better. Otherwise, it's an easy way to introduce flakes to our CI.

CaseyHillers (approved):
flutter-lgtm

CaseyHillers:
I'm not sure how this changed the ordering of the build dashboard, but LGTM (I suspect it relates to the `==` operator using timestamps). Otherwise, LGTM

XilaiZhang:
Thanks for the guidances and approval! I thought the variable $index is random so the ordering is random as well? or maybe my understanding is off. Let me double check

CaseyHillers:
`index` should be an indicator in where in the for-loop it was generated (which may not be respective to where it is in the task grid due to the weight algorithm).

PR #1590, opened 2022-02-04
[dashboard] Fix race condition in updating commit statuses

CaseyHillers:
You can reproduce this when using the settings dialogs to switch repos when first loading Cocoon. It'll show data for the other repo, which is incorrect.


keyonghan (approved):
Ha, just hit the issue, and was about to file a bug. Thanks for the fix.

PR #1592, opened 2022-02-04
[app_dart] Quick fix to base64 decode messages in scheduler requests

CaseyHillers:
I filed https://github.com/flutter/flutter/issues/97828 as the tech-debt clean up to make this API easier to use in the future.


XilaiZhang (approved):
image

XilaiZhang:
LGTM! not sure if i am qualified enough to review, but i just googled all these functions and the usages are legit

PR #1594, opened 2022-02-07
[dashboard] practicing on 87661, add sign in, error message and stacktrace

XilaiZhang:
This is a starter project that aims to address the issues mentioned in https://github.com/flutter/flutter/issues/87661 . 
Currently this commit shows more specific notifications when asking the user to sign in, before the user is allowed to perform a [rerun] operation. This commit also presents error and stack trace information to users in the form of snack bar message, if the rerun request gives a bad status code or throws an exception.

TODO(xilaizhang): 
As addressed in 87661, ideally we would write the error and stack trace to a file that's publicly viewable via HTTP, then linking to that location with a "more details" link in the snackbar notification. I am not sure if there is a publicly hosted url / pastebin etc. within flutter org we can use for this issue. Can someone provide references so that I can address in upcoming commits? Thanks a lot!

Warnings thrown while testing:
For some of tests warnings in similar formats to `../../../.pub-cache/hosted/pub.dartlang.org/url_launcher_platform_interface-2.0.3/lib/link.dart:91:22: Warning: Operand of null-aware operation '?.' has type 'WidgetsBinding' which excludes null.` were thrown when running `flutter test`.

Release dashboard errors shown in local IDE:
I am currently on the latest version of master branch of flutter, and have done plenty `flutter pub cache repair` / `flutter upgrade` / `flutter packages get` etc. Some of the errors (while not affecting passing all tests) still show up: 1. `Undefined name 'KReleaseIncrements'`. shows up in release_dashboard/lib/widgets/create_release_substeps.dart  2. `Missing concrete implementation of 'NextContext.pushWorkingBranch'.` shows up in release_dashboard/test/fakes/fake_next_context.dart.


dashboard/lib/service/appengine_cocoon.dart
```diff
@@ -146,23 +147,36 @@ class AppEngineCocoonService implements CocoonService {
...
+          }));
+      final bool success = response.statusCode == HttpStatus.ok;
+      return success
+          ? const CocoonResponse<bool>.data(true)
+          : CocoonResponse<bool>.error('bad status code is ' +
+              response.statusCode.toString() +
+              ', response body is ' +
+              response.body.toString() +
+              '\n');
+    } catch (error, stacktrace) {
```
CaseyHillers, line 177: Catching errors should be specific (we should specify the type like `HttpException error`. https://pub.dev/documentation/http/latest/http/post.html indicates post will not return exceptions, so what are we trying to catch?

dashboard/lib/service/google_authentication.dart
```diff
@@ -73,7 +73,6 @@ class GoogleSignInService extends ChangeNotifier {
     }
 
     final String idToken = await user?.authentication?.then((GoogleSignInAuthentication key) => key.idToken);
-    assert(idToken != null && idToken.isNotEmpty);
```
CaseyHillers, line 76: Why was this removed?

dashboard/lib/widgets/task_overlay.dart
```diff
@@ -316,8 +317,12 @@ class TaskOverlayContents extends StatelessWidget {
...
   Future<void> _rerunTask() async {
-    final bool success = await buildState.rerunTask(task);
-    final Text snackBarText = success ? const Text(rerunSuccessMessage) : const Text(rerunErrorMessage);
+    final CocoonResponse<bool> rerunResponse = await buildState.rerunTask(task);
+    final Text snackBarText = rerunResponse.data == null
+        ? Text(rerunResponse.error)
+        : rerunResponse.data
+            ? const Text(rerunSuccessMessage)
+            : const Text(rerunErrorMessage);
     showSnackBarCallback(
```
CaseyHillers, line 326: We'll want to update this so the snackbar is red for errors. This can probably hook into an error brook instead https://github.com/flutter/cocoon/blob/f5d0471044e3d66d2e6e1aabbbeeecff5c0c1f5a/dashboard/lib/widgets/error_brook_watcher.dart

dashboard/test/service/appengine_cocoon_test.dart
```diff
@@ -2,6 +2,8 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'dart:io';
```
CaseyHillers, line 5: dart:io shouldn't be added (it isn't supported for the web platform)

dashboard/lib/logic/error_to_string.dart
```diff
@@ -0,0 +1,12 @@
+// Copyright 2014 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/// Converts the error and stack trace objects to strings for display.
+String errorToString(Object error, StackTrace stackTrace) {
```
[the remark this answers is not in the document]
XilaiZhang: Oh this file I copied entirely from under the release_dashboard (cocoon/release_dashboard/lib/logic/error_to_string.dart), Might not be correct, based on my understanding I think it has the benefit of incorporating stack trace along side error?

dashboard/lib/service/google_authentication.dart
XilaiZhang: just based on my personal testing in scenarios like, if I click rerun and close the pop up sign in dialogue, then the system will throw an exception due to this implementation, and a generic error information will be shown. I removed it so that I can control sign in related error message display in other functions outside google auth. Maybe there is a better way to do it instead of removing it?

dashboard/lib/service/appengine_cocoon.dart
XilaiZhang: got it. I think in 87661 discussion it was mentioned that we want to retrieve stack trace as well. I thought we would get stack trace if we encounter an exception. Umm maybe my understanding is off, so if exceptions are not thrown then we wouldn't get any stack trace?

dashboard/lib/widgets/task_overlay.dart
XilaiZhang: for sure. I will get this part incorporated.

dashboard/lib/service/google_authentication.dart
CaseyHillers: If you're running this on web, try using `flutter run -d chrome --web-port=8080`. The Google Sign In app will only run on that port for local runs

dashboard/lib/service/appengine_cocoon.dart
CaseyHillers: There's no underlying issues we're not catching. The issue is we don't surface this information (so moving this from `bool` to `CocoonResponse<bool` is enough. Additionally, we'll need to add a description of the issue if there is an error

dashboard/lib/service/google_authentication.dart
XilaiZhang: umm yeah i did that and on https://flutter-dashboard.appspot.com/#/build, if i click on a box for rerun and close the sign in dialogue, it shows generic error (which from my understanding is caused by exception thrown). On local web I tested by printing out the exception. I think it is rare that people would close a sign in dialogue after it pops up, umm are you able to reproduce the scenario i described?

dashboard/lib/service/appengine_cocoon.dart
XilaiZhang: Thanks for explaining! So essentially we don't need exceptions, and merely describing the various kinds of bad http return code would be enough?

dashboard/lib/logic/error_to_string.dart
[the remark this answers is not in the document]
XilaiZhang: Thanks for providing background on this file! Will discard this part and adopt better topology. And I guess all the IDE errors reported from release dashboard I could safely ignore

keyonghan (commented):
One related nice to have, not required in this PR though. It would be good to give a warning message when no retry is allowed. 
Check `/api/reset-prod-task/`. When a success run occurs, no further retry would be triggered. But the frontend shows nothing and people have no idea what happens.

XilaiZhang:
> One related nice to have, not required in this PR though. It would be good to give a warning message when no retry is allowed. Check `/api/reset-prod-task/`. When a success run occurs, no further retry would be triggered. But the frontend shows nothing and people have no idea what happens.

Sure, i guess that can be part of errorbrook as well. I am still looking at errorbrook and trying to understand how it works.

CaseyHillers:
> Sure, i guess that can be part of errorbrook as well. I am still looking at errorbrook and trying to understand how it works.

https://www.youtube.com/watch?v=nQBpOIHE4eE (and generally the whole series on async coding) to be one of the best guides on this type of logic.

XilaiZhang:
> > Sure, i guess that can be part of errorbrook as well. I am still looking at errorbrook and trying to understand how it works.
> 
> https://www.youtube.com/watch?v=nQBpOIHE4eE (and generally the whole series on async coding) to be one of the best guides on this type of logic.

Cool thanks for the resources!

XilaiZhang:
This newer commit is based on my understanding of how errorBrook works (correctness is not guaranteed). I moved the part of showing rerun failure, from previously displayed by regular task bar, to be handled by errorBrook which is triggered in state/build.dart. I removed the test for checking failed rerun display in task_overlay_test, because I think the error snack bar display is already checked in build_dashboard_page_test. And since I am not displaying error in regular snack bar I can't check it through task_overlay. According to the effect I observed on dev cocoon dashboard I think the errorBrook works correctly.

XilaiZhang:
As for the TODO section: (which is “writing the error and stack trace to a file that's publicly viewable via HTTP, then linking to that location with a 'more details' link in the snack bar notification.”)

As Casey suggested earlier we can use a github gist to be the publicly viewable HTTP. And my understanding is that in order to use the github service we need a config initialization, and we register the service in a wrapped request_handler and triggers it through cron jobs. As Keyong pointed out that since the error we get from rerunning the task would  fit in one line of error message. (given the assumption that we only return a bad status code from failed rerun request) So maybe it's not worth the effort and we can simply print a single line of error message?

dashboard/lib/service/appengine_cocoon.dart
```diff
@@ -146,23 +146,33 @@ class AppEngineCocoonService implements CocoonService {
   }
 
   @override
-  Future<bool> rerunTask(Task task, String idToken) async {
-    assert(idToken != null);
+  Future<CocoonResponse<bool>> rerunTask(Task task, String idToken) async {
+    if (idToken == null || idToken.isEmpty) {
+      return const CocoonResponse<bool>.error('Need to sign in to perform rerun operation');
```
CaseyHillers, line 151:
```suggestion
      return const CocoonResponse<bool>.error('Sign in to trigger reruns');
```

Error messages should be actionable. https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#error-messages-should-be-useful

dashboard/lib/service/appengine_cocoon.dart
```diff
@@ -146,23 +146,33 @@ class AppEngineCocoonService implements CocoonService {
...
+    }
 
     final QualifiedTask qualifiedTask = QualifiedTask.fromTask(task);
     assert(qualifiedTask.isLuci);
 
-    /// This endpoint only returns a status code.
+    //this endpoint returns CocoonResponse<bool> which reflects whether the api
+    //call returns success, or returns bad status code.
```
CaseyHillers, line 158: `apiEndpoint` is for REST calls to Cocoon's backend. Since we didn't change anything in app_dart, this comment doesn't need any updates

dashboard/lib/service/appengine_cocoon.dart
```diff
@@ -146,23 +146,33 @@ class AppEngineCocoonService implements CocoonService {
...
         },
         body: jsonEncode(<String, String>{
           'Key': task.key.child.name,
         }));
-
-    return response.statusCode == HttpStatus.ok;
+    final bool success = response.statusCode == HttpStatus.ok;
+    return success
+        ? const CocoonResponse<bool>.data(true)
+        : CocoonResponse<bool>.error('bad status code is ' +
```
CaseyHillers, line 171: ditto about useful error messages

For now, let's use the error message being sent from the backend

https://cs.opensource.google/flutter/cocoon/+/main:app_dart/lib/src/request_handlers/reset_prod_task.dart

dashboard/lib/service/appengine_cocoon.dart
```diff
@@ -146,23 +146,33 @@ class AppEngineCocoonService implements CocoonService {
...
+            response.statusCode.toString() +
+            ', response body is ' +
+            response.body.toString() +
+            '\n');
```
CaseyHillers, line 175: In Dart, you can string interpolation to get a nicer result of:

`HTTP Code: ${response.statusCode}, ${response.body}'`

https://dart.dev/guides/language/language-tour#:~:text=String%20interpolation%3A%20including%20a%20variable%20or%20expression%E2%80%99s%20string%20equivalent%20inside%20of%20a%20string%20literal.%20For%20more%20information%2C%20see%20Strings.

dashboard/lib/service/google_authentication.dart
CaseyHillers: This looks to be caused by the fake data (we could talk a lot more about this but tl;dr we can support prod data only on the desktop platforms, but you can't sign in as the plugin doesn't have desktop support, and web we can support prod sign in, but can't support prod data due to CORS)

dashboard/test/widgets/task_overlay_test.dart
```diff
@@ -373,51 +373,8 @@ void main() {
     expect(find.text(TaskOverlayContents.rerunSuccessMessage), findsNothing);
   });
 
-  testWidgets('failed rerun shows error snackbar message', (WidgetTester tester) async {
```
CaseyHillers, line 376: We should then be migrating this test to validate error brook surfaces the error

dashboard/lib/service/dev_cocoon.dart
```diff
@@ -135,8 +135,8 @@ class DevelopmentCocoonService implements CocoonService {
   }
 
   @override
-  Future<bool> rerunTask(Task task, String accessToken) async {
-    return false;
+  Future<CocoonResponse<bool>> rerunTask(Task task, String accessToken) async {
+    return const CocoonResponse<bool>.error('view errorBrook effect from dev cocoon');
```
CaseyHillers, line 139: I don't understand this error message.

Why not just `Unable to retry against fake data. Try building the app to use prod data.`

dashboard/test/widgets/task_overlay_test.dart
XilaiZhang: Thanks a lot for taking the time to review the code! I tried various approaches but was not very successful in getting failed rerun tests to work. My current approach is (https://gist.github.com/XilaiZhang/2a962269c010d1650f4915de3f87f355): I tried to wrap the test widget with a errorBrook inside a valueProvider of buildState, but there is overflow and the super.disposal() from errorBrook throws an exception. I have tried other approaches such as borrowing code from build_dashboard_page.test but they create more errors. Could you provide some guidance on how I should get errorBrooks incorporated into my test? Thank you!

dashboard/lib/service/appengine_cocoon.dart
XilaiZhang: Thanks for the instructions on error messages! I will use more verbs, make them shorter, and explain more accurately.

dashboard/lib/service/appengine_cocoon.dart
XilaiZhang: I might be wrong, my understanding is that the error message sent from this handler is automatically encoded in response.body, so that we can retrieve from response.body? I am not 100 percent clear on the underlying mechanisms of how the request_handlers are registered and managed

CaseyHillers: This sgtm! This is to acknowledge that we'll need a follow up PR in the backend to make the error messages there more useful (this PR will only surface the messages, not make them useful)

XilaiZhang: Oh Casey it's so kind of you to take time to review while still on jury duty. Feel free to reply when you are more convenient. And thanks a lot for the guidance!

dashboard/test/widgets/task_overlay_test.dart
CaseyHillers: I added some comments. I recommend taking a look at the error brook tests to see how they work.

XilaiZhang:
Awesome, major technical issues has been solved and we now have error brook. I will clean up other issues soon and I am expecting to close this saga soon!

CaseyHillers:
> Awesome, major technical issues has been solved and we now have error brook. I will clean up other issues soon and I am expecting to close this saga soon!

Was it the null safety issues? Or was there an env issue?

XilaiZhang:
> > Awesome, major technical issues has been solved and we now have error brook. I will clean up other issues soon and I am expecting to close this saga soon!
> 
> Was it the null safety issues? Or was there an env issue?

I think it's some of the recent commits that miraculously fixed it, probably not null safety because I didn't apply the migration brach to this branch

dashboard/lib/service/appengine_cocoon.dart
```diff
@@ -167,12 +167,10 @@ class AppEngineCocoonService implements CocoonService {
         }));
     final bool success = response.statusCode == HttpStatus.ok;
```
CaseyHillers, line 168: Prefer not nesting code, as it makes it harder to read.

Additionally, assume if HTTP 200 was returned that there was no error

```dart
if (response.statusCode == HttpStatus.ok) {
  return const CocoonResponse<bool>.data(true);
}

return CocoonResponse<bool>.error('HTTP ${response.statusCode}: ${response.body}');
```

CaseyHillers:
Let's wait for https://github.com/flutter/cocoon/pull/1616 to merge, and then rebase this with the nnbd changes (as it'll be less time consuming than the reverse)

dashboard/lib/service/appengine_cocoon.dart
XilaiZhang: For sure, I will rewrite the logic. Just a few contexts on this block, it is hoping to addresses the feedback that if a task has already succeeded and could not be rerun, instead of showing nothing or other notification, we now prompt a notification to say succeeded task cannot rerun. It checks that aside from 200 code, whether the returned response body is empty. If it's empty then it's a task that has already succeeded and could not be rerun.

CaseyHillers: I don't think you can rely on the response. For example, in https://github.com/flutter/cocoon/pull/1610 is removing that logic, and it'll return an error when trying to rerun a task that shouldn't be rerun

XilaiZhang: umm I might be wrong, but I thought 1610 kept L76-83 (https://github.com/flutter/cocoon/pull/1610/files#diff-8b18660d0517d62995b30d9277a6cd6601f428ccd5def5d4535f668cd6ba8cb3L76-L83), and my understanding is that this part will be responsible for returning an empty response body if task has already succeeded?

CaseyHillers: No, it'll return an exception in that case (it's being moved in checkRerunBuilder)

XilaiZhang:
In retrospect it would probably be way easier to start all over again, than resolving all these merge conflicts. But anyways here we go, an end to another saga!

CaseyHillers (approved):
flutter-lgtm

PR #1595, opened 2022-02-07
[dashboard] Update USER_GUIDE.md

CaseyHillers:
Refresh this doc with the latest information

### DIFF
--- a/dashboard/USER_GUIDE.md
+++ b/dashboard/USER_GUIDE.md
@@ -4,7 +4,7 @@
 
 The build page is accessible at https://flutter-dashboard.appspot.com/#/build.
 
-Build statuses of commits on this page are sync'ed with the flutter/flutter repo.
+Build statuses of commits on this page are sync'ed with most repos in https://github.com/flutter
 
 ### Tree Closures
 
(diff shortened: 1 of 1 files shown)


dashboard/USER_GUIDE.md
```diff
@@ -13,36 +13,23 @@ at least one test has failed against the top of tree and flutter/flutter is
...
-
-When the DeviceLab migrates to LUCI, logs will be available without sign in.
 
 ### Why is a task stuck on "new task" status?
 
 The dashboard aggregates build results from multiple build environments,
-including Cirrus, LUCI, and DeviceLab. DeviceLab and Cirrus test every commit
-that goes into the `master` branch. However, LUCI may skip commits when they
-come in too fast.
+including Cirrus, LUCI, and DeviceLab. Due to limited capacity to run tests on
```
keyonghan, line 23: DeviceLab is now also LUCI.

dashboard/USER_GUIDE.md
```diff
@@ -13,36 +13,23 @@ at least one test has failed against the top of tree and flutter/flutter is
...
 provided from a more recent commit.
 
-DeviceLab tests are eventually run against every commit to ensure benchmark
-data is collected from every commit. This helps with triaging where performance
-regressions started.
-
 ### Why are some tasks outlined instead of solid?
 
-Outlined tasks indicate a task that is running experimentally. This can be for
+Outlined tasks indicate a task that is running as bringup. This can be for
```
keyonghan, line 32: Nit: a quick link to bringup.

keyonghan (approved):
LGTM with nits.

dashboard/USER_GUIDE.md
CaseyHillers: I removed this as there's info inside the legend of the app. Additionally, I linked to the general ci yaml docs explaining that it generates this dashboard

dashboard/USER_GUIDE.md
CaseyHillers: I understand, but I think it takes away the the complexity here (we have multiple ways of validating tests, and each of them can be different). I could split this into LUCI (VM) and LUCI (DeviceLab), but that seems verbose. WDYT?

keyonghan: Actually we do not have any cirrus on post submit tests for different repos. Could we rephrase it to something like below? 
```
The dashboard aggregates build results from LUCI, including hostonly and devicelab.
```

CaseyHillers: My understanding is we still want to support cirrus for post-submit, but we don't actively use it today - See https://cs.opensource.google/flutter/cocoon/+/main:app_dart/lib/src/request_handlers/refresh_cirrus_status.dart and there's quite a few other places in the code that support it

keyonghan: These are for presubmit tests. @godofredoc is it expected we will not support cirrus post-submit tests in our CI?

godofredoc: Packages and plugins are still using cirrus for post-submit and they have no plans on migrating everything to LUCI. If we can convince them to move all the post-submit testing to LUCI then we can safely assume cirrus will be running only on presubmit.

CaseyHillers: Thanks Godofredo. With that said, I'm going to leave this as is. SG Keyong?

keyonghan: SGTM!

PR #1598, opened 2022-02-10
Add isBenchmark to task

KristinBi:
Added isBenchmark to the Task class to show whether this task is benchmark.

Fixed the issue: https://github.com/flutter/flutter/issues/98215

### DIFF
--- a/app_dart/lib/src/model/appengine/task.dart
+++ b/app_dart/lib/src/model/appengine/task.dart
@@ -42,7 +42,7 @@ class Task extends Model<int> {
     this.buildNumberList,
     this.builderName,
     this.luciBucket,
-    this.isBenchmark,
+    this.isBenchmark = false,
     String? status,
   }) : _status = status {
     if (status != null && !legalStatusValues.contains(status)) {
@@ -71,7 +71,6 @@ class Task extends Model<int> {
       stageName: 'chromebot',
       status: Task.statusNew,
       timeoutInMinutes: 0,
-      isBenchmark: builder.benchmark,
     );
   }
 
@@ -92,7 +91,7 @@ class Task extends Model<int> {
       stageName: target.value.scheduler.toString(),
       status: Task.statusNew,
       timeoutInMinutes: target.value.timeout,
-      isBenchmark: target.isBenchmark(),
+      isBenchmark: target.isBenchmark,
     );
   }
 
--- a/app_dart/lib/src/model/appengine/task.dart
+++ b/app_dart/lib/src/model/appengine/task.dart
@@ -271,6 +271,9 @@ class Task extends Model<int> {
   @JsonKey(name: 'Properties')
   String? serializedProperties;
 
+  Map<String, dynamic> get properties =>
+      serializedProperties != null ? jsonDecode(serializedProperties!) as Map<String, dynamic> : <String, dynamic>{};
+
   /// Update [Task] related fields based on LUCI's [BuildPushMessage].
   String updateFromBuildPushMessage(BuildPushMessage pushMessage) {
     if (pushMessage.build?.status == Status.started) {
--- a/app_dart/lib/src/model/appengine/task.dart
+++ b/app_dart/lib/src/model/appengine/task.dart
@@ -42,6 +42,7 @@ class Task extends Model<int> {
     this.buildNumberList,
     this.builderName,
     this.luciBucket,
+    this.isBenchmark,
     String? status,
   }) : _status = status {
     if (status != null && !legalStatusValues.contains(status)) {
@@ -70,6 +71,7 @@ class Task extends Model<int> {
       stageName: 'chromebot',
       status: Task.statusNew,
       timeoutInMinutes: 0,
+      isBenchmark: builder.benchmark,
     );
   }
 
@@ -90,6 +92,7 @@ class Task extends Model<int> {
       stageName: target.value.scheduler.toString(),
       status: Task.statusNew,
       timeoutInMinutes: target.value.timeout,
+      isBenchmark: target.isBenchmark(),
     );
   }
 
@@ -262,6 +265,11 @@ class Task extends Model<int> {
     _status = value;
   }
 
+  /// Whether the task is a benchmark test
+  @BoolProperty(propertyName: 'Benchmark')
+  @JsonKey(name: 'Benchmark')
+  bool? isBenchmark;
+
   /// Update [Task] related fields based on LUCI's [BuildPushMessage].
   String updateFromBuildPushMessage(BuildPushMessage pushMessage) {
     if (pushMessage.build?.status == Status.started) {
@@ -314,6 +322,7 @@ class Task extends Model<int> {
       ..write(', buildNumberList: $buildNumberList')
       ..write(', builderName: $builderName')
       ..write(', luciBucket: $luciBucket')
+      ..write(', isBenchmark: $isBenchmark')
       ..write(')');
     return buf.toString();
   }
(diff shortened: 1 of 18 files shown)


app_dart/lib/src/model/appengine/task.dart
```diff
@@ -262,6 +265,11 @@ class Task extends Model<int> {
     _status = value;
   }
 
+  /// Whether the task is a benchmark test
+  @BoolProperty(propertyName: 'Benchmark')
+  @JsonKey(name: 'Benchmark')
+  bool? isBenchmark;
```
CaseyHillers, line 271: We don't want to add this here. Instead we should be adding a Map<String, dynamic> properties (the equivalent of what's in models/ci_yaml/target.dart -> getProperties())

app_dart/lib/src/model/ci_yaml/target.dart
```diff
@@ -141,6 +141,19 @@ class Target {
   String getBucket() {
     return value.bringup ? 'staging' : 'prod';
   }
+
+  /// Check if this [Target] is benchmark.
+  ///
+  /// keyword benchmark is extracted from the properties.
+  bool isBenchmark() {
```
CaseyHillers, line 148:
```suggestion
static const String kBenchmarkPropertyKey = 'benchmark';

bool get isBenchmark => getProperties()['benchmark'] == true ?? false;
```

app_dart/lib/src/request_handlers/flaky_handler_utils.dart
```diff
@@ -455,9 +455,9 @@ List<dynamic>? _getTags(String? builderName, YamlMap ci) {
 }
 
 String _getTestNameFromBuilderName(String builderName) {
-  // The builder names is in the format '<platform> <test name>'.
+  // The builder names is in the format '<platform> <test name>' or '<platform> benchmark <test name>'.
```
CaseyHillers, line 458: These seem to be old changes. Should they be reverted?

app_dart/lib/src/service/luci.dart
```diff
@@ -229,6 +229,7 @@ class LuciBuilder {
     this.enabled,
     this.runIf,
     this.taskName,
+    this.benchmark,
```
CaseyHillers, line 232: Changes to this file should not be necessary at this time. Generally, only commit changes that are going to be used in this PR

app_dart/test/model/task_test.dart
```diff
@@ -55,6 +57,11 @@ void main() {
       );
       expect(Task.chromebot(commitKey: commitKey, createTimestamp: 123, builder: builder).isFlaky, isFalse);
     });
+
+    test('benchmark defaults to null', () {
+      final Task task = generateTask(1, attempts: 5);
+      expect(task.isBenchmark, null);
```
CaseyHillers, line 63: With null safety, we should only have true or false. So this should default to false.

app_dart/lib/src/model/appengine/task.dart
keyonghan: This is to save the benchmark flag to datastore. Do you want to save the entire properties? which I don't think is needed and will make the query complicated. Or do i miss anything?

app_dart/lib/src/model/appengine/task.dart
```diff
@@ -70,6 +71,7 @@ class Task extends Model<int> {
       stageName: 'chromebot',
       status: Task.statusNew,
       timeoutInMinutes: 0,
+      isBenchmark: builder.benchmark,
```
keyonghan, line 74: This seems not needed.

app_dart/lib/src/model/appengine/task.dart
```diff
@@ -90,6 +92,7 @@ class Task extends Model<int> {
       stageName: target.value.scheduler.toString(),
       status: Task.statusNew,
       timeoutInMinutes: target.value.timeout,
+      isBenchmark: target.isBenchmark(),
```
keyonghan, line 95: update here based on @CaseyHillers 's one comment below.

app_dart/lib/src/model/ci_yaml/target.dart
keyonghan: Nit on @CaseyHillers comment above: `getProperties()['benchmark']` => `_getTargetProperties()['benchmark']`

As this is only defined on target level.

app_dart/lib/src/model/appengine/task.dart
```diff
@@ -42,6 +42,7 @@ class Task extends Model<int> {
     this.buildNumberList,
     this.builderName,
     this.luciBucket,
+    this.isBenchmark,
```
keyonghan, line 45: Let's define a default `false`, which should be okay considering old data is null. (front end is based on tot task status).

keyonghan (commented):
Please regenerate json serialization codes: https://github.com/flutter/cocoon/tree/main/app_dart#json

app_dart/lib/src/model/appengine/task.dart
CaseyHillers: Yes, let's save the entire properties. This should be straight forward by passing it from Target to Task.

Part of the migration to microservices is to simplify what data we store in entities. Adding columns to the datastore isn't sustainable when only a few tests would benefit from it. \cc @yusuf-goog is likely to remove a lot of these entries in the results processing refactor

app_dart/lib/src/model/ci_yaml/target.dart
CaseyHillers: `getProperties` is the properties defined for a target (as its combination of local and platform properties). I would avoid looking at local or platform specific properties separately.

app_dart/lib/src/model/appengine/task.dart
keyonghan: Thanks for the explanation. That makes sense to me. 
@KristinBi since datastore doesn't support map (https://cloud.google.com/datastore/docs/concepts/entities#properties_and_value_types), you may need to save the properties as string via jsonEncode, and use jsonDecode when retrieving the values.

CaseyHillers:
Can you push an empty commit? Sometimes force pushes break the presubmit testing

app_dart/lib/src/model/ci_yaml/target.dart
```diff
@@ -159,6 +159,11 @@ class Target {
   String getBucket() {
     return value.bringup ? 'staging' : 'prod';
   }
+
+  /// Check if this [Target] is benchmark.
+  ///
+  /// keyword benchmark is extracted from the properties.
+  bool get isBenchmark => getProperties()['benchmark'] == true ?? false;
```
keyonghan, line 166:
```suggestion
  bool get isBenchmark => getProperties()['benchmark'] as bool? ?? false;
```

app_dart/lib/src/model/appengine/task.dart
```diff
@@ -42,6 +44,7 @@ class Task extends Model<int> {
     this.buildNumberList,
     this.builderName,
     this.luciBucket,
+    this.serializedProperties,
```
CaseyHillers, line 47: This doesn't appear to be set anywhere. Additionally, you'll likely want this to take in `properties` and create the `serializedProperties` in the constructor from it

CaseyHillers: Talked offline. `Task.fromTarget` is correctly parsing the properties.

We should still remove `serializedProperties` from here as it's easy to mess up (or to pass in Map<String, dynamic> properties and encode it in the constructor)

CaseyHillers: More offline discussion. Here's a quick example

```dart
class Task {
  Task(Map<String, dynamic> properties) {
    serializedProperties = jsonEncode(properties);
  }
  
  String? serializedProperties;
}
```

PR #1599, opened 2022-02-11
Test dimensions followup

keyonghan:
This is a follow PR of https://github.com/flutter/cocoon/pull/1597 to move `getDimensions` function to Target.

### DIFF
--- a/app_dart/lib/src/model/ci_yaml/target.dart
+++ b/app_dart/lib/src/model/ci_yaml/target.dart
@@ -35,7 +35,6 @@ class Target {
   /// Target prefixes that indicate it will run on an ios device.
   static const List<String> iosPlatforms = <String>['mac_ios', 'mac_ios32'];
 
-
   /// Gets dimensions for this [pb.Target].
   List<RequestedDimension> getDimensions() {
     final Map<String, Object> properties = getProperties();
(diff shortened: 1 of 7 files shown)


app_dart/lib/src/service/config.dart
```diff
@@ -97,7 +97,7 @@ class Config {
   }
 
   // Dimension list defined in .ci.yaml.
-  List<String> get dimensionList => <String>['os', 'device_os', 'device_type', 'mac_model'];
+  static List<String> get dimensionList => <String>['os', 'device_os', 'device_type', 'mac_model'];
```
CaseyHillers, line 100: Should this be moved into `Target`? I can't see a reason why these need to be accessible outside of `Target`.

app_dart/lib/src/model/ci_yaml/target.dart
```diff
@@ -33,6 +35,18 @@ class Target {
   /// Target prefixes that indicate it will run on an ios device.
   static const List<String> iosPlatforms = <String>['mac_ios', 'mac_ios32'];
 
+  /// Gets dimensions for this [pb.Target].
```
CaseyHillers, line 38: Can you expand on what dimensions are? Possibly a link to the swarming documentation

CaseyHillers (commented):
Thank you for the quick follow up!

app_dart/lib/src/service/config.dart
keyonghan: Done.

CaseyHillers (approved):
flutter-lgtm

CaseyHillers:
FYI tests are failing due to formatting issues

PR #1601, opened 2022-02-11
[app_dart] Add logs for postsubmit build results

CaseyHillers:
This is to help with debugging why the Cocoon build dashboard isn't updating.

https://flutter-dashboard.appspot.com/#/build?repo=cocoon&branch=main

### DIFF
--- a/app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
+++ b/app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
@@ -8,6 +8,7 @@ import 'package:appengine/appengine.dart';
 import 'package:gcloud/db.dart';
 import 'package:meta/meta.dart';
 
+import '../model/appengine/commit.dart';
 import '../model/appengine/key_helper.dart';
 import '../model/appengine/task.dart';
 import '../model/luci/push_message.dart';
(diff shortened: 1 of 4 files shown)


app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
```diff
@@ -53,20 +53,28 @@ class PostsubmitLuciSubscription extends SubscriptionHandler {
     final BuildPushMessage buildPushMessage =
         BuildPushMessage.fromJson(json.decode(String.fromCharCodes(base64.decode(data))) as Map<String, dynamic>);
     log.fine(buildPushMessage.userData);
+    log.fine('Updating buildId=${buildPushMessage.build?.id} for builder=${buildPushMessage.build?.result}');
```
yusuf-goog, line 56: does .result contain the builder name?

app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
```diff
@@ -53,20 +53,28 @@ class PostsubmitLuciSubscription extends SubscriptionHandler {
...
+    if (buildPushMessage.userData == null) {
+      log.fine('User data is empty');
+      return Body.empty;
+    }
     final Map<String, dynamic> userData = jsonDecode(buildPushMessage.userData!) as Map<String, dynamic>;
     final String? taskKey = userData['task_key'] as String?;
     if (taskKey == null) {
       throw const BadRequestException('userData does not contain task_key');
     }
+    log.fine('Looking up key...');
```
yusuf-goog, line 70: Maybe print taskKey, or some value to track which key its trying to lookup?

yusuf-goog (approved):
LGTM, with couple comments.

app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
CaseyHillers: It's logged in the userData line above

app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
CaseyHillers: No, it's an enum of cancelled, success, infra_failure, failure

CaseyHillers:
I did a test version with these logs and was able to debug why the postsubmit builds aren't being processed. I added the quick fix in https://github.com/flutter/cocoon/pull/1601/commits/54bcf35f6af360783dd825571f34ca73c837e820

tl;dr It was expecting encoded keys, but I decided to simplify the process by using the raw id (so we can eventually remove the KeyHelper code)

app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
yusuf-goog: Then change `...for builder=` to something like `...with result=`?

CaseyHillers: Ah, thanks! Done

PR #1603, opened 2022-02-14
Shift flake bot to run on Wed.

keyonghan:
Fixes: https://github.com/flutter/flutter/issues/98430 by shifting auto flake bot run from Mon. to Wed.

(will land this after Wed.)


keyonghan:
> LGTM, just want to mention I originally set Monday because it is also close to cocoon deployment.

Cocoon is being deployed daily, so this should not be an issue:-)

PR #1606, opened 2022-02-16
[autosubmit] Initial project creation

CaseyHillers:


auto_submit/bin/server.dart
```diff
@@ -0,0 +1,60 @@
...
+Future main() async {
+  final projectId = await currentProjectId();
+  print('Current GCP project id: $projectId');
+
+  final authClient = await clientViaApplicationDefaultCredentials(
+    scopes: [FirestoreApi.datastoreScope],
+  );
+
+  try {
+    final api = FirestoreApi(authClient);
```
keyonghan, line 22: Could you please change the example API to a different one, like the BigqueryApi? Our current project does not support Cloud Firestore API (Below is the error I hit when running locally). This would make it easier to avoid any local run errors for @KristinBi to start with.
```
Error thrown by handler.
DetailedApiRequestError(status: 400, message: The Cloud Firestore API is not available for Datastore Mode projects.)
```

CaseyHillers: I removed the firestore project. I'll let @KristinBi start with an empty codebase and decide what dependencies we should pull in

auto_submit/bin/server.dart
```diff
@@ -0,0 +1,24 @@
...
+      headers: {
+        'content-type': 'application/json',
+      },
+    );
+  }
+
+  final router = Router()..get('/', emptyHandler);
+
+  await serveHandler(router);
+}
```
keyonghan, line 24: Nit: extra line.

auto_submit/bin/server.dart
keyonghan: SG. Thanks!

keyonghan (approved):
LGTM with nit.

auto_submit/bin/server.dart
CaseyHillers: Done

PR #1609, opened 2022-02-18
[dashboard] microseconds for start time and end time (for now)

XilaiZhang:
A maybe-not-so-good fix to circumvent https://github.com/flutter/flutter/issues/98676 and https://github.com/flutter/flutter/issues/92854. Currently the entities stored in our datastore has start time (need to verify since they are all 0s) and end time in microseconds since epoch, while create time is stored in milliseconds since epoch. In this fix, dashboard displays accordingly based on current assumptions. After infra has been migrated to a new scheduler in the future, we may need to change milli / micro seconds based on datastore again.

future work: find out root cause of start and end time being stored in microseconds in datastore.

### DIFF
--- a/dashboard/lib/widgets/task_overlay.dart
+++ b/dashboard/lib/widgets/task_overlay.dart
@@ -226,7 +226,6 @@ class TaskOverlayContents extends StatelessWidget {
         task.startTimestamp == 0 ? now.difference(createTime) : startTime.difference(createTime);
     final Duration runDuration = task.endTimestamp == 0 ? now.difference(startTime) : endTime.difference(startTime);
 
-    print('queue duration is $queueDuration and runDuration is $runDuration');
     /// There are 2 possible states for queue time:
     ///   1. Task is waiting to be scheduled (in queue)
     ///   2. Task has been scheduled (out of queue)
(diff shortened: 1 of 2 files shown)


XilaiZhang:
> Is the issue that tasks with stage == `luci` are in millseconds, and those with `cocoon` are in microseconds?

I think that is very likely to be true but stage == luci jobs currently have start and end time being 0 so I can't really tell if they are in mill or micro.

keyonghan:
For timestamp saved in datastore:
`create time` is inserted based on millisecond (from commit create time) (https://github.com/flutter/cocoon/blob/main/app_dart/lib/src/service/scheduler.dart#L129, https://github.com/flutter/cocoon/blob/main/app_dart/lib/src/model/appengine/commit.dart#L31

`end time` seems also use milliseconds though: https://github.com/flutter/cocoon/blob/main/app_dart/lib/src/request_handlers/update_task_status.dart#L61

XilaiZhang:
verified on go/cocoon-build that after https://github.com/flutter/cocoon/pull/1611 has been merged, tasks happening after the merge have timestamps in sync, in the units of microseconds. Although we process the data sources assuming they are in milliseconds, since all times are computed based on relative difference, the computed results are correct. I am expecting https://flutter-dashboard.appspot.com/#/build to have the same correct results as go/cocoon-build soon, and we can close this PR for now?

PR #1611, opened 2022-02-18
Fix push message not processing in microseconds

CaseyHillers:
Timestamp is expected to be in microseconds from LUCI.

This is affecting the plugins, cocoon, and packages build dashboards from showing an accurate time.

### DIFF
--- a/app_dart/lib/src/model/common/json_converters.dart
+++ b/app_dart/lib/src/model/common/json_converters.dart
@@ -86,6 +86,27 @@ class Base64Converter implements JsonConverter<String, String> {
   }
 }
 
+/// A converter for "timestamp" fields encoded as microseconds since epoch.
+class MicrosecondsSinceEpochConverter implements JsonConverter<DateTime?, String?> {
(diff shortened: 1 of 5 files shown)


XilaiZhang:
umm maybe my understanding is off, but should we change pushmessage.g.dart e.g. L79 as well. I am new to metadata constructors and my understanding is that metadata declaration does not do the actual work?

dnfield (approved):

XilaiZhang:
> > umm maybe my understanding is off, but should we change pushmessage.g.dart e.g. L79 as well. I am new to metadata constructors and my understanding is that metadata declaration does not do the actual work?
> 
> That file should get updated by the build process

Thanks a lot for the hint! I am trying to understand it and currently I thought it might be expanded through KeyHandler by its  _populateTypes function, but I could not find the exact place where the content of .g.dart files got built. Could you provide some references to the parts of the code that generates .g.dart file and expand the metadata constructors? Thank you!

PR #1615, opened 2022-02-18
[autosubmit] Add webhook handler

KristinBi:
Created the webhook handler to handle the request.

This intends to solve the issue: https://github.com/flutter/flutter/issues/98703


auto_submit/bin/server.dart
```diff
@@ -9,25 +9,27 @@ import 'package:auto_submit/helpers.dart';
 import 'package:shelf/shelf.dart';
 import 'package:shelf_router/shelf_router.dart';
 
-class AutosubmitServer {
-  AutosubmitServer();
+Future<Response> webhookHandler(Request request) async {
+  final String? reqHeader = jsonEncode(request.headers);
+  print('Header: $reqHeader');
```
CaseyHillers, line 14: Import `package:logging` and use logs instead

https://dart-lang.github.io/linter/lints/avoid_print.html

auto_submit/bin/server.dart
```diff
@@ -9,25 +9,27 @@ import 'package:auto_submit/helpers.dart';
 import 'package:shelf/shelf.dart';
 import 'package:shelf_router/shelf_router.dart';
 
-class AutosubmitServer {
-  AutosubmitServer();
+Future<Response> webhookHandler(Request request) async {
+  final String? reqHeader = jsonEncode(request.headers);
```
CaseyHillers, line 13: You can directly log request.headers, not need to json encode it then print it

auto_submit/bin/server.dart
```diff
@@ -9,25 +9,27 @@ import 'package:auto_submit/helpers.dart';
...
+  print('Header: $reqHeader');
+  return Response.ok(
+    jsonEncode(<String, String>{}),
```
CaseyHillers, line 16: I would have this parrot back the response body to make it more testable, and drop the headers. There isn't much functionality beyond auth that comes from headers.

auto_submit/bin/server.dart
```diff
@@ -9,25 +9,27 @@ import 'package:auto_submit/helpers.dart';
 import 'package:shelf/shelf.dart';
 import 'package:shelf_router/shelf_router.dart';
 
-class AutosubmitServer {
-  AutosubmitServer();
+Future<Response> webhookHandler(Request request) async {
```
CaseyHillers, line 12: This should be moved into it's own file under lib/ like `lib/requests/github_webhook.dart`

auto_submit/app.yaml
```diff
@@ -0,0 +1,3 @@
+runtime: custom
```
keyonghan, line 1: Do we need this file?

auto_submit/test/server_test.dart
```diff
@@ -0,0 +1,20 @@
+// Copyright 2021 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:shelf/shelf.dart';
+import 'package:test/test.dart';
```
keyonghan, line 6: Nit: add an extra space.

auto_submit/test/server_test.dart
```diff
@@ -0,0 +1,21 @@
+// Copyright 2021 The Flutter Authors. All rights reserved.
```
CaseyHillers, line 1: This should be moved test/requests/github_webhook_test.dart

auto_submit/test/server_test.dart
```diff
@@ -0,0 +1,21 @@
...
+import 'package:test/test.dart';
+import 'package:auto_submit/requests/github_webhook.dart';
+
+void main() {
+  Request req = Request('POST', Uri.parse('http://localhost/'),
+      headers: {
+        'header1': 'header value1',
+      },
+      body: "{\"label1\": \"label1cotent\"}");
+  GithubWebhook githubWebhook = GithubWebhook();
```
CaseyHillers, line 15: Avoid adding logic directly in main. It makes it harder to debug and find what code is being run. Adding basic data models is fine

This should either be in test(...); or in a `setUp(...);`

auto_submit/lib/requests/github_webhook.dart
```diff
@@ -0,0 +1,24 @@
...
+// found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:logging/logging.dart';
+import 'package:shelf/shelf.dart';
+
+final Logger logger = Logger('GithubWebhook');
+
+class GithubWebhook {
```
CaseyHillers, line 12: dartdoc?

auto_submit/lib/requests/github_webhook.dart
```diff
@@ -0,0 +1,24 @@
...
+  githubWebhook() {}
+
+  Future<Response> webhookHandler(Request request) async {
```
CaseyHillers, line 15: Maybe call this method `post` instead?

auto_submit/lib/requests/github_webhook.dart
```diff
@@ -0,0 +1,24 @@
...
+
+import 'dart:async';
+
+import 'package:logging/logging.dart';
+import 'package:shelf/shelf.dart';
+
+final Logger logger = Logger('GithubWebhook');
+
+class GithubWebhook {
+  githubWebhook() {}
```
CaseyHillers, line 13:
```suggestion
  const GithubWebhook();
```

auto_submit/lib/helpers.dart
```diff
@@ -7,6 +7,9 @@ import 'dart:io';
 
 import 'package:shelf/shelf.dart';
 import 'package:shelf/shelf_io.dart';
+import 'package:logging/logging.dart';
+
+final Logger logger = Logger('helper');
```
CaseyHillers, line 12: Let's move the loggers into their own file, like `service/log.dart` and instantiate it like `final Logger log = Logger('auto_submit')`

Changing the tag in every file will make it hard to read the logs and understand what was happening for a request

auto_submit/lib/auto_submit.dart
```diff
@@ -0,0 +1,5 @@
+// Copyright 2019 The Flutter Authors. All rights reserved.
```
keyonghan, line 1: Consistent to be 2022.

auto_submit/bin/server.dart
```diff
@@ -18,7 +20,10 @@ Future main() async {
     );
   }
 
-  final router = Router()..get('/', emptyHandler);
+  GithubWebhook githubWebhook = GithubWebhook();
 
+  final router = Router()
+    ..get('/', emptyHandler)
```
keyonghan, line 26: We can remove the emptyHandler now.

keyonghan (approved):
LGTM with test fix: the new 2022 should be added here: https://github.com/flutter/cocoon/blob/main/licenses/check_licenses.dart#L35

auto_submit/lib/requests/github_webhook.dart
CaseyHillers: No, dartdoc refers to adding a class doc here like

```dart
/// Handler for processing GitHub webhooks.
///
/// On events where a label was added to a pull request, ...
```

See https://dart.dev/guides/language/effective-dart/documentation for the general style guide and suggestions on writing effective documentation

KristinBi: Gotcha, will add it now. Thanks

CaseyHillers: FYI this would be above line 10 on the `class GithubWebhook`

CaseyHillers (approved):
LGTM after the nit is resolved with the doc

PR #1616, opened 2022-02-22
[dashboard] migrate entire dashboard to null safety

XilaiZhang:
Migrates cocoon dashboard to null safety.

I proposed ways (some of them probably not the best) to pass flutter test, and those sections are marked with Comments that start with "FOR REVIEW".


XilaiZhang:
finally, let's go!

XilaiZhang:
More contexts on some of the major parts marked with "FOR REVIEW": 
1. for each url launch, I also registered it with the channel `MethodChannel('plugins.flutter.io/url_launcher')` and launched again in that channel, so that this method call gets captured inside test. I might be wrong, but it seems like after null safety migration, `setMockMethodCallHandler` can no longer intercept method calls that are not registered to its channel. 
2. Commit.message would be nullable if it take its definition from commit.pb.dart. However, in commit_box.dart, lint would consider commit.message as not nullable. But this is the only such case, on which I disagree with lint. 
3.  Other secondary stubbings, logics and types rewrites.

CaseyHillers:
I'll set aside a few hours today to review this, but this is very impressive! Thank you Xilai!

dashboard/test/navigation_drawer_test.dart
```diff
@@ -72,20 +72,9 @@ void main() {
...
-            'useSafariVC': true,
-            'useWebView': false,
-            'enableJavaScript': false,
-            'enableDomStorage': false,
-            'universalLinksOnly': false,
-            'headers': <String, String>{}
-          })
-        ],
-      );
+      expect(log[0].runtimeType, equals(MethodCall));
```
CaseyHillers, line 75: When debugging this code, does log contain any entries (without you explicitly adding a mock method channel everywhere).

Otherwise, we'll need to look how to do this the null safe way

dashboard/test/state/build_test.dart
```diff
@@ -309,8 +311,13 @@ void main() {
...
   testWidgets('sign in functions call notify listener', (WidgetTester tester) async {
-    final MockGoogleSignInPlugin mockSignInPlugin = MockGoogleSignInPlugin();
-    when(mockSignInPlugin.onCurrentUserChanged).thenAnswer((_) => Stream<GoogleSignInAccount>.value(null));
+    final MockGoogleSignIn mockSignInPlugin = MockGoogleSignIn();
+    when(mockSignInPlugin.signIn()).thenAnswer((_) => Future<GoogleSignInAccount?>.value(null));
+    when(mockSignInPlugin.signOut()).thenAnswer((_) => Future<GoogleSignInAccount?>.value(null));
+    when(mockSignInPlugin.signInSilently()).thenAnswer((_) => Future<GoogleSignInAccount?>.value(null));
+    // FOR REVIEW:
+    // probably not the right way to stub null for signin and signout, but it passes the test
+    when(mockSignInPlugin.onCurrentUserChanged).thenAnswer((_) => Stream<GoogleSignInAccount?>.value(null));
```
CaseyHillers, line 320: Stream.empty is useful for tests

https://api.dart.dev/stable/2.10.4/dart-async/Stream/Stream.empty.html

dashboard/test/state/index_test.dart
```diff
@@ -13,8 +13,13 @@ import '../utils/mocks.dart';
 
 void main() {
   testWidgets('IndexState sign in functions call notify listener', (WidgetTester tester) async {
-    final MockGoogleSignInPlugin mockSignInPlugin = MockGoogleSignInPlugin();
-    when(mockSignInPlugin.onCurrentUserChanged).thenAnswer((_) => Stream<GoogleSignInAccount>.value(null));
+    final MockGoogleSignIn mockSignInPlugin = MockGoogleSignIn();
+    when(mockSignInPlugin.onCurrentUserChanged).thenAnswer((_) => Stream<GoogleSignInAccount?>.value(null));
+    when(mockSignInPlugin.signIn()).thenAnswer((_) => Future<GoogleSignInAccount?>.value(null));
```
CaseyHillers, line 18: https://api.dart.dev/stable/2.10.4/dart-async/Stream/Stream.empty.html

dashboard/test/state/build_test.dart
```diff
@@ -309,8 +311,13 @@ void main() {
   });
 
   testWidgets('sign in functions call notify listener', (WidgetTester tester) async {
-    final MockGoogleSignInPlugin mockSignInPlugin = MockGoogleSignInPlugin();
-    when(mockSignInPlugin.onCurrentUserChanged).thenAnswer((_) => Stream<GoogleSignInAccount>.value(null));
+    final MockGoogleSignIn mockSignInPlugin = MockGoogleSignIn();
+    when(mockSignInPlugin.signIn()).thenAnswer((_) => Future<GoogleSignInAccount?>.value(null));
+    when(mockSignInPlugin.signOut()).thenAnswer((_) => Future<GoogleSignInAccount?>.value(null));
+    when(mockSignInPlugin.signInSilently()).thenAnswer((_) => Future<GoogleSignInAccount?>.value(null));
+    // FOR REVIEW:
```
CaseyHillers, line 318: nit: rm comment

dashboard/test/utils/fake_google_account.dart
```diff
@@ -19,11 +20,33 @@ class FakeGoogleSignInAccount implements GoogleSignInAccount {
...
 
   @override
-  Future<void> clearAuthCache() => null;
+  late Future<GoogleSignInAuthentication> authentication;
+
+  @override
+  Future<void> clearAuthCache() => Future<void>.value(null);
+
+  @override
+  bool operator ==(dynamic other) {
```
CaseyHillers, line 35: For a test, keep it simple and do

```dart
  @override
  bool operator ==(dynamic other) {
    final GoogleSignInAccount otherAccount = other;
    return email == otherAccount.email;
  }
```

dashboard/test/utils/fake_google_account.dart
```diff
@@ -19,11 +20,33 @@ class FakeGoogleSignInAccount implements GoogleSignInAccount {
...
+    final GoogleSignInAccount otherAccount = other;
+    return displayName == otherAccount.displayName &&
+        email == otherAccount.email &&
+        id == otherAccount.id &&
+        photoUrl == otherAccount.photoUrl &&
+        serverAuthCode == otherAccount.serverAuthCode;
+  }
+
+  @override
+  int get hashCode => hashValues(displayName, email, id, photoUrl, serverAuthCode);
```
CaseyHillers, line 51: Same here.

```dart
  @override
  int get hashCode => email.hasCode;
```

dashboard/test/utils/wrapper.dart
```diff
@@ -3,29 +3,41 @@
...
+    final MockGoogleSignInService fakeAuthService = MockGoogleSignInService();
+    if (signedIn) {
+      when(fakeAuthService.isAuthenticated).thenAnswer((_) => Future<bool>.value(true));
+    } else {
+      when(fakeAuthService.isAuthenticated).thenAnswer((_) => Future<bool>.value(false));
+    }
+
+    when(fakeAuthService.user).thenReturn(FakeGoogleSignInAccount());
+
+    //final GoogleSignInService authService = MockGoogleSignInService();
```
CaseyHillers, line 33: nit: rm comment

dashboard/test/widgets/sign_in_button_test.dart
```diff
@@ -25,7 +25,7 @@ final Widget testApp = MaterialApp(
 );
 
 void main() {
-  GoogleSignInService mockAuthService;
+  GoogleSignInService? mockAuthService;
```
CaseyHillers, line 28: nit: make this late instead of nullable

dashboard/lib/logic/links.dart
```diff
@@ -24,25 +25,36 @@ List<CocoonLink> createCocoonLinks(BuildContext context) {
...
     CocoonLink(
-      name: 'Repository',
-      icon: const Icon(Icons.info_outline),
-      action: () => launch('/repository.html'),
-    ),
+        name: 'Repository',
+        icon: const Icon(Icons.info_outline),
+        action: () {
+          launch('/repository.html');
+          const MethodChannel('plugins.flutter.io/url_launcher').invokeMethod<void>('launch', '/repository.html');
```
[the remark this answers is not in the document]
XilaiZhang: Thanks a lot for all the comments! Yeah this the part I am mostly concerned. Because after migrating to null safety, if I keep it as it was with only the launch statement, it does not get captured by `setMockMethodCallHandler`and will fail a ton of tests. Would be great if I can keep launch as it as and still pass all the tests, but I wasn't successful in doing that.

CaseyHillers:
For the url launcher issue with the update to null safety, try updating the tests to use something like:

```dart
class FakeUrlLauncher extends UrlLauncherPlatform {
  final List<String> launches = <String>[];
  
  @override
  Future<bool> canLaunch(String url) {
    throw UnimplementedError('canLaunch() has not been implemented.');
  }

  @override
  Future<bool> launch(
    String url, {
    required bool useSafariVC,
    required bool useWebView,
    required bool enableJavaScript,
    required bool enableDomStorage,
    required bool universalLinksOnly,
    required Map<String, String> headers,
    String? webOnlyWindowName,
  }) {
    launches.add(url);
  }

  @override
  Future<void> closeWebView() {
    throw UnimplementedError('closeWebView() has not been implemented.');
  }
}

void main() {
  FakeUrlLauncher urlLauncher;
  setUp(() {
    urlLauncher = FakeUrlLauncher();
    UrlLauncherPlatform.setInstance(urlLauncher);
  });
  
  wigetTest('example', () {
    // do some actions
    expect(urlLauncher.launches, isNotEmpty);
    expect(urlLauncher.launches.single, 'https://google.com');
  });
}
```

dashboard/lib/index_page.dart
```diff
@@ -44,9 +44,7 @@ class IndexPage extends StatelessWidget {
                   Column(children: <Widget>[
                     IntrinsicWidth(
                       child: ElevatedButton.icon(
-                          icon: link.icon!,
-                          label: Text(link.name!.toUpperCase()),
-                          onPressed: link.action as void Function()?),
+                          icon: link.icon!, label: Text(link.name!.toUpperCase()), onPressed: link.action),
```
CaseyHillers, line 47: nit: add trailing commas so the formatting looks nice

dashboard/lib/state/build.dart
```diff
@@ -125,23 +127,26 @@ class BuildState extends ChangeNotifier {
...
   /// Request the latest [repos] from [CocoonService].
   Future<void> _fetchRepos() async {
-    final CocoonResponse<List<String>> response = await cocoonService!.fetchRepos();
+    final CocoonResponse<List<String>> response = await cocoonService.fetchRepos();
     if (response.error != null) {
       _errors.send('$errorMessageFetchingBranches: ${response.error}');
     } else {
-      _repos = response.data;
+      _repos = response.data!;
+      //FOR REVIEW:
```
CaseyHillers, line 147: Can you clarify the question

dashboard/lib/widgets/lattice.dart
```diff
@@ -296,27 +296,25 @@ class _LatticeBodyElement extends RenderObjectElement implements _LatticeDelegat
...
-    }
-    // [ FOR REVIEW ]
-    // null safety rewrite: visitChildren enforce not null and visitor can't take nullable value
-    // original : (_newChildrenByCoordinate.values.toList()..sort(_compareChildren)).forEach(visitor);
-  }
-
-  int _compareChildren(Element? a, Element? b) {
-    final _Coordinate aSlot = a!.slot as _Coordinate;
-    final _Coordinate bSlot = b!.slot as _Coordinate;
+    (_newChildrenByCoordinate.values.whereType<Element>().toList()..sort(_compareChildren)).forEach(visitor);
```
CaseyHillers, line 299: Is `toList()` needed?

dashboard/test/navigation_drawer_test.dart
```diff
@@ -57,9 +58,8 @@ void main() {
     });
 
     testWidgets('skia perf links opens skia perf url', (WidgetTester tester) async {
-      const MethodChannel urlLauncherChannel = MethodChannel('plugins.flutter.io/url_launcher');
-      final List<MethodCall> log = <MethodCall>[];
-      urlLauncherChannel.setMockMethodCallHandler((MethodCall methodCall) async => log.add(methodCall));
+      final FakeUrlLauncher urlLauncher = FakeUrlLauncher();
```
CaseyHillers, line 61: This should get added to `setUp`

dashboard/test/state/build_test.dart
```diff
@@ -316,7 +316,8 @@ void main() {
     when(mockSignInPlugin.signOut()).thenAnswer((_) => Future<GoogleSignInAccount?>.value(null));
     when(mockSignInPlugin.signInSilently()).thenAnswer((_) => Future<GoogleSignInAccount?>.value(null));
     // FOR REVIEW:
-    // probably not the right way to stub null for signin and signout, but it passes the test
+    // I was trying to use stream<GoogleSignInAccount?>.empty() here but I only found how to cast future to stream
+    // didn't seem viable to cast stream to Future<GoogleSignInAccount?> type?
```
CaseyHillers, line 320: Did you try adding `async`? Like `async => Stream.empty`

dashboard/test/utils/fake_url_launcher.dart
```diff
@@ -0,0 +1,48 @@
...
+    return Future<bool>.value(true);
+  }
+
+  @override
+  Future<void> closeWebView() {
+    throw UnimplementedError('closeWebView() has not been implemented.');
+  }
+}
+
+// void main() {
```
CaseyHillers, line 40: rm?

CaseyHillers (commented):
Nice work! There's just a few cleanup comments left

dashboard/test/navigation_drawer_test.dart
XilaiZhang: umm I thought if we move them to setUp then several tests will share the same fake url launcher instance, and therefore multiple urls will be added to the same list that collects the launches. Therefore I was expecting the urlLauncher.launches.single test to always fail. Much to my surprise, after moving them all to setUp, all individual tests pass smoothly. Maybe my understanding is wrong, but I thought there should be corruption between tests since they are sharing the same fake url launcher instance?

CaseyHillers: setUp is per-test. setUpAll is for all tests in the file, and generally should be avoided

XilaiZhang: Thanks for the explanation! everything makes sense now! as for the other trivial parts, async did not work, to Lists() is needed etc. I will just push another commit that have everything sorted out now

dashboard/test/navigation_drawer_test.dart
```diff
@@ -57,9 +65,8 @@ void main() {
     });
 
     testWidgets('skia perf links opens skia perf url', (WidgetTester tester) async {
-      const MethodChannel urlLauncherChannel = MethodChannel('plugins.flutter.io/url_launcher');
-      final List<MethodCall> log = <MethodCall>[];
-      urlLauncherChannel.setMockMethodCallHandler((MethodCall methodCall) async => log.add(methodCall));
+      // final FakeUrlLauncher urlLauncher = FakeUrlLauncher();
```
CaseyHillers, line 68: nit: rm dead code

dashboard/test/navigation_drawer_test.dart
```diff
@@ -72,26 +79,13 @@ void main() {
...
-      );
+      expect(urlLauncher.launches, isNotEmpty);
+      expect(urlLauncher.launches.single, 'https://flutter-engine-perf.skia.org/');
     });
 
     testWidgets('repository opens repository html url', (WidgetTester tester) async {
-      const MethodChannel urlLauncherChannel = MethodChannel('plugins.flutter.io/url_launcher');
-      final List<MethodCall> log = <MethodCall>[];
-      urlLauncherChannel.setMockMethodCallHandler((MethodCall methodCall) async => log.add(methodCall));
+      // final FakeUrlLauncher urlLauncher = FakeUrlLauncher();
```
CaseyHillers, line 87: nit: rm dead code

dashboard/test/navigation_drawer_test.dart
```diff
@@ -103,26 +97,13 @@ void main() {
...
-      );
+      expect(urlLauncher.launches, isNotEmpty);
+      expect(urlLauncher.launches.single, '/repository.html');
     });
 
     testWidgets('source code opens github cocoon url', (WidgetTester tester) async {
-      const MethodChannel urlLauncherChannel = MethodChannel('plugins.flutter.io/url_launcher');
-      final List<MethodCall> log = <MethodCall>[];
-      urlLauncherChannel.setMockMethodCallHandler((MethodCall methodCall) async => log.add(methodCall));
+      // final FakeUrlLauncher urlLauncher = FakeUrlLauncher();
```
CaseyHillers, line 105: nit: rm dead code

dashboard/test/state/build_test.dart
```diff
@@ -272,9 +272,9 @@ void main() {
       when(
         mockCocoonService.fetchCommitStatuses(branch: 'master', repo: 'flutter'),
       ).thenAnswer(
-        (_) => Future<CocoonResponse<List<CommitStatus>>>.value(
+        (dynamic _) => Future<CocoonResponse<List<CommitStatus>>>.value(
```
CaseyHillers, line 275: this shouldn't need a type

CaseyHillers (approved):
LGTM! Thanks for yak shaving this!

flutter-lgtm

XilaiZhang:
> LGTM! Thanks for yak shaving this!
> 
> flutter-lgtm

Thanks a ton for reviewing! (learning new idioms every day, I will soon be a champion on slangs 🥇 )

PR #1621, opened 2022-02-25
[dashboard] Google test icon and source url

XilaiZhang:
Template code for step 4 and 5 of https://github.com/flutter/cocoon/blob/main/CI_YAML.md#external-tests. Need to update this PR, if there are url addresses that host commit and log pages for frob tasks.


dashboard/lib/logic/qualified_task.dart
```diff
@@ -14,12 +14,14 @@ class StageName {
   static const String cocoon = 'cocoon';
   static const String legacyLuci = 'chromebot';
   static const String luci = 'luci';
+  static const String frob = 'frob';
```
keyonghan, line 17: Do we expect to have more tests other than `google test` in frob? Team is more familiar with the term google-test rather than frob.

XilaiZhang: thanks for the advice! Shall I go ahead and change every frob naming to be google-test instead?

dashboard/lib/widgets/task_icon.dart
```diff
@@ -32,6 +32,13 @@ class TaskIcon extends StatelessWidget {
     // different assets.
     final Color? blendFilter = brightness == Brightness.dark ? Colors.white : null;
 
+    if (qualifiedTask.isFrob) {
+      return Image.asset(
+        'assets/ACX.png',
```
CaseyHillers, line 37: Let's change this to a google logo.

These tests aren't just ACX, and no one outside of Google will know what ACX is

dashboard/lib/logic/qualified_task.dart
CaseyHillers: Google test sgtm

dashboard/lib/logic/qualified_task.dart
```diff
@@ -14,12 +14,14 @@ class StageName {
...
 }
 
 /// Base URLs for various endpoints that can relate to a [Task].
 const String _cirrusUrl = 'https://cirrus-ci.com/github/flutter/flutter';
 const String _cirrusLogUrl = 'https://cirrus-ci.com/build/flutter/flutter';
 const String _luciUrl = 'https://ci.chromium.org/p/flutter';
+const String _frobUrl = 'https://flutter-rob.corp.google.com';
```
CaseyHillers, line 24: We'll need to update the log url as well

CaseyHillers (approved):
LGTM. We can follow up with the log url later

XilaiZhang:
cool, shall we merge this for now?

CaseyHillers:
Yep, feel free to merge

PR #1626, opened 2022-02-26
[dashboard] add and use lint package for dashboard

XilaiZhang:
use flutter_lint package for analysis, and only list the rules we would like to exclude from the package
fixes https://github.com/flutter/flutter/issues/99169


dashboard/analysis_options.yaml
```diff
@@ -29,68 +29,4 @@ include: package:flutter_lints/flutter.yaml
 
 linter:
```
CaseyHillers, line 30: Can we remove the linter block here since it's empty?

XilaiZhang: for sure

CaseyHillers (approved):

PR #1628, opened 2022-02-26
exclude rules from lint package for cocoon backend

XilaiZhang:
migrate cocoon backend rules to use lint packages
fixes https://github.com/flutter/flutter/issues/99168

### DIFF
--- a/app_dart/analysis_options.yaml
+++ b/app_dart/analysis_options.yaml
@@ -42,155 +42,69 @@ analyzer:
     - "**/*.pbenum.dart"
     - "test/src/utilities/mocks.mocks.dart"
 
+include: package:flutter_lints/flutter.yaml
+
 linter:
   rules:
     # these rules are documented on and in the same order as
     # the Dart Lint rules page to make maintenance easier
-    # https://github.com/dart-lang/linter/blob/master/example/all.yaml
-    - always_declare_return_types
-    - always_put_control_body_on_new_line
-    # - always_put_required_named_parameters_first # we prefer having parameters in the same order as fields https://github.com/flutter/flutter/issues/10219
-    - always_require_non_null_named_parameters
-    - always_specify_types
-    - annotate_overrides
-    # - avoid_annotating_with_dynamic # conflicts with always_specify_types
-    # - avoid_as # incompatible with implicit-casts: false
-    - avoid_bool_literals_in_conditional_expressions
-    # - avoid_catches_without_on_clauses # we do this commonly
-    # - avoid_catching_errors # we do this commonly
-    - avoid_classes_with_only_static_members
-    # - avoid_double_and_int_checks # only useful when targeting JS runtime
-    - avoid_empty_else
-    - avoid_field_initializers_in_const_classes
-    - avoid_function_literals_in_foreach_calls
-    # - avoid_implementing_value_types # not yet tested
-    - avoid_init_to_null
-    # - avoid_js_rounded_ints # only useful when targeting JS runtime
-    - avoid_null_checks_in_equality_operators
-    # - avoid_positional_boolean_parameters # not yet tested
-    # - avoid_private_typedef_functions # we prefer having typedef (discussion in https://github.com/flutter/flutter/pull/16356)
-    - avoid_relative_lib_imports
-    - avoid_renaming_method_parameters
-    - avoid_return_types_on_setters
-    # - avoid_returning_null # there are plenty of valid reasons to return null
-    # - avoid_returning_null_for_future # not yet tested
-    - avoid_returning_null_for_void
-    # - avoid_returning_this # there are plenty of valid reasons to return this
-    # - avoid_setters_without_getters # not yet tested
-    # - avoid_shadowing_type_parameters # not yet tested
-    # - avoid_single_cascade_in_expression_statements # not yet tested
-    - avoid_slow_async_io
-    - avoid_types_as_parameter_names
-    # - avoid_types_on_closure_parameters # conflicts with always_specify_types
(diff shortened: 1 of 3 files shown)


app_dart/analysis_options.yaml
```diff
@@ -42,155 +42,70 @@ analyzer:
...
-    # - unsafe_html # not yet tested
-    - use_full_hex_values_for_flutter_colors
-    # - use_function_type_syntax_for_parameters # not yet tested
-    - use_rethrow_when_possible
-    # - use_setters_to_change_properties # not yet tested
-    # - use_string_buffers # has false positives: https://github.com/dart-lang/sdk/issues/34182
-    # - use_to_and_as_if_applicable # has false positives, so we prefer to catch this by code-review
-    - valid_regexps
-    # - void_checks # not yet tested
+    # https://dart-lang.github.io/linter/lints/index.html
```
CaseyHillers, line 51: We should remove all the rules below as we want to move to using the recommended rules.

After removing, I think dart fix will get most of the trivial issues

XilaiZhang: Thanks for the reply! Umm some of the rules were stated in the comments section that they prefer to be excluded, like for avoid_private_typedef_functions, the comment says `we prefer having typedef (discussion in https://github.com/flutter/flutter/pull/16356)`. Should I go ahead and remove these ones as well?

CaseyHillers: Yes, these are a copy and paste from flutter/flutter's repo before we had package:lint to enforce the style guide.

app_dart/analysis_options.yaml
```diff
@@ -42,155 +42,11 @@ analyzer:
...
-    # - use_function_type_syntax_for_parameters # not yet tested
-    - use_rethrow_when_possible
-    # - use_setters_to_change_properties # not yet tested
-    # - use_string_buffers # has false positives: https://github.com/dart-lang/sdk/issues/34182
-    # - use_to_and_as_if_applicable # has false positives, so we prefer to catch this by code-review
-    - valid_regexps
-    # - void_checks # not yet tested
+    # a few rules listed below are the ones we would like to exclude from flutter_lint package, for app_dart
+    # reasons for exclusions are provided in the comments to the right
+    avoid_print: false # we have necessary print calls in the code
```
CaseyHillers, line 51: Can you link where? By now, they should be changed to use log

app_dart/test/request_handlers/refresh_chromebot_status_test.dart
```diff
@@ -189,19 +189,19 @@ void main() {
...
-                        const LuciTask(
-                            commitSha: 'abc',
-                            ref: 'refs/heads/master',
-                            status: Task.statusSucceeded,
-                            buildNumber: 1,
-                            builderName: 'abc')
-                      ],
-                    });
+        final Map<BranchLuciBuilder, Map<String, List<LuciTask>>> luciTasks = {
+          for (var builder in builders)
```
CaseyHillers, line 193: @keyonghan I prefer this use of var (we don't have to explicitly list the type, but it's implicit from above). WDYT?

app_dart/analysis_options.yaml
XilaiZhang: Screen Shot 2022-02-28 at 1 12 23 PM

XilaiZhang: umm yeah I am not sure if logs have been created for each of these files. Some sample print statement looks like this
Screen Shot 2022-02-28 at 1 13 50 PM

CaseyHillers: I think all of these are fine to switch with `log.fine`, you'll just need to add an import

XilaiZhang: cool for sure

app_dart/test/request_handlers/refresh_chromebot_status_test.dart
keyonghan: Any reason why not specify the `LuciBuilder` type explicitly? https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#avoid-using-var-and-dynamic

app_dart/analysis_options.yaml
XilaiZhang: Oh I think we actually need to keep the print statements, as they are eventually being directed by useLoggingPackageAdaptor() function in server.dart, to actual logging implementations such as setupAppEngineLogging()  function in logging_package_adaptor.dart. The print statements align with the examples listed in https://pub.dev/packages/logging, where print statements are eventually captured by log listeners

CaseyHillers: package:logging should default to stdout if there's no other instances

XilaiZhang: umm My understanding is that the logging is initialized once for all in server.dart

CaseyHillers: Ah, I see what you mean. We can follow up with this in a different PR

app_dart/test/request_handlers/refresh_chromebot_status_test.dart
CaseyHillers: @XilaiZhang can you re-add back the variable types to the lint?

XilaiZhang: yes sir

CaseyHillers (approved):

CaseyHillers:
Thank you for the clean up!

XilaiZhang:
nah you are too polite!

PR #1629, opened 2022-02-26
[release dashboard] suppress 3k errors in release dashboard to manifest the real ones

XilaiZhang:
With dart lint enabled, problems view would list all problems, including the 3k dependency and syntax errors under release dashboard folder, which is cluttering the actual problems of the opened project (e.g. app_dart, dashboard etc.). This commit suppresses lint in release dashboard, so we have a cleaner view of the actual problems.


CaseyHillers (approved):
RSLGTM

PR #1631, opened 2022-03-01
Add cirrus logs to help triage

keyonghan:
This is to help triage: https://github.com/flutter/flutter/issues/99306

### DIFF
--- a/app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
+++ b/app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
@@ -367,6 +367,7 @@ class CheckForWaitingPullRequests extends ApiRequestHandler<Body> {
 
     // The first build of cirrusGraphQL query always reflects the latest test statuses of the PR.
     final List<Map<String, dynamic>>? cirrusStatuses = cirrusResults.first.tasks;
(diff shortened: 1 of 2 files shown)


app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart
```diff
@@ -367,6 +367,7 @@ class CheckForWaitingPullRequests extends ApiRequestHandler<Body> {
 
     // The first build of cirrusGraphQL query always reflects the latest test statuses of the PR.
     final List<Map<String, dynamic>>? cirrusStatuses = cirrusResults.first.tasks;
+    log.info('First cirrus searchBuild id for flutter/$name, sha: $sha: ${cirrusResults.first.id}');
```
fkorotkov, line 370: Maybe it will be better to log all the build ids in order? Just to double check how many there are.

BTW will the log statement contain unix timestamp? Might be helpful if there is a race condition.

keyonghan: Yeah, please see the log in the other file `refresh_cirrus_status.dart`, where we log the id sequentially as the order returned by cirrus graphQL. 

These logs are saved in GCP with exact timestamp on each entry.

CaseyHillers (approved):

PR #1633, opened 2022-03-02
Call for the pull request and check whether should be merged

KristinBi:
After checking the pull request has the autosubmit label, send rest calls to github to get more data for this pull request to help check whether this pull request should merge or not.
This pr is intended to solve the issue: https://github.com/flutter/flutter/issues/98707.


CaseyHillers:
Rebasing on tip of tree will fix the dashboard failures

auto_submit/lib/requests/github_webhook.dart
```diff
@@ -25,19 +33,432 @@ class GithubWebhook {
...
-    hasAutosubmit = pullRequest.labels!.any((label) => label.name == 'autosubmit');
+    hasAutosubmit =
+        pullRequest.labels!.any((label) => label.name == 'autosubmit');
 
     if (hasAutosubmit) {
-      // TODO(kristinbi): Check if PR can be submitted. https://github.com/flutter/flutter/issues/98707
+      // TODO(kristin): hardcode the githubToken this time for test, will add it to env later.
+      // final String githubToken = Platform.environment['AUTOSUBMIT_TOKEN']!;
+
+      final String githubToken = 'f4d8bc081a5f3ad57f3df3a99ec0417b269cec90';
```
keyonghan, line 49: You don't want to submit with the real token. I believe this is for local test purpose, and for this PR I suggest bring line 47 back.

auto_submit/lib/requests/github_webhook.dart
```diff
@@ -25,19 +33,432 @@ class GithubWebhook {
...
+      final GithubService gitHub =
+          config.createGithubServiceWithToken(githubToken);
+
+      if (!(body.containsKey('repository') &&
+          body['repository'].containsKey('full_name') &&
+          body.containsKey('number'))) {
+        return Response.ok(jsonEncode(<String, String>{}));
+      }
```
keyonghan, line 57: Is this logic necessary?

auto_submit/lib/requests/github_webhook.dart
```diff
@@ -25,19 +33,432 @@ class GithubWebhook {
...
+      RepositorySlug slug =
+          RepositorySlug.full(body['repository']['full_name']);
+      int number = body['number'];
```
keyonghan, line 60: Are these variables expected to change later? Otherwise specify `final` keyword.

auto_submit/lib/requests/github_webhook.dart
```diff
@@ -25,19 +33,432 @@ class GithubWebhook {
...
+
+      // Use github Rest API to get this single pull request.
+      final PullRequest pr =
+          await gitHub.getPullRequest(slug, prNumber: number);
+      logger.info('Get the pull request $pr');
+
+      _AutoMergeQueryResult queryResult =
+          await _parseQueryData(pr, gitHub, body);
+      if (await shouldMergePullRequest(queryResult, slug, gitHub)) {
+        // TODO(Kristin): publish the pr with 'autosubmit' labek to pubsub. https://github.com/flutter/flutter/issues/98704
```
keyonghan, line 70: This logic should be in the cronjob API to keep pulling pubsub queues. Here, we can simply add a note: `log.info('adding pr to pubsub')`.

auto_submit/bin/server.dart
```diff
@@ -5,11 +5,13 @@
 import 'dart:async';
 
 import 'package:auto_submit/helpers.dart';
-import 'package:shelf_router/shelf_router.dart';
 import 'package:auto_submit/requests/github_webhook.dart';
+import 'package:auto_submit/service/config.dart';
```
CaseyHillers, line 9: This is great progress!

Can you split this PR into several smaller ones. In it's current state, its hard to give feedback.

I think something like:
1. Github app authentication
2. GraphQL query
3. Have the webhook work just with an approval, and merge it
4. Later we can follow up with all the extra logic

This will ensure we have high quality tests for your project

keyonghan (commented):
Tried reviewing the PR but as @CaseyHillers mentioned, this is hard to comment in a single PR.
Let's split into smaller ones for a better review and quality.

auto_submit/bin/server.dart
KristinBi: Sure Casey, l will try to split them.

auto_submit/lib/requests/github_webhook.dart
KristinBi: Here I think it's necessary to check if we can get the slug and prNumber from the webhook paylaod, because without them we cannot call rest api to get this pull request, let alone continue with the following logics.

auto_submit/lib/requests/github_webhook.dart
KristinBi: Yes, thanks Keyong, I will add final to all variables which are not supposed change later.

auto_submit/lib/requests/github_webhook.dart
KristinBi: Hi Keyong, do you mean that we should move the logic of  TODO to the cronjob API? I think line 67- 69 here is usde to check wheter this pr should be merged here.

keyonghan: Ideally, we should push whatever PRs that are available to the pubsub queue, and leave the check logic in the cronjob API.

Considering what @CaseyHillers suggested above, we can leave the logic here to check immediately whether a PR should merge, prior the pubsub is ready.

KristinBi: Oh I see, I will leave them here prior the pubsub is ready.

KristinBi:
After discussion, we decided to split this whole big PR to several small PRs, so close this one.

PR #1634, opened 2022-03-02
[dashboard] suppress overscroll notifications elided from inner vertical scrollbar

XilaiZhang:
Might not be the best approach, I am suppressing the left and right and movement from inner vertical scrollbar to fix https://github.com/flutter/flutter/issues/99418. After flutter upgrade, the tot version is currently throwing
https://github.com/flutter/flutter/issues/99418 and blocking other PRs from landing.


XilaiZhang:
My guess is that code like `await tester.drag(find.byType(TaskGrid), const Offset(-5000.0, 0.0));` caused a horizontal overscoll on the inner vertical scroller

CaseyHillers (approved):
flutter-lgtm

PR #1637, opened 2022-03-03
[app_dart] Add ci.yaml validation to ensure new targets aren't added that skip bringup

XilaiZhang:
this is an attempt to validate new builder and check whether they have bringup:true attached. as detailed in https://github.com/flutter/flutter/issues/90166. Some major things that might need further review:
1. I split the check into two functions, one of which regular, another async. So for functions that call upon scheduler but are not interested in new builders(e.g. validate ownership), they can use the old regular function. And for functions that would want to check for new builders, they would take time for async function to resolve. 
2. I am passing in slug as optional named parameter and Cofig to help retrieve repo information.


app_dart/test/service/scheduler/graph_test.dart
```diff
@@ -179,4 +179,93 @@ targets:
...
+  group('scheduler config tested and compared with tip of tree builders', () {
+    late SchedulerConfig? totConfig;
+
+    setUp(() {
+      YamlMap? totYaml = loadYaml('''
+enabled_branches:
+  - master
+targets:
+  - name: A
+  - name: B
```
CaseyHillers, line 192: Try to keep tests as simple as possible.

In this case, we only need 1 target to validate the functionality

CaseyHillers (commented):
It looks like you cherry-picked the upstream commits to get your branch on the latest changes. See go/git-workflows, but generally you want to do:

```sh
git fetch --all
git merge upstream/main
```

In it's current state, this is hard to review as I'm reviewing many changes.

At this point, you may need to do `git rebase -i`

app_dart/lib/src/service/scheduler/graph.dart
```diff
@@ -2,20 +2,51 @@
...
   final SchedulerConfig config = SchedulerConfig();
   config.mergeFromProto3Json(yamlConfig);
-  _validateSchedulerConfig(config);
+  validateSchedulerConfig(config);
 
   return config;
 }
 
-void _validateSchedulerConfig(SchedulerConfig schedulerConfig) {
+Future<SchedulerConfig> schedulerConfigFromYamlWithBuilderCheck(YamlMap? yamlConfig, {RepositorySlug? slug}) async {
```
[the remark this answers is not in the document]
XilaiZhang: Thanks for the advice!

Sorry for the github commits, I have squashed them now, but I guess the patches from other contributors still pertain in the overall change lists, which makes it difficult to review. Sorry about that. 

FOR REVIEW:

Yeah I was duplicating schedulerConfigFromYaml because with one declaration of schedulerConfigFromYaml, the process of pulling from the tip of tree makes the function async. And if schedulerConfigFromYaml is async, every function that depend on it become async. I have now removed the duplicate declarations, and instead make all functions that depend on it to be async as well. 

for _validateSchedulerConfig I removed the _ for testing purpose, so that I can supply yaml files in tests, instead of invoking the actual pulling from tip of tree. 

I have changed the optional parameter of slug to be a required positional parameter, because an optional parameter needs [Default values of an optional parameter must be constant] check, and since we cannot have a const constructor for RepositorySlug, I changed it to be positional.

auto_submit/pubspec.lock
```diff
@@ -163,6 +163,16 @@ packages:
     source: hosted
     version: "1.3.0"
   grpc:
+<<<<<<< HEAD
```
CaseyHillers, line 166: nit: merge conflict

XilaiZhang: Thanks a lot! didn't notice that part

app_dart/bin/generate_jspb.dart
```diff
@@ -66,6 +66,6 @@ Future<void> main(List<String> args) async {
   final YamlMap configYaml = loadYaml(configContent) as YamlMap;
   // There's an assumption that we're only generating builder configs from commits that
   // have already landed with validation. Otherwise, this will fail.
-  final pb.SchedulerConfig schedulerConfig = schedulerConfigFromYaml(configYaml);
+  final pb.SchedulerConfig schedulerConfig = await schedulerConfigFromYaml(configYaml, null);
```
CaseyHillers, line 69: Seeing how this is used, I'd recommend instead of passing branch we add a parameter like

```
schedulerConfigFromYaml(pb.SchedulerConfig config, {
  bool ensureBringupTargets = true,
});
```

Then in the classes impacted we can ignore the check.

As it stands, it's hard to understand the reasoning in the API

XilaiZhang: That's a very good point! Sorry for the delay, I guess I end up spending too much time on the shuttle bus commute today.

I have added the named parameter ensureBringupTargets. Eventually I kept the named parameter slug, because my understanding is that slug and the branch information it encodes are necessary for the githubFileContent function to work, and currently I am using githubFileContent to pull down tip of tree targets information.

I wasn't very sure if we can get schedulerConfigFromYaml to work with one named boolean parameter. Currently I am supplying two named parameters to the schedulerConfigFromYaml function. Maybe there is a way to pull information of tip of tree targets without supplying the branch information?

app_dart/lib/src/foundation/utils.dart
```diff
@@ -200,10 +200,10 @@ Future<void> insertBigquery(String tableName, Map<String, dynamic> data, Tableda
 }
 
 /// Validate test ownership defined in [testOwnersContent] for tests configured in `ciYamlContent`.
-List<String> validateOwnership(String ciYamlContent, String testOwnersContent) {
+Future<List<String>> validateOwnership(String ciYamlContent, String testOwnersContent) async {
   final List<String> noOwnerBuilders = <String>[];
   final YamlMap? ciYaml = loadYaml(ciYamlContent) as YamlMap?;
-  final pb.SchedulerConfig schedulerConfig = schedulerConfigFromYaml(ciYaml);
+  final pb.SchedulerConfig schedulerConfig = await schedulerConfigFromYaml(ciYaml, ensureBringupTargets: false);
```
CaseyHillers, line 206: Since there's only one case where `ensureBringupTargets = true`, we should set the default value to false.

app_dart/lib/src/service/scheduler.dart
```diff
@@ -231,7 +231,9 @@ class Scheduler {
       retryOptions: retryOptions,
     );
     final YamlMap configYaml = loadYaml(configContent) as YamlMap;
-    return schedulerConfigFromYaml(configYaml).writeToBuffer();
+    pb.SchedulerConfig schedulerConfig =
+        await schedulerConfigFromYaml(configYaml, ensureBringupTargets: true, slug: commit.slug);
```
CaseyHillers, line 235: Instead of passing `RepositorySlug`, we should actually have this validate against `CiYaml` which contains this protobuf. `CiYaml` already contains the extra state necessary to do this validation, like repository slug and branch.

XilaiZhang: Umm maybe my understanding is not correct, I think the most common way to obtain a CiYaml class is to call the scheduler.getCiYaml(commit) function in scheduler.dart. But I think `getCiYaml` function calls on the underlying `_downloadCiYaml` function, which relies on `githubFileContent` function to generate a scheduler config. My understanding is that the dependency goes like `getCiYaml` --depends on--> `_downloadCiYaml` --depends on--> `githubFileContent`. where `getCiYaml` produces CiYaml and `_downloadCiYaml` produces schedulerConfig.

Maybe I am wrong, but if we are supplying CiYaml class in `_downloadCiYaml`, I thought we are using CiYaml class which is produced by scheduler config which is eventually produced by the function itself. And therefore we have a dependency loop.

CaseyHillers: One way would be to move this `schedulerConfigFromYaml` directly into `CiYaml` as a named constructor. `CiYaml` is a relatively new class, so that's why it wasn't in there in the first place. You can still keep the http requests outside of `CiYaml` by having `CiYaml.fromYaml(YamlMap yaml);`.

For the cache section, you can still access the underlying proto via `ciYaml.value`

XilaiZhang: Thanks for the advice! I experimented around and umm my current plan is to move the entire logic of getCiYaml and _downloadCiYaml into under CiYaml class. I figured that I would have a new `.fromYaml` function in CiYaml which is a wrapper of githubFileContent and takes in parameters such a slug and branch. I am not entirely sure about the cache section, my understanding is that we create a `.value` function in CiYaml which wraps the caching logic from getCiYaml, and therefore move everything from getCiYaml and _downloadCiYaml to CiYaml class as well. 

For schedulerConfigFromYaml as a named constructor inside CiYaml, I am not sure how to supply a CiYaml parameter from within CiYaml. My current plan is to still use slug information, and supply it with the slug and branch information obtained from within the CiYaml class. And for other classes depending on just the schedulerConfig portion of the CiYaml class, I am planning to provide a CiYaml class with every field set to null except the schedulerConfig part. Maybe I misunderstood, does it sound correct to move everything into CiYaml, and to still supply branch and slug information to these functions, after they have become CiYaml class functions? Thank you!

CaseyHillers: We want to minimize deps in models (that ways its easier to share the models in other code bases), so avoid adding the http work to `CiYaml` The caching logic will remain in `Scheduler`.

I don't understand your second point. Can you clarify, or show a snippet of the issue you're running into?

XilaiZhang:
The newest commit has the following implementations: 
the schedulerConfigFromYaml function resides in the CiYaml class, and the parameters it takes are 1 YamlMap of currentConfigYaml, and 2 optional parameter: a boolean of ensureBringupTargets, and a YamlMap of totConfigYaml. The logic of verifications that schedulerConfigFromYaml relies on are also moved to be under CiYaml class.

The implementation is based on the following considerations:
1. We want to keep the http request outside of CiYaml class even though schedulerConfigFromYaml is in CiYaml. If we pass in slug or branch information or another CiYaml parameter to the schedulerConfigFromYaml function inside CiYaml class. Then the http request will be happening inside schedulerConfigFromYaml, and therefore inside CiYaml class, which contradicts with our assumption. So I am proposing the workaround to have the YamlMap information pre-fetched before passing into schedulerConfigFromYaml, and therefore schedulerConfigFromYaml takes in only YamlMap parameters and a boolean.
2.  I combined the suggestions and kept caching inside scheduler. Did not implement `.formYaml` and `.value` functions for CiYaml classes, and did not implement schedulerConfigFromYaml as a named constructor. (Mostly because I am not very sure if my understandings of these suggestions are correct. And further review is needed on the current approach)

XilaiZhang:
Thanks for the advice! I have refactored schedulerConfigYaml and its friends into a named constructor under CiYaml class. I might have talked about this before, (but just to verify again and sorry for the verbosity,) for other classes interested in only using the schedulerConfig portion of the CiYaml class, I provide a CiYaml class through the named constructor, with fields set to dummy values except the schedulerConfig part. i.e., currently for named constructor schedulerConfigYaml, I supply 'branch' as a default empty string, 'slug' as defaulted to 'flutter/fake' (some gibberish I made up), and only schedulerConfigYaml is something meaningful.

CaseyHillers:
>  for other classes interested in only using the schedulerConfig portion of the CiYaml class,

This is tech debt. They should be migrated to use `CiYaml`. Very rarely do they need the underlying proto values

app_dart/lib/src/model/ci_yaml/ci_yaml.dart
```diff
@@ -18,14 +19,35 @@ class CiYaml {
...
 
   /// The [RepositorySlug] that [config] is from.
   final RepositorySlug slug;
 
   /// The git branch currently being scheduled against.
   final String branch;
 
+  /// Load [currentConfigYaml] (and optionally [totConfigYaml] ) to [pb.SchedulerConfig] and validate the dependency graph.
+  /// where totYaml means tip of tree Yaml.
+  CiYaml.schedulerConfigFromYaml(YamlMap? currentConfigYaml,
```
CaseyHillers, line 32:
```dart
  CiYaml.fromYaml(YamlMap? currentConfigYaml,
```

app_dart/lib/src/model/ci_yaml/ci_yaml.dart
```diff
@@ -18,14 +19,35 @@ class CiYaml {
...
+      {YamlMap? totConfigYaml, bool ensureBringupTargets = false})
```
CaseyHillers, line 33: If we have to pass in `totConfigYaml`, I would just remove `ensureBringupTarget` as run the check whenever the tot config is passed

app_dart/lib/src/model/ci_yaml/ci_yaml.dart
```diff
@@ -97,4 +119,61 @@ class CiYaml {
 
     return regexp.hasMatch(branch);
   }
+
+  void _validateSchedulerConfig(pb.SchedulerConfig schedulerConfig, {pb.SchedulerConfig? totConfig}) {
```
CaseyHillers, line 123: Can you add a dart doc here with an overview of the validations?

CaseyHillers: A dart doc is a comment in the code

See https://dart.dev/guides/language/effective-dart/documentation

XilaiZhang: Sorry for the misunderstanding, I have removed the google doc link and added in a real dart doc

app_dart/lib/src/model/ci_yaml/ci_yaml.dart
```diff
@@ -18,14 +19,34 @@ class CiYaml {
...
-  final pb.SchedulerConfig config;
+  pb.SchedulerConfig config;
 
   /// The [RepositorySlug] that [config] is from.
   final RepositorySlug slug;
 
   /// The git branch currently being scheduled against.
   final String branch;
 
+  /// Load [currentConfigYaml] (and optionally [totConfigYaml] ) to [pb.SchedulerConfig] and validate the dependency graph.
```
CaseyHillers, line 30: The format for dart docs is a single, clear sentence. Then follow up with specific details

```dart
/// Creates [CiYaml] from a [YamlMap].
///
/// If [totConfigYaml] is passed, the validation will verify no new targets have been added that may temporarily break the LUCI infrastructure (such as new prod or presubmit targets).
```

app_dart/lib/src/model/ci_yaml/ci_yaml.dart
```diff
@@ -18,14 +19,34 @@ class CiYaml {
...
+  /// where totYaml means tip of tree Yaml.
+  CiYaml.fromYaml(YamlMap? currentConfigYaml, {YamlMap? totConfigYaml})
```
CaseyHillers, line 32: Would it be easier to pass totConfig as a CiYaml instead of the YamlMap?

app_dart/lib/src/model/ci_yaml/ci_yaml.dart
```diff
@@ -18,14 +19,34 @@ class CiYaml {
...
+      : branch = '',
+        slug = RepositorySlug('flutter', 'fake'),
```
CaseyHillers, line 34: Don't make these fake values. These should be required (following something similar as the existing constructor)

app_dart/lib/src/model/ci_yaml/ci_yaml.dart
```diff
@@ -18,14 +19,34 @@ class CiYaml {
...
+        config = pb.SchedulerConfig() {
+    config.mergeFromProto3Json(currentConfigYaml);
+
+    // check for new builders and compare to tip of tree,
```
CaseyHillers, line 38: This comment reads like it should be deleted as it was written in _validateSchedulerConfig

app_dart/lib/src/model/ci_yaml/ci_yaml.dart
```diff
@@ -97,4 +118,74 @@ class CiYaml {
 
     return regexp.hasMatch(branch);
   }
+
+  /// validate [schedulerConfig], extracted from ci.yaml files provided by user
```
CaseyHillers, line 122: Use full sentences and proper grammar in dart docs

CaseyHillers, line 122: In this case, it would be clearer to refer to the data structure instead of the instance

[schedulerConfig] -> [pb.SchedulerConfig]

app_dart/lib/src/model/ci_yaml/ci_yaml.dart
```diff
@@ -97,4 +118,74 @@ class CiYaml {
...
+  ///   2. It has at least one [branch] in [schedulerConfig.enabledBranches]
+  ///   3. If [totConfig] is passed in, compare current list of [pb.Target] with the list of
+  ///   [pb.Target] from tip of the tree config [totConfig]. If a [pb.Target] is indentified
+  ///   as a new target compared to target list from tip of the tree. The new target should
+  ///   have its field [pb.Target.bringup] set to true.
+  ///   4. no cycle should exist in the dependency graph, as tracked by map [targetGraph]
+  ///   5. [pb.Target] should not depend on self
+  ///   6. [pb.Target] cannot have more than 1 dependency
+  ///   7. [pb.Target] should depend on target that already exist in depedency graph, and already recorded in map [targetGraph]
+  void _validateSchedulerConfig(pb.SchedulerConfig schedulerConfig, {pb.SchedulerConfig? totConfig}) {
```
CaseyHillers, line 135: Since this is in CiYaml, we should simplify the method call to `_validate`

XilaiZhang:
The fromYaml constructor now takes a Yaml Map and an optional CiYaml class, to create a CiYaml class. The optional CiYaml class contains tip of tree targets information, and if necessary, the input CiYaml parameter is constructed with dummy slug and branch values. Other minor modifications are included for review as well.

app_dart/bin/generate_jspb.dart
```diff
@@ -52,20 +53,18 @@ String getLocalConfigContent(String path) {
 }
```
CaseyHillers, line 53: getLocalFileContent should be deleted now that it's no longer used

app_dart/integration_test/validate_all_ci_configs_test.dart
```diff
@@ -26,6 +26,13 @@ final List<SupportedConfig> configs = <SupportedConfig>[
 ];
 
 Future<void> main() async {
+  late CiYaml? totConfig;
+
+  setUp(() {
+    totConfig = CiYaml(
```
CaseyHillers, line 32: Prefer using the test helper `generateCiYaml` from entity generators.

app_dart/lib/src/model/ci_yaml/ci_yaml.dart
```diff
@@ -119,20 +115,21 @@ class CiYaml {
     return regexp.hasMatch(branch);
   }
 
-  /// validate [schedulerConfig], extracted from ci.yaml files provided by user
+  /// validates [pb.SchedulerConfig] extracted from [CiYaml] files.
```
CaseyHillers, line 118: nit: Capital V to start a sentence

app_dart/lib/src/request_handlers/update_existing_flaky_issues.dart
```diff
@@ -38,8 +38,9 @@ class UpdateExistingFlakyIssue extends ApiRequestHandler<Body> {
     final GithubService gitHub = config.createGithubServiceWithToken(await config.githubOAuthToken);
     final BigqueryService bigquery = await config.createBigQueryService();
     final YamlMap? ci = loadYaml(await gitHub.getFileContent(slug, kCiYamlPath)) as YamlMap?;
-    // current pull is already from default branch, and therefore no need to check for new builders
-    final pb.SchedulerConfig schedulerConfig = CiYaml.fromYaml(ci).config;
+    final CiYaml totConfig = CiYaml(
+        config: pb.SchedulerConfig(), slug: Config.flutterSlug, branch: Config.defaultBranch(Config.flutterSlug));
+    final pb.SchedulerConfig schedulerConfig = CiYaml.fromYaml(ci, totConfig).config;
```
CaseyHillers, line 44: In this case, we don't need to validate the tot config

app_dart/lib/src/service/scheduler.dart
```diff
@@ -229,6 +229,7 @@ class Scheduler {
     );
     final YamlMap configYaml = loadYaml(configContent) as YamlMap;
     pb.SchedulerConfig schedulerConfig = pb.SchedulerConfig();
+    pb.SchedulerConfig totSchedulerConfig = pb.SchedulerConfig();
```
CaseyHillers, line 232: This should get passed in directly as a CiYaml so we're not duplicating code in this method

app_dart/lib/src/service/scheduler.dart
```diff
@@ -376,12 +382,14 @@ class Scheduler {
   ///
   /// In the case there is an issue getting the diff from GitHub, all targets are returned.
   Future<List<Target>> getPresubmitTargets(github.PullRequest pullRequest) async {
+    //
```
CaseyHillers, line 385: nit: rm empty comment

app_dart/bin/generate_jspb.dart
XilaiZhang: umm I added that part back in to pass an integration test that checks for local yaml validation. I was thinking of replacing the old integration test with a new test that validates against tip of tree, but then realized we would have to keep the jspb expectation file updated all the time, since tip of tree yaml file could change over time. Shall we remove the entire integration test that validates local yaml as well?

app_dart/lib/src/service/scheduler.dart
XilaiZhang: umm sorry for asking for clarification, by "pass directly as CiYaml" are we trying to concatenate smaller pieces of code into a one liner? Maybe my understanding is wrong, I thought that the slug, branch and schedulerConfig have to be initialized beforehand to create a CiYaml class.

XilaiZhang:
The newest commit refactors and restructures files around the following ideology:
```dart
totCiYaml = getCiYaml(totCommit);
ciYaml = getCiYaml(commit, totCiYaml: totCiYaml);
```
1. CiYaml is the major class type to carry information around.
CiYaml.fromYaml now takes only CiYaml type parameters, and also returns CiYaml type.
2. [getYaml] is the wrapper function of the underlying validation logic. 
getYaml takes in commit and CiYaml, and calls upon CiYaml.fromYaml to perform the validation check.
If the commit passed into [getYaml] has an empty sha, getYaml will pull CiYaml information against default branch of tip of tree.
Otherwise, getYaml pulls CiYaml information from the given sha.
Functions that obtain CiYaml through [getYaml], such as addPullRequests and checkPresubmitTargets, will perform necessary checks to determine if the current branch is a release branch, and then determine whether to pull against tip of tree.
3. helper function [generateTotCommit] is added under lib/src/foundation/utils to assist [getYaml].
[generateTotCommit] generates commit with an empty sha. A commit with empty sha instructs [getYaml] function to use default branch to pull afainst tip of tree, instead of using a sha.
4. [fake_scheduler] has access to the real [scheduler.getYaml] function.
files such as generate_jspb.dart and validate_scheduler_config.dart, use a fake scheduler to invoke the [realGetYaml] function. While other files use fake scheduler to invoke a dummy [getYaml] function.
5. backward compatibility to maintain functions to process local files
Some files, such as generate_jspb and validate_scheduler_config, have integration test that validate against local CiYaml files. For this reason, we maintain the ability to construct CiYaml file from YamlMap parameters, to process local CiYaml files when they are supplied as YamlMap types.

app_dart/lib/src/service/scheduler.dart
XilaiZhang: Thanks for the example! This is very helpful! I have refactored files implementing this topology

app_dart/bin/generate_jspb.dart
```diff
@@ -52,20 +58,46 @@ String getLocalConfigContent(String path) {
...
     configContent = getLocalConfigContent(args[0]);
   }
-  final YamlMap configYaml = loadYaml(configContent) as YamlMap;
-  // There's an assumption that we're only generating builder configs from commits that
-  // have already landed with validation. Otherwise, this will fail.
-  final pb.SchedulerConfig schedulerConfig = schedulerConfigFromYaml(configYaml);
+
+  pb.SchedulerConfig schedulerConfig;
+  if (args.length == 3) {
+    final FakeScheduler scheduler = FakeScheduler(config: FakeConfig());
```
CaseyHillers, line 75: Don't use fakes in tooling

app_dart/bin/generate_jspb.dart
```diff
@@ -52,20 +58,46 @@ String getLocalConfigContent(String path) {
...
+    // FOR REVIEW:
+    // requiring branch information from user as well to check whether current branch is a release branch.
+    if (args[2] == Config.defaultBranch(RepositorySlug('flutter', args[0]))) {
+      // FOR REVIEW:
+      // supply 0 in generateCommit, to singal that we generate a commit with an empty sha
+      Commit totCommit =
+          generateTotCommit(0, repo: args[1], branch: Config.defaultBranch(RepositorySlug('flutter', args[0])));
+      // There's an assumption that we're only generating builder configs from commits that
+      // have already landed with validation. Otherwise, this will fail.
+      CiYaml totConfig = await scheduler.getRealCiYaml(totCommit);
```
CaseyHillers, line 87: I don't see a need for a `getRealCiYaml`. Why can't we use `getCiYaml`?

CaseyHillers (commented):
This mostly lgtm. I would say:
1. Remove the need for `getRealCiYaml` and using a fake scheduler
2. Only pass a tot ciyaml in the case where we're running presubmit

app_dart/bin/generate_jspb.dart
```diff
@@ -52,20 +58,46 @@ String getLocalConfigContent(String path) {
...
+  if (args.length == 3) {
+    final FakeScheduler scheduler = FakeScheduler(config: FakeConfig());
+    Commit currentCommit = generateCommit(1, repo: args[0], sha: args[1], branch: args[2]);
+
+    // FOR REVIEW:
+    // requiring branch information from user as well to check whether current branch is a release branch.
+    if (args[2] == Config.defaultBranch(RepositorySlug('flutter', args[0]))) {
+      // FOR REVIEW:
+      // supply 0 in generateCommit, to singal that we generate a commit with an empty sha
+      Commit totCommit =
```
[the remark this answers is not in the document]
XilaiZhang: Thanks for the reply! 

Umm the reason I am using a fake scheduler is that, currently [getCiYaml] function is under [Scheduler] class. To access [getCiYaml] in tooling files, I am avoiding creating a real scheduler which requires cache, luckBuildService etc., which might also need fakes values to get initialized. And based on previous learnings I did not to convert [getCiYaml] to a static function. There might be a better way, but eventually I decided to use a fakeScheduler to give tooling files access to [getCiYaml].

For the validate_scheduler_config.dart file, it is also a tooling but from the name of it I am assuming it would need to trigger the actual validation tests by calling upon getCiYaml. Currently I am avoiding creating a real scheduler (which needs required parameters such as cache, luciBuildServices etc.), and instead using fakeScheduler.getRealCiYaml to run the validation check. The getCiYaml function under fakeScheduler does not run the underlying validation checks, and I am using it for other files such as update_existing_flaky_issues. 

umm I guess I was not very sure how to give tooling files access to [scheduelr.getCiYaml] without creating fakes or static functions.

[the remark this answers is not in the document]
XilaiZhang: umm just in case it gets buried in other emails. Would be great if you can take a look the above comment from yesterday as well. Thank you!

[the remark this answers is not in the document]
CaseyHillers: Again, I want to emphasize never to use fakes or to create dummy values to work around issues. It's a code smell / red flag, and we can talk about it more in our next 1:1

We should leave the logic as is (downloading the yaml map and pass it ourselves). This tooling does not need the extra caching or datastore logic that `Scheduler` introduces (and it wouldn't work on LUCI bots). So instead we should directly use `CiYaml`

`validate_scheduler_config.dart` is a local way to run the validation, but it's not used in CI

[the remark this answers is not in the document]
XilaiZhang: Thanks a lot for the reply! I have a better understanding of it now and will go ahead and fix it. Thank you!

XilaiZhang:
Thanks a lot for providing the insights! The newest commit cleans up toolings. Previous "FOR REVIEW" sections have been removed, assuming the approaches taken are legit.

app_dart/bin/generate_jspb.dart
```diff
@@ -63,9 +64,11 @@ Future<void> main(List<String> args) async {
...
     configContent = getLocalConfigContent(args[0]);
   }
+
+  pb.SchedulerConfig schedulerConfig;
   final YamlMap configYaml = loadYaml(configContent) as YamlMap;
-  // There's an assumption that we're only generating builder configs from commits that
-  // have already landed with validation. Otherwise, this will fail.
-  final pb.SchedulerConfig schedulerConfig = schedulerConfigFromYaml(configYaml);
+  CiYaml currentYaml = generateCiYamlFromYamlMap(configYaml);
+  schedulerConfig = CiYaml.fromYaml(currentYaml).config;
```
CaseyHillers, line 71: Prefer making variables fine:

```dart
  final CiYaml currentYaml = generateCiYamlFromYamlMap(configYaml);
  final pb.SchedulerConfig schedulerConfig = CiYaml.fromYaml(currentYaml).config;
```

app_dart/bin/validate_scheduler_config.dart
```diff
@@ -4,17 +4,23 @@
...
   final File configFile = File(configPath);
   if (!configFile.existsSync()) {
-    print('validate_scheduler_config.dart configPath');
+    print('please provide a valid file path as configPath');
     exit(1);
   }
 
   final YamlMap configYaml = loadYaml(configFile.readAsStringSync()) as YamlMap;
-  print(schedulerConfigFromYaml(configYaml));
+  CiYaml currentConfig = generateCiYamlFromYamlMap(configYaml);
```
CaseyHillers, line 24: ditto here about final

app_dart/lib/src/foundation/utils.dart
```diff
@@ -222,3 +226,28 @@ class Pair<S, T> {
   final S first;
   final T second;
 }
+
+CiYaml generateCiYamlFromYamlMap(YamlMap? yamlConfig) {
```
CaseyHillers, line 230: Avoid global functions. Code should call the CiYaml constructors directly

XilaiZhang:
Thanks a lot for the previous reviews! Was a bit caught up in the release but did the newest refactor along the side. The newest commit have CiYaml constructor in its ultimate form, explicit flutterSlug setups, real and fake totCommit generations and necessary accommodations to pass tests.

XilaiZhang:
A placeholder comment to help this PR stay on top of emails notifications, would like to get the newest commit reviewed, thank you!

app_dart/lib/src/service/scheduler.dart
```diff
@@ -183,9 +191,17 @@ class Scheduler {
...
   Future<CiYaml> getCiYaml(
     Commit commit, {
+    CiYaml? totCiYaml,
     RetryOptions retryOptions = const RetryOptions(maxAttempts: 3),
   }) async {
-    final String ciPath = '${commit.repository}/${commit.sha!}/$kCiYamlPath';
+    String ciPath;
+
+    /// Use default branch instead of sha when pulling against tip of tree.
+    if (commit.sha == null) {
```
CaseyHillers, line 200: `commit.sha` should in practice never be null. Can we throw an exception instead?

The gcloud datastore entities in dart require all fields to be nullable so they can be lazily set, but by the time they're used in cocoon it's non-nullable

app_dart/lib/src/service/scheduler.dart
```diff
@@ -198,6 +214,10 @@ class Scheduler {
     ))!;
     final pb.SchedulerConfig schedulerConfig = pb.SchedulerConfig.fromBuffer(ciYamlBytes);
     log.fine('Retrieved .ci.yaml for $ciPath');
+    // If totCiYaml is not null, we assume upper level function has verified that current branch is not a release branch.
+    if (totCiYaml != null) {
```
CaseyHillers, line 218: You can remove this if and add the totConfig directly below

app_dart/lib/src/service/scheduler.dart
```diff
@@ -227,11 +247,14 @@ class Scheduler {
       commit.slug,
       '.ci.yaml',
       httpClientProvider: httpClientProvider,
-      ref: commit.sha!,
+      ref: commit.sha == null ? Config.defaultBranch(commit.slug) : commit.sha!,
```
CaseyHillers, line 250: ditto: sha shouldn't be null. If it is we need to throw an exception

app_dart/lib/src/service/scheduler.dart
```diff
@@ -227,11 +247,14 @@ class Scheduler {
...
       retryOptions: retryOptions,
     );
     final YamlMap configYaml = loadYaml(configContent) as YamlMap;
-    return schedulerConfigFromYaml(configYaml).writeToBuffer();
+    pb.SchedulerConfig schedulerConfig = pb.SchedulerConfig();
```
CaseyHillers, line 254: ditto on cascading notation

app_dart/lib/src/service/scheduler.dart
```diff
@@ -124,7 +124,15 @@ class Scheduler {
       return;
     }
 
-    final CiYaml ciYaml = await getCiYaml(commit);
+    late CiYaml ciYaml;
```
CaseyHillers, line 127: We don't need to pass the tot config in this case (only for the presubmit targets)

app_dart/lib/src/service/scheduler.dart
```diff
@@ -369,7 +392,15 @@ class Scheduler {
       repository: pullRequest.base!.repo!.fullName,
       sha: pullRequest.head!.sha,
     );
-    final CiYaml ciYaml = await getCiYaml(commit);
+    late CiYaml ciYaml;
```
CaseyHillers, line 395: Generally, you want to avoid late as loosens the restrictions on the compiler and can lead to bugs. We can rework this to instead have a nullable toCiYaml

```dart
CiYaml? totCiYaml;
if (commit.branch == Config.defaultBranch(commit.slug)) {
  final Commit totCommit = await generateTotCommit(slug: commit.slug, branch: Config.defaultBranch(commit.slug));
  totCiYaml = await getCiYaml(totCommit);
}
final CiYaml ciYaml = getCiYaml(commit, totCiYaml: totCiYaml);
```

app_dart/lib/src/service/scheduler.dart
```diff
@@ -456,4 +487,19 @@ class Scheduler {
       log.warning('Failed to add commits to BigQuery: $ApiRequestError');
     }
   }
+
+  Future<Commit> generateTotCommit({required String branch, required RepositorySlug slug}) async {
```
CaseyHillers, line 491: dartdoc?

app_dart/lib/src/service/scheduler.dart
```diff
@@ -456,4 +487,19 @@ class Scheduler {
...
+    datastore = datastoreProvider(config.db);
+    final BuildStatusService buildStatusService = buildStatusProvider(datastore);
+    final Commit totCommit = (await buildStatusService
+        .retrieveCommitStatus(
+          limit: 1,
+          branch: branch,
+          slug: slug,
+        )
+        .map<Commit>((CommitStatus status) => status.commit)
+        .toList())[0];
```
CaseyHillers, line 501:
```suggestion
        .toList()).single;
```

app_dart/lib/src/service/scheduler.dart
XilaiZhang: this is very artful! impressive!

CaseyHillers (approved):
flutter-lgtm

PR #1647, opened 2022-03-08
[auto_submit] Use SecretManager

CaseyHillers:
This is to support storing the secrets without Datastore.

# Test

I did some local runs and deploys to verify the APIs connected.

# Future work

Upstream this into `package:gcloud`


auto_submit/lib/service/secrets.dart
```diff
@@ -0,0 +1,74 @@
...
+
+  Future<String> get(String key);
+
+  void put(String key, [String? value]);
+}
+
+class CloudSecretManager implements SecretManager {
+  CloudSecretManager();
+
+  final String projectId = Platform.environment['APPLICATION_ID'] ?? 'flutter-dashboard';
```
keyonghan, line 25: Nit: simply use `flutter-dashboard` as this is not local run and these secrets are in this project only.

keyonghan (approved):
LGTM with nit.

auto_submit/lib/service/secrets.dart
CaseyHillers: I want to upstream this into `package:gcloud`, and I'm leaving this as a breadcrumb to sort out how to find this (googleapis surfaces it somewhere). SG?

keyonghan (approved):
LGTM

PR #1648, opened 2022-03-08
Update flaky bugs with both prod and staging stats when `bringup: true`

keyonghan:
Fixes: https://github.com/flutter/flutter/issues/99442

This PR does:
1) update flaky bugs with both prod and staging stats when the builder is `bringup: true`
2) skip build number check, it is useful to present any available test results even we have limited number
3) update tests

### DIFF
--- a/app_dart/lib/src/request_handlers/flaky_handler_utils.dart
+++ b/app_dart/lib/src/request_handlers/flaky_handler_utils.dart
@@ -135,6 +135,8 @@ class IssueUpdateBuilder {
 
   bool get isBelow => statistic.flakyRate < threshold;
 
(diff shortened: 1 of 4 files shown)


chunhtai (approved):
LGTM, if i remember correctly, the bring up test is only ran in stage pool? if so, wouldn't the prod pool states never change?

keyonghan:
> LGTM, if i remember correctly, the bring up test is only ran in stage pool? if so, wouldn't the prod pool states never change?

That's correct. The update here is to catch potential flaky runs in the gap period where the flaky bug has been filed whereas the flake PR hasn't landed (which usually takes days).

PR #1650, opened 2022-03-08
[dashboard] Change canvaskit source to source from build

CaseyHillers:
Fixes http://b/215905909

# Test

Ran a local run and verified the wasm request was to my localhost


yjbanov (approved):
lgtm

PR #1651, opened 2022-03-08
[autosubmit] Add check statuses logic

KristinBi:
To check whether this pull request should be merged or not, here we get more necessary information including author, author assoiciation, and labels. And we will follow the checkStatuses() logic to check whether all statuses are successful.

This pr is the forth step of https://github.com/flutter/cocoon/pull/1633.
This pr solved the issue https://github.com/flutter/flutter/issues/99718, and is intended to partly solve the issue https://github.com/flutter/flutter/issues/98707.


auto_submit/lib/requests/refresh_cirrus_status.dart
```diff
@@ -0,0 +1,85 @@
+// Copyright 2019 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:async';
```
CaseyHillers, line 5: Instead of importing this, we should update the GitHub graphql code to query the check suites directly

See https://github.com/flutter/cocoon/pull/1548 (which was later reverted)

auto_submit/lib/requests/check_pull_request.dart
```diff
@@ -6,12 +6,14 @@ import 'dart:async';
 import 'dart:convert';
 
 import 'package:github/github.dart';
+import 'package:graphql/client.dart' hide Request, Response;
```
CaseyHillers, line 9: See https://github.com/SpinlockLabs/github.dart/blob/master/lib/src/common/checks_service.dart#L146 for listing check runs

auto_submit/lib/requests/refresh_cirrus_status.dart
keyonghan: As we are migrating from graphQL to RestAPI, we should rely on RestAPI to get the Suites/checkruns list.

Actually the `checkruns` API returns all check runs including the cirrus ones. So we are good to remove all the cirrus GraphQL logics.

One TODO is distinguish check runs from `cirrus` or `flutter-dashboard`, so that we can have correct task links in the `_FailureDetail`.

auto_submit/lib/requests/check_pull_request.dart
```diff
@@ -88,6 +105,152 @@ class CheckPullRequest extends RequestHandler {
...
+        if (status.state == 'failure' && !notInAuthorsControl.contains(name)) {
+          failures.add(_FailureDetail(name!, status.targetUrl as String));
+        }
+      }
+    }
+
+    log.info('Validating name: $name, checks: $checkRuns');
+    for (CheckRun checkRun in checkRuns) {
+      final String? name = checkRun.name;
+      if (checkSuite!.conclusion == CheckRunConclusion.success) {
```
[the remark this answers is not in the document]
KristinBi: Hi keyong, thank you for the advice. I also considered to use checkRun!.conclusion previously, but the issue here is that the getter 'conclusion' isn't defined for the type 'CheckRun'.
I searched in the CheckRun class in the github package, and discovered that only CheckSuite class contains conclusion. This is also the reason that I call for the checkSuite here.

auto_submit/lib/requests/check_pull_request.dart
```diff
@@ -88,6 +105,152 @@ class CheckPullRequest extends RequestHandler {
...
+  }
+
+  for (PullRequestReview review in reviews) {
+    // Ignore reviews from non-members/owners.
+    if (!allowedReviewers.contains(review.authorAssociation)) {
+      continue;
+    }
+    // Reviews come back in order of creation.
+    final String? state = review.state;
+    final String? authorloggerin = review.user.login;
```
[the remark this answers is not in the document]
KristinBi: Thank you Keyong, I've changed the name for this variable.

auto_submit/lib/requests/check_pull_request.dart
```diff
@@ -88,6 +105,152 @@ class CheckPullRequest extends RequestHandler {
...
+  /// Returns whether all statuses are successful.
+  ///
+  /// Also fills [failures] with the names of any status/check that has failed.
+  Future<bool> _checkStatuses(
+    RepositorySlug slug,
+    String sha,
+    Set<_FailureDetail> failures,
+    List<RepositoryStatus> statuses,
+    List<CheckRun> checkRuns,
+    CheckSuite? checkSuite,
```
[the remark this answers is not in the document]
KristinBi: As I explained in the following comment, I used checkSuite here to get the conclusion.

auto_submit/lib/requests/check_pull_request.dart
```diff
@@ -60,23 +62,38 @@ class CheckPullRequest extends RequestHandler {
       checkRuns.addAll(await gitHub.getCheckRuns(slug, pr.head!.sha!));
```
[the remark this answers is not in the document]
KristinBi: Sure, the bug for this issue is https://github.com/flutter/flutter/issues/99804. I will also take time to study these two possible solutions to solve this issue.

auto_submit/lib/requests/refresh_cirrus_status.dart
KristinBi: I created a bug to track this TODO: https://github.com/flutter/flutter/issues/99805.

auto_submit/lib/requests/check_pull_request.dart
```diff
@@ -88,6 +105,151 @@ class CheckPullRequest extends RequestHandler {
...
+        failures.add(_FailureDetail(name!, checkRun.detailsUrl as String));
+      }
+      allSuccess = false;
+    }
+
+    // Validate cirrus
+    const List<String> _failedStates = <String>['FAILED', 'ABORTED'];
+    const List<String> _succeededStates = <String>['COMPLETED', 'SKIPPED'];
+
+    //TODO(Kristin): Distinguish check runs from cirrus or flutter-dashboard. https://github.com/flutter/flutter/issues/99805.
```
keyonghan, line 179: Let's put this comment above line 165: `for (CheckRun checkRun in checkRuns) {`, and remove all below cirrus logics.

The distinction part should happen inside the checkrun loop.

KristinBi: Sure, done.

auto_submit/lib/requests/check_pull_request.dart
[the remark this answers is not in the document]
KristinBi: Gotcha Keyong, I will also add a TODO here and submit PRs to the github package.

auto_submit/lib/requests/check_pull_request.dart
keyonghan: Yes. These logics will be obsolete if we switch the check results from cirrus GraphQL to Rest API.

I would recommend to add cirrus related logics when solving: https://github.com/flutter/flutter/issues/99805, where you only need to distinguish the FailureDetail. The check run result check has been taken care of in the line 165 loop.

KristinBi: Gotcha Keyong, thanks. I will remove them here and add related logics later in the correct places.

auto_submit/lib/requests/check_pull_request.dart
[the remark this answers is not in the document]
keyonghan: The current lines 166-175 is not correct, I would prefer to skip it now and wait until the upstream PR lands to create the new logic.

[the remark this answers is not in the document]
KristinBi: Deleted these lined and add a TODO to solve this later. https://github.com/flutter/flutter/issues/99873.

keyonghan (approved):
Thanks for the changes. LGTM!

auto_submit/lib/requests/refresh_cirrus_status.dart
KristinBi: Hi @keyonghan, I discussed with Casey once more on the check runs and found out that it's not necessary to distinguish cirrus to flutter-dashboard.
In the previous logic, the only difference for processing checkRuns and the cirrus is their task links in the _FailureDetail. However, I took a look at the checkRuns script https://api.github.com/repos/flutter/packages/commits/68d068dfda0cf5580290c278bfa3f3b7ff657654/check-runs, and found out that the correct cirrus task link has already been included into the checkRun's details_url. Therefore, we don't need to hardcode its task links as previous logic, we can instead use the checkRun's details_url, the same as line 357 https://github.com/flutter/cocoon/blob/main/app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart#L357

keyonghan: That makes it easier. Thanks for the follow up:-)

PR #1653, opened 2022-03-09
implement shouldMerge logic

KristinBi:
Here we implement the logic of check whether this pull request should merge or not.

This pr is the fifth step of https://github.com/flutter/cocoon/pull/1633.
This pr solved the issue https://github.com/flutter/flutter/issues/98707, and is intended to partly solve the issue https://github.com/flutter/flutter/issues/98707.

# Future work
If this pull request doesn't meet the requirement to be merged, we want to remove the 'autosubmit' label from it. Implement the logic of removing the label is needed.

### DIFF
--- a/auto_submit/lib/requests/check_pull_request.dart
+++ b/auto_submit/lib/requests/check_pull_request.dart
@@ -27,13 +27,16 @@ class CheckPullRequest extends RequestHandler {
     final String rawBody = await request.readAsString();
     final body = json.decode(rawBody) as Map<String, dynamic>;
     final PullRequest pullRequest = PullRequest.fromJson(body['pull_request']);
+    final RepositorySlug slug = pullRequest.base!.repo!.slug();
 
     final GithubService gitHub = await config.createGithubService();
     final _AutoMergeQueryResult queryResult = await _parseQueryData(pullRequest, gitHub);
-    if (await shouldMergePullRequest(queryResult)) {
+    if (await shouldMergePullRequest(queryResult, slug, gitHub)) {
       // TODO(Kristin): Keep pulling pubsub queue, https://github.com/flutter/flutter/issues/98704
 
-    } else {
+    } else if (queryResult.shouldRemoveLabel) {
+      log.info('Removing label for commit: ${queryResult.sha}');
+      await _removeLabel();
       return Response.ok(jsonEncode(<String, String>{}));
     }
 
@@ -41,9 +44,35 @@ class CheckPullRequest extends RequestHandler {
   }
 
   /// Check if the pull request should be merged.
-  Future<bool> shouldMergePullRequest(_AutoMergeQueryResult queryResult) async {
-    // TODO(Kristin): Implement shouldMergePullRequest, https://github.com/flutter/flutter/issues/98707
+  Future<bool> shouldMergePullRequest(
+      _AutoMergeQueryResult queryResult, RepositorySlug slug, GithubService github) async {
+    if (queryResult.shouldMerge) {
+      return true;
+    }
+    // If the PR is a revert of the tot commit, merge without waiting for checks passing.
+    return await isTOTRevert(queryResult.sha, slug, github);
+  }
 
+  /// Check if the `commitSha` is a clean revert of TOT commit.
+  ///
+  /// By comparing the current commit with second TOT commit, an empty `files` in
+  /// `GitHubComparison` validates a clean revert of TOT commit.
+  ///
+  /// Note: [compareCommits] expects base commit first, and then head commit.
+  Future<bool> isTOTRevert(String headSha, RepositorySlug slug, GithubService github) async {
+    final RepositoryCommit secondTotCommit = await github.getRepoCommit(slug, 'HEAD~');
+    log.info('Current commit is: $headSha');
+    log.info('Second TOT commit is: ${secondTotCommit.sha}');
+    final GitHubComparison githubComparison = await github.compareTwoCommits(slug, secondTotCommit.sha!, headSha);
+    final bool filesIsEmpty = githubComparison.files!.isEmpty;
+    if (filesIsEmpty) {
+      log.info('This is a TOT revert. Merge ignoring tests statuses.');
+    }
+    return filesIsEmpty;
+  }
+
+  Future<bool> _removeLabel() async {
+    // TODO(Kristin): Implement the logic to remove the label. https://github.com/flutter/flutter/issues/99877
     return true;
   }
 
--- a/auto_submit/lib/service/github_service.dart
+++ b/auto_submit/lib/service/github_service.dart
@@ -41,4 +41,14 @@ class GithubService {
   ) async {
     return await github.repositories.listStatuses(slug, ref).toList();
   }
+
+  /// Fetches the specified commit.
+  Future<RepositoryCommit> getRepoCommit(RepositorySlug slug, String sha) async {
+    return await github.repositories.getCommit(slug, sha);
+  }
+
+  /// Compares two commits
+  Future<GitHubComparison> compareTwoCommits(RepositorySlug slug, String refBase, String refHead) async {
+    return await github.repositories.compareCommits(slug, refBase, refHead);
+  }
 }


auto_submit/lib/service/github_service.dart
```diff
@@ -26,19 +26,21 @@ class GithubService {
...
   /// Retrieves the statuses of a repository at the specified reference.
   Future<List<RepositoryStatus>> getStatuses(
     RepositorySlug slug,
     String ref,
   ) async {
     return await github.repositories.listStatuses(slug, ref).toList();
   }
+
+  /// Fetches the specified commit.
+  Future<RepositoryCommit> getRepoCommit(RepositorySlug slug, String sha) async {
```
CaseyHillers, line 38: `getRepoCommit` -> `getCommit`

auto_submit/lib/service/github_service.dart
```diff
@@ -26,19 +26,21 @@ class GithubService {
...
+    return await github.repositories.getCommit(slug, sha);
+  }
+
+  /// Compares two commits
```
CaseyHillers, line 42: Can you add some more context?

auto_submit/lib/requests/check_pull_request.dart
```diff
@@ -27,23 +27,52 @@ class CheckPullRequest extends RequestHandler {
...
+  }
+
+  /// Check if the `commitSha` is a clean revert of TOT commit.
+  ///
+  /// By comparing the current commit with second TOT commit, an empty `files` in
+  /// `GitHubComparison` validates a clean revert of TOT commit.
+  ///
+  /// Note: [compareCommits] expects base commit first, and then head commit.
+  Future<bool> isTOTRevert(String headSha, RepositorySlug slug, GithubService github) async {
+    final RepositoryCommit secondTotCommit = await github.getRepoCommit(slug, 'HEAD~');
```
CaseyHillers, line 63: Why are we using `HEAD~` instead of `HEAD`?

We want to compare the pullRequest.base.sha to the repositories HEAD

auto_submit/lib/service/github_service.dart
KristinBi: Thank you Casey, done.

auto_submit/lib/requests/check_pull_request.dart
KristinBi: Hi Casey, I have the same doubts before, i also thought here should be HEAD. However, I noticed that the previous logic https://github.com/flutter/cocoon/blob/main/app_dart/lib/src/request_handlers/check_for_waiting_pull_requests.dart#L117 here used the HEAD~, I thought this was a special case so I just retained what he used.

auto_submit/lib/service/github_service.dart
KristinBi: Sure Casey, I added more context to explain what compareCommits() is doing and what its response means.

auto_submit/lib/requests/check_pull_request.dart
CaseyHillers: Can we add clear documentation? I remember this being confusing last time I reviewed it, and I'm not sure I still understand. It's an issue with the compare files, right?

Here's an example I did against flutter/flutter that shows why I would expect HEAD:

```sh
$ git log -n 2
* 57aa49028d (HEAD -> master, upstream/master) Migrate NavigationRail to Material 3. (#99171)  Darren Austin, 79 minutes ago
* 9291c8ba31 Remove expired ThemeData deprecations (#98578)  Kate Lovett, 80 minutes ago
$ git log HEAD -n 1
* 57aa49028d (HEAD -> master, upstream/master) Migrate NavigationRail to Material 3. (#99171)  Darren Austin, 79 minutes ago
$ git log HEAD~ -n 1
* 9291c8ba31 Remove expired ThemeData deprecations (#98578)  Kate Lovett, 80 minutes ago
```

CaseyHillers:
FYI, the pull request body is not clear (you're linking to same issues, and I can't find what you're referring to by the fifth step)

auto_submit/lib/requests/check_pull_request.dart
keyonghan: To clarify, there is no issue with the compareCommits API. Let's take the provided 2 commits for example:
If we have a clean revert of `57aa49028d`, and suppose the revert has commit sha `abcdefg`:
1) `files` of `github.repositories.compareCommits(slug, '57aa49028d', 'abcdefg')` contains all the changed files from `57aa49028d`, but in the opposite way (`+` => `-`, and `-` => `+)
2) files` of `github.repositories.compareCommits(slug, '9291c8ba31', 'abcdefg')` will be empty, as revert commit `abcdefg` reverts all changes from the TOT `57aa49028d`. And this is what we want to validate.

KristinBi:
> FYI, the pull request body is not clear (you're linking to same issues, and I can't find what you're referring to by the fifth step)

Hi Casey, I updated the body to make it more clear.

auto_submit/lib/requests/check_pull_request.dart
```diff
@@ -27,23 +27,56 @@ class CheckPullRequest extends RequestHandler {
...
 
     return Response.ok(rawBody);
   }
 
   /// Check if the pull request should be merged.
-  Future<bool> shouldMergePullRequest(_AutoMergeQueryResult queryResult) async {
-    // TODO(Kristin): Implement shouldMergePullRequest, https://github.com/flutter/flutter/issues/98707
+  ///
+  /// A pull request should be merged on either cases:
+  /// 1) All tests have finished running and satified basic merge requests
```
CaseyHillers, line 49: I don't think (1) has been added in this PR

auto_submit/lib/requests/check_pull_request.dart
```diff
@@ -27,23 +27,56 @@ class CheckPullRequest extends RequestHandler {
...
+    if (queryResult.shouldMerge) {
+      return true;
+    }
+    // If the PR is a revert of the tot commit, merge without waiting for checks passing.
+    return await isTOTRevert(queryResult.sha, slug, github);
+  }
+
+  /// Check if the `commitSha` is a clean revert of TOT commit.
+  ///
+  /// By comparing the current commit with second TOT commit, an empty `files` in
```
CaseyHillers, line 62: It's still unclear to me why we need the second commit, and not the tot

auto_submit/lib/requests/check_pull_request.dart
KristinBi: Hi Casey, I think queryResult.shouldMerge implements 1). It checks the checkRuns and statuses and also some other basic merge requests like whether CI has run successfully and whether the PR has conflicts or not.
The checkRuns part is still a TODO.

CaseyHillers: Thanks, I didn't see it in the diff (but it was added in a previous PR)

auto_submit/lib/requests/check_pull_request.dart
KristinBi: Hi Casey, I will give you an example here. 
Let's assume we have 3 commits. Commit1 is the TOT, commit2 is the second commit, and commit0 is the commit to be check here and is also the revert of commit1(TOT).
Let's then assume commit2 made a change to add file A,  while commit1 made a change to add file B. Since commit0 is the top, it will include all changes made in previous commits. Since commit0 is the revert commit of TOT, it will revert the changes made by commit1, that is, revert adding file B. As a result, commit0 now only contains the previous changes of adding file A which is made by commit2.
Then, when we compare commit0 to commit1, since commit1 added file B but commit0 didn't, the changes shown in the file will be in the opposite way. When we compare commit0 and commit2, since they both added file A and commit0 didn't contain any other changes made by commit1 now, the file will be empty and we can get to know that commit0 is a revert of TOT.

CaseyHillers: Can you put this in a chart? To recap, you're saying we have:

Tree: 2 -> 1
PR has 0

If the PR is a ToT revert, then when merged we would have
Tree: 2 -> 1 -> 0
In essence, the tree should be 2 equivalent to what is with 2

PS: I'm not sure i'm capturing the content correctly, but just to give you and idea on how to clearly communicate this

KristinBi: Hi Casey, thank you fro giving me an example here. I think you got my idea, but I will also try to use chart to explain it more clearly:

Tree: 2 -> 1
Changes: 2( + fileA), 1(+ fileB)
PR has 0

If the PR 0 is a clean ToT revert:
changes for 0 is: 0(revert +fileB)

then when merged we would have
Tree: 2 -> 1 -> 0
In essence, the status of 0 is equivalent to what is with 2, but different with 1

CaseyHillers (approved):

PR #1654, opened 2022-03-10
[cloudbuild] Allow for 20 versions

CaseyHillers:
Cloud build has recently been failing as we added auto_submit service, and there can be versions stored there. This allows each service to have ~5 versions (we're allotted 30 total). Doubling the limit here to prevent the timeouts.


keyonghan (approved):
LGTM

PR #1655, opened 2022-03-10
[backend] request handler that returns green commits

XilaiZhang:
An attempt to address https://github.com/flutter/flutter/issues/99869, and to create a request handler that returns commit shas of green commits.

A green commit satisfies rules:
1. A commit had all its tests run (at least those that are not in bringup)
2. All those blocking tasks were green

Currently I am returning commits that have all its stages with status [TaskSucceeded]. My understanding of rule 1 is that commit doesn't have [Task.statusInprogress]. And I wasn't very sure what "blocking" tasks are, and assumed that we would fulfill rule 2 if all tasks succeed.

My understanding of the rules could be very off, and I am more than happy to rework.


app_dart/lib/src/request_handlers/get_green_commits.dart
```diff
@@ -0,0 +1,57 @@
...
+    final int lastCommitTimestamp = DateTime.now().millisecondsSinceEpoch;
+
+    final List<String?> greenCommits = await buildStatusService
+        .retrieveCommitStatus(
+          limit: commitNumber,
+          timestamp: lastCommitTimestamp,
+          branch: branch,
+          slug: slug,
+        )
+        .where((CommitStatus status) => status.stages.every((Stage s) => s.taskStatus == Task.statusSucceeded))
```
CaseyHillers, line 49: As is, we'll get 0 commits from this.

You need to filter out the tasks that are marked in bringup

app_dart/bin/server.dart
```diff
@@ -216,6 +216,33 @@ Future<void> main() async {
         delegate: GetStatus(config),
       ),
 
+      /// Returns a list of the commit shas that had green runs.
```
CaseyHillers, line 219: This doc should be in the class.

See https://github.com/flutter/flutter/issues/97675

app_dart/bin/server.dart
```diff
@@ -216,6 +216,33 @@ Future<void> main() async {
...
+      /// 1. A commit had all its tests run (at least those that are not in bringup)
+      /// 2. All those blocking tasks were green
+      /// Green commit shas are returned in the order of commit timestamp.
+      ///
+      /// Parameters:
+      ///   repo: default: 'flutter'. Name of the repository.
+      ///
+      /// GET: /api/public/get-green-commits?repo=$repo
+      ///
+      /// Response: Status: 200 OK
```
CaseyHillers, line 231: I would avoid hard coding a response here, but instead say it's a List<String> of the commit shas

app_dart/bin/server.dart
```diff
@@ -216,6 +216,33 @@ Future<void> main() async {
...
+      /// {
+      ///   "greenCommits":[
+      ///     "d5b0b3c8d1c5fd89302089077ccabbcfaae045e4",
+      ///     "ea28a9c34dc701de891eaf74503ca4717019f829"
+      ///   ]
+      /// }
+      ///
+
+      '/api/public/get-green-commits': CacheRequestHandler<Body>(
```
CaseyHillers, line 240: This isn't going to get called very often, so we can remove the cache wrapper

app_dart/lib/src/request_handlers/get_green_commits.dart
```diff
@@ -0,0 +1,57 @@
...
+        .map<String?>((CommitStatus status) => status.commit.sha)
+        .toList();
+
+    return Body.forJson(<String, List<String?>>{
+      'greenCommits': greenCommits,
```
CaseyHillers, line 54: Why return a Map? You can just return a List<String> which is easier to use

app_dart/test/request_handlers/get_green_commits_test.dart
```diff
@@ -0,0 +1,156 @@
...
+      keyHelper = FakeKeyHelper(applicationContext: clientContext.applicationContext);
+      tester = RequestHandlerTester();
+      config = FakeConfig(keyHelperValue: keyHelper);
+      buildStatusService = FakeBuildStatusService(commitStatuses: <CommitStatus>[]);
+      handler = GetGreenCommits(
+        config,
+        datastoreProvider: (DatastoreDB db) => DatastoreService(config.db, 5),
+        buildStatusProvider: (_) => buildStatusService,
+      );
+      commit1 = Commit(
```
CaseyHillers, line 61: Take advantage of the `entity_generators.dart` class. It allows you to do `generateCommit` and `generateTask` to remove the boiler plate

app_dart/test/request_handlers/get_green_commits_test.dart
```diff
@@ -0,0 +1,156 @@
...
+        status: Task.statusSucceeded,
+      );
+      task5 = Task(
+        key: commit2.key.append(Task, id: 789),
+        commitKey: commit2.key,
+        name: 'Linux C',
+        status: Task.statusSucceeded,
+      );
+
+      stage1 = Stage('cocoon', commit1, [task1, task2], Task.statusSucceeded);
```
CaseyHillers, line 107: Can we avoid using stages? We're looking to remove them soon

app_dart/lib/src/request_handlers/get_green_commits.dart
XilaiZhang: umm Maybe my understanding is not correct, but I think the get() methods in request handlers have to be a valid override of the `Future<Body> get()` function, which requires a `Future<Body>` as a return type. And to create a Body type, the three valid ways are through `Body.forString`, `Body.forStream` and `Body.forJson`, which requires input type as String, Stream<Uint8List?> and dynamic(json) respectively. I am not sure how to cast List<String> into a String or Stream<Uint8List?> type, and so I ended up making it a valid json type (a map, sort of), and then feed it into Body.forJson to create a Body type. umm Would there be a valid creation of Body type if we use List<String> type?

app_dart/bin/server.dart
XilaiZhang: umm Maybe my understanding is not correct, but I think the get() methods in request handlers override the Future<Body> get() function, and returns a Body type, which from my understanding represents a HTTP response body. Correct me if I am wrong, I think if we use a HTTP get request and it succeeds with a response body, my understanding is that we are not hard coding it?

app_dart/test/request_handlers/get_green_commits_test.dart
XilaiZhang: Sure. I have updated validation logic in request handler to be based on individual [Task] instead of [stage]. umm In tests, I might be wrong, but I think given the current topology of accessing tasks from commitStatus: [commitStatus] -> List<stage> -> List<task> , it would be necessary to create [stage] variables as intermediate bridges to retrieve the underlying [Task] variables?

XilaiZhang:
Thanks a lot for the instructions! I resolved part of them and left comments for the rest. Would be great if you can take a look at the comments and provide some insights, thank you!

app_dart/lib/src/request_handlers/get_green_commits.dart
CaseyHillers: List<String> is valid json. You can use Body.fromJson(listStringVar)

app_dart/bin/server.dart
CaseyHillers: You're hard coding an example response in the documentation. Prefer giving the type returned instead as it's more concise and obvious to update when we change the logic here

app_dart/test/request_handlers/get_green_commits_test.dart
CaseyHillers: The tests are fine then

XilaiZhang:
Thanks a lot for the reply! I have now removed example response and used List type

app_dart/lib/src/request_handlers/get_green_commits.dart
XilaiZhang: Thanks for the hint! I have now casted Stream.single to List<dynamic> and then to List<String> to make it work for List type.

app_dart/bin/server.dart
XilaiZhang: I see. My bad, I misunderstood it as my response was hard coded instead of generated.

app_dart/lib/src/request_handlers/get_green_commits.dart
```diff
@@ -0,0 +1,74 @@
...
+import 'package:github/github.dart';
+import 'package:meta/meta.dart';
+
+import '../request_handling/body.dart';
+import '../request_handling/request_handler.dart';
+import '../service/build_status_provider.dart';
+import '../service/config.dart';
+import '../service/datastore.dart';
+
+/// Returns a list of the commit shas that had green runs.
```
CaseyHillers, line 18: Take advantage of the dart docs format

```suggestion
/// Returns [List<String>] of commit shas that had all passing tests.
```

app_dart/lib/src/request_handlers/get_green_commits.dart
```diff
@@ -0,0 +1,74 @@
...
+///
+/// A green commit is used to help the release tooling find commits Flutter infrastructure has validated. The rules are:
```
CaseyHillers, line 20: The terminology here should use `[CommitStatus]` as it's an aggregate of the `[List<Task>]` for a `[Commit]`

app_dart/lib/src/request_handlers/get_green_commits.dart
```diff
@@ -0,0 +1,74 @@
...
+/// 1. A commit had all its tests run (at least those that are not in bringup)
+/// 2. All those blocking tasks were green
+/// Green commit shas are returned in the order of commit timestamp.
```
CaseyHillers, line 23: What order?

app_dart/test/request_handlers/get_green_commits_test.dart
```diff
@@ -0,0 +1,126 @@
...
+void main() {
+  group('GetStatus', () {
+    late FakeConfig config;
+    FakeClientContext clientContext;
+    FakeKeyHelper keyHelper;
+    FakeBuildStatusService buildStatusService;
+    late RequestHandlerTester tester;
+    late GetGreenCommits handler;
+
+    late Commit commit1;
```
CaseyHillers, line 31: Since this test doesn't change data, we should make these final instead of late

app_dart/test/request_handlers/get_green_commits_test.dart
```diff
@@ -0,0 +1,126 @@
...
+import 'package:test/test.dart';
+
+import '../src/datastore/fake_config.dart';
+import '../src/request_handling/fake_authentication.dart';
+import '../src/request_handling/request_handler_tester.dart';
+import '../src/service/fake_build_status_provider.dart';
+import '../src/utilities/entity_generators.dart';
+
+void main() {
+  group('GetStatus', () {
```
CaseyHillers, line 23: nit: GetGreenCommits

XilaiZhang:
Thanks for the reply! I have moved late variables out of `Setup()` and make them final. I have added explanation on the order of returned shas, replaced the wording of `green` with `passing all tests`, and added more brackets to detail the data types.

app_dart/lib/src/request_handlers/get_green_commits.dart
```diff
@@ -0,0 +1,78 @@
...
+
+  @override
+  Future<Body> get() async {
+    final String repoName = request!.uri.queryParameters[kRepoParam] ?? Config.flutterSlug.name;
+    final RepositorySlug slug = RepositorySlug('flutter', repoName);
+    final String branch = Config.defaultBranch(slug);
+    final DatastoreService datastore = datastoreProvider(config.db);
+    final BuildStatusService buildStatusService = buildStatusProvider(datastore);
+    final int commitNumber = config.commitNumber;
+    final int lastCommitTimestamp = DateTime.now().millisecondsSinceEpoch;
```
CaseyHillers, line 57: Is `lastCommitTimestamp` needed?

app_dart/test/request_handlers/get_green_commits_test.dart
```diff
@@ -0,0 +1,113 @@
...
+        buildStatusProvider: (_) => buildStatusService,
+      );
+    });
+
+    test('no green commits', () async {
+      final List<String?> result = (await decodeHandlerBody())!;
+      expect(result, isEmpty);
+    });
+
+    test(
```
CaseyHillers, line 74: It's best practice to keep your tests small, and only validate a specific aspect. For this, you can tell this test is doing too much by having "and" and "but" mulitple times in the description. In these cases, either split into smaller tests or wrap in ` group`

app_dart/lib/src/request_handlers/get_green_commits.dart
```diff
@@ -54,12 +53,11 @@ class GetGreenCommits extends RequestHandler<Body> {
     final DatastoreService datastore = datastoreProvider(config.db);
     final BuildStatusService buildStatusService = buildStatusProvider(datastore);
     final int commitNumber = config.commitNumber;
-    final int lastCommitTimestamp = DateTime.now().millisecondsSinceEpoch;
 
     final List<String?> greenCommits = await buildStatusService
         .retrieveCommitStatus(
           limit: commitNumber,
-          timestamp: lastCommitTimestamp,
+          timestamp: DateTime.now().millisecondsSinceEpoch,
```
CaseyHillers, line 60: You don't need to pass timestamp (by default it'll be the current time).

The reason we want to avoid this is using DateTime.now() will cause flakiness in tests

app_dart/test/request_handlers/get_green_commits_test.dart
```diff
@@ -86,11 +84,48 @@ void main() {
...
+      );
+
+      final List<String?> result = (await decodeHandlerBody())!;
+
       expect(result.length, 1);
       expect(result, <String>['d5b0b3c8d1c5fd89302089077ccabbcfaae045e4']);
     });
 
-    test('Also select green commits that include failed tasks but have bringup: true label', () async {
+    test('Also return commits with both flaky and succeeded tasks', () async {
```
CaseyHillers, line 128: nit: rm Also

CaseyHillers (commented):
LGTM. Just some last comments

CaseyHillers (approved):

XilaiZhang:
thanks for the previous approval! I added in a few lines to clean up the side effect of removing timestamp, would like to get them verified before I merge, thank you!

CaseyHillers:
flutter-lgtm

PR #1656, opened 2022-03-10
[scheduler] Increment task attempts and reset status on reruns

CaseyHillers:
This is to fix the attempts on Cocoon scheduled tasks.

## Example

image

## Issue

https://github.com/flutter/flutter/issues/92300


keyonghan (approved):
LGTM

PR #1657, opened 2022-03-10
[scheduler] Add auto-retries on pubsub notifications

CaseyHillers:
https://github.com/flutter/flutter/issues/92300

This is dependent on https://github.com/flutter/cocoon/pull/1656 (the test will pass after it lands)

### DIFF
--- a/app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
+++ b/app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
@@ -4,6 +4,8 @@
 
 import 'dart:convert';
 
+import 'package:cocoon_service/ci_yaml.dart';
+import 'package:cocoon_service/src/service/luci_build_service.dart';
 import 'package:gcloud/db.dart';
 import 'package:meta/meta.dart';
 
@@ -18,6 +20,7 @@ import '../service/cache_service.dart';
 import '../service/config.dart';
 import '../service/datastore.dart';
 import '../service/logging.dart';
+import '../service/scheduler.dart';
 
 /// An endpoint for listening to build updates for postsubmit builds.
 ///
@@ -28,11 +31,13 @@ import '../service/logging.dart';
 @immutable
 class PostsubmitLuciSubscription extends SubscriptionHandler {
   /// Creates an endpoint for listening to LUCI status updates.
-  const PostsubmitLuciSubscription(
-    CacheService cache,
-    Config config, {
+  const PostsubmitLuciSubscription({
+    required CacheService cache,
+    required Config config,
     AuthenticationProvider? authProvider,
     @visibleForTesting this.datastoreProvider = DatastoreService.defaultProvider,
+    required this.luciBuildService,
+    required this.scheduler,
   }) : super(
           cache: cache,
           config: config,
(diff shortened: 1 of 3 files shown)


app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
```diff
@@ -81,6 +88,20 @@ class PostsubmitLuciSubscription extends SubscriptionHandler {
     await datastore.insert(<Task>[task]);
     log.fine('Updated datastore');
 
+    if (task.status == Task.statusFailed || task.status == Task.statusInfraFailure) {
+      log.fine('Trying to auto-retry...');
```
keyonghan, line 92: Does this also handle the manual retry?

app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
```diff
@@ -28,11 +31,13 @@ import '../service/logging.dart';
...
-    CacheService cache,
-    Config config, {
+  const PostsubmitLuciSubscription({
+    required CacheService cache,
+    required Config config,
     AuthenticationProvider? authProvider,
     @visibleForTesting this.datastoreProvider = DatastoreService.defaultProvider,
+    required this.luciBuildService,
+    required this.scheduler,
   }) : super(
```
keyonghan, line 41: Are we using this topic `luci-postsubmit` anywhere now?

app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
CaseyHillers: `ResetProdTask` is for manual retries, but it uses the same `LuciBuildService.checkRerunBuilder` call (granted it can ignore the default checks)

app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
CaseyHillers: Yes, it's used for targets that are using the cocoon scheduler (so cocoon, plugins, packages, and the engine's ci.yaml roller)

keyonghan: Could you point me to the locations referencing this topic? I am not able to find..

CaseyHillers: For scheduling post-submit builds, we add it here : https://cs.opensource.google/flutter/cocoon/+/main:app_dart/lib/src/service/luci_build_service.dart;l=503

keyonghan: The topic here is `luci-postsubmit`, where the referenced line uses topic `luci-builds-prod`. Is this intended?

CaseyHillers: There's two parts to pubsub:

1. Topics. This is generic, and requires a subscription
2. Subscription. There can be N of these per topic to support multiple consumers

With that said, BuildBucket publishes the messages to the `luci-postsubmit` topic, and Cocoon has a single subscription of `luci-builds-prod` to get the messages. In the future, someone might add another subscription dedicated for the results processing aspect

keyonghan: Ah, that makes sense. Thanks for the explanations.

keyonghan (approved):
LGTM

PR #1658, opened 2022-03-11
[dashboard] fix apply state, non-nullable fieldname for filter property

XilaiZhang:
An attempt to fix https://github.com/flutter/flutter/issues/99937#event-6221498414. When filtering by properties, the fieldName is now required to be non nullable. In code, this is manifested by changing fieldName of ValueFilterProperty in filter_property_sheet.dart to be a required non nullable String. Tested on dev dashboard, and queries, filters and apply button works fine. However, duplicate commits are detected when querying on dev dashboard. I am assuming duplicate commits are unrelated to this issue and could be related to the way cocoon generate fakes.


XilaiZhang:
fix https://github.com/flutter/flutter/issues/99938#event-6221509121 on top of this pr as well, so that we don't have to wait until the previous pr gets merge. In the newest commit, we move branch and repo dropdowns to appbar when the screen size is decently large, and keep the slug selection dropdowns under settings icon when screen size is small.

dashboard/test/build_dashboard_page_test.dart
```diff
@@ -30,8 +30,19 @@ void main() {
...
+    items: const <DropdownMenuItem<String>>[],
+  ).runtimeType;
+
   setUp(() {
-    binding.window.physicalSizeTestValue = const Size(1000, 2000);
+    binding.window.devicePixelRatioTestValue = 1.0;
+    binding.window.physicalSizeTestValue = const Size(1080, 2280);
+    // FOR REVIEW:
+    // device pixel ratio of 1.0 works well on web app and emulator
+    // If not set, flutter test uses a Pixel 4 device pixel ratio of roughly 2.75, which doesn't quite work
```
CaseyHillers, line 43: Can you send a follow up PR to remove the FOR REVIEW?

dashboard/lib/build_dashboard_page.dart
```diff
@@ -107,56 +109,18 @@ class BuildDashboardPageState extends State<BuildDashboardPage> {
...
-                ),
+                if (_smallScreen) ..._slugSelection(context, _buildState),
                 TextButton(
                   child: const Text('Vacuum GitHub Commits'),
                   onPressed: _buildState.refreshGitHubCommits,
                 ),
-                FilterPropertySheet(_filter),
+                Row(
+                  children: [
+                    Expanded(child: Center(child: FilterPropertySheet(_filter))),
```
flar, line 121: Is this Expanded the reason that the filter dialog now covers the entire page? This means that you can't mouse around or click on items (particular row/column headers) with the dialog open.

XilaiZhang: yeah sorry about that, was a side effect of me trying to center the dialog but now it covers the entire screen. Was wondering if there is a good way to center it

flar: Isn't Center enough? Its child is a Center.

XilaiZhang: Yeah I thought that way previously but Center itself does not quite work, not sure why

flar: Are you looking into this or should I file an issue so we don't forget? (Would that be a flutter/flutter issue?)

flar: Hmmm, it would be nice if it was repositionable.

XilaiZhang: yes sir feel free to file an issue so that we can address it later. Oh I am not looking at it now. I am currently doing a release. The last time I looked at it I wasn't very successful in getting it reposition-able. Might need some help from my mentor @CaseyHillers

XilaiZhang: Hello flar, I am back from release. I have updated a pr at https://github.com/flutter/cocoon/pull/1667 to address your suggestions.

PR #1667, opened 2022-03-16
[dashboard] make settings dialog take a portion of the screen per request

XilaiZhang:
context: https://github.com/flutter/cocoon/pull/1658#discussion_r827506164


XilaiZhang:
As per request, settings dialog will take just a portion of the entire screen as it used to be. sample screen shot will look like:
Screen Shot 2022-03-15 at 6 03 46 PM

dashboard/lib/build_dashboard_page.dart
```diff
@@ -109,43 +109,47 @@ class BuildDashboardPageState extends State<BuildDashboardPage> {
...
-                        }
-                        _removeSettingsDialog();
-                      },
-                    ),
-                  ],
-                ),
-              ],
+            child: FractionallySizedBox(
+              widthFactor: 0.5,
+              heightFactor: 0.7,
```
flar, line 114: Fractional view makes it unnecessarily large. The goldens have the dialog disappear off the bottom of the image and it's mostly blank space.

Was the original change to move from a Column to a ListView intended to keep it from overflowing a screen? The problem with ListView is that it always expands to fill its space. You can limit the main-axis (vertical) size using "shrinkwrap: true", but that doesn't stop it from filling all of the cross-axis (horizontal) space.

This widget looks like it will size itself to its children, but scroll when necessary, perhaps that is a better fit: https://api.flutter.dev/flutter/widgets/SingleChildScrollView-class.html

flar (changes requested):

dashboard/lib/build_dashboard_page.dart
CaseyHillers: We could simplify it to a sized box. There isn't much variation for the layout here, and a 300 px by 300px box is likely enough to fit the content

XilaiZhang: Thanks a lot for the suggestions! (Wow everyone gets up so early :)

flar: LGTM

flar (approved):
LGTM

PR #1673, opened 2022-03-17
[cloudbuild] Extend timeout and create separate auto_submit step

CaseyHillers:
https://github.com/flutter/flutter/issues/100185


keyonghan:
This is not needed as https://github.com/flutter/flutter/issues/100185 separated the build configs.

PR #1674, opened 2022-03-17
Separate cloud builds of app_dart and auto_submit

keyonghan:
Follow up of https://github.com/flutter/flutter/issues/100185

Instead of deploying both apps in a single cloudbuild file, we are separating into two config files. This way we can achieve:
1) independent build
2) Separate timeout control

This needs separate triggers config setup from gcloud side.


cloudbuild_app_dart.yaml
```diff
@@ -1,6 +1,6 @@
 # Provide instructions for google Cloud Build to auto-build flutter
 # dashboard to flutter-dashboard project. Auto-build will be triggered
```
CaseyHillers, line 2: Are we able to move these yamls into the app_dart and auto_submit directories?

keyonghan: Let me play around.

keyonghan: Now this works.

CaseyHillers (approved):

PR #1684, opened 2022-03-25
[app_dart] Remove extra dependencies in generate_jspb

CaseyHillers:
The extra dependencies make it fail in environments where you can't import `dart:mirrors`


XilaiZhang (approved):
unnamed

PR #1686, opened 2022-03-26
[autosubmit] Add installation to generate github token

KristinBi:
This PR fixed the service account issue https://github.com/flutter/flutter/issues/100502 by formatting the privateKey in correct format and adding the installation id to generate the github token.


auto_submit/lib/service/config.dart
```diff
@@ -41,24 +41,42 @@ class Config {
...
       // To ensure no expired tokens are used, set this to 10 - 1, with an extra buffer of a duplicate request.
       const Duration(minutes: 8),
     );
-
+    final String token = String.fromCharCodes(cacheValue!);
     return GitHub(auth: Authentication.withToken(token));
   }
 
+  Future<String> _getInstallationId() async {
+    final String jwt = await _generateGithubJwt();
```
CaseyHillers, line 55: Can we add a test?

auto_submit/lib/service/config.dart
```diff
@@ -41,24 +41,42 @@ class Config {
...
+    final Map<String, String> headers = <String, String>{
+      'Authorization': 'Bearer $jwt',
+      'Accept': 'application/vnd.github.machine-man-preview+json'
+    };
+    final Uri githubInstallationUri = Uri.https('api.github.com', 'app/installations');
```
CaseyHillers, line 60: Can we add a TODO to upstream into `package:github`?

auto_submit/lib/service/config.dart
KristinBi: Sure Casey, I added a unit test for the config class.

auto_submit/test/service/config_test.dart
```diff
@@ -0,0 +1,51 @@
...
+    setUp(() {
+      cacheProvider = Cache.inMemoryCacheProvider(kCacheSize);
+      config = Config(
+        cacheProvider: cacheProvider,
+        secretManager: CloudSecretManager(),
+      );
+    });
+
+    test('githubAppInstallationId ', () async {
+      await withAppEngineServices(() async {
```
CaseyHillers, line 31: Avoid using `withAppEngineServices` in tests. Instead, we should inject the httpProvider into the config so it's testable

KristinBi: Hi Casey, when I tried to delete withAppEngineServices and use httpProvider, I got an error: Bad state: Not running inside a service scope zone.
I think this is because it failed to look up to get authClientService. Do you have any suggestions on this error?

CaseyHillers: The test is incorrectly using `CloudSecretManager`. This is for prod workflows only. You'll want `LocalSecretManager()` for tests

KristinBi: Oh I see, gotcha, thanks Casey!

auto_submit/test/service/config_test.dart
```diff
@@ -0,0 +1,51 @@
...
+import 'package:test/test.dart';
+
+/// Number of entries allowed in [Cache].
+const int kCacheSize = 1024;
+void main() {
+  group('Config', () {
+    late CacheProvider cacheProvider;
+    late Config config;
+    const int kCacheSize = 1024;
+    final HttpProvider httpProvider = Providers.freshHttpClient;
```
CaseyHillers, line 21: Use `MockClient` instead. Unit tests should not use http

https://github.com/dart-lang/http/blob/master/lib/testing.dart

KristinBi: Thank you Casey! I studied some example codes and changed our unit test to use mock client here.

CaseyHillers (approved):

PR #1680, opened 2022-03-23
[release] Full stack logic to store, retrieve, update and display release branches

XilaiZhang:
This is the first few steps of addressing https://github.com/flutter/flutter/issues/100491. When a create branch webhook event happens, cocoon will parse the event and check the branch, if the branch is a new branch that did not previously exist, cocoon will create a datastore entity of the new branch and store it.

PR #1688, opened 2022-03-28
[app_dart] github webhook for create events and branch class initialization

XilaiZhang:
This smaller PR contains how we create branch event from github webhook, parse the event and store the event in the form of a 'branch' class into google cloud datastore.

This is a split of PR 1680.  (some files are later modified, e.g. branch.proto). Context: https://github.com/flutter/cocoon/pull/1680#pullrequestreview-923180297


CaseyHillers (commented):
Can you update the PR title and body to describe the changes being done? Stating it's part of larger work is great, but the issue should be tracking what small piece this PR is adding to the larger project. This gives the reviewers context of what to give feedback on versus what will be done in the future.

XilaiZhang:
For sure, I have updated the title and description. Thank you!

app_dart/lib/src/model/appengine/branch.dart
```diff
@@ -0,0 +1,66 @@
...
+  /// when current branch had activity.
+  @IntProperty(propertyName: 'lastActivity', required: true)
+  int? lastActivity;
+
+  /// [RepositorySlug] of where this commit exists.
+  RepositorySlug get slug => RepositorySlug.full(repository);
+
+  String get repository => key.id!.substring(0, key.id!.lastIndexOf('/'));
+
+  String get branch => key.id!.substring(key.id!.lastIndexOf('/') + 1);
```
CaseyHillers, line 29: Can we add a channel field as well? In the future, we can use this to get the current candidate branch for CPs

app_dart/lib/src/model/appengine/branch.dart
```diff
@@ -0,0 +1,66 @@
...
+import 'package:gcloud/db.dart';
+import 'package:github/github.dart';
+import 'package:json_annotation/json_annotation.dart';
+
+part 'branch.g.dart';
+
+@Kind(name: 'Branch', idType: IdType.String)
+class Branch extends Model<String> {
+  Branch({Key<String>? key, required int lastActivity}) {
+    parentKey = key?.parent;
```
CaseyHillers, line 15: There's no parents for `Branch`, so this is dead code

app_dart/lib/src/model/appengine/branch.dart
```diff
@@ -0,0 +1,66 @@
...
+    id = key?.id;
+  }
+
+  /// The timestamp (in milliseconds since the Epoch) of the last time
+  /// when current branch had activity.
+  @IntProperty(propertyName: 'lastActivity', required: true)
```
CaseyHillers, line 21: We should avoid making values required, especially for values that are likely to come from GitHub

app_dart/lib/src/model/github/create_event.dart
```diff
@@ -0,0 +1,33 @@
...
+/// See more:
+///  * https://docs.github.com/en/developers/webhooks-and-events/events/github-event-types#createevent
+import 'package:github/github.dart' show Repository, User;
+import 'package:github/hooks.dart' show HookEvent;
+import 'package:json_annotation/json_annotation.dart';
+
+part 'create_event.g.dart';
+
+@JsonSerializable(fieldRename: FieldRename.snake)
+class CreateEvent extends HookEvent {
```
CaseyHillers, line 16: This should get upstreamed in `package:github` first. Feel free to add me as a reviewer there

app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -86,6 +98,30 @@ class GithubWebhook extends RequestHandler<Body> {
     }
   }
 
+  Future<void> _handleCreateRequest(String rawRequest, DatastoreService datastore) async {
```
CaseyHillers, line 101: Avoid adding code to the `RequestHandler` classes as they're harder to test.

This should be moved into a `BranchService` class

dashboard/lib/model/branch.proto
```diff
@@ -0,0 +1,12 @@
+// Copyright 2021 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+syntax = "proto3";
+
+import "lib/model/key.proto";
+
+message Branch {
+    optional RootKey key = 1;
```
CaseyHillers, line 10: This is missing the branch name

dashboard/lib/model/branch.proto
```diff
@@ -0,0 +1,12 @@
+// Copyright 2021 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+syntax = "proto3";
+
+import "lib/model/key.proto";
+
+message Branch {
```
CaseyHillers, line 9: Avoid adding code that isn't used yet. We can add this in the frontend PR

app_dart/lib/src/model/appengine/branch.dart
```diff
@@ -0,0 +1,66 @@
...
+      ..write(', branch: $branch')
+      ..write(', repository: $repository')
+      ..write(')');
+    return buf.toString();
+  }
+}
+
+/// The serialized representation of a [Branch].
+// TODO(tvolkert): Directly serialize [Branch] once frontends migrate to new serialization format.
+@JsonSerializable(createFactory: false, ignoreUnannotated: true)
```
CaseyHillers, line 46: Avoiding adding code that isn't used. We could simplify by having the backend adapt this to the proto

app_dart/lib/src/model/appengine/branch.dart
XilaiZhang: umm maybe I am wrong, but my understanding is it if we keep the code this way, when we initialize the key using `final Key<String> key = datastore.db.emptyKey.append<String>(Branch, id: id);`, the parentKey will be set to `datastore.db.emptyKey` , which is a bit different from a `null` parentKey (what we would obtain if we remove this line).

When we are looking up and deleting entities, and providing a `currentBranch.key`, we are calling on the getter function of `Key<T> get key => parentKey!.append(runtimeType, id: id);` (which is not modifiable) part of the code to retrieve the key. This line of code asserts on parentKey not being `null`. If we initialized the value to be `datastore.db.emptyKey` we can pass this check. If we remove this line, parentKey will be defaulted to `null`, and causing the assertion of `parentKey!` to fail when we lookup or delete entities. 

My understanding is based on how we currently generate and lookup keys. Maybe there is a way to workaround this. I could be wrong but I thought with the current topology we need this line to properly initialize a non null parentKey.

CaseyHillers: What you're describing is an issue in the tests. We should update the fake datastore to handle null parent keys

XilaiZhang: umm so I have the logic at L115 - L120 of githubwebhook.dart currently looking like this https://github.com/flutter/cocoon/pull/1688/files#diff-3a0003489aad220cc50ef5c753a82ed536839ae6ed0ada1ac8b45c12e3ad1f2aR115-R123 . Maybe there is a workaround for this part but I am not sure how to use null as a base key for `append` to work

app_dart/lib/src/request_handlers/github_webhook.dart
XilaiZhang: Sure, moved

app_dart/lib/src/model/github/create_event.dart
XilaiZhang: sure, raised pr at https://github.com/SpinlockLabs/github.dart/pull/304

app_dart/lib/src/model/appengine/branch.dart
XilaiZhang: Sure, in later commits I have a get_branches request handler that relies on the 'lastactivity' timestamp to compare and filter branches later. I will add assertion of not nullable to those

app_dart/lib/src/model/appengine/branch.dart
XilaiZhang: Sure, added. I wasn't very sure about the use of channel. I thought if we have a channel name such as beta, then it would already manifest itself as being 'beta' in the branch name field? umm I am just wondering what's the difference between branch name and channel

XilaiZhang:
Thanks for the reply! The newest commit addresses the feedbacks except the logic to create a key from null base key, which I wasn't very sure about and left a comment in the comment section. Thank you!

app_dart/lib/src/model/appengine/branch.dart
CaseyHillers: The linked pasted doesn't work.

Can you share the snippet?

XilaiZhang: For sure, it looks like 
```    final String id = '$repository/$branch';
    final Key<String> key = datastore.db.emptyKey.append<String>(Branch, id: id);
    final Branch currentBranch = Branch(key: key, lastActivity: lastActivity);
    try {
      await datastore.lookupByValue<Branch>(currentBranch.key);
    } on KeyNotFoundException {
      await datastore.insert(<Branch>[currentBranch]);
    }
```

CaseyHillers: I'm not sure what the error here is. That's expected, and I don't see issues with it. There's no need to save the parent key as it's always going to be empty/null

app_dart/lib/src/service/branch_service.dart
```diff
@@ -0,0 +1,55 @@
...
+import 'package:cocoon_service/src/service/datastore.dart';
+import 'package:gcloud/db.dart';
+
+import '../model/appengine/branch.dart';
+import '../model/github/create_event.dart';
+import '../request_handling/exceptions.dart';
+
+class RetryException implements Exception {}
+
+/// Parse a create github webhook event, and add it to datastore.
```
CaseyHillers, line 17: This is more a method doc for `handleCreateRequest`. A class doc would mention this handles GitHub branches and manages them in relation to the release process

app_dart/lib/src/model/appengine/branch.dart
XilaiZhang: My understanding is that when we are calling `currentBranch.key`, we are calling on the getter method `Key<T> get key => parentKey!.append(runtimeType, id: id);`. In this getter method, it first creates a parentKey, assert its not null (but could be empty), and then append. So if we have a null value for parentKey it will fail, but an empty key is different from null in that it is a Key instance

XilaiZhang: one of the sample errors looks like this 
```:00:03 +139 -1: test/request_handlers/github_webhook_test.dart: github webhook create branch event should add branch to db if db is empty [E]
  Null check operator used on a null value
  package:gcloud/src/db/models.dart 106:30                               Model.key
  package:cocoon_service/src/service/branch_service.dart 42:59           BranchService.handleCreateRequest
  ===== asynchronous gap ===========================
  package:cocoon_service/src/request_handlers/github_webhook.dart 88:11  GithubWebhook.post
  ===== asynchronous gap ===========================
  test/request_handlers/github_webhook_test.dart 2053:7                  main.<fn>.<fn>
```

XilaiZhang: the line 106 it points to is from the key model class which is not modifiable
```
abstract class Model<T> {
  T? id;
  Key? parentKey;

  Key<T> get key => parentKey!.append(runtimeType, id: id);       //this is the line 106
}
```

app_dart/lib/src/model/github/create_event.dart
CaseyHillers: SGTM

app_dart/lib/src/model/appengine/branch.dart
CaseyHillers: SGTM

app_dart/lib/src/model/appengine/branch.dart
CaseyHillers: Since this isn't used, it should be removed from this PR

app_dart/test/request_handlers/github_webhook_test.dart
```diff
@@ -2037,6 +2041,73 @@ void foo() {
     });
   });
 
+  group('github webhook create branch event', () {
+    test('should add branch to db if db is empty', () async {
```
CaseyHillers, line 2045: Most of these tests should be moved into the test for the branch service. This should verify that branch service is called only for the create event

XilaiZhang: Sure, I moved the main testing logics to branch_service_test. 

It is kind of hard to verify branchService is called for create event in github_webhook_test for the following reasons: 
1. verify calls need to be called on a mockito object, and if we create a mock branchService object in test, it is a different instance of branchService from the one created in github_webhook.dart.
2. if we add branch service as a field to github_webhook, since the class is immutable we have to make branch service a final variable and provide it with a const default value. The final field constraint would prevent us from creating a branchService later in the code. And we don't have a good default const constructor for branchService because it relies on a datastore service.
3. If we mark the key function in branch service as visible for testing, the function will be constrained to the class and tests, and it can't be called in github_webhook.
4. If we create a static handling function in github_webhook.dart, we will need a mock instance of github_webhook, which doesn't seem ideal on top of the already non ideal static function.

I ended up creating a default error case for, the switch statement on event types pullrequest, checkrun and create. Therefore if we encounter an event type not conforming to the fore mentioned three types, the program will throw an error. And for github_webhook test, I made create event test similar to checkrun test in that, it won't check for how many times the function is called. Instead it only checks if the test runs without causing an exception. And we trust the switch statement categorizes event types correctly without adding a test to it.

CaseyHillers: Throwing an error on the webhooks will cause a large number of errors in our backend.

1 is the right idea, and you in the case no branch service was passed, it can be constructed like the previously logic had.

XilaiZhang: Thanks for the reply! umm does `in the case no branch service was passed` imply that we are passing branchService as a field variable to github_webhook? I think the challenge is that github_webhook is an immutable class which requires the branchService passed in to be final. If that is the case, we would need something like `GithubWebhook(branchService: BranchService(DatastoreService(config.db, maxentitygroup), requestString)` when we are initializing GithubWebhook, but at that point we can't properly provide the information of maxentitygroup and requestString?

CaseyHillers: Correct. Feel free to remove the immutability status of the handlers. For server handlers, ti doesn't matter much, but it's helpful for models

XilaiZhang: Thanks for the explanations! I removed the immutability of request handler class to better serve tests

app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -37,14 +36,13 @@ const Set<String> kNeedsCheckLabelsAndTests = <String>{
...
-    Config config, {
-    required this.scheduler,
-    this.githubChecksService,
-    this.datastoreProvider = DatastoreService.defaultProvider,
-  }) : super(config: config);
+  GithubWebhook(Config config,
+      {required this.scheduler,
+      this.githubChecksService,
+      this.datastoreProvider = DatastoreService.defaultProvider,
+      this.branchService})
```
CaseyHillers, line 44: nit: trailing comma

app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -84,11 +83,9 @@ class GithubWebhook extends RequestHandler<Body> {
           }
           break;
         case 'create':
-          final BranchService branchService = BranchService(datastore, rawRequest: stringRequest);
-          await branchService.handleCreateRequest();
+          branchService = branchService ?? BranchService(datastore, rawRequest: stringRequest);
```
CaseyHillers, line 86:
```suggestion
branchService ??= BranchService(datastore, rawRequest: stringRequest);
```

CaseyHillers (approved):
flutter-lgtm

PR #1689, opened 2022-03-29
[release] Test exemption bot should not run on release branch PRs

XilaiZhang:
being lazy and directing context to issue https://github.com/flutter/flutter/issues/100923 , welp, if you click on this link it will direct you to another context link :)

### DIFF
--- a/app_dart/lib/src/request_handlers/github_webhook.dart
+++ b/app_dart/lib/src/request_handlers/github_webhook.dart
@@ -487,9 +487,13 @@ class GithubWebhook extends RequestHandler<Body> {
     if (baseName == defaultBranchName) {
       return false;
     }
+    // Check if branch name confroms to the format flutter-x.x-candidate.x,
+    // A pr with conforming branch name is likely to be intended
+    // for a release branch, whereas a pr with non conforming name is likely
+    // caused by user misoperations, in which case bot
+    // will suggest open pull request against default branch instead.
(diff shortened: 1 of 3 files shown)


app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -482,6 +480,21 @@ class GithubWebhook extends RequestHandler<Body> {
     }
   }
 
+  bool _isReleaseBranch(PullRequest pr) {
+    final String defaultBranchName = Config.defaultBranch(pr.base!.repo!.slug());
+    final String baseName = pr.base!.ref!;
+
+    if (baseName == defaultBranchName) {
```
CaseyHillers, line 487: This is enough. The connotation in infra is that a release branch is anything that isn't the default branch

XilaiZhang: Thanks for the reply! umm this would short circuit the logic of `_getWrongBaseComment`. Previously branches of the format not conforming to flutter-release-x will receive `_getWrongBaseComment`. If we change the logic to only check for default branch alignment, then the block of code of `_getWrongBaseComment` should be removed?

CaseyHillers: No, let's leave it in. Can you leave a comment explaining why we explicitly check the format?

CaseyHillers (approved):
LGTM once comment is added

CaseyHillers (approved):

PR #1692, opened 2022-03-30
[autosubmit] Inject latency

KristinBi:
This PR intends to solve https://github.com/flutter/flutter/issues/101010. We injected latency per repository, so we will skip the pull request whose newest commits are less than injected latency.

### DIFF
--- a/auto_submit/lib/requests/check_pull_request.dart
+++ b/auto_submit/lib/requests/check_pull_request.dart
@@ -15,6 +15,14 @@ import '../service/github_service.dart';
 import '../service/log.dart';
 import '../server/request_handler.dart';
 
+/// Injected latency per repository. Engine and Flutter use an injected latency of 1h meaning
+/// that the bot skips any commits younger than 1h. However 1h is too long for some repositories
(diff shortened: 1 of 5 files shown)


auto_submit/lib/requests/check_pull_request.dart
```diff
@@ -51,25 +59,44 @@ class CheckPullRequest extends RequestHandler {
...
+      log.info('The pull request $prNumber has unfinished tests,'
           'push to pubsub and check later.');
-      await pubsub.publish('auto-submit-queue', pullRequest);
     }
-    return Response.ok('Does not merge the pull request ${queryResult.number}.');
+    return Response.ok('Does not merge the pull request $prNumber.');
+  }
+
+  /// Skip commits that are less than injected latency.
+  Future<bool> shouldSkip(RepositorySlug slug, int number, GithubService gitHub) async {
```
CaseyHillers, line 89: Check pull request is starting to contain a lot of logic. We should split this logic into a set of composable rules that we pull from a yaml config first.

KristinBi:
Since it's not necessary to inject latency and we can get to know that all checks are scheduled from one specific check, I will close this PR.

PR #1694, opened 2022-03-31
Add cancelled to task status

CaseyHillers:
Fixes https://github.com/flutter/flutter/issues/101072


XilaiZhang (approved):
Today I learnt the word cancelled can be spelled with either one 'L' or two 'LL's :)
download

PR #1695, opened 2022-03-31
[release] update release branches through cron jobs, and retrieve release branches for frontend display

XilaiZhang:
This is the second split of breaking up https://github.com/flutter/cocoon/pull/1680, which updates release branches through cron jobs, and retrieve release branches for frontend display. The original linked issue was https://github.com/flutter/flutter/issues/100491.


app_dart/lib/src/service/build_status_provider.dart
```diff
@@ -137,6 +138,25 @@ class BuildStatusService {
     }
   }
 
+  /// Returns a map from [Branch.key] to [Branch.lastActivity], i.e., all the active branches Ids and their associated timestamps.
+  Future<Map<String, int>> retrieveActiveBranchIds({
```
CaseyHillers, line 142: This should be in `BranchService`

app_dart/lib/src/service/build_status_provider.dart
```diff
@@ -137,6 +138,25 @@ class BuildStatusService {
...
+    int? timestamp,
+  }) async {
+    Map<String, int> branchIdToActivity = {};
+    await for (Commit commit in datastoreService.queryActiveCommits(
```
CaseyHillers, line 146: We shouldn't rely on datastore as we don't know the push date. Instead, use the github client to list the most recent commit on the branch

dashboard/lib/model/branch.proto
```diff
@@ -0,0 +1,10 @@
+// Copyright 2021 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+syntax = "proto3";
+
+message Branch {
+    optional string branch = 1;
+    optional string repository = 2;
```
CaseyHillers, line 9: This is missing channel and latest activity

app_dart/test/request_handlers/get_repos_test.dart
```diff
@@ -2,39 +2,43 @@
...
-      final Body body = await tester.get(handler);
-      final Map<String, dynamic> result = await utf8.decoder
-          .bind(body.serialize() as Stream<List<int>>)
-          .transform(json.decoder)
-          .single as Map<String, dynamic>;
-
-      expect(result['Branches'], <String>['flutter-1.1-candidate.1', 'master']);
-    });
-  });
+//   group('GetBranches', () {
```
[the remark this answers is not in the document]
XilaiZhang: umm I was thinking about deleting it because the test description says the test `returns branches matching regExps`, but I didn't see any regex matching in this test. My understanding is that it is setting flutterBranchesValue to be flutter-1.1-candidate.1, and return that value to check for flutter-1.1-candidate.1, which isn't very useful?

app_dart/lib/src/service/build_status_provider.dart
XilaiZhang: I looked around and it seems like after we obtain the list of recent commits, there is no github api to map commit to the branch it belongs to. Also since we have cherrypicks, we can't compare sha with branch commit to check for inclusion. ( https://stackoverflow.com/questions/23899329/list-of-branches-a-commit-appears-on/23970412#23970412,  https://github.community/t/how-to-get-all-branches-which-contain-a-commit-from-sha-using-githubapi/14067/6) The online community suggested using graphQLs to workaround, would that be an approach we would like to pursuit? (I studied a few examples in cocoon but they look different from the online examples, I will look into it more if we would indeed like to do graphQL)

XilaiZhang: Another topology I just finished implementing was that: 1. we first retrieve all the branches, 2. for each branch, find its associated commit sha, 3. from the sha, use another commit api request to retrieve the last activity time for the commit, 4. collect all the last activity associated with commit -> branch sha -> branch, and update the branches. But after implementing this topology I realized that since this topology sends too many url requests to github, it gets "API rate limit exceeded". umm in this case, I am not sure if we have a way to resolve the design flow using a limited number of url requests? Thank you!

XilaiZhang: Based on discussion with Keyong, I think our best bet is to forfeit the github api approach (attached below), and instead pursuit a graphQL approach.
```
  Future<void> _updateBranchesForAllRepos (DatastoreService datastore) async{
    final List<String> repos = Config.supportedRepos.map((RepositorySlug slug) => slug.name).toList();
    DateTime timeNow = DateTime.now();
    for(String repo in repos){
      await _updateBranchesForRepo(repo, timeNow, datastore);
    }

    return;
  }

  Future<void> _updateBranchesForRepo (String repo, DateTime timeNow, DatastoreService datastore) async{
    int pageNumber = 1;
    List<Branch> updatedBranches = []; 

    while(true){
      final Uri url = Uri.https(baseUrl, '/repos/flutter/$repo/branches', <String, String>{'page': pageNumber.toString(),});
      
      String allBranchesOfRepo = await getUrl(url, () => http.Client(),  );
      List<dynamic> branchArray = jsonDecode(allBranchesOfRepo) as List<dynamic>;//as List<Map<String, dynamic>>;
      if(branchArray.isEmpty){
        break;
      }

      print('runtimetype is ${branchArray[0].runtimeType}');
      for(dynamic branch in branchArray){
        Map<String, dynamic> branchCommitInfo = branch['commit'] as Map<String, dynamic>;
        String branchSha = branchCommitInfo['sha'] as String;
        int branchLatestUpdate = await _branchLastUpdateFromSha(branchSha, repo); 
        
        if ( branchLatestUpdate > timeNow.subtract(const Duration(days: kActiveBranchActivityPeriod)).millisecondsSinceEpoch){
          final String id = '$repo/$branch';
          final Key<String> key = datastore.db.emptyKey.append<String>(Branch, id: id);
          updatedBranches.add(Branch(key: key, lastActivity: branchLatestUpdate));
        }
      }
      pageNumber += 1;
    }

    await datastore.insert(updatedBranches);
  }

  Future<int> _branchLastUpdateFromSha (String branchSha, String repo) async{
    final Uri url = Uri.https(baseUrl, '/repos/flutter/$repo/commits/$branchSha');
    String branchTipInfo = await getUrl(url, () => http.Client(), );
    Map<String, dynamic> branchTipCommit = jsonDecode(branchTipInfo) as Map<String, dynamic>;
    String lastCommitDate = branchTipCommit['commit']['author']['date'] as String;
    return DateTime.parse(lastCommitDate).millisecondsSinceEpoch;
  } 

}
```

CaseyHillers: You can use the `git` cli to check the list of commits via https://pub.dev/documentation/io/latest/io/ProcessManager-class.html

`git log upstream/flutter-2.12-candidate.3 -n 5`

Then you can use `--pretty=...'` with the format to show the push date for the most recent activity

XilaiZhang: umm I studied release_dashboard, and since its logic relies on conductor tool I had a chat with Christopher. If my understanding is correct, release_dashboard first clones the upstream into a local filesystem and then runs from the filesystem. For release_dashboard tests they clone the git repo first and then carry out testing logics. My understanding is that it is somehow inevitable to clone the git repository first, in order to gain ability to perform further operations such git log, git branch etc. My understanding is that we can't run git commands outside git repositories because we don't have access to .git context. cc @christopherfujino in case I am missing anything

CaseyHillers: `git ls-remote --heads git@github.com:flutter/flutter`

CaseyHillers: Then you can use https://api.github.com/repos/flutter/flutter/branches to get the latest info on a branch, then look up the commit

XilaiZhang: Thanks for the reply! This is impressive for listing the branches. umm maybe there is another magic command, but I examined all flags that ls-remote can take, and the closest I can get is using sort with something like `git ls-remote --heads --sort=committerdate git@github.com:flutter/flutter`. But that still requires a clone. It seems like we need a clone to be able to retrieve date information and sort, according to https://stackoverflow.com/questions/65729722/git-ls-remote-tags-how-to-get-date-information. I might be wrong, but I still think we need a clone to be able to do computations such as looking up and sorting on date?

CaseyHillers: Can you add a reproduction step for what you're seeing? I can do the following outside a git repo:

```sh
chillers$ git ls-remote --heads git@github.com:flutter/flutter
38ea03e585263cef744bcb959aa2a68e7fc17737	refs/heads/Hixie-patch-1
5c931b769b581d536ee5b5d58677d39a110da3d0	refs/heads/beta
...
```

In this case, you already have the branch, so you don't need ls-remote. You only need to parse https://api.github.com/repos/flutter/flutter/branches

XilaiZhang: Oh I think I understand now. Previously I was still hitting the api limit because I was running it through `Uri.https`, which puts me in the category as unauthenticated user. I think if I use `githubClient` instead and exploit the multiple wrapper classes I should be categorized as an authenticated user. My understanding is that it should work if I use the original api approach, but as an authenticated user. Thanks a lot for the help!

app_dart/cron.yaml
```diff
@@ -58,3 +58,7 @@ cron:
 - description: check flaky builders to either deflake the builder or file a new flaky bug
   url: /api/check_flaky_builders
   schedule: every wednesday 16:00
+
+- description: update branches to reflect most recent commit activity
+  url: '/api/public/get-branches?update=true'
```
CaseyHillers, line 63: /api/update-branches

app_dart/lib/src/request_handlers/update_branches.dart
```diff
@@ -0,0 +1,124 @@
...
+        updatedBranches.add(md.Branch(key: key, lastActivity: lastUpdate));
+      }
+    }
+    await datastore.insert(updatedBranches);
+  }
+}
+
+// FOR REVIEW: Probably related to how keyhelper is implementated,
+// but I have to add a wrapper class to avoid `Converting object to an encodable object failed: Instance of 'SerializableBranch'` error
+class BranchWrapper {
```
CaseyHillers, line 112: Why do we have BranchWrapper and SerializableBranch? Wrapper and serializable are doing the same function

app_dart/lib/src/model/appengine/branch.dart
```diff
@@ -42,3 +46,27 @@ class Branch extends Model<String> {
...
+
+/// The serialized representation of a [Branch].
+// TODO(tvolkert): Directly serialize [Branch] once frontends migrate to new serialization format.
+@JsonSerializable(createFactory: false, ignoreUnannotated: true)
+class SerializableBranch {
+  const SerializableBranch(this.branch);
+
+  final Branch branch;
+
+  @JsonKey(name: 'Key')
```
CaseyHillers, line 58: Avoid adding `JsonKey`. The reason we have these littered about is we used to run a Go codebase at the same time where the default generated it that way

app_dart/test/request_handlers/update_branches_test_data.dart
```diff
@@ -0,0 +1,125 @@
+// Copyright 2021 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+const List<int> stderrArray = [];
+const List<int> stdoutArray = [
```
CaseyHillers, line 6: Putting bytes as test data makes it hard to reason what's going on. Instead, use something like

`'my stdout'.codeUnits`

dashboard/lib/build_dashboard_page.dart
```diff
@@ -78,7 +79,13 @@ class BuildDashboardPageState extends State<BuildDashboardPage> {
       queryParameters.addAll(_filter!.toMap(includeDefaults: false));
     }
     queryParameters['repo'] = repo!;
-    queryParameters['branch'] = branch!;
+
+    Set<String> validOptions =
+        buildState.branches.where((Branch b) => b.repository == repo).map((Branch b) => b.branch).toSet();
+    branch = validOptions.contains(branch) ? branch : 'master';
+    queryParameters['branch'] = validOptions.contains(branch) ? branch! : 'master';
```
CaseyHillers, line 86: Don't hard code it as 'master'. You should fall back to the `defaultBranches`

dashboard/lib/service/appengine_cocoon.dart
```diff
@@ -113,22 +114,31 @@ class AppEngineCocoonService implements CocoonService {
...
-      final List<String>? branches = jsonResponse['Branches'].cast<String>();
-      return CocoonResponse<List<String>>.data(branches);
+
+      final List<Branch> branches = <Branch>[];
+      for (final Map<String, dynamic> jsonBranch in jsonResponse['Branches']) {
+        final Map<String, dynamic> branchInfo = jsonBranch['branch'];
+        branches.add(Branch()
+          ..branch = branchInfo['branch']
+          ..repository = branchInfo['repository'].split('/')[1]);
+        // FOR REVIEW : currently only supports flutter/repo since the function name is fetchFlutterBranches
```
CaseyHillers, line 136: This is fine. The build dashboard can be hardcoded to only work for flutter/* repos.

app_dart/lib/src/request_handlers/get_branches.dart
```diff
@@ -4,24 +4,115 @@
...
+    Config config, {
+    this.datastoreProvider = DatastoreService.defaultProvider,
+    BuildStatusServiceProvider? buildStatusProvider,
+  })  : buildStatusProvider = buildStatusProvider ?? BuildStatusService.defaultProvider,
+        super(config: config);
+
+  final DatastoreServiceProvider datastoreProvider;
+  final BuildStatusServiceProvider buildStatusProvider;
+
+  static const String kUpdateBranchParam = 'update';
```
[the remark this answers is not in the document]
CaseyHillers: Is this still necessary now that it's been moved to its own handler?

app_dart/test/request_handlers/get_repos_test.dart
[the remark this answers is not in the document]
CaseyHillers: Thanks! Then let's delete it

app_dart/test/request_handlers/update_branches_test.dart
```diff
@@ -0,0 +1,134 @@
...
+import '../src/datastore/fake_config.dart';
+import '../src/datastore/fake_datastore.dart';
+import '../src/request_handling/fake_authentication.dart';
+import '../src/request_handling/fake_http.dart';
+import '../src/request_handling/request_handler_tester.dart';
+import '../src/utilities/mocks.dart';
+import 'update_branches_test_data.dart';
+
+void main() {
+  group('GetBranches', () {
```
CaseyHillers, line 26: nit: update branches

app_dart/test/request_handlers/update_branches_test.dart
```diff
@@ -0,0 +1,134 @@
...
+      int lastActivity = DateTime.tryParse("2019-05-15T15:20:56Z")!.millisecondsSinceEpoch;
+      final Key<String> branchKey = db.emptyKey.append<String>(Branch, id: id);
+      final Branch currentBranch = Branch(
+        key: branchKey,
+        lastActivity: lastActivity,
+      );
+      db.values[currentBranch.key] = currentBranch;
+    });
+
+    test('should not retrieve branches older than a week', () async {
```
CaseyHillers, line 73: nit: This is more like 2 months

app_dart/test/request_handlers/update_branches_test.dart
```diff
@@ -0,0 +1,134 @@
...
+      expect(db.values.values.whereType<Branch>().length, 1);
+
+      when(mockRepositoriesService.getCommit(any, any)).thenAnswer((Invocation invocation) {
+        return Future<RepositoryCommit>.value(RepositoryCommit(
+            sha: testBranchSha,
+            commit: GitCommit(
+                committer: GitCommitUser(
+              'xilaizhang',
+              'xilaizhang@google.com',
+              DateTime.tryParse("2020-05-15T15:20:56Z"),
```
CaseyHillers, line 126: nit: we use single quotes

app_dart/lib/src/request_handlers/update_branches.dart
XilaiZhang: umm I spent another hour looking at this, and guess that I probably need some help on understanding the internals. the closest example from getStatus was that it uses a `keyHelper.encode(status.commit.key)` , and had a commitStatus class wrapped around the commit class which has an internal key. If I remove JsonKey or keyHelper or wrapper class in code, my variables become empty or  json.decode would fail. The only way I can make current code run is to do an exact copy of getstatus (copying both keyhelper to help with the underlying key in branch class, and the wrapper class). I think you are right that there should be a better way to do this, but I haven't figure that out yet

CaseyHillers: Skip keyhelper, and directly use `key.id`

XilaiZhang:
TLDR; the newest commit gets rid of wrapper class and tackles the root cause of serializing non-simple fields. Also incorporated other feedbacks.

A summary of what i learnt from json serialization after studying all the files:
1. it is possible to serialize a custom class with nested class fields, but eventually the smallest class units have to resolve to all simple fields (string, int etc.) In our case, the wrapper class wasn't the main blocker. The main blocker was that we have a branch class which has a Key\<String\> type field, and Key\<T\> cannot be further resolved to simple fields. A class that is simple-field-resolvable can be serialized using approaches outlined in https://stackoverflow.com/questions/57662221/how-to-parse-a-complexnested-object-to-json-and-send-it-to-server-using-http-i. KeyHelper is a helper class that we wrote for cocoon that specializes in serialize Key\<T\> type fields.
2. the reason I was seeing empty serialization was because the `ignoreUnannotated: true` option. This essentially tells the serializer that ignore all custom fields that you don't know how to serialize. with this option the branch class will be serialized to empty json object because its non-simple fields are omitted.
Therefore in my newest commit I used keyHelper to pinpoint serialize the Key<String> field ( the only non-simple field), and got rid of other wrapper class.

app_dart/lib/src/model/appengine/branch.dart
```diff
@@ -44,10 +44,10 @@ class Branch extends Model<String> {
 }
 
 class SerializableBranch {
-  const SerializableBranch(this.branch, this.key);
+  const SerializableBranch(this.branch, this.id);
```
CaseyHillers, line 47: Do you even need `SerializableBranch`? If `Branch` can be json encoded, why not use that?

XilaiZhang: Ohh this is very true. Thanks a lot for pointing this out!

app_dart/lib/src/model/appengine/branch.dart
```diff
@@ -41,25 +41,32 @@ class Branch extends Model<String> {
...
-      'branch': facade,
+      'branch': <String, dynamic>{
+        'branch': branch,
+        'repository': repository,
+      },
     };
   }
 }
+
+// class SerializableBranch {
```
CaseyHillers, line 56: nit: dead code?

app_dart/lib/src/request_handlers/get_branches.dart
```diff
@@ -4,24 +4,61 @@
...
+    Config config, {
+    this.datastoreProvider = DatastoreService.defaultProvider,
+    this.processRunner,
+  }) : super(config: config);
+
+  final DatastoreServiceProvider datastoreProvider;
+  ProcessRunner? processRunner;
+
+  static const String kUpdateBranchParam = 'update';
+  static const int kActiveBranchActivityPeriod = 7;
```
CaseyHillers, line 50: This should be a `Duration`

app_dart/lib/src/request_handlers/update_branches.dart
```diff
@@ -0,0 +1,105 @@
+// Copyright 2020 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:cocoon_service/src/model/appengine/branch.dart' as md;
+import 'package:gcloud/db.dart';
+import 'package:github/github.dart' show Branch, GitHub, RepositoryCommit, RepositorySlug;
+import 'package:process_runner/process_runner.dart';
```
CaseyHillers, line 10: We should use `package:process`. We don't need the extra complexity of process_runner, and it's already in our dep tree

app_dart/lib/src/request_handlers/update_branches.dart
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  Future<void> _updateBranchesForAllRepos(Config config, DatastoreService datastore) async {
+    DateTime timeNow = DateTime.now();
+
+    processRunner ??= ProcessRunner();
+    final Set<RepositorySlug> slugs = Config.supportedRepos;
+    for (RepositorySlug slug in slugs) {
+      ProcessRunnerResult result =
+          await processRunner!.runProcess(['git', 'ls-remote', '--heads', 'git@github.com:flutter/${slug.name}']);
```
CaseyHillers, line 77: You can also use `${slug.owner}` so we don't hard code flutter

app_dart/test/request_handlers/update_branches_test.dart
```diff
@@ -0,0 +1,138 @@
...
+      request = FakeHttpRequest();
+      keyHelper = FakeKeyHelper(applicationContext: clientContext.applicationContext);
+      tester = RequestHandlerTester(request: request);
+
+      final MockGitHub mockGitHubClient = MockGitHub();
+      mockRepositoriesService = MockRepositoriesService();
+      when(mockGitHubClient.repositories).thenReturn(mockRepositoriesService);
+      when(processRunner.runProcess(any)).thenAnswer(
+          (Invocation invocation) => Future.value(ProcessRunnerResult(0, stdoutArray, stderrArray, outputArray)));
+      // this encodes '72fe8a9ec3af4d76097f09a9c01bf31c62a942aa' and 'refs/heads/main'
```
CaseyHillers, line 58: nit: Is this comment still necesary?

app_dart/test/request_handlers/update_branches_test.dart
```diff
@@ -0,0 +1,138 @@
...
+      int lastActivity = DateTime.tryParse("2019-05-15T15:20:56Z")!.millisecondsSinceEpoch;
+      final Key<String> branchKey = db.emptyKey.append<String>(Branch, id: id);
+      final Branch currentBranch = Branch(
+        key: branchKey,
+        lastActivity: lastActivity,
+      );
+      db.values[currentBranch.key] = currentBranch;
+    });
+
+    test('should not retrieve branches older than 2 months', () async {
```
CaseyHillers, line 77: We should a test for when the git command fails

dashboard/lib/build_dashboard_page.dart
```diff
@@ -78,7 +79,13 @@ class BuildDashboardPageState extends State<BuildDashboardPage> {
       queryParameters.addAll(_filter!.toMap(includeDefaults: false));
     }
     queryParameters['repo'] = repo!;
-    queryParameters['branch'] = branch!;
+
+    Set<String> validOptions =
+        buildState.branches.where((Branch b) => b.repository == repo).map((Branch b) => b.branch).toSet();
+    branch = validOptions.contains(branch) ? branch : 'master';
+    queryParameters['branch'] = validOptions.contains(branch) ? branch! : defaultBranches[repo]!;
+    // FOR REVIEW: If we switch to a repo where the current release branch does not exist, I am updating the branch to be master
```
CaseyHillers, line 87: Don't do this. Always assume the user knows more than the system.

For example, I can query directly https://flutter-dashboard.appspot.com/#/build?repo=engine&branch=flutter-2.13-candidate.0 but this change would break that

dashboard/lib/build_dashboard_page.dart
```diff
@@ -218,14 +225,18 @@ class BuildDashboardPageState extends State<BuildDashboardPage> {
         ),
         onChanged: (String? selectedBranch) {
           branch = selectedBranch;
-          _updateNavigation(context);
+          _updateNavigation(context, _buildState);
         },
-        items: _buildState.branches.map<DropdownMenuItem<String>>((String value) {
+        items: _buildState.branches
+            .where((Branch b) => b.repository == _buildState.currentRepo)
```
CaseyHillers, line 231: Is the `where` needed? The get-branches called should only return relevant data

dashboard/test/utils/appengine_cocoon_test_data.dart
```diff
@@ -57,14 +57,24 @@ const String luciJsonGetStatsResponse = '''
...
-const String jsonGetBranchesResponse = '''
-  {
-    "Branches": [
-      "master",
-      "flutter-0.0-candidate.1"
-    ]
-  }
-''';
+const String jsonGetBranchesResponse = '''{
+   "Branches":[
```
CaseyHillers, line 61: Does this need to be updated as well? It's missing the id field

dashboard/lib/build_dashboard_page.dart
XilaiZhang: Yeah currently get-branches will return all branches with recent activities. Does it sound right if we add a url parameter 'repo', so that get-branches will filter based on repository during web request?

dashboard/lib/build_dashboard_page.dart
XilaiZhang: umm my understanding is that branches flutter-2.13-candidate.0 will be reachable when the repository is valid, such as engine. Currently I am trying to prevent an access of branch from an invalid repository, such as from impeller repository. This happens when the branches dropdown stays the same, and when users toggles the repository dropdown. I am trying to use the current logic to help branches dropdown get updated when the repository dropdown is updated.

CaseyHillers: It's a better UX to have this fail by showing an empty dashboard than to have it lock someone out.

For example, if this was in place last week I wouldn't have been able to drive the google3 CPs as the branch didn't exist according to flutter-dashboard

app_dart/lib/src/request_handlers/update_branches.dart
XilaiZhang: It might not be ideal to import dart:io, but it seems like process/process.dart itself wasn't quite enough. I looked at the examples at https://github.com/google/process.dart/blob/master/test/src/interface/common_test.dart and imported dart:io to get process to work. but for testing the mock of a Process class, i.e., a mockProcess would not have a valid mockProcess.run function because the run function is abstract. I am not sure if there is a valid way for us to test process and provide a fake result in tests? Thank you!

christopherfujino: Here's an example of testing with `package:process`: https://github.com/flutter/cocoon/blob/main/device_doctor/test/src/host_utils_test.dart

dashboard/lib/build_dashboard_page.dart
CaseyHillers: That SGTM

CaseyHillers:
Any blockers here?

XilaiZhang:
> Any blockers here?

haha that's true let me context switch back to this commit. Yeah I was code signing and rolling and forgot about this.

app_dart/lib/src/request_handlers/update_branches.dart
XilaiZhang: Thanks a lot! imported dart:io and updated accordingly.

dashboard/lib/build_dashboard_page.dart
XilaiZhang: Our current logic is we run fetchBranches() once during setup, and retrieve all the branches we want. Later for frontend display, we just use `where` to select the branches we interested in from the pre-retrieved branches. So in total we only need one api request.

If we remove the `where` logic, we would have to send an api request whenever the user changes branch, and let updateBranch trigger fetchBranch(). Maybe i am missing something, I think it might be more efficient if we just pre-retrieve the branches and serve the user with what they need, instead of sending multiple api requests?

dashboard/lib/build_dashboard_page.dart
XilaiZhang: Sure, I handled the transition state by keeping the latest branch selection for the user, along with all the valid branch options that would come from the newest repository selection.

dashboard/lib/build_dashboard_page.dart
CaseyHillers: You can update `updateCurrentRepoBranch` to make the call for `_fetchBranches`

XilaiZhang: I guess one of the things removing `where` statements will also affect, is the logic to transit states. Currently, in order for dashboard to not break when user toggle the repo dropdown, (in case where user's current branch selection might be invalid for the new repo selection), I stored and carried over the latest branch selection for user, and combine with all the valid branch options that would come from the newest repository selection. This approach depends on a where statement to remove duplicate branches when combining the branch options.

If we remove the `where` logic, we would need to find another way for dashboard to not break. We can also list all branch options for all repos in the branch dropdown, but I think that would be less ideal.

CaseyHillers: I wouldn't worry about this case. If users are manually overriding the branch, it's fine to reset it when the repo is swapped. Granted, I don't see why we can't keep the branch the same.

XilaiZhang: umm previously I reset it to the default branch when switching repos, but that broke the dashboard from supporting invalid combination of repo+branch. To keep branch the same while swapping repositories, we need to support invalid combinations. 

And to add more context, I am trying to advocate for a `where` statement because of how dropdown button is built. The options in the dropdown button list at any time should always have an entry that matches current branch. Therefore if we want to keep the branch same after swapping repository, along with all the valid options brought by the new repository, we keep track of user's previous selection of branch (this selection is carried over from repo to repo), and combine user's previous selection of branch (which might be invalid in the new settings) with the currently valid list of options. And during this combination, we want to check if there are any duplicates in dropdownMenuItem, and I think we need a where statement to check that.

XilaiZhang: I implemented this logic in the newest commit, maybe the code in  https://github.com/flutter/cocoon/pull/1695/commits/c7dbc53a13111e388a23c4ec212bca6ef2b51137#diff-795424fcac01b97b0d9ed4cff372972e6c6280b81e293264a85d3c8123106886R226-R251 would help explain why I think a `where` statement is necessary

dashboard/lib/build_dashboard_page.dart
```diff
@@ -218,17 +221,33 @@ class BuildDashboardPageState extends State<BuildDashboardPage> {
...
         onChanged: (String? selectedBranch) {
           branch = selectedBranch;
-          _updateNavigation(context);
+          _updateNavigation(context, _buildState);
         },
-        items: _buildState.branches.map<DropdownMenuItem<String>>((String value) {
-          return DropdownMenuItem<String>(
-            value: value,
+        items:
+            // FOR REVIEW: to ensure a valid transition no matter what invalid combinations of repo + branch user inputed,
```
CaseyHillers, line 227: nit: These would work better as standalone comments instead of intending them for the reviewer. GitHub comments should be used for conversations

CaseyHillers (approved):

CaseyHillers:
Any blockers I can help with here?

XilaiZhang:
Thanks for checking in! Yep I need to find time to work on this. Once I figure out why notarytool won't take my 2FA password I will get back to this.

XilaiZhang:
Newest commit adds tests to ensure transitions from one invalid combination of repo+branch to another invalid combination of repo+branch do not raise exceptions. The test also make sure delayed network requests are handled correctly.

dashboard/lib/build_dashboard_page.dart
```diff
@@ -223,22 +223,22 @@ class BuildDashboardPageState extends State<BuildDashboardPage> {
...
         },
-        items:
-            // FOR REVIEW: to ensure a valid transition no matter what invalid combinations of repo + branch user inputed,
-            // we preserve that invalid combination as a single entry, along with the other valid combinations
-            [
+        items: [
+          // To ensure a valid transition no matter what invalid combinations of repo + branch user inputed,
+          // we preserve that invalid combination as a single entry, along with the other valid combinations
+          // corresponding test is "ensure smooth transition between invalid state"
           DropdownMenuItem<String>(
```
CaseyHillers, line 230: nit: we don't leave comments like this. Well written tests should make it obvious the functionality they're testing

CaseyHillers: Fwiw, the test.looks well documented to me

CaseyHillers (approved):

PR #1698, opened 2022-04-01
[scheduler] Process build messages with null task key

CaseyHillers:
Fixes alerts in https://pantheon.corp.google.com/monitoring/alerting/incidents/0.mg89mnnl5se8?project=flutter-dashboard&pageState=(%22interval%22:(%22d%22:%22P1D%22,%22i%22:(%22s%22:%222022-03-31T22:48:04.000Z%22,%22e%22:%222022-04-01T16:58:24.320Z%22)))

Fixes https://github.com/flutter/flutter/issues/101205

This:
1. Updates the pubsub subscription to be able to handle null task keys
2. Fixes the scheduler logic to write tasks to datastore to get task keys, which can then be passed to LUCI


CaseyHillers:
I did a prod test and this fixed it.

app_dart/lib/src/service/scheduler.dart
```diff
@@ -154,6 +154,10 @@ class Scheduler {
       log.severe('Failed to add commit ${commit.sha!}: $error');
     }
 
+    await luciBuildService.schedulePostsubmitBuilds(commit: commit, toBeScheduled: toBeScheduled);
```
CaseyHillers, line 157: The fake datastore makes it hard to write a regression test as if I was to write one it would be very brittle and only work in this case

app_dart/lib/src/service/scheduler.dart
```diff
@@ -154,6 +154,10 @@ class Scheduler {
...
+    // Update tasks for those that were scheduled
+    await datastore.insert(tasks);
```
keyonghan, line 159: why do we need to insert again?

CaseyHillers: LuciBuildService updates the task statuses once triggered

keyonghan: I may miss it, but which section updates the task status?

CaseyHillers: My bad, it's in https://cs.opensource.google/flutter/cocoon/+/main:app_dart/lib/src/service/scheduler.dart;l=140

The logic is:
1. Insert commit and initial targets to datastore to generate keys
2. Trigger the builds based on priority
3. Update datastore again with the final status (some builds will be scheduled due to batch scheduling)

keyonghan: but `tasks` are already updated at line 149?

CaseyHillers: This PR moves it around. In `schedulePostsubmitBuilds` these entities need to be written beforehand to get their keys (since they're deterministic, datastore generates them). See https://github.com/flutter/flutter/issues/78876

keyonghan: I think the confusion here ` await luciBuildService.schedulePostsubmitBuilds(commit: commit, toBeScheduled: toBeScheduled);`.

As I see the logic, it only pushes requests to pub sub where I don't see any task status update.

CaseyHillers: Correct. We preemptively need to mark these tasks as in progress as there's no gurantees when LUCI will send a notification acking them as scheduled. This is to ensures batching is maintained

keyonghan: To wrap up what I am talking about:

https://github.com/flutter/cocoon/blob/66667f15396ee7d6fd46d27bc47dd62752f3f93e/app_dart/lib/src/service/scheduler.dart#L133-L159

line140: task status changes
line149: tasks updated in datastore
line157: no task status change
line159: why do we need to update tasks in datastore again?

CaseyHillers: Thanks for your patience on this, Friday brain. I removed the extra write.

app_dart/lib/src/request_handlers/scheduler/gob_poller.dart
```diff
@@ -0,0 +1,64 @@
+// Copyright 2019 The Flutter Authors. All rights reserved.
```
keyonghan, line 1: Is this change intended to be in this PR?

CaseyHillers: Removed

keyonghan (approved):
LGTM

PR #1696, opened 2022-03-31
[app_flutter] wrap AppBar.title in a Tooltip

christopherfujino:
Fixes https://github.com/flutter/flutter/issues/101137

PR #1701, opened 2022-04-04
Revert "[app_flutter] wrap AppBar.title in a Tooltip"

christopherfujino:
Reverts flutter/cocoon#1696

There appears to be a bug in this that causes the text to not always be updated (missing key in tooltip?)

Fixes https://github.com/flutter/flutter/issues/101311


CaseyHillers (approved):
What a weird bug

Piinks (approved):
LGTM

PR #1708, opened 2022-04-07
[autosubmit] Get correct installation id

KristinBi:
For the installation Id used to generate github token, we should pick the one belongs to flutter to get the write permission.
This PR also solves the issue: https://github.com/flutter/flutter/issues/100808. So now we can track the installation Id by repo to ensure every repo's quota is independent.

### DIFF
--- a/auto_submit/lib/service/config.dart
+++ b/auto_submit/lib/service/config.dart
@@ -56,7 +56,7 @@ class Config {
     return String.fromCharCodes(cacheValue!);
   }
 
-  Future<String?> getInstallationId() async {
+  Future<String> getInstallationId() async {
     final String jwt = await _generateGithubJwt();
     final Map<String, String> headers = <String, String>{
       'Authorization': 'Bearer $jwt',
@@ -70,14 +70,14 @@ class Config {
       githubInstallationUri,
       headers: headers,
     );
-    final list = json.decode(response.body).map((data) => (data) as Map<String, dynamic>).toList();
-    String? installatinId;
+    final List<dynamic> list = json.decode(response.body).map((data) => (data) as Map<String, dynamic>).toList();
+    late String installationId;
     for (Map<String, dynamic> installData in list) {
       if (installData['account']!['login']!.toString() == 'flutter') {
-        installatinId = installData['id']!.toString();
+        installationId = installData['id']!.toString();
       }
     }
-    return installatinId;
+    return installationId;
   }
 
   Future<GraphQLClient> createGitHubGraphQLClient() async {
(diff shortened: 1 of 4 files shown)


auto_submit/lib/service/config.dart
```diff
@@ -71,7 +71,12 @@ class Config {
       headers: headers,
     );
     final list = json.decode(response.body).map((data) => (data) as Map<String, dynamic>).toList();
-    final String installatinId = list[0]['id'].toString();
+    String? installatinId;
```
keyonghan, line 74: The `installationId` is expected to always have a value. Let's use `late String installationId` instead.

nit: `installatinId` -> `installationId`
nit: add type for the `list` variable in line 73.

KristinBi: Sorry for the error. Fixed them.

auto_submit/lib/service/config.dart
```diff
@@ -70,9 +70,14 @@ class Config {
       githubInstallationUri,
       headers: headers,
     );
-    final list = json.decode(response.body).map((data) => (data) as Map<String, dynamic>).toList();
-    final String installatinId = list[0]['id'].toString();
-    return installatinId;
+    final List<dynamic> list = json.decode(response.body).map((data) => (data) as Map<String, dynamic>).toList();
+    late String installationId;
+    for (Map<String, dynamic> installData in list) {
+      if (installData['account']!['login']!.toString() == 'flutter') {
```
CaseyHillers, line 76: Instead of hard coding 'flutter', can we pass the `RepositorySlug` in `getInstallationId`?

That way this works for teams outside of flutter

CaseyHillers (approved):

auto_submit/lib/service/config.dart
KristinBi: Sure! I added the repository slug.

auto_submit/lib/service/config.dart
```diff
@@ -35,28 +35,28 @@ class Config {
...
-    String token = await generateGithubToken();
+  Future<GitHub> createGithubClient(RepositorySlug slug) async {
+    String token = await generateGithubToken(slug);
     return GitHub(auth: Authentication.withToken(token));
   }
 
-  Future<String> generateGithubToken() async {
+  Future<String> generateGithubToken(RepositorySlug slug) async {
     // GitHub's secondary rate limits are run into very frequently when making auth tokens.
     final Uint8List? cacheValue = await cache['githubToken'].get(
```
CaseyHillers, line 50: You'll need to add the repo.owner to the cache key

KristinBi: Hi Casey, do you mean that we should store the related repo info into Cache, and get the value by calling cache.get()?

CaseyHillers: `cache['githubToken-${slug.owner}']`

CaseyHillers:
```suggestion
    final Uint8List? cacheValue = await cache['githubToken-${slug.owner}'].get(
```

KristinBi: Gotcha, thanks Casey! I updated the code.

CaseyHillers (approved):

PR #1710, opened 2022-04-07
[autosubmit] test mergePR

KristinBi:
This PR is to test whether we can use API to merge PR, I just modified some lines of codes which won't break our current codes in main.


keyonghan (approved):
Let's have the bot land this PR.

PR #1711, opened 2022-04-07
Increase label node request limit when graphql quering waiting PRs

keyonghan:
Existing logic returns top 5 labels for any PR, but PRs may have more than that, e.g. https://github.com/flutter/flutter/pull/101345.

When this happens, it will cause API failure and stop merging PRs.

This PR increases the limit to 10 from 5. The issue may still happen if a PR has more than 10 labels, but this will not be a problem when we move to `auto-submit` bot later this quarter.

This PR also adds logs to make the error message more clear.

Fixes https://github.com/flutter/flutter/issues/101539

### DIFF
--- a/app_dart/lib/src/request_handlers/check_for_waiting_pull_requests_queries.dart
+++ b/app_dart/lib/src/request_handlers/check_for_waiting_pull_requests_queries.dart
@@ -61,7 +61,7 @@ query LabeledPullRequestsWithReviews($sOwner: String!, $sName: String!, $sLabelN
             state
           }
         }
-        labels(first: 5) {
+        labels(first: 10) {
           nodes {
             name
             id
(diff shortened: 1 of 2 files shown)


app_dart/lib/src/request_handlers/check_for_waiting_pull_requests_queries.dart
```diff
@@ -61,7 +61,7 @@ query LabeledPullRequestsWithReviews($sOwner: String!, $sName: String!, $sLabelN
             state
           }
         }
-        labels(first: 5) {
+        labels(first: 10) {
```
godofredoc, line 64: Does this impact the query time?

keyonghan: I don't see much difference when trying in the graphql explorer. This makes sense as this just adds more info based on same filters.

godofredoc (approved):

PR #1713, opened 2022-04-08
[app_dart] Add flutter-2.13-candidate.0 to release branches

CaseyHillers:
This is necessary as next engine beta will use the Cocoon scheduler. As a hack, you can add the branch manually in the URL encode, this is to make it easier to surface these results to the rest of the team.

This is temporary until https://github.com/flutter/flutter/issues/100491 is implemented


yusuf-goog (approved):
Can you add a TODO comment to change this once the other bug is fixed.

PR #1714, opened 2022-04-08
[app_dart] Add branch support to get green commits

CaseyHillers:
This is so I can set a bash script to ping me when to create the Flutter PR that rolls an Engine CP.

Long term, this will be used by rapid

### DIFF
--- a/app_dart/test/request_handlers/get_green_commits_test.dart
+++ b/app_dart/test/request_handlers/get_green_commits_test.dart
@@ -15,6 +15,7 @@ import 'package:test/test.dart';
 
 import '../src/datastore/fake_config.dart';
 import '../src/request_handling/fake_authentication.dart';
+import '../src/request_handling/fake_http.dart';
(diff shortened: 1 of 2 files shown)


XilaiZhang (approved):
cool thanks for replacing the commit shas with commit.sha!
images (1)

app_dart/test/request_handlers/get_green_commits_test.dart
```diff
@@ -86,8 +93,8 @@ void main() {
 
       expect(result.length, 2);
       expect(result, <String>[
-        'd5b0b3c8d1c5fd89302089077ccabbcfaae045e4',
-        'ea28a9c34dc701de891eaf74503ca4717019f829',
+        commit2.sha!,
```
yusuf-goog, line 96: Minor nit: speaking of null safety, should we make commit1/2 null safe and not have the

yusuf-goog (approved):

PR #1715, opened 2022-04-08
[app_dart] Approve release PRs

CaseyHillers:
Fixes https://github.com/flutter/flutter/issues/101397

### DIFF
--- a/app_dart/lib/src/request_handlers/github_webhook.dart
+++ b/app_dart/lib/src/request_handlers/github_webhook.dart
@@ -7,7 +7,8 @@ import 'dart:convert';
 
 import 'package:cocoon_service/src/service/branch_service.dart';
 import 'package:crypto/crypto.dart';
-import 'package:github/github.dart' show PullRequest, RepositorySlug, GitHub, PullRequestFile, IssueComment;
+import 'package:github/github.dart'
+    show CreatePullRequestReview, GitHub, IssueComment, PullRequest, PullRequestFile, RepositorySlug;
 import 'package:github/hooks.dart';
 
 import '../model/github/checks.dart' as cocoon_checks;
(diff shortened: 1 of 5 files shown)


XilaiZhang (approved):
images (1)

app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -181,6 +183,29 @@ class GithubWebhook extends RequestHandler<Body> {
     await scheduler.triggerPresubmitTargets(pullRequest: pr);
   }
 
+  /// Release tooling generates cherrypick pull requests that should be granted an approval.
+  Future<void> _tryReleaseApproval(
```
godofredoc, line 187: Part of the release process requires that two people get involved in release PRs this change will allow unilateral changes. @Hixie thoughts?

Hixie: i don't mind this for the cherrypick step, so long as the actual release (once the cherrypicks are known) requires two people to sign off.

CaseyHillers: I'm less concerned about getting reviews on beta and stable releases as those are reviewed weekly. This is to unblock the g3 roller where we're CPing changes into the internal release branches.

My long term plan here is there would be no human accounts that would get sign off, but instead automated bot accounts creating CP PRs would get approved.

godofredoc (approved):

app_dart/test/src/datastore/fake_config.dart
```diff
@@ -258,4 +258,7 @@ class FakeConfig implements Config {
 
   @override
   String get defaultRecipeBundleRef => 'refs/heads/main';
+
+  @override
+  Future<List<String>> get releaseAccounts async => <String>['dash-releaser'];
```
godofredoc, line 263: s/dash/dart-flutter?

PR #1718, opened 2022-04-11
[autosubmit] add HMAC auth

KristinBi:
To protect the webhook endpoints, we add the HMAC auth.


auto_submit/lib/service/config.dart
```diff
@@ -165,4 +166,12 @@ class Config {
 
   /// The autosubmit label.
   String get autosubmitLabel => 'autosubmit';
+
+  /// Get the webhook key
+  Future<String> getWebhookKey() async {
+    final Uint8List? cacheValue = await cache['webhookKey'].get(
```
CaseyHillers, line 172: nit: Make this `cache[webhookKey]`

CaseyHillers (approved):

auto_submit/lib/service/config.dart
KristinBi: Changed.

PR #1729, opened 2022-04-13
[dashboard] Enable semantics

CaseyHillers:
This lets me use vimium to navigate the dashboard (granted I can't open tasks which seems like a bug with `Lattice`)


keyonghan (approved):
LGTM

PR #1737, opened 2022-04-16
make devicelab tests count as tests

christopherfujino:
Fixes https://github.com/flutter/flutter/issues/101948

### DIFF
--- a/app_dart/lib/src/request_handlers/github_webhook.dart
+++ b/app_dart/lib/src/request_handlers/github_webhook.dart
@@ -262,12 +262,7 @@ class GithubWebhook extends RequestHandler<Body> {
         needsTests = !_allChangesAreCodeComments(file);
       }
 
-      if ((filename.endsWith('_test.dart') ||
-              filename.endsWith('.expect') ||
-              filename.contains('test_fixes') ||
-              filename.startsWith('dev/bots/test.dart') ||
-              filename.startsWith('dev/bots/analyze.dart')) &&
-          !kNotActuallyATest.any(filename.endsWith)) {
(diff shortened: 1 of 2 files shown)


jmagman (approved):

app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -262,12 +262,7 @@ class GithubWebhook extends RequestHandler<Body> {
         needsTests = !_allChangesAreCodeComments(file);
```
jmagman, line 262: Can the above be changed to `!_isATest(filename)`
```
          !filename.startsWith('dev/devicelab/bin/tasks') &&
          !filename.startsWith('dev/devicelab/lib/tasks') &&
```
?
At a minimum it those two lines could be removed.

jmagman:
Formatting issues
```
+ echo '############# files that require formatting ###########'
############# files that require formatting ###########
+ dart format --set-exit-if-changed --line-length=120 .
Formatted lib/src/request_handlers/github_webhook.dart
```

app_dart/lib/src/request_handlers/github_webhook.dart
christopherfujino: I think we want to check `_allChangesAreCodeComments`, because if the file isn't one of the above file types, but the only diffs are in comments, then that doesn't not make `needsTests`.

But you're right, I can remove the checks for the devicelab files, because even if that marks this `needs` test, we will certainly also mark the PR as having tests.

jmagman (approved):
Still LGTM, thank you!

PR #1741, opened 2022-04-19
[cron] Increase autosubmit frequency to 2 minutes

CaseyHillers:
I've been using the bot. It would be nice if it was every 2 minutes, which roughly matches the existing logic.

This would allow 2 PRs to be submitted every 4 minutes versus the 2 PRs every 5 minutes today.


keyonghan (approved):
LGTM

PR #1743, opened 2022-04-20
Update modify mode when creating PRs via auto flake bot

keyonghan:
Fixes: https://github.com/flutter/flutter/issues/102258

### DIFF
--- a/app_dart/lib/src/request_handlers/flaky_handler_utils.dart
+++ b/app_dart/lib/src/request_handlers/flaky_handler_utils.dart
@@ -38,7 +38,7 @@ const String kCiYamlTargetTagsFramework = 'framework';
 const String kCiYamlTargetTagsHostonly = 'hostonly';
 
 const String kMasterRefs = 'heads/master';
-const String kModifyMode = '100755';
+const String kModifyMode = '100644';
 const String kModifyType = 'blob';
 
 const int kSuccessBuildNumberLimit = 3;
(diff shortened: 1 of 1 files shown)


christopherfujino (approved):
LGTM

app_dart/lib/src/request_handlers/flaky_handler_utils.dart
```diff
@@ -38,7 +38,7 @@ const String kCiYamlTargetTagsFramework = 'framework';
 const String kCiYamlTargetTagsHostonly = 'hostonly';
 
 const String kMasterRefs = 'heads/master';
-const String kModifyMode = '100755';
```
chunhtai, line 41: what does the new number do? maybe add a comment as well

chunhtai (approved):
LGTM

app_dart/lib/src/request_handlers/flaky_handler_utils.dart
keyonghan: Added a comment to explain.

Context: the bot is using `100755` which updates the .ci.yaml file to be executable. And this causes `analyze` failing with recent PR: https://github.com/flutter/flutter/pull/101853

jmagman (commented):
Why is the bot modifying the mode in the first place?

keyonghan:
> Why is the bot modifying the mode in the first place?

/cc @chunhtai may have a better idea.

PR #1775, opened 2022-04-22
[device doctor] Migrate to null safety

keyonghan:
https://github.com/flutter/flutter/issues/102350

### DIFF
--- a/device_doctor/lib/src/health.dart
+++ b/device_doctor/lib/src/health.dart
@@ -4,7 +4,6 @@
 
 import 'dart:async';
 import 'dart:io';
-import 'dart:convert';
 
 import 'package:path/path.dart' as path;
 import 'package:platform/platform.dart' as platform;
--- a/device_doctor/lib/src/health.dart
+++ b/device_doctor/lib/src/health.dart
@@ -4,6 +4,7 @@
 
 import 'dart:async';
 import 'dart:io';
+import 'dart:convert';
 
 import 'package:path/path.dart' as path;
 import 'package:platform/platform.dart' as platform;
@@ -16,7 +17,7 @@ import 'utils.dart';
 /// The dialogs often cause test flakiness and performance regressions.
 Future<HealthCheckResult> closeIosDialog({
   ProcessManager pm = const LocalProcessManager(),
-  String deviceId,
+  String? deviceId,
   platform.Platform pl = const platform.LocalPlatform(),
   String infraDialog = 'infra-dialog',
 }) async {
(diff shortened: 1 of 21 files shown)


device_doctor/lib/src/health.dart
```diff
@@ -43,9 +44,6 @@ Future<HealthCheckResult> closeIosDialog({
     }
     Process proc = await pm.start(command, workingDirectory: dialogDir.path);
     logger.info('Executing: $command');
-    // Discards stdout and stderr as they are too large.
-    await proc.stdout.drain<Object>();
-    await proc.stderr.drain<Object>();
```
keyonghan, line 48: These two lines block several unit tests. Removing as we are not using this logic at all.

CaseyHillers: +1, I'm always happier if we can remove Stream logic as it's usually over complicated for our use cases

device_doctor/test/src/ios_device_test.dart
```diff
@@ -46,154 +46,154 @@ void main() {
...
-      expect(results['ios-device-abcdefg'][2].succeeded, false);
-      expect(results['ios-device-abcdefg'][3].name, kDevicePairCheckKey);
-      expect(results['ios-device-abcdefg'][3].succeeded, false);
-      expect(results['ios-device-abcdefg'][4].name, kUserAutoLoginCheckKey);
-      expect(results['ios-device-abcdefg'][4].succeeded, false);
-      expect(results['ios-device-abcdefg'][5].name, kDeviceProvisioningProfileCheckKey);
-      expect(results['ios-device-abcdefg'][5].succeeded, false);
-      expect(results['ios-device-abcdefg'][6].name, kBatteryLevelCheckKey);
-      expect(results['ios-device-abcdefg'][6].succeeded, false);
+      expect(results['ios-device-abcdefg']?.length, 7);
```
CaseyHillers, line 56: nit: You could do `expect(results['ios-device-abcdefg'], isNotNull);` first, then you won't need the extra `?` checks.

Also, doing a single `!` will assert it's not null for the rest of the code.

device_doctor/test/src/utils.dart
```diff
@@ -3,29 +3,57 @@
...
 import 'dart:io';
+import 'dart:convert';
 
 import 'package:mockito/mockito.dart';
 import 'package:process/process.dart';
 
 class MockPlatform extends Mock implements Platform {}
 
-class MockProcessManager extends Mock implements ProcessManager {}
+class MockProcessManager extends Mock implements ProcessManager {
```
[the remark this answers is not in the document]
CaseyHillers: +1 - I think a lot of Dart codebases are benefitting in creating their own mocks or writing good fakes. In the future it would be a good thing to consider here. For example, the tools team doesn't use mockito, and I've been writing fakes for app_dart which lets me validate various parts of the codebase interacting.

CaseyHillers (approved):

device_doctor/test/src/ios_device_test.dart
keyonghan: Updated.

PR #1776, opened 2022-04-22
[scheduler] Add backfilling

CaseyHillers:
https://github.com/flutter/flutter/issues/100793

### DIFF
--- a/app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
+++ b/app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
@@ -2,44 +2,123 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-import 'dart:convert';
-
+import 'package:cocoon_service/src/foundation/utils.dart';
+import 'package:cocoon_service/src/model/appengine/task.dart';
+import 'package:cocoon_service/src/request_handling/body.dart';
+import 'package:cocoon_service/src/service/buildbucket.dart';
+import 'package:cocoon_service/src/service/datastore.dart';
+import 'package:cocoon_service/src/service/scheduler/policy.dart';
+import 'package:github/github.dart';
 import 'package:meta/meta.dart';
-import 'package:retry/retry.dart';
 
-import '../../../cocoon_service.dart';
-import '../../model/luci/buildbucket.dart';
-import '../../request_handling/exceptions.dart';
+import '../../model/appengine/commit.dart';
+import '../../model/ci_yaml/ci_yaml.dart';
+import '../../model/ci_yaml/target.dart';
+import '../../request_handling/request_handler.dart';
+import '../../service/config.dart';
 import '../../service/logging.dart';
+import '../../service/luci_build_service.dart';
+import '../../service/scheduler.dart';
 
-/// Subscription for making requests to BuildBucket.
+/// Cron request handler for scheduling targets when capacity becomes available.
 ///
-/// The PubSub subscription is set up here:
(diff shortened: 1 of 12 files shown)


app_dart/lib/src/service/config.dart
```diff
@@ -46,7 +46,7 @@ class Config {
   /// This adds support for the `waiting for tree to go green label` to the repo.
   ///
   /// Relies on the GitHub Checks API being enabled for this repo.
-  static Set<RepositorySlug> supportedRepos = <RepositorySlug>{
+  Set<RepositorySlug> supportedRepos = <RepositorySlug>{
```
CaseyHillers, line 49: This needed to be changed so tests could override it. Otherwise, the data would look weird where there would be 6 backfills instead of 1 (1 for each supported repo)

cron.yaml
```diff
@@ -27,6 +27,10 @@ cron:
   url: /api/vacuum-github-commits
   schedule: every 1 hours
 
+- description: backfills builds
```
yusuf-goog, line 30: Please add a todo to remove this code once the move to cocoon is complete.

app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
```diff
@@ -0,0 +1,127 @@
...
+  ///
+  /// Otherwise, returns null indicating nothing should be backfilled.
+  FullTask? _backfillTask(Target target, List<FullTask> tasks) {
+    final List<FullTask> relevantTasks = tasks.where((FullTask task) => task.task.name == target.value.name).toList();
+    if (relevantTasks.any((FullTask task) => task.task.status == Task.statusInProgress)) {
+      // Don't schedule more builds where there is already a running task
+      return null;
+    }
+
+    final List<FullTask> backfillTask =
```
keyonghan, line 118: It would be interesting to use this logic to rerun failed tests to surface flakes. Not needed in this PR though.

app_dart/test/request_handlers/scheduler/batch_backfiller_test.dart
```diff
@@ -0,0 +1,99 @@
...
+];
+
+void main() {
+  late BatchBackfiller handler;
+  late RequestHandlerTester tester;
+  late FakeDatastoreDB db;
+  late FakePubSub pubsub;
+  late FakeScheduler scheduler;
+
+  group('BatchBackfiller', () {
```
keyonghan, line 55: could you add tests for batch scheduling cases: more than 1 target are scheduled when they all need to schedule; only one target is scheduled when itself needs to schedule, etc?

cron.yaml
```diff
@@ -27,6 +27,10 @@ cron:
...
+  url: /api/scheduler/batch-backfiller
+  schedule: every 5 minutes
```
keyonghan, line 32: I believe this is a no-op for LUCI triggered builders?

CaseyHillers: Correct. Those targets have a [OmitPolicy](https://cs.opensource.google/flutter/cocoon/+/main:app_dart/lib/src/service/scheduler/policy.dart;l=72) which means Cocoon doesn't do anything.

app_dart/test/request_handlers/scheduler/batch_backfiller_test.dart
CaseyHillers: Thanks, good point on having multiple columns being tested at a time.

To clarify, I added tests for when:
1. Only one column needs to be backfilled
2. 2 columns need to be backfilled

keyonghan (approved):
LGTM

PR #1785, opened 2022-04-25
Improve flutter-gold pending message

guidezpl:
Before | After 
--- | ---
Screen Shot 2022-04-25 at 13 36 32 | Screen Shot 2022-04-25 at 13 37 28


app_dart/lib/src/service/config.dart
```diff
@@ -228,7 +228,7 @@ class Config {
       '(https://github.com/flutter/flutter/wiki/Tree-hygiene#how-to-review-code) '
       'and make sure this patch meets those guidelines before LGTMing.';
 
-  String get flutterGoldPending => 'Pending.';
+  String get flutterGoldPending => 'Waiting for all other checks to be successful.';
```
christopherfujino, line 231: Is this always true?

Also, this would need a test.

Piinks: This looks ok to me, although in the past when we had it this way folks did not like it much better.

It is not always true. It is equally frustrating to see a message that says "waiting for all other checks to complete" when all of the checks are completed. Once the checks complete, it can take up to 5 minutes to update since it is a scheduled cron job. The actual check itself it just one request to Gold for a yes/no response about new images, it does not really run for a period of time.

christopherfujino: If this is not guaranteed to be true, let's not say. What about something like "This check is pending (have all other checks completed?)"

guidezpl: Could also be "Waiting for all other checks to be successful before querying Gold."

guidezpl: @Piinks WDYT?

Piinks: Oh sorry I missed the last comment. 

> Waiting for all other checks to be successful before querying Gold.

That sounds good, let's try it. We have changed it before, we can change it again if someone does not like it. :)

guidezpl: Done :)

Piinks (approved):
LGTM

PR #1793, opened 2022-04-27
[app_dart] Add exe to postsubmit builds

CaseyHillers:


app_dart/test/src/utilities/mocks.mocks.dart
```diff
@@ -155,87 +156,89 @@ class _FakeClientContext_49 extends _i1.Fake implements _i19.ClientContext {}
 
 class _FakePullRequestMerge_50 extends _i1.Fake implements _i9.PullRequestMerge {}
 
-class _FakePullRequestReview_51 extends _i1.Fake implements _i9.PullRequestReview {}
+class _FakePullRequestComment_51 extends _i1.Fake implements _i9.PullRequestComment {}
```
CaseyHillers, line 159: I'm not sure why I needed to regenerate these to get the tests to pass, but it might be related to us updating dependabot to run on cocoon

godofredoc: Yeah, I noticed that dependabot may be using a different version from what Dockerfile is using.

godofredoc (approved):

PR #1796, opened 2022-04-27
[app_dart] Fix release auto-approvals to only work on branches

CaseyHillers:
Fixes the issues with:
1. Was looking at HEAD and not the base (base is what the PR wants to merge into)
2. The check was looking for a null branch AND the branch to contain the default branch to skip reviewing it

Fixes https://github.com/flutter/flutter/issues/102017


keyonghan (approved):
LGTM

PR #1797, opened 2022-04-27
[test_utilities] Remove flutter upgrade from ci scripts

CaseyHillers:
This is logic that should be contained in the recipe.

Part of https://flutter-review.googlesource.com/c/recipes/+/29360 to unblock the fix for making cocoon post-submit green again


godofredoc (approved):

godofredoc:
Obsolete

PR #1807, opened 2022-04-29
Use cocoon default branch when fetching GOB file

keyonghan:
Fixes https://github.com/flutter/flutter/issues/102825


CaseyHillers (approved):

zanderso:
Can this change have a test?

christopherfujino:
> Can this change have a test? Are there any other places that would also need to be updated like this?

Yeah, since this just GETing a static web asset, maybe we should have an integration test that calls this function and verifies there's something there?

PR #1816, opened 2022-05-04
Revert "[scheduler] Process build messages with null task key (#1698)"

CaseyHillers:
This was temporarily added to recover from an outage. Removing as it seems to be causing some bugs in the release builds on go/engine-build

Fixes https://github.com/flutter/flutter/issues/102998


keyonghan (approved):
LGTM

PR #1827, opened 2022-05-06
[signing bot] add bot_type for existing post submit builds

XilaiZhang:
This is to distinguish regular building bots from code signing bots. 

Regular building bots and code signing bots will have different value for `bot_type`.

Context: go/code-signing-bot. Context of planned introduction of code signing bot: https://chrome-internal-review.googlesource.com/c/infradata/config/+/4685734/7

### DIFF
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -12,26 +12,27 @@ platform_properties:
   linux:
     properties:
       os: Linux
+      bot_type: "regular"
   mac:
     properties:
       os: Mac-10.15
+      bot_type: "regular"
   windows:
     properties:
       os: Windows
+      bot_type: "regular"
 
 targets:
   - name: Linux Cocoon
     recipe: cocoon/cocoon
     properties:
       add_recipes_cq: "true"
-      bot_type: "regular"
 
   - name: Linux device_doctor
     postsubmit: false
     recipe: cocoon/device_doctor
     properties:
       add_recipes_cq: "true"
-      bot_type: "regular"
     runIf:
       - device_doctor/**
 
@@ -40,7 +41,6 @@ targets:
     recipe: cocoon/device_doctor
     properties:
       add_recipes_cq: "true"
-      bot_type: "regular"
     runIf:
       - device_doctor/**
 
@@ -49,7 +49,6 @@ targets:
     recipe: cocoon/device_doctor
     properties:
       add_recipes_cq: "true"
-      bot_type: "regular"
     runIf:
       - device_doctor/**
 
@@ -57,4 +56,3 @@ targets:
     recipe: infra/ci_yaml
     properties:
       add_recipes_cq: "true"
-      bot_type: "regular"
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -12,15 +12,15 @@ platform_properties:
   linux:
     properties:
       os: Linux
-      bot_type: "regular"
+      signing_certs: "none"
   mac:
     properties:
       os: Mac-10.15
-      bot_type: "regular"
+      signing_certs: "none"
   windows:
     properties:
       os: Windows
-      bot_type: "regular"
+      signing_certs: "none"
 
 targets:
   - name: Linux Cocoon
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -24,12 +24,14 @@ targets:
     recipe: cocoon/cocoon
     properties:
       add_recipes_cq: "true"
+      bot_type: "regular"
 
   - name: Linux device_doctor
     postsubmit: false
     recipe: cocoon/device_doctor
     properties:
       add_recipes_cq: "true"
+      bot_type: "regular"
     runIf:
       - device_doctor/**
 
@@ -38,6 +40,7 @@ targets:
     recipe: cocoon/device_doctor
     properties:
       add_recipes_cq: "true"
+      bot_type: "regular"
     runIf:
       - device_doctor/**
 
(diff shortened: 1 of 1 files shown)


.ci.yaml
```diff
@@ -24,12 +24,14 @@ targets:
     recipe: cocoon/cocoon
     properties:
       add_recipes_cq: "true"
+      bot_type: "regular"
```
CaseyHillers, line 27: I don't think this is the approach we should go with. By default 99.9% of tasks are going to run on regular bots.

To get this working, you would need to update every ci.yaml and every flutter/infra starlark config

XilaiZhang: Right I am not sure if this is the optimal approach. This is an implementation based on https://chrome-internal-review.googlesource.com/c/infradata/config/+/4685734/7 . The gerrit commit has been stale for a while and I am not sure if there are better approaches.

I started a discussion on this approach in the design doc at https://docs.google.com/document/d/1EacFm_abJIzRj95J92f_VyHwvG_d0HWlQRsxAUSp4yc/edit?resourcekey=0-fA7GHZRgrN1eH6hfepveJg&disco=AAAAYxIzxwc. Would be great if people could drop some thoughts. Thank you!

XilaiZhang: update: updated to top level properties

.ci.yaml
```diff
@@ -12,12 +12,15 @@ platform_properties:
   linux:
     properties:
       os: Linux
+      bot_type: "regular"
```
CaseyHillers, line 15: Revisiting this, can we make the property more descriptive? Regular is vague. Maybe we could do something like:

signing_certs: None

Then we can add this in flutter/infra at the builder template

XilaiZhang: Good advice! I will update the robot configs to use this as well

.ci.yaml
```diff
@@ -12,15 +12,15 @@ platform_properties:
   linux:
     properties:
       os: Linux
-      bot_type: "regular"
+      signing_certs: "none"
```
CaseyHillers, line 15: nit: yaml can do `signing_certs: None`

CaseyHillers (changes requested):
https://cs.opensource.google/flutter/infra/+/main:config/lib/common.star;l=171 is where this change should be, not in the ci.yaml.

You may need to add this to the Cocoon scheduler as well as a default dimension, but we could follow up afterwards. I think the default flutter/infra dimension will be respected, but I'm not sure what the swarming merge logic looks like.

.ci.yaml
```diff
@@ -12,12 +12,15 @@ platform_properties:
...
       os: Linux
+      signing_certs: "none"
   mac:
     properties:
       os: Mac-10.15
+      signing_certs: "none"
   windows:
     properties:
       os: Windows
+      signing_certs: "none"
```
keyonghan, line 23: We only need to add this dimension for `Mac` bots.

XilaiZhang: Thanks for the comments! I think if I add it as a common property it will apply to all luci builders?

keyonghan: I was thinking about adding the dimension in the .ci.yaml, but `common` sounds easier to proceed. In that case, no extra work is needed to separate different platforms but the shared `none`. I am okay with `none` the default for all platforms.

XilaiZhang: sounds good! pushed up https://flutter-review.googlesource.com/c/infra/+/29800 (this infra change will wait after this week's flutter 3 release). I will abort the changes on .ci.yaml files for cocoon/engine/framework, and use common.star to request luci builders to auto populate the properties to ci.yaml.

.ci.yaml
```diff
@@ -12,12 +12,15 @@ platform_properties:
   linux:
     properties:
       os: Linux
+      signing_certs: "none"
```
godofredoc, line 15: Why we need a property?

XilaiZhang: For sure just a quick recap on what happened. The initial plan was to use a property to distinguish a code signing bot from a build*-m9 chromium bot. A code signing bot and a m9 bot share the same configs, and a code signing bot might pick up and run post-submit tests if not differentiated. 

There were two approaches on how to add a property to distinguish the bots. 
1. One was to modify ci.yaml directly, either through platform properties or through build properties. 
2. the other was to modify the luci scheduler in common.star in infra, so that the scheduler adds in a default property to each build, and thus change all builds in ci.yaml

The current plan is to use the second approach, because the change required is smaller in scale. We are planning to change infra after this week's flutter 3 release in case it would break anything.

godofredoc: Is this because the scheduler does not recognize dimensions and we need to map/translate properties to dimensions?
if this is the case I would suggest adding proper support for dimensions instead of work around it. This will prevent having to change the scheduler/infra-configs and .ci.yaml whenever we need to introduce a new dimension.

CaseyHillers: +1 - This would be a good yak shave. I filed https://github.com/flutter/flutter/issues/103361

XilaiZhang: Right my understanding is that dimensions are the things we pass into luci to tell it which bots should be scheduled to pick up certain builds. And properties are the attributes passed into individual builds and bots, which is related to how the build should be executed.

Oh I didn't know we were trying to circumvent a property/dimension mismatch by supplying dimensions as properties in ci.yaml. Sure I will research and take on https://github.com/flutter/flutter/issues/103361

PR #1831, opened 2022-05-09
[release] exclude forks for branch datastore

XilaiZhang:
fixes https://github.com/flutter/flutter/issues/103363.

We will not store branches that are created in a forked flutter repository.

### DIFF
--- a/app_dart/lib/src/service/branch_service.dart
+++ b/app_dart/lib/src/service/branch_service.dart
@@ -35,8 +35,12 @@ class BranchService {
       return;
     }
     final String? branch = createEvent.ref;
-    final String? repository = createEvent.repository!.slug().fullName;
+    final String repository = createEvent.repository!.slug().fullName;
     final int lastActivity = createEvent.repository!.pushedAt!.millisecondsSinceEpoch;
+
+    if (repository.split('/')[0] != 'flutter') {
+      return;
+    }
 
     final String id = '$repository/$branch';
     final Key<String> key = datastore.db.emptyKey.append<String>(Branch, id: id);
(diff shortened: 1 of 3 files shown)


app_dart/lib/src/service/branch_service.dart
```diff
@@ -35,9 +35,13 @@ class BranchService {
       return;
     }
     final String? branch = createEvent.ref;
-    final String? repository = createEvent.repository!.slug().fullName;
+    final String repository = createEvent.repository!.slug().fullName;
     final int lastActivity = createEvent.repository!.pushedAt!.millisecondsSinceEpoch;
 
+    if (repository.split('/')[0] != 'flutter') {
```
CaseyHillers, line 41: We may want to support future orgs like dart-sdk. Is there a way we can avoid hard coding this to only support Flutter?

XilaiZhang: yes sir

app_dart/lib/src/service/branch_service.dart
```diff
@@ -37,8 +37,9 @@ class BranchService {
     final String? branch = createEvent.ref;
     final String repository = createEvent.repository!.slug().fullName;
     final int lastActivity = createEvent.repository!.pushedAt!.millisecondsSinceEpoch;
+    final bool forked = createEvent.repository!.isFork;
 
-    if (repository.split('/')[0] != 'flutter') {
+    if (forked) {
```
CaseyHillers, line 42: Neat! Much better

CaseyHillers (approved):

PR #1832, opened 2022-05-09
[app_dart] Remove branches.txt

CaseyHillers:
This is no longer used as we're storing them in datastore


XilaiZhang (approved):
images (1)

PR #1834, opened 2022-05-10
Catch exception to ack pub/sub message

keyonghan:
Fix: https://github.com/flutter/flutter/issues/103168

This PR catches exception when retry limit is hit to batch schedule builds.

### DIFF
--- a/app_dart/lib/src/request_handlers/scheduler/request_subscription.dart
+++ b/app_dart/lib/src/request_handlers/scheduler/request_subscription.dart
@@ -57,16 +57,25 @@ class SchedulerRequestSubscription extends SubscriptionHandler {
       throw BadRequestException(e.toString());
     }
 
-    await retryOptions.retry(
-      () async {
-        final List<Request> requestsToRetry = await _sendBatchRequest(request);
-        request = BatchRequest(requests: requestsToRetry);
-        if (requestsToRetry.isNotEmpty) {
-          throw const InternalServerError('Failed to schedule builds');
-        }
-      },
-      retryIf: (Exception e) => e is InternalServerError,
-    );
+    /// Retry scheduling builds upto 3 times.
+    ///
+    /// Log error message when still failing after retry. Avoid endless rescheduling
+    /// by acking the pub/sub message without throwing an exception.
+    try {
+      await retryOptions.retry(
+        () async {
+          final List<Request> requestsToRetry = await _sendBatchRequest(request);
+          request = BatchRequest(requests: requestsToRetry);
+          if (requestsToRetry.isNotEmpty) {
+            throw const InternalServerError('Failed to schedule builds.');
+          }
+        },
+        retryIf: (Exception e) => e is InternalServerError,
+      );
+    } catch (e) {
+      log.warning('Failed to schedule builds.');
+      return Body.forString('Failed to schedule builds.');
+    }
 
     return Body.empty;
   }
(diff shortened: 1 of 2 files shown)


app_dart/lib/src/request_handlers/scheduler/request_subscription.dart
```diff
@@ -57,16 +57,25 @@ class SchedulerRequestSubscription extends SubscriptionHandler {
...
+        () async {
+          final List<Request> requestsToRetry = await _sendBatchRequest(request);
+          request = BatchRequest(requests: requestsToRetry);
+          if (requestsToRetry.isNotEmpty) {
+            throw const InternalServerError('Failed to schedule builds.');
+          }
+        },
+        retryIf: (Exception e) => e is InternalServerError,
+      );
+    } catch (e) {
```
CaseyHillers, line 75: I don't think this is what we want. It removes the error retrying from pubsub completely, which is one of the main reasons we use it. For example, if GitHub has an outage, we'll throw away LUCI updates during that period, and have a hard recovery.

We should update the results subscription to ack in the failing case, and to try requeuing the message with only the builds that were failed to schedule.

app_dart/lib/src/request_handlers/scheduler/request_subscription.dart
```diff
@@ -57,16 +57,25 @@ class SchedulerRequestSubscription extends SubscriptionHandler {
...
+      log.warning('Failed to schedule builds.');
```
godofredoc, line 76: Let's print the builds that are failing to ensure that we can debug any not expected failures.

app_dart/lib/src/request_handlers/scheduler/request_subscription.dart
CaseyHillers: Another case, we could alleve the existing issues by acking schedule build failures, and not throw an exception. Generally, I would avoid this logic in the generic subscription handler.

godofredoc: We may need to put in a balance the pros and cons of using pub/sub messages as a retry mechanism. Deterministic signals are preferred over unexpected eventual consistent  behavior.

That being said if we really want to continue with this approach I'd recommend the following refactoring:

* Create one message per build, this way we can remove the retry logic from here and avoid duplicated builds as a single passed/failed schedule attempt will be associated with the message
* Ensure that messages are not being duplicated - this requires to update the subscription to ensure a single message is delivered. This will prevent the message queue to grow exponentially when there is an outage and prevent multiple builds for the same commit/builder combination.
* Update the configurations to ensure the messages eventually expire, there will be messages that will fail forever until they expire.

keyonghan: > Generally, I would avoid this logic in the generic subscription handler.

This is for the batch scheduling handler only.

CaseyHillers: Ah, you're right. The auto_submit bot confused my sense of naming for these. I was thinking this was https://cs.opensource.google/flutter/cocoon/+/main:app_dart/lib/src/request_handling/subscription_handler.dart, which is very close.

This LGTM, but I'd like to see a TODO for moving this so every request is a single build

CaseyHillers (approved):

app_dart/lib/src/request_handlers/scheduler/request_subscription.dart
keyonghan: https://github.com/flutter/flutter/issues/103443 to track.

app_dart/lib/src/request_handlers/scheduler/request_subscription.dart
keyonghan: Done.

PR #1839, opened 2022-05-11
[ciyaml] add dimension support

XilaiZhang:
part of https://github.com/flutter/flutter/issues/103361. related change on the infra side https://flutter-review.googlesource.com/c/infra/+/29880.

A target should return aggregated dimensions of platform dimensions and target specific dimensions.


app_dart/lib/src/model/ci_yaml/target.dart
```diff
@@ -40,12 +40,30 @@ class Target {
...
+
+    final Map<String, Object> targetDimensions = _getTargetDimensions();
+    for (String key in targetDimensions.keys) {
+      String value = targetDimensions[key].toString();
+      dimensions.add(RequestedDimension(key: key, value: value));
+    }
+
+    final Map<String, Object> properties = getProperties();
+
     for (String dimension in dimensionList) {
```
keyonghan, line 67: We should deprecate dimensions defined in properties to make the logic more clear. With that, this would not be needed any more.

To remove this logic, we need to make sure .ci.yaml have been updated to move `dimension` values under `properties` to the newly supported `dimensions` section.

Let's add a TODO to remove this logic when .ci.yaml are updated.

app_dart/lib/src/model/proto/internal/scheduler.proto
```diff
@@ -20,6 +20,7 @@ message SchedulerConfig {
     message Properties {
         // Generic key, value that can be passed anywhere.
         map<string, string> properties = 1;
+        map<string, string> dimensions = 2;
```
keyonghan, line 23: It's a bit confusing to read dimensions under `Properties`.

app_dart/lib/src/model/proto/internal/scheduler.proto
```diff
@@ -20,6 +20,7 @@ message SchedulerConfig {
     message Properties {
```
keyonghan, line 20: Nit: update `Next ID` value.

app_dart/lib/src/model/proto/internal/scheduler.proto
XilaiZhang: sure, renamed to PropertiesOrDimensions

keyonghan: How about we call it `map<string, PlatformArgs> platform_args = 3`? Doc also needs updates.

app_dart/lib/src/model/ci_yaml/target.dart
keyonghan: Nit: a TODO comes with a bug created to track.

app_dart/lib/src/model/proto/internal/scheduler.proto
XilaiZhang: sounds good! umm is there a related doc for this proto?

XilaiZhang: umm this would also involve updating every ci.yaml file to use platform_args instead of platform_properties. Do we want to proceed with changing every ci.yaml file?

app_dart/lib/src/model/ci_yaml/target.dart
```diff
@@ -40,12 +40,32 @@ class Target {
...
+
+    final Map<String, Object> targetDimensions = _getTargetDimensions();
+    for (String key in targetDimensions.keys) {
+      String value = targetDimensions[key].toString();
+      dimensions.add(RequestedDimension(key: key, value: value));
+    }
+
+    final Map<String, Object> properties = getProperties();
+
+    // TODO(xilaizhang): remove this logic after dimensions such as 'cpu' in ci.yaml files
```
CaseyHillers, line 67: nit: https://dart-lang.github.io/linter/lints/flutter_style_todos.html

TODOs should be actionable, and the best way to do it is by linking to a GitHub issue

app_dart/lib/src/model/ci_yaml/target.dart
```diff
@@ -40,12 +40,32 @@ class Target {
...
+    final Map<String, Object> platformDimensions = _getPlatformDimensions();
+    for (String key in platformDimensions.keys) {
+      String value = platformDimensions[key].toString();
+      dimensions.add(RequestedDimension(key: key, value: value));
+    }
+
+    final Map<String, Object> targetDimensions = _getTargetDimensions();
+    for (String key in targetDimensions.keys) {
+      String value = targetDimensions[key].toString();
+      dimensions.add(RequestedDimension(key: key, value: value));
```
CaseyHillers, line 62: This doesn't handle the case where the target overrides the platform dimension.

If platform's cpu is x86, and the target wants arm64 it'll request both and cause an infra failure

app_dart/lib/src/model/proto/internal/scheduler.proto
CaseyHillers: I'm against adding `Or` to the name as it's an anti-pattern (limits us to two fields). The next person to change it will need to generalize it. I think calling it `PlatformProperties` is fine, as that's what the key is already called

app_dart/test/model/ci_yaml/target_test.dart
```diff
@@ -240,19 +256,35 @@ void main() {
         expect(target.getDimensions().length, 0);
       });
 
-      test('dimensions exit', () {
-        final Target target = generateTarget(1, properties: <String, String>{'os': 'abc', 'cpu': 'x64'});
+      test('platform dimensions and target dimensions are combined', () {
```
CaseyHillers, line 259: Please add a test for the case where the target overrides the platform

app_dart/lib/src/model/proto/internal/scheduler.proto
keyonghan: I see. In that case, let's update the doc on line 16-17 to reflect the dimension `property`.

XilaiZhang: sounds good! changed naming, updated TODO, added override and updated tests.

keyonghan (approved):
LGTM

app_dart/lib/src/model/proto/internal/scheduler.proto
```diff
@@ -14,17 +14,19 @@ message SchedulerConfig {
...
-    // Next ID: 2
-    message Properties {
-        // Generic key, value that can be passed anywhere.
+    // Keys are the platforms and values are the PlatformProperties (properties, dimensions etc.).
+    map<string, PlatformProperties> platform_properties = 3;
+    // Next ID: 3
+    message PlatformProperties {
+        // Generic key, value pairs to set platform-wise properties
         map<string, string> properties = 1;
+        // Generic key, value pairs to set platform-wise dimensions
```
CaseyHillers, line 23: Can you add a doc linking to the swarming dimensions docs?

app_dart/test/model/ci_yaml/target_test.dart
```diff
@@ -13,6 +13,22 @@ import '../../src/utilities/entity_generators.dart';
 void main() {
   group('Target', () {
     group('properties', () {
+      test('properties with ignore_flakiness true', () {
```
CaseyHillers, line 16: This diff doesn't seem related your pr. Can you try merging with upstream?

app_dart/test/model/ci_yaml/target_test.dart
```diff
@@ -240,19 +256,50 @@ void main() {
...
+        expect(dimensions.length, 3);
+        expect(dimensions[0].key, 'signing_cert');
+        expect(dimensions[0].value, 'none');
+        expect(dimensions[1].key, 'os');
+        expect(dimensions[1].value, 'abc');
+        expect(dimensions[2].key, 'cpu');
+        expect(dimensions[2].value, 'x64');
+      });
+
+      test('target specific dimensions overrides platform dimensions', () {
```
CaseyHillers, line 278: Can you add a test that the legacy dimensions are still used (such as property cpu), and ensure that the new dimensions fields are prioritized over them

app_dart/test/model/ci_yaml/target_test.dart
```diff
@@ -290,6 +274,19 @@ void main() {
...
+          1,
+          platform: 'Windows',
+          dimensions: <String, String>{'cpu': 'x64'},
+          properties: <String, String>{'cpu': 'x32'},
+        );
+        final List<RequestedDimension> dimensions = target.getDimensions();
+        expect(dimensions.length, 1);
+        expect(dimensions[0].key, 'cpu');
+        expect(dimensions[0].value, 'x64');
+      });
```
CaseyHillers, line 288: Can we also add a test for the case of targets being able to override the legacy platform property dimensions

XilaiZhang: for sure! added

PR #1861, opened 2022-05-17
Codesigning standalone

XilaiZhang:
example usage: `dart run bin/main.dart --commit=<commit number> --production=false --filepath=darwin-x64/FlutterMacOS.framework.zip#ios/artifacts.zip#dart-sdk-darwin-arm64.zip`

TODO: add tests to detect exception messages.

06/10: finalized tests and added documentations. made a first split to https://github.com/flutter/cocoon/pull/1899 for easier review.

06/09: Complex, nested, mixed file system structures tests. Can test an intricate combination of folder, zip, and binary file types. Rewrote and overrode code sign logic and classes. cleanup and re-factor is needed.

06/08: Changes include the initializations of test variables and test classes, the simulation of code sign workflow through fakeCommand and fakeProcesses, and the test of basic recursive folder structures through overriding classes with alternative logics. Some of the helper functions I added are long and not obvious to understand, I will document them later in the standalone design doc (apart from the main one). 

06/06: There is a lot to talk about. So I wrote go/code-signing-standalone to document the changes. I also inherited the cool designs from https://github.com/flutter/flutter/pull/101732, but didn't talk about them in the doc. In the doc I mainly focused on the changes.

06/03: Code sign functionality is now complete (can still be optimized). Based on file types, we can now identify unsigned binary files and undocumented binary files as we recursively traverse the file directory. We detect symlink files and symlink folders along the way, and ignore them when necessary. Added protection flag to address https://github.com/flutter/flutter/issues/105041, so we only upload back to GCS when absolutely necessary. Restored multi-upload and multi-notarize logic. 

05/17: after several rounds of debugging, I eventually was able to get the current app recursively examine every file in a directory. And based on file type of zip file/binary file/remote zip (using mime and ls), pick out binaries to code sign. This is the first major step to remove dependency of a hard coded archive structure. 

It is currently unfinished and not fully cleaned up(so there are unoptimized code all over the place). Future work will be to add in and parse a txt file of entitlements, which is extracted form downloaded zip.

current working version: can run `dart run bin/main.dart --commit=fce741dd83cf29624bdbed4c663a9a022fb1f559 --filepath=ios/artifacts.zip#darwin-x64/FlutterMacOS.framework.zip`. which takes in an engine hash, and a list of file paths to be code signed.


XilaiZhang:
cc @CaseyHillers @christopherfujino @godofredoc

codesign/.gitignore
```diff
@@ -0,0 +1,6 @@
+# Files and directories created by pub.
+.dart_tool/
+.packages
+
+# Conventional directory for build output.
+build/
```
CaseyHillers, line 6: Let's split this into several smaller PRs.

For example, my initial thoughts is codesign isn't hooked up into the test logic, so Linux Cocoon isn't running it. We can discuss that more in the initial codesign project creation PR with no extra code added.

codesign/LICENSE
```diff
@@ -0,0 +1,25 @@
+Copyright 2020 The Flutter Authors. All rights reserved.
```
christopherfujino, line 1: You don't need to add a new LICENSE file, this would be covered by the repo's existing license file: https://github.com/flutter/cocoon/blob/main/LICENSE

XilaiZhang: sounds good! I have newer version that is almost ready for christopher review. Currently stuck up with flutter g3 roll but once that is in a good state i will push the new version up.

codesign/test/cli_test.dart
```diff
@@ -0,0 +1,7 @@
+// Copyright 2022 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:test/test.dart';
+
+void main() {}
```
CaseyHillers, line 7: Please add tests.

This is a prod tool, and it should have a robust test suite to prevent breakages.

XilaiZhang: sure, I will add tests

XilaiZhang: tests are now complete!

CaseyHillers (changes requested):
Please split this PR into smaller PRs (aim for <300 LOC, but an initial PR for just a `dart create` is fine). In its current state, 3k LOC isn't reviewable.

PR #1878, opened 2022-05-19
[dashboard] Migrate to flutter_app_icon

codebird:
@CaseyHillers I've started writing the plugin. favicon was easy, android is a bit more complicated... 2 tests will fail, but have no idea why. Let me know what you think.

### DIFF
--- a/dashboard/lib/build_dashboard_page.dart
+++ b/dashboard/lib/build_dashboard_page.dart
@@ -16,6 +16,7 @@ import 'widgets/error_brook_watcher.dart';
 import 'widgets/filter_property_sheet.dart';
 import 'widgets/task_box.dart';
 import 'widgets/task_grid.dart';
+import 'package:icon_plugin/icon_plugin.dart';
 
 /// Shows information about the current build status of flutter/flutter.
 ///
@@ -40,6 +41,7 @@ class BuildDashboardPageState extends State<BuildDashboardPage> {
   TaskGridFilter? _settingsBasis;
   bool _smallScreen = false;
   double screenWidthThreshold = 600;
+  final _iconPlugin = IconPlugin();
 
   /// Current Flutter repository to view.
   String? repo;
@@ -66,6 +68,12 @@ class BuildDashboardPageState extends State<BuildDashboardPage> {
     _filter!.addListener(() {
       setState(() {});
     });
+  }
+
+  @override
+  void dispose() {
+    _iconPlugin.setIcon(icon: 'favicon.png');
+    super.dispose();
   }
 
   /// Convert the fields from this class into a URL.
@@ -315,8 +323,10 @@ class BuildDashboardPageState extends State<BuildDashboardPage> {
       return 'Loading...';
     }
     if (buildState.isTreeBuilding!) {
+      _iconPlugin.setIcon(icon: 'favicon.png');
       return 'Tree is Open';
     } else {
+      _iconPlugin.setIcon(icon: 'favicon-failure.png');
       if (buildState.failingTasks.isNotEmpty) {
         return 'Tree is Closed (failing: ${buildState.failingTasks.join(', ')})';
       } else {
--- a/dashboard/lib/build_dashboard_page.dart
+++ b/dashboard/lib/build_dashboard_page.dart
@@ -16,7 +16,7 @@ import 'widgets/error_brook_watcher.dart';
 import 'widgets/filter_property_sheet.dart';
 import 'widgets/task_box.dart';
 import 'widgets/task_grid.dart';
-import 'package:icon_plugin/icon_plugin.dart';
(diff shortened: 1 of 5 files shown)


dashboard/pubspec.lock
```diff
@@ -268,6 +268,13 @@ packages:
       url: "https://pub.dartlang.org"
     source: hosted
     version: "4.0.0"
+  icon_plugin:
```
CaseyHillers, line 271: Unfortunately, we can't pull this into this project if it's licensed with GNU GPL v3

dashboard/pubspec.yaml
```diff
@@ -30,6 +30,7 @@ dependencies:
   protobuf: ^2.0.1
   provider: ^6.0.2
   url_launcher: ^6.0.6
+  icon_plugin: ^0.0.2
```
CaseyHillers, line 33: I think calling it icon is going to cause confusion. See https://api.flutter.dev/flutter/material/Icons-class.html where the namespace is dedicated for font like icons.

Maybe flutter_app_icons would give better visibility in the pub.dev ecosystem?

CaseyHillers (changes requested):

dashboard/pubspec.yaml
CaseyHillers: Just to follow up, I do appreciate you writing the plugin for this! This is great, and I'm happy to see this functionality get added to the ecosystem

codebird: :) I won't take it badly even if the pull request doesn't land. I used flutter for a small project, and I'll be using it for a much bigger project. So trying to give back as much as I can.

codebird: > I think calling it icon is going to cause confusion. See https://api.flutter.dev/flutter/material/Icons-class.html where the namespace is dedicated for font like icons.
> 
> Maybe flutter_app_icons would give better visibility in the pub.dev ecosystem?

True concerning the naming. I'll change it. Can you clue me concerning the failing tests too pls?

dashboard/pubspec.lock
CaseyHillers: As a disclaimer, I'm not a lawyer :-)

Most popular open source licenses will work, this is one of the few that prevents us from including it. You could probably swipe through our existing pubspec to see, but I think the majority are MIT.

dashboard/lib/build_dashboard_page.dart
```diff
@@ -16,6 +16,7 @@ import 'widgets/error_brook_watcher.dart';
 import 'widgets/filter_property_sheet.dart';
 import 'widgets/task_box.dart';
 import 'widgets/task_grid.dart';
+import 'package:icon_plugin/icon_plugin.dart';
```
CaseyHillers, line 19: re the failing test:

```
The following MissingPluginException was thrown running a test (but after the test had completed):
```

You'll need to update the plugin so that if there's no implementation for the current platform, it no-ops all calls.

There's docs in https://docs.flutter.dev/development/packages-and-plugins/developing-packages#plugin and it's generally referred to as Flutter's federated plugin structure. You'll define a generic interface that then all the platforms implement.

dashboard/pubspec.lock
codebird: hahaha thought you are! All the plugins I looked at are made by flutter team :D. I'll take a look again in few.

CaseyHillers: https://pub.dev/packages/github is an example we use a lot in Cocoon.

dashboard/pubspec.yaml
```diff
@@ -30,7 +30,7 @@ dependencies:
   protobuf: ^2.0.1
   provider: ^6.0.2
   url_launcher: ^6.0.6
-  flutter_app_icons: ^0.0.2
+  flutter_app_icons: ^0.0.3
```
CaseyHillers, line 33: When I go to pub.dev's site for this, it redirects to https://github.com/codebird/flutter_app_icons for the source code which is an empty repo. Can you upload the source code?

codebird: Change to master branch

codebird: Updated default branch

codebird:
> flutter-lgtm
> 
> Thanks for making a plugin!

Updated to work for Android too, but the issue is that flutter doesn't take into consideration activity aliases so making it work is a bit of a pain. Check the Readme on github

godofredoc:
@codebird is this ready to land?

CaseyHillers:
> @codebird is this ready to land?

@godofredoc this needs a second flutter hacker to review

godofredoc (approved):

godofredoc:
Thanks for improving the application!

PR #1879, opened 2022-05-19
[app_dart] branch creation: catch exceptions and provide better logging

XilaiZhang:
context: https://github.com/flutter/flutter/issues/103996.

This commit adds more catch blocks to potentially catch previously foreseeable exceptions, and prints more logging information in appengine for better analysis.

Tried to fix the issue and investigated appengine logs multiple times with different queries but the logs are all legit. Also created different tests for different json webhook responses and tested locally, and the tests all passed. Couldn't dig out valuable information from current setup. Decided to add more catch blocks and loggings to help diagnose the issue.


keyonghan (approved):
LGTM

CaseyHillers (approved):

PR #1883, opened 2022-05-26
Validate ci.yaml dependencies are pinned to a specific version

ricardoamador:
## Description of The Issue Being Addressed
The builder config is fragile when dependencies are not pinned to specific versions. Any upstream changes may cause breakage. We should create a test to validate all .ci.yaml dependencies are pinned. This also requires that we pin specific versions for all dependencies.

## Issues Addressed
#101712 - https://github.com/flutter/flutter/issues/101712

## Pre-launch Checklist

- [x] I signed the [CLA].


app_dart/lib/src/model/ci_yaml/ci_yaml.dart
```diff
@@ -176,3 +184,34 @@ class CiYaml {
...
+        if (version.isEmpty || version == 'latest') {
+          exceptions.add('ERROR: dependency ${depMap['dependency']} must have a non empty, non "latest" version supplied.');
+        }
+      }
+    }
+
+    // if (exceptions.isNotEmpty) {
+    //   final String fullException = exceptions.reduce((String exception, _) => exception + '\n');
+    //   throw FormatException(fullException);
+    // }
```
keyonghan, line 215: Nit: remove.

app_dart/lib/src/model/ci_yaml/ci_yaml.dart
```diff
@@ -176,3 +184,34 @@ class CiYaml {
     }
   }
 }
+
+/// Class to expose the PinnedVersion method for testing.
+class DependencyValidator {
+  /// Dependency is guaranteed to be non empty as it must be found before this
```
keyonghan, line 190: Nit: Let's update the doc a bit to follow the style: https://dart.dev/guides/language/effective-dart/documentation. Like, use sentence, capital for the first word, etc.

app_dart/lib/src/model/ci_yaml/ci_yaml.dart
```diff
@@ -176,3 +184,34 @@ class CiYaml {
...
+  /// Checks a dependency string for a pinned version.
+  /// If a version is found then it must not be empty or 'latest.'
+  static void hasVersion({required String dependencyJsonString}) {
+    final List<String> exceptions = <String>[];
+    // decoded will contain a list of maps
+    dynamic decoded = json.decode(dependencyJsonString);
+
+    for (Map<String, dynamic> depMap in decoded) {
+    
+      if (!depMap.containsKey('version')) {
```
keyonghan, line 202: Do we need a skip list, like `xcode`?

keyonghan (commented):
This PR should wait until versions are pinned from .ci.yaml side.

app_dart/lib/src/model/ci_yaml/ci_yaml.dart
ricardoamador: My fault. I mean to keep this. Good catch though.

app_dart/lib/src/model/ci_yaml/ci_yaml.dart
ricardoamador: Can you please clarify here? Do you mean we skip certain dependencies in this check or is it something else?

keyonghan: Yeah, I meant if we need to skip certain dependencies in this check, such as `xcode`, whose dependency version would not be counted (we rely on property value instead).
After a rethink, I think it would be good to keep all dependencies consistent with a version.
Let's keep the current logic as you have right now.

app_dart/lib/src/model/ci_yaml/ci_yaml.dart
```diff
@@ -165,6 +167,13 @@ class CiYaml {
           }
         }
       }
+
+      /// Check the dependencies for the current target if it is viable and to
+      /// be added to graph.
+      final String? dependencyJson = target.properties['dependencies'];
+      if (dependencyJson != null) {
```
CaseyHillers, line 174: This will need to temporarily skip on release branches. When the ci.yaml changes roll to stable (O(months)), we can make.it the default everywhere

ricardoamador: Is the branch something that can be gathered from the target.proto?

CaseyHillers: You should be able to pass it to _validate

https://cs.opensource.google/flutter/cocoon/+/main:app_dart/lib/src/model/ci_yaml/ci_yaml.dart;l=23

ricardoamador: How would I know based on the branch string if it is a release branch?

CaseyHillers:
FYI - Merging with ToT won't fix the Linux Cocoon failures

One of the integration tests is failing:

```
app_dart/integration_test/generate_jspb_test.dart: validate cocoon ci.yaml generates jspb
```

ricardoamador:
@CaseyHillers Yeah I missed a push to the packages change and it will pass once that gets merged. Thanks Casey!

CaseyHillers (approved):
flutter-lgtm

keyonghan (approved):
LGTM

PR #1884, opened 2022-05-26
remove package logging

christopherfujino:
this was not actually logging anywhere anyway, so remove the package. some of the logs were redundant to what was printed to stderr, I simply deleted these. Some seemed useful diagnostics to show, I ported these to `stdout.writeln()`. And printing the executed command seemed unimportant so I simply deleted (I can port it to stdout.writeln() if it would be useful).

No tests were affected by this change.


keyonghan (approved):
Thanks for the clean up @christopherfujino ! LGTM.

PR #1888, opened 2022-06-01
[app_dart] branch logic fix: make 100 percent sure request is passed in

XilaiZhang:
continuing effort to tackle (https://github.com/flutter/flutter/issues/103996).

Casey and Godofredo have both observed branches not being correctly added to datastore. 
This PR removes the assumption that GithubWebhook instance is re-created upon every request, and assures us a better guarantee that the string request is passed in.


app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -85,7 +85,7 @@ class GithubWebhook extends RequestHandler<Body> {
           }
           break;
         case 'create':
-          branchService ??= BranchService(datastore, rawRequest: stringRequest);
```
XilaiZhang, line 88: After many thinkings and testings, my deduction is that:
Cocoon might not be creating a new webhook instance every time we send a request. If the webhook instance is reused over time, then the string request carried by the new event will not make it into the old instance.

CaseyHillers (approved):

PR #1893, opened 2022-06-03
[dashboard] Update tree favicons

codebird:
Fixes flutter/flutter#105035 

I detached the favicon update from AnimationBuilder, as AnimationBuilder doesn't update when the tab is inactive...


CaseyHillers (approved):

gaaclarke (approved):
LGTM, do you think we should update the docstring for AnimationBuilder to mention this?

codebird:
> LGTM, do you think we should update the docstring for AnimationBuilder to mention this?

I guess yes. It is logical but you don't think about it unless it happens.

PR #1895, opened 2022-06-06
do not re-trigger builds when marking ready for review

christopherfujino:
We already test when drafts are opened, so we don't need to re-trigger when marked ready for review.

### DIFF
--- a/app_dart/test/request_handlers/github_webhook_test.dart
+++ b/app_dart/test/request_handlers/github_webhook_test.dart
@@ -273,6 +273,80 @@ void main() {
       )).called(1);
     });
 
+    // We already schedule checks when a draft is opened, don't need to re-test
+    // just because it was marked ready for review
+    test('Does nothing on ready_for_review', () async {
+      const int issueNumber = 123;
+      request.headers.set('X-GitHub-Event', 'pull_request');
+      request.body = generatePullRequestEvent(
+        'ready_for_review',
+        issueNumber,
+        kDefaultBranchName,
+      );
+      final Uint8List body = utf8.encode(request.body!) as Uint8List;
+      final Uint8List key = utf8.encode(keyString) as Uint8List;
+      final String hmac = getHmac(body, key);
+      request.headers.set('X-Hub-Signature', 'sha1=$hmac');
+
+      bool batchRequestCalled = false;
+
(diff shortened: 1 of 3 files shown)


app_dart/test/src/service/fake_buildbucket.dart
```diff
@@ -60,6 +60,9 @@ class FakeBuildBucketClient extends BuildBucketClient {
 
   @override
   Future<BatchResponse> batch(BatchRequest request) async {
+    if (batchResponse != null) {
```
christopherfujino, line 63: Previously some tests were setting this field but never actually used it.

app_dart/test/src/service/fake_buildbucket.dart
```diff
@@ -22,7 +22,7 @@ class FakeBuildBucketClient extends BuildBucketClient {
 
   Future<Build>? scheduleBuildResponse;
   Future<SearchBuildsResponse>? searchBuildsResponse;
-  Future<BatchResponse>? batchResponse;
+  Future<BatchResponse> Function()? batchResponse;
```
christopherfujino, line 25: one test was asserting that a closure was being invoked; however, that closure was actually being invoked before it was set here. Change this type to a callback so that it gets invoked in place doing the actual `batch()` call.

app_dart/test/request_handlers/github_webhook_test.dart
```diff
@@ -2051,8 +2125,7 @@ void foo() {
       expect(db.values.values.whereType<Commit>().length, 1);
     });
 
-    test('Does not test pest draft pull requests.', () async {
-      //FOR REVIEW : umm is pest a typo here or maybe some form of pr
+    test('Does not comment about needing tests on draft pull requests.', () async {
```
christopherfujino, line 2128: "test pest" was correct (means commenting that the author needs to write a test) but confusing.

XilaiZhang (approved):
images (1)

app_dart/test/request_handlers/github_webhook_test.dart
XilaiZhang: haha thanks for explaining! I have always wondered what a test pest is

PR #1896, opened 2022-06-07
Differentiate flake bot cron jobs time

keyonghan:
There are three cron jobs running for flake bot, and they are at the same time. This caused github issue when calling APIs too frequently.

This is a mitigation to differentiate their time.

Issue: https://github.com/flutter/flutter/issues/105277


ricardoamador (approved):
LGTM

PR #1900, opened 2022-06-13
[codesign] initialization of functions and inherited helpers

XilaiZhang:
This is the second split of https://github.com/flutter/cocoon/pull/1861.

This split include the entrance point of the app, `main.dart`, function definitions (without implementations), and inherited helper functions from conductor codesign written by Christopher.


XilaiZhang:
cc @CaseyHillers  . feel free to ignore since [go/chillers-ooo](https://goto.google.com/chillers-ooo)

codesign/lib/codesign.dart
```diff
@@ -2,6 +2,8 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-int calculate() {
-  return 6 * 7;
-}
+export 'src/globals.dart';
```
jmagman, line 5: Are you planning on making this package available to something outside of main.dart? 
I think you should remove this and explicitly import the dart files you need where you need them.

codesign/lib/src/codesign_context.dart
```diff
@@ -0,0 +1,92 @@
...
+    print('checking Xcode version...');
+    final ProcessResult result = processManager.runSync(
+      <String>[
+        'xcodebuild',
+        '-version',
+      ],
+    );
+    final List<String> outArray = (result.stdout as String).split('\n');
+    final int xcodeVersion = int.parse(outArray[0].split(' ')[1].split('.')[0]);
+    if (xcodeVersion <= 12) {
```
jmagman, line 55: I think we're on Xcode 13+ everywhere, if you want to remove all the `isNotaryTool` fallback logic.

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -0,0 +1,112 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:async';
+import 'package:codesign/codesign.dart';
+import 'package:file/file.dart';
+import 'package:process/process.dart';
+
+/// Interface for classes that interact with files nested inside of [RemoteZip]s.
+abstract class FileVisitor {
```
jmagman, line 11: Maybe you have this abstract for some future usage, but as-written I think you should just use `FileCodesignVisitor`, the interface isn't really necessary.

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -0,0 +1,112 @@
...
+      required this.codesignCertName,
+      required this.codesignPrimaryBundleId,
+      required this.codesignUserName,
+      required this.appSpecificPassword,
+      required this.codesignAppstoreId,
+      required this.codesignTeamId,
+      required this.stdio,
+      required this.isNotaryTool,
+      required this.filepaths,
+      this.production = false});
```
jmagman, line 40:
```suggestion
  FileCodesignVisitor({
    required this.tempDir,
    required this.commitHash,
    required this.processManager,
    required this.codesignCertName,
    required this.codesignPrimaryBundleId,
    required this.codesignUserName,
    required this.appSpecificPassword,
    required this.codesignAppstoreId,
    required this.codesignTeamId,
    required this.stdio,
    required this.isNotaryTool,
    required this.filepaths,
    this.production = false,
  });
```

codesign/bin/main.dart
```diff
@@ -0,0 +1,113 @@
...
+      help: 'Apple-id for connecting to app store. Used by notary service for xcode version 13+.',
+    )
+    ..addOption(
+      kCodesignTeamId,
+      help: 'Team-id is used by notary service for xcode version 13+.',
+    )
+    ..addOption(
+      kCodesignFilepath,
+      help:
+          'the # deliminated zip file paths to be codesigned. e.g. darwin-x64/font-subset.zip#darwin-x64-release/gen_snapshot.zip',
```
jmagman, line 74: Why `#` delineated? Can you use `addMultiOption` instead?

codesign/bin/main.dart
```diff
@@ -0,0 +1,113 @@
...
+    )
+    ..addOption(kCommit,
+        help: 'the commit hash of flutter/engine github pr used for google cloud storage bucket indexing')
+    ..addOption(kProduction, help: 'whether we are going to upload the artifacts back to GCS for production');
```
jmagman, line 78: Use `addFlag` for booleans.

codesign/bin/main.dart
```diff
@@ -0,0 +1,113 @@
...
+  return CodesignContext(
+          codesignCertName: codesignCertName,
+          codesignPrimaryBundleId: codesignPrimaryBundleId,
+          codesignUserName: codesignUserName,
+          commitHash: commit,
+          appSpecificPassword: appSpecificPassword,
+          codesignAppstoreId: codesignAppstoreId,
+          codesignTeamId: codesignTeamId,
+          codesignFilepath: codesignFilepath,
+          production: production)
```
jmagman, line 111: Use trailing commas.

codesign/lib/src/globals.dart
```diff
@@ -0,0 +1,52 @@
...
+
+import 'package:args/args.dart';
+
+class ConductorException implements Exception {
+  ConductorException(this.message);
+
+  final String message;
+
+  @override
+  String toString() => 'Exception: $message';
```
jmagman, line 15: I don't think you need to override `toString()` here.

codesign/lib/src/codesign_context.dart
```diff
@@ -0,0 +1,92 @@
...
+      appSpecificPassword: appSpecificPassword,
+      stdio: stdio,
+      codesignAppstoreId: codesignAppstoreId,
+      codesignTeamId: codesignTeamId,
+      isNotaryTool: isNotaryTool,
+      production: production,
+      filepaths: filepaths,
+    );
+
+    await codesignVisitor!.validateAll();
```
jmagman, line 89: Delete `tempDir` in a `finally`?
> /// This file will be deleted if [validateAll] completes successfully.

codesign/lib/codesign.dart
XilaiZhang: umm yeah I am planning to make it visible to tests and other dart files. would this structure be less ideal if I am am trying to build something like https://github.com/flutter/cocoon/pull/1861?

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: yep the trailing commas helped

codesign/bin/main.dart
XilaiZhang: for sure, switched to addMultiOption and updated document

codesign/bin/main.dart
XilaiZhang: sounds good, I switched to flag

codesign/bin/main.dart
XilaiZhang: sure, thanks for the advice!

codesign/lib/src/globals.dart
XilaiZhang: umm I did a quick test on removing it. I think overriding has the effect of showing 
`Exception: i am a test message` instead of `Instance of 'ConductorException'`, which is a bit more verbose.

codesign/lib/src/codesign_context.dart
XilaiZhang: good point! I added a clean up in production mode.

XilaiZhang:
Thanks for the advice! The newest commit incorporated the feedbacks.

codesign/bin/main.dart
```diff
@@ -0,0 +1,121 @@
...
+      help: 'Apple-id for connecting to app store. Used by notary service for xcode version 13+.',
+    )
+    ..addOption(
+      kCodesignTeamId,
+      help: 'Team-id is used by notary service for xcode version 13+.',
+    )
+    ..addMultiOption(
+      kCodesignFilepath,
+      help: 'the zip file paths to be codesigned. e.g. --filepath=darwin-x64/font-subset.zip'
+          ' --filepath=darwin-x64-release/gen_snapshot.zip --filepath=dart-sdk-darwin-arm64.zip',
```
jmagman, line 76: You can use `valueHelp` and change the text to something like:
```suggestion
      help: 'The zip file paths to be codesigned. Pass this option multiple times to codesign ',
          'multiple zip files'.
      valueHelp: 'darwin-x64/font-subset.zip',
```
The help text will look like:
```
--filepath=<darwin-x64/font-subset.zip>
```

codesign/lib/codesign.dart
jmagman: I think you have it the suggested way then:
https://dart.dev/guides/libraries/create-library-packages#organizing-a-library-package

codesign/lib/src/codesign_context.dart
```diff
@@ -0,0 +1,79 @@
...
+    stderr: stderr,
+    stdin: stdin,
+  );
+
+  void createTempDirectory() {
+    tempDir ??= fileSystem.systemTempDirectory.createTempSync('conductor_codesign');
+  }
+
+  Future<void> run() async {
+    print('production is $production, and its type is ${production.runtimeType}');
```
jmagman, line 49: Did you mean to leave this in?

codesign/lib/src/codesign_context.dart
```diff
@@ -0,0 +1,79 @@
...
+    // assume file path is passed in as connected by # sign. e.g. path1#path2#path3
```
jmagman, line 50: This isn't true anymore since you switch to a multi option.

codesign/lib/src/codesign_context.dart
```diff
@@ -0,0 +1,79 @@
...
+  final String codesignCertName;
+  final String codesignPrimaryBundleId;
+  final String codesignUserName;
+  final String appSpecificPassword;
+  final String codesignAppstoreId;
+  final String codesignTeamId;
+  final List<String> codesignFilepaths;
+  final String commitHash;
+  final bool production;
+  Directory? tempDir;
```
jmagman, line 33: Does this need to be an instance variable? Looks like it can just be a nonnull local set in `run`.

codesign/lib/src/file_codesign_visitor.dart
jmagman: I don't really have a preference, will let Chris chime in.

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -0,0 +1,111 @@
...
+  final String appSpecificPassword;
+  final String codesignAppstoreId;
+  final String codesignTeamId;
+  final Stdio stdio;
+  final bool production;
+  final List<String> filepaths;
+  // Utility utility = Utility();
+  // TODO(xilaizhang): add back utitlity in later splits
+  Set<String>? fileWithEntitlements;
+  Set<String>? fileWithoutEntitlements;
```
jmagman, line 61: I'm guessing these will be used in a later PR.  To avoid checking for these being empty all the time it's probably better to initialize them as empty, unless them being null is meaningful.
```suggestion
  final Set<String> fileWithEntitlements = <String>{};
  final Set<String>? fileWithoutEntitlements = <String>{};
```

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -0,0 +1,111 @@
...
+  Set<String> fileConsumed = <String>{};
+
+  late final File entitlementsFile = tempDir.childFile('Entitlements.plist')
+    ..writeAsStringSync(_entitlementsFileContents);
+
+  late final Directory remoteDownloadsDir = tempDir.childDirectory('downloads')..createSync();
+  late final Directory codesignedZipsDir = tempDir.childDirectory('codesigned_zips')..createSync();
+
+  int _remoteDownloadIndex = 0;
+  int get remoteDownloadIndex => _remoteDownloadIndex++;
```
jmagman, line 71: This side effect is unexpected...

codesign/bin/main.dart
```diff
@@ -0,0 +1,121 @@
...
+    )
+    ..addOption(
+      kCommit,
+      help: 'the commit hash of flutter/engine github pr used for google cloud storage bucket indexing',
+    )
+    ..addFlag(
+      kProduction,
+      help: 'whether we are going to upload the artifacts back to GCS for production',
+    );
```
jmagman, line 85: This defaults to `false`, is that what you intended here? Just checking.

codesign/lib/src/codesign_context.dart
XilaiZhang: Yep l put it as instance variable because later in tests I override `codesignVisitor` with `fakeCodesignVisitor`, and set the tempDir of the fakeCodesignVisitor with `codesignContext.tempDir`.

codesign/bin/main.dart
XilaiZhang: yep

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -0,0 +1,111 @@
...
+import 'dart:async';
+import 'package:codesign/codesign.dart';
+import 'package:file/file.dart';
+import 'package:process/process.dart';
+
+/// Interface for classes that interact with files nested inside of [RemoteZip]s.
+abstract class FileVisitor {
+  const FileVisitor();
+
+  Future<void> visitEmbeddedZip(EmbeddedZip file, String parent, String entitlementParentPath);
```
godofredoc, line 14: Docs for the interface methods?

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -0,0 +1,111 @@
...
+  final String codesignCertName;
+  final String codesignPrimaryBundleId;
+  final String codesignUserName;
+  final String appSpecificPassword;
+  final String codesignAppstoreId;
+  final String codesignTeamId;
+  final Stdio stdio;
+  final bool production;
+  final List<String> filepaths;
+  // Utility utility = Utility();
```
godofredoc, line 58: rm?

codesign/bin/main.dart
```diff
@@ -0,0 +1,122 @@
...
+const String kCommit = 'commit';
+const String kProduction = 'production';
+const String kCodesignCertName = 'codesign-cert-name';
+const String kCodesignPrimaryBundleId = 'codesign-primary-bundle-id';
+const String kCodesignUserName = 'codesign-username';
+const String kAppSpecificPassword = 'app-specific-password';
+const String kCodesignAppStoreId = 'codesign-appstore-id';
+const String kCodesignTeamId = 'codesign-team-id';
+const String kCodesignFilepath = 'filepath';
+
```
itsjustkevin, line 23: Is there a specific reason we are prefacing constants with a k?  The dart style guide recommends against using prefixes.  [Effective Dart: Don't use prefix letters](https://dart.dev/guides/language/effective-dart/style#dont-use-prefix-letters).

CaseyHillers: Cocoon generally follows the Flutter repo style guide.

https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#begin-global-constant-names-with-prefix-k

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -0,0 +1,111 @@
...
+    ..writeAsStringSync(_entitlementsFileContents);
+
+  late final Directory remoteDownloadsDir = tempDir.childDirectory('downloads')..createSync();
+  late final Directory codesignedZipsDir = tempDir.childDirectory('codesigned_zips')..createSync();
+
+  int _remoteDownloadIndex = 0;
+  int get remoteDownloadIndex => _remoteDownloadIndex++;
+
+  static const String _entitlementsFileContents = '''
+<?xml version="1.0" encoding="UTF-8"?>
```
CaseyHillers, line 74: Using something like https://pub.dev/packages/mustache_template would be cleaner

codesign/lib/src/globals.dart
```diff
@@ -0,0 +1,52 @@
...
+}
+
+/// Translate CLI arg names to env variable names.
+///
+/// For example, 'state-file' -> 'STATE_FILE'.
+String fromArgToEnvName(String argName) {
+  return argName.toUpperCase().replaceAll(r'-', r'_');
```
CaseyHillers, line 22: Is the raw string necessary?

```suggestion
  return argName.toUpperCase().replaceAll('-', '_');
```

codesign/lib/src/globals.dart
```diff
@@ -0,0 +1,52 @@
...
+  if (argValue != null) {
+    return argValue;
+  }
+
+  if (allowNull) {
+    return null;
+  }
+  throw ConductorException('Expected either the CLI arg --$name or the environment variable $envName '
+      'to be provided!');
+}
```
CaseyHillers, line 52: Can we add tests for this funciton?

codesign/lib/src/stdio.dart
```diff
@@ -0,0 +1,105 @@
+// Copyright 2019 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:io' as io;
+
+import 'package:meta/meta.dart';
+
+abstract class Stdio {
```
CaseyHillers, line 9: Does the codesign tool need its own stdio? Can we use package:logging instead?

codesign/lib/src/zip_archive.dart
```diff
@@ -0,0 +1,66 @@
+// Copyright 2019 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:codesign/codesign.dart';
+import 'package:file/file.dart';
+
+/// A zip file that contains files that must be codesigned.
+abstract class ZipArchive extends ArchiveFile {
```
CaseyHillers, line 9: Is this logic needed yet, or should we move it to a future PR?

codesign/bin/main.dart
```diff
@@ -0,0 +1,122 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:io';
+
+import 'package:args/args.dart';
+import 'package:codesign/codesign.dart';
+import 'package:platform/platform.dart';
+
+const String filepathFlag = 'filepath';
```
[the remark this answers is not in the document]
XilaiZhang: yes the help texts of what each variable means, are encoded in `addOption` and `addFlag` calls a little bit below.

codesign/bin/main.dart
```diff
@@ -0,0 +1,122 @@
...
+    ..addFlag(
+      kProduction,
+      help: 'whether we are going to upload the artifacts back to GCS for production',
+    );
+
+  final ArgResults argResults = parser.parse(args);
+
+  final Platform platform = LocalPlatform();
+
+  final String commit = getValueFromEnvOrArgs(kCommit, argResults, platform.environment)!;
```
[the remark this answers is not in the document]
XilaiZhang: yep the conversion was handled by `getValueFromEnvOrArgs` function, which lives in lib/src/globals.dart

codesign/bin/main.dart
```diff
@@ -0,0 +1,122 @@
...
+///
+/// Usage:
+/// dart run bin/main.dart --commit=a5967ed309ef2beb9625f128571f7060597b5eda
+/// --filepath=darwin-x64/FlutterMacOS.framework.zip --filepath=ios/artifacts.zip
+/// --filepath=dart-sdk-darwin-arm64.zip
+/// ( add `--production` if this is intended for production)
+Future<void> main(List<String> args) async {
+  final ArgParser parser = ArgParser();
+  parser
+    ..addFlag(helpFlag, help: 'Prints usage info.', callback: (bool value) {
```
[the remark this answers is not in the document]
XilaiZhang: Yes this is part of the `getValueFromEnvOrArgs` function. It lives in globals.dart and will ask user for variable if not provided.

codesign/lib/src/codesign_context.dart
```diff
@@ -0,0 +1,77 @@
...
+
+  ProcessManager processManager = LocalProcessManager();
+  FileSystem fileSystem = LocalFileSystem();
+  Stdio stdio = VerboseStdio(
+    stdout: stdout,
+    stderr: stderr,
+    stdin: stdin,
+  );
+
+  void createTempDirectory() {
```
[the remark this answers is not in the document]
XilaiZhang: Sure, removed function.

codesign/lib/src/codesign_context.dart
```diff
@@ -0,0 +1,77 @@
...
+    tempDir ??= fileSystem.systemTempDirectory.createTempSync('conductor_codesign');
```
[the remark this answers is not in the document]
XilaiZhang: Yeah I override this variable later in test with `MemoryFileSystem.test().systemTempDirectory.createTempSync()`. I can remove the function and set the instance variable instead.

codesign/lib/src/codesign_context.dart
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  Future<void> run() async {
+    createTempDirectory();
+
+    codesignVisitor ??= FileCodesignVisitor(
```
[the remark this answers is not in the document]
XilaiZhang: Yeah in tests I created temp directory rand0 and passed it down to the fake subclasses. and temp directory would be created with a random number 0, etc.
If I let `create temp directory` run when code context runs, another temp directories will be created. And the names of multiple temp directories would not match.  So I set ??= to only allow one temp directory to be created.

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: Sure, I removed these functions and saved them for later PRs, based on Casey's comment.

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: Sure, removed.

codesign/bin/main.dart
```diff
@@ -0,0 +1,122 @@
+// Copyright 2019 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:io';
```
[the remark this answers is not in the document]
XilaiZhang: Sure, changed.

tests.yaml
```diff
@@ -20,6 +20,11 @@ tasks:
   - task: licenses_check
     script: test_utilities/bin/licenses.sh
 
+  #temporarily move to front in PRs, so that it can report errors earlier
```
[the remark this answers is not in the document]
XilaiZhang: Done.

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: umm this will be a constant string and unmodified. I might be wrong, but looking at mustache's template it is good for dynamic args?

codesign/lib/src/globals.dart
XilaiZhang: Done.

codesign/lib/src/globals.dart
XilaiZhang: umm this utility is called in bin/codesign.dart. Is there a way I can make file under directory bin visible, and importable by tests? Thank you!

codesign/lib/src/stdio.dart
XilaiZhang: umm Yeah currently we are overriding stdio, and we can use test Logger to override logger similar to https://github.com/flutter/cocoon/blob/dc6bde54efcb0310670823f3bd38006ad15b651b/device_doctor/test/src/utils.dart#L83-L120. shall we override stdio or logger?

codesign/lib/src/zip_archive.dart
XilaiZhang: Yeah they are for later PRs. I removed them and the corresponding function declarations, and replaced with a TODO.

codesign/bin/codesign.dart
```diff
@@ -2,8 +2,99 @@
...
+import 'dart:io';
 
-void main(List<String> arguments) {
-  print('Hello world: ${codesign.calculate()}!');
+import 'package:args/args.dart';
+import 'package:codesign/codesign.dart';
+import 'package:platform/platform.dart';
+
+/// Definitions of variables are included in help texts below.
+const String helpFlag = 'help';
```
CaseyHillers, line 12:
```suggestion
const String kHelpFlag = 'help';
```

codesign/bin/codesign.dart
```diff
@@ -2,8 +2,99 @@
...
-import 'package:codesign/codesign.dart' as codesign;
+import 'dart:io';
 
-void main(List<String> arguments) {
-  print('Hello world: ${codesign.calculate()}!');
+import 'package:args/args.dart';
+import 'package:codesign/codesign.dart';
+import 'package:platform/platform.dart';
+
+/// Definitions of variables are included in help texts below.
```
CaseyHillers, line 11: Are these all cli flags? If so, we should add `Flag` at the end of each variable's name to make it clearer

codesign/bin/codesign.dart
```diff
@@ -2,8 +2,99 @@
...
+/// back to google cloud storage.
+/// Otherwise, nothing will be uploaded back for production. default value is
+/// false.
+///
+/// For `--commit`, asks for the engine commit to be code signed.
+///
+/// For `--filepath`, provides the artifacts zip paths to be code signed.
+///
+/// Usage:
+/// dart run bin/main.dart --commit=a5967ed309ef2beb9625f128571f7060597b5eda
```
CaseyHillers, line 33: Avoid putting real data in here. It will make it a bit too easy to accidentally do the wrong thing.

```suggestion
/// ```
///   dart run bin/main.dart --commit=$commitSha \
///   --filepath=darwin-x64/FlutterMacOS.framework.zip \
///   --filepath=ios/artifacts.zip \
///   --filepath=dart-sdk-darwin-arm64.zip
/// ```
```

codesign/lib/src/codesign_context.dart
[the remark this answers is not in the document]
CaseyHillers: You can create `FileCodesignVisitor` in the constructor to gurantee it's set.

```dart
class CodesignContext {
  CodesignContext(this.x): codesignVisitor = FileCodesignVisitor(x);
    
  final int x;
  final FileCodesignVisitor codesignVisitor;
}

class FileCodesignVisitor {
  const FileCodesignVisitor(this.y);
  
  final int y;
}
````

File: codesign/lib/src/file_codesign_visitor.dart (hunk 3 above)
Comment (CaseyHillers) [reply codesign/lib/src/file_codesign_visitor.dart:11]: Revisiting this. What's the value having an abstract class is giving us? Do we have other versions planned?

File: codesign/test/codesign_test.dart [hunk 36]
```diff
@@ -2,11 +2,32 @@
...
+
+  FakeCodesignContext codesignContext = FakeCodesignContext(
+      codesignCertName: randomString,
+      codesignUserName: randomString,
+      appSpecificPassword: randomString,
+      codesignAppstoreId: randomString,
+      codesignTeamId: randomString,
+      commitHash: randomString);
+
+  test('nothing', () async {
```
CaseyHillers, line 30: Implementing a basic process execution on each codesign visitor would make this testable. This would let you ensure the correct nodes are processed.

PR #1917, opened 2022-06-16
Check target existence before scheduling

keyonghan:
Potential fix: https://github.com/flutter/flutter/issues/102998
Fixes: https://github.com/flutter/flutter/issues/100706

This PR
1) adds data model support for LUCI builders RPC
2) adds logic to check whether a target exists in LUCI before scheduling.

PR #1927, opened 2022-06-22
Fix urls in buildbucket RPCs

keyonghan:
Fixes: https://github.com/flutter/flutter/issues/106429

Context: https://github.com/flutter/cocoon/pull/1917 used incorrect `builder` url for `builds` related requests, and caused buildbucket RPC failures.


ricardoamador (approved):
LGTM

PR #1938, opened 2022-06-24
Add triaging logs for github secondary rate limit error

keyonghan:
Help triage: https://github.com/flutter/flutter/issues/102402


godofredoc (approved):

CaseyHillers (approved):
In the future, we may need to move the GitHub client to a singleton. The redis caching of the token is only a patch to the underlying issue of Cocoon always creating a new client.

PR #1953, opened 2022-06-30
Skip gold diff check for autoroller in Engine

keyonghan:
Fixes: https://github.com/flutter/flutter/issues/106673

### DIFF
--- a/app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
+++ b/app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
@@ -158,16 +158,10 @@ class PushGoldStatusToGithub extends ApiRequestHandler<Body> {
           assert(!pr.draft!);
           // Get Gold status.
           final String goldStatus = await _getGoldStatus(slug, pr);
-          statusRequest = _createStatus(
-              goldStatus,
-              goldStatus == GithubGoldStatusUpdate.statusRunning
-                  ? config.flutterGoldChanges
-                  : config.flutterGoldSuccess,
-              slug,
-              pr.number!);
+          final String description = getStatusDescription(goldStatus, pr);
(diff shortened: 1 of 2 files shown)


mdebbar (approved):

Piinks (commented):
When will new images be approved and checked in? I think @mdebbar implemented this similar to the framework. When the engine rolls into the framework, any new images are approved in presubmit. If the skia roller introduces new images, where in the workflow do they get approved? If left untriaged, they can cause test failures and tree breakages.

keyonghan:
@mdebbar Could you help take another look? I return `complete` earlier for auto roller to simplify the logic.

keyonghan:
> When will new images be approved and checked in? I think @mdebbar implemented this similar to the framework. When the engine rolls into the framework, any new images are approved in presubmit. If the skia roller introduces new images, where in the workflow do they get approved? If left untriaged, they can cause test failures and tree breakages.

Will this be handled by engine sheriff? @zanderso

Piinks:
> Will this be handled by engine sheriff? @zanderso

How will the engine sheriff be notified? There is no system to inform them currently AFAIK.

Also, if the workflow is going to change, it should definitely be documented. An announcement would not hurt either, especially if there is going to be different processes  between the framework and the engine.

https://github.com/flutter/flutter/wiki/Writing-a-golden-file-test-for-package:flutter

Piinks:
Full disclosure, I am being a bit critical here because any time golden failures start rolling in, or trees break etc, I am blown up on all channels. 😁

keyonghan:
Thank you @Piinks for the context.
Just F.Y.I. I will be ooo starting tomorrow. Please feel free to grab this PR to move forward if we have a clear answer of how.

keyonghan:
Closing based on https://github.com/flutter/flutter/issues/106673.

PR #1955, opened 2022-07-02
[cipd] yaml file with cipd build information

XilaiZhang:
context: https://flutter-review.googlesource.com/c/recipes/+/31240

This comes from the code review of https://flutter-review.googlesource.com/c/recipes/+/31240, that we should have a generic recipe to build and upload cipd packages. The recipe will read cipd package information from this file.


cipd.yaml
```diff
@@ -0,0 +1,28 @@
...
+#
+# This file defines the cipd packages that are built from cocoon
+# It uses the following schema:
+#
+# targets:
+#   - target: <platform name> <project name>
+#     script: <name of build script> 
+
+targets:
+  - target: Mac codesign
```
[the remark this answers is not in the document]
XilaiZhang: umm my understanding is that a CI.yaml target property will be parsed through the starlark file config/lib/ci_yaml/ci_yaml.star, and in starlark we might not have easy access to recipe functions such as cipd.build, cipd.register etc. 

Or maybe we are looking at parsing and building as part of cocoon/cocoon.py recipe, and make the cipd property part of tests.yaml. This way the logic is still part of recipe.

I might be wrong, but I think it would be easy to build through recipe than through starlark?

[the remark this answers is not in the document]
CaseyHillers: Cocoon is on the cocoon scheduler, so the builds properties (api.properties.get) will have the latest ci.yaml property values for the commit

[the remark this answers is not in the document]
XilaiZhang: Sounds good, I will add "script" properties for codesign and device_doctor targets, and retrieve them in cipd.py recipe through api.properties.get. Thanks for explaining!

.ci.yaml
```diff
@@ -31,6 +31,8 @@ targets:
     recipe: cocoon/device_doctor
     properties:
       add_recipes_cq: "true"
+      script: "build.sh"
```
CaseyHillers, line 34: nit: you don't need the quotation marks, yaml should convert this to a string

.ci.yaml
```diff
@@ -40,6 +42,8 @@ targets:
...
```
CaseyHillers, line 45: It would help to add the full path here (e.g. `device_doctor/tools/build.sh`). This will reduce the hard coding in the recipe.

Similar to below, you may want to add the full cipd name (like flutter/device_doctor/mac-amd64)

.ci.yaml
XilaiZhang: sure, updated

.ci.yaml
XilaiZhang: Sure, updated. 

Yeah in recipe I would need to extract information such as `device_doctor` from the full path, and concatenate it with `.zip` etc. I am planning to extract platform information from full path as well, so that I can compare with the three character platform name ('win', 'mac' and 'lin') from api.platform.name in recipe

CaseyHillers: Sounds good to me, feel free to remove the platform from the end of cipd_name (up to you, but I don't think it'll change often). Might be good for cases where the dimensions aren't correctly assigned, and a mac arm is assigned instead of a mac intel

CaseyHillers (approved):
flutter-lgtm

CaseyHillers:
FYI - Tests are failing as you need to regenerate the expected ci.yaml json for cocoon in the integration test

PR #1971, opened 2022-07-12
Return all tasks for a commit when querying recent tasks

keyonghan:
Fixes: https://github.com/flutter/flutter/issues/100793

### DIFF
--- a/app_dart/lib/src/service/datastore.dart
+++ b/app_dart/lib/src/service/datastore.dart
@@ -130,9 +130,6 @@ class DatastoreService {
   /// [commitLimit] argument specifies how many commits to consider when
   /// retrieving the list of recent tasks.
   ///
-  /// The [taskLimit] argument specifies how many tasks to retrieve for each
-  /// commit that is considered.
-  ///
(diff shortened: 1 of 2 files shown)


app_dart/lib/src/service/datastore.dart
```diff
@@ -142,14 +139,11 @@ class DatastoreService {
   Stream<FullTask> queryRecentTasks({
     String? taskName,
     int commitLimit = 20,
-    int taskLimit = 20,
     String? branch,
     required RepositorySlug slug,
   }) async* {
     await for (Commit commit in queryRecentCommits(limit: commitLimit, branch: branch, slug: slug)) {
-      final Query<Task> query = db.query<Task>(ancestorKey: commit.key)
-        ..limit(taskLimit)
```
CaseyHillers, line 151: TIL we've had this for 3 years, and I don't think anything used it

CaseyHillers (approved):
flutter-lgtm

app_dart/lib/src/service/datastore.dart
keyonghan: `refresh_chromebot_status` is using it as well.

PR #1984, opened 2022-07-14
Update task status as `inProgress` whenever build is scheduled

keyonghan:
Solves: https://github.com/flutter/flutter/issues/107599

### DIFF
--- a/app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
+++ b/app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
@@ -84,18 +84,6 @@ class BatchBackfiller extends RequestHandler {
     log.fine('Backfilling ${backfill.length} builds');
     log.fine(backfill.map<String>((Tuple<Target, Task, Commit> tuple) => tuple.first.value.name));
 
-    // Datastore must be written to generate task keys
-    try {
-      await datastore.withTransaction<void>((Transaction transaction) async {
-        transaction.queueMutations(inserts: backfillTasks);
-        await transaction.commit();
-        log.fine(
-            'Updated ${backfillTasks.length} tasks: ${backfillTasks.map((e) => e.name).toList()} when backfilling.');
-      });
-    } catch (error) {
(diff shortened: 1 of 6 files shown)


app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
```diff
@@ -97,6 +100,18 @@ class BatchBackfiller extends RequestHandler {
     }
     // Schedule all builds asynchronously
     await Future.wait<void>(futures);
+
+    // Update tasks status as in progress to avoid duplicate scheduling.
+    try {
```
CaseyHillers, line 105: Instead of adding a new field to track separately, you can use map to get it from the backfill tuple

```dart
final List<Task> backfillTasks = backfill.map((Tuple<Target, Task, Commit> tuple) => tuple.second).toList();
```

CaseyHillers (approved):

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -595,7 +595,8 @@ class LuciBuildService {
     await pubsub.publish('scheduler-requests', request);
 
     task.attempts = (task.attempts ?? 0) + 1;
-    task.status = Task.statusNew;
+    // Mark task as in progress to ensure it isn't scheduled over
+    task.status = Task.statusInProgress;
```
ricardoamador, line 599: Is this to prevent rescheduling? Or starting it over?

ricardoamador (approved):
LGTM - Just have a question for my own understanding.

app_dart/lib/src/service/luci_build_service.dart
keyonghan: Prevent duplicate re-scheduling from backfiller.

app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
keyonghan: Updated.

PR #1987, opened 2022-07-18
Add logs for autosubmit triage

keyonghan:
For triage: https://github.com/flutter/flutter/issues/107788

### DIFF
--- a/auto_submit/lib/requests/check_pull_request.dart
+++ b/auto_submit/lib/requests/check_pull_request.dart
@@ -45,6 +45,7 @@ class CheckPullRequest extends AuthenticatedRequestHandler {
       log.info('There are no requests in the queue');
       return Response.ok('No requests in the queue.');
     }
+    log.info('Processing ${receivedMessages.length} messages');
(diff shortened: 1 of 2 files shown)


ricardoamador:
Make sure that if any of the objects you are using in the logging are from the github package we import they support the toString method for interpolation. Otherwise looks good. I like that we are adding more logging.

ricardoamador (approved):
LGTM

keyonghan:
> Make sure that if any of the objects you are using in the logging are from the github package we import they support the toString method for interpolation. Otherwise looks good. I like that we are adding more logging.

Thanks. We used pullRequest here, which has the toString set up.

keyonghan:
> > Make sure that if any of the objects you are using in the logging are from the github package we import they support the toString method for interpolation. Otherwise looks good. I like that we are adding more logging.
> 
> Thanks. We used pullRequest here, which has the toString set up.

Oops, there is only a `toJson` function. I am updating the log here: https://github.com/flutter/cocoon/pull/1988. Sorry for that.

PR #1990, opened 2022-07-19
[test_utiltiies] Add analyze for trailing whitespace

CaseyHillers:
Add a static analysis for trailing whitespace and executable files.

Long term, any type of static analysis can be added to this project.


analyze/analyze.dart
```diff
@@ -0,0 +1,249 @@
...
+    if (isExecutable && !kExecutableAllowlist.contains(relativePath)) {
+      unexpectedExecutableCount += 1;
+      print('$relativePath is executable: ${(stat.mode & 0x1FF).toRadixString(2)}');
+    }
+  }
+  if (unexpectedExecutableCount > 0) {
+    throw Exception(
+      'found $unexpectedExecutableCount unexpected executable file'
+      '${unexpectedExecutableCount == 1 ? '' : 's'}! If this was intended, you '
+      'must add this file to kExecutableAllowlist in dev/bots/analyze.dart',
```
christopherfujino, line 238: nit: update path

christopherfujino (approved):
LGTM with a nit about the source path in an exception message

CaseyHillers:
I tried extending this to enforce spaces over tabs, but it broke a lot of the generated files. I'll skip this for now as it requires checking if there's any specific conversions we should use for the various file types.

PR #1998, opened 2022-07-21
Remove cronjob relying on `/api/refresh-chromebot-status`

keyonghan:
Now everything is on cocoon scheduler (https://github.com/flutter/cocoon/pull/1998), and API `/api/refresh-chromebot-status` is not used anymore.

This PR removes cronjobs replying on this API.

### DIFF
--- a/cron.yaml
+++ b/cron.yaml
@@ -3,26 +3,6 @@
 # To manually deploy this config, run:
 #   gcloud app deploy --project flutter-dashboard cron.yaml
 cron:
-- description: refresh chromebot build status (cocoon)
-  url: /api/refresh-chromebot-status?repo=cocoon
-  schedule: every 15 minutes
-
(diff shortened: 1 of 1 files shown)


yusuf-goog (approved):

cron.yaml
```diff
@@ -3,26 +3,6 @@
 # To manually deploy this config, run:
 #   gcloud app deploy --project flutter-dashboard cron.yaml
 cron:
-- description: refresh chromebot build status (cocoon)
-  url: /api/refresh-chromebot-status?repo=cocoon
```
yusuf-goog, line 7: Can we remove these handlers as well?

yusuf-goog (approved):

cron.yaml
keyonghan: Yeah, I will remove after couple of days validation.

PR #1999, opened 2022-07-22
Send batch requests when scheduling new commit builds

keyonghan:
With cocoon scheduling, a single schedule request of all builds will be created. This causes unpredicted builds schedule failure.

This PR use batch requests instead of a single request of ~300 builds: each batch is with default 5 builds.

Validation: a test version of cocoon wit this change has been running since yesterday.

Related: https://github.com/flutter/flutter/issues/103443, https://github.com/flutter/flutter/issues/92300

### DIFF
--- a/app_dart/lib/src/service/scheduler.dart
+++ b/app_dart/lib/src/service/scheduler.dart
@@ -161,7 +161,7 @@ class Scheduler {
 
   /// Schedule all builds in batch requests instead of a single request.
   ///
-  /// Each batch request contains `kBatchSize` builds to be scheduled.
+  /// Each batch request contains [Config.batchSize] builds to be scheduled.
   Future<void> _batchScheduleBuilds(Commit commit, List<Tuple<Target, Task, int>> toBeScheduled) async {
     final List<Future> futures = <Future>[];
     for (int i = 0; i < toBeScheduled.length; i += config.batchSize) {
--- a/app_dart/lib/src/service/scheduler.dart
+++ b/app_dart/lib/src/service/scheduler.dart
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'dart:math';
 import 'dart:typed_data';
 
 import 'package:cocoon_service/src/service/build_status_provider.dart';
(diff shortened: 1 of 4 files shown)


app_dart/test/src/datastore/fake_config.dart
```diff
@@ -132,6 +134,9 @@ class FakeConfig implements Config {
   @override
   int get schedulingShardSize => 5;
 
+  @override
+  int get batchSize => batchSizeValue ?? 5;
```
CaseyHillers, line 138: Making this const in config.dart would prevent need to add this stub in the test

app_dart/lib/src/service/scheduler.dart
```diff
@@ -154,11 +155,24 @@ class Scheduler {
...
 
-    await luciBuildService.schedulePostsubmitBuilds(commit: commit, toBeScheduled: toBeScheduled);
-
+    await _batchScheduleBuilds(commit, toBeScheduled);
     await _uploadToBigQuery(commit);
   }
 
+  /// Schedule all builds in batch requests instead of a single request.
+  ///
+  /// Each batch request contains `kBatchSize` builds to be scheduled.
```
CaseyHillers, line 164:
```suggestion
  /// Each batch request contains [Config.batchSize] builds to be scheduled.
```

CaseyHillers (approved):

app_dart/lib/src/service/scheduler.dart
keyonghan: Done

app_dart/test/src/datastore/fake_config.dart
keyonghan: I need to have customized `batchSize` values to validate from test.

app_dart/lib/src/service/scheduler.dart
```diff
@@ -154,11 +155,24 @@ class Scheduler {
...
 
+  /// Schedule all builds in batch requests instead of a single request.
+  ///
+  /// Each batch request contains [Config.batchSize] builds to be scheduled.
+  Future<void> _batchScheduleBuilds(Commit commit, List<Tuple<Target, Task, int>> toBeScheduled) async {
+    final List<Future> futures = <Future>[];
+    for (int i = 0; i < toBeScheduled.length; i += config.batchSize) {
+      futures.add(luciBuildService.schedulePostsubmitBuilds(
+        commit: commit,
+        toBeScheduled: toBeScheduled.sublist(i, min(i + config.batchSize, toBeScheduled.length)),
```
ricardoamador, line 170: is sublist upper bound index non-inclusive?

keyonghan: That's correct.

ricardoamador (approved):
LGTM

PR #2006, opened 2022-07-25
Remove refresh chromebot status API

keyonghan:
With cocoon scheduling, this API is no longer needed.

Part of https://github.com/flutter/flutter/issues/92300


ricardoamador (approved):
LGTM

PR #1891, opened 2022-06-02
Refactoring of auto-submit validations.

godofredoc:
This is to simplify the application architecture in a way that it is
easy to extend.

Bug: https://github.com/flutter/flutter/issues/99717
Bug: https://github.com/flutter/flutter/issues/96896

## Pre-launch Checklist

- [X] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [X] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [X] I read the [Flutter Style Guide] _recently_, and have followed its advice.
- [X] I signed the [CLA].
- [X] I listed at least one issue that this PR fixes in the description above.
- [X] I updated/added relevant documentation (doc comments with `///`).
- [X] I added new tests to check the change I am making, or this PR is [test-exempt].
- [X] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[test-exempt]: https://github.com/flutter/flutter/wiki/Tree-hygiene#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat

PR #2009, opened 2022-07-26
Land ToT revert immediately without waiting for validations

keyonghan:
This is a reland of an earlier logic about TOT revert:
Earlier logic in `waiting for tree to go green`: https://github.com/flutter/cocoon/pull/1498
Earlier logic in `autosubmit`: https://github.com/flutter/cocoon/pull/1653/files, which was removed when refactoring validations https://github.com/flutter/cocoon/pull/1891/files

Fixes: https://github.com/flutter/flutter/issues/108374

### DIFF
--- a/auto_submit/lib/service/validation_service.dart
+++ b/auto_submit/lib/service/validation_service.dart
@@ -79,10 +79,59 @@ class ValidationService {
     await processPullRequest(config, queryResult, messagePullRequest, ackId, pubsub);
   }
 
-  /// Processes a PullRequest running several validations to decide whether to
-  /// land the commit or remove the autosubmit label.
+  /// Processes a PullRequest.
+  ///
+  /// If it is a ToT revert, the PR will be landed without validations.
+  /// If it is not a ToT revert, the PR will be landed if all validations pass.
   Future<void> processPullRequest(
       Config config, QueryResult result, github.PullRequest messagePullRequest, String ackId, PubSub pubsub) async {
+    github.RepositorySlug slug = messagePullRequest.base!.repo!.slug();
+    final PullRequest pullRequest = result.repository!.pullRequest!;
+    Commit commit = pullRequest.commits!.nodes!.single.commit!;
+    final String? sha = commit.oid;
+    final GithubService gitHubService = await config.createGithubService(slug);
+    final bool isTotRevert = await checkIsTotRevert(sha!, slug, gitHubService);
+    bool shouldMergePR = false;
+    if (!isTotRevert) {
+      shouldMergePR = await checkShoudMergePR(config, result, messagePullRequest, ackId, pubsub, slug, gitHubService);
+      if (!shouldMergePR) {
+        return;
+      }
+    }
+    // If we got to this point it means we are ready to submit the PR.
+    bool processed = await processMerge(config, result, messagePullRequest, sha);
+    if (processed) {
+      await pubsub.acknowledge('auto-submit-queue-sub', ackId);
+      log.info('Acked the processed message : $ackId.');
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


auto_submit/lib/service/validation_service.dart
```diff
@@ -79,19 +79,66 @@ class ValidationService {
...
+  }
+
+  /// Check if the `commitSha` is a clean revert of TOT commit.
+  ///
+  /// A clean revert of TOT commit only reverts all changes made by TOT, thus should be
+  /// equivalent to the second TOT commit. When comparing the current commit with second
+  /// TOT commit, empty `files` in `GitHubComparison` validates a clean revert of TOT commit.
+  ///
+  /// Note: [compareCommits] expects base commit first, and then head commit.
+  Future<bool> checkIsTotRevert(String headSha, github.RepositorySlug slug, GithubService githubService) async {
```
godofredoc, line 116: Can we move this logic to its own validation rather than implementing it here?

auto_submit/lib/service/validation_service.dart
```diff
@@ -79,19 +79,66 @@ class ValidationService {
...
+    }
+    // If we got to this point it means we are ready to submit the PR.
+    bool processed = await processMerge(config, result, messagePullRequest, sha);
+    if (processed) {
+      await pubsub.acknowledge('auto-submit-queue-sub', ackId);
+      log.info('Acked the processed message : $ackId.');
+    }
+  }
+
+  /// Check if the `commitSha` is a clean revert of TOT commit.
```
zanderso, line 109: This should allow reverting anything within the past 10-15 commits or so since several commits could be landed on top of the bad commit that needs to be reverted.

godofredoc: That functionality will be implemented once we have a revert API, github does not provide a way to identify a PR is a revert.

zanderso: As written this PR is not useful. It is very rare that the commit that needs to be reverted is directly at the top of the tree. As it turns out, at this moment, the tree is closed on the second commit back, and the revert is being blocked waiting for review.

godofredoc: @ricardoamador @keyonghan in terms of effort how difficult would it be to auto-approve the PR if it was created by a member of `flutter-hackers` and it contains the `warning: land on red to fix tree breakage` label?

if that functionality is implemented would we need the revert API?

auto_submit/lib/service/validation_service.dart
ricardoamador: This is essentially the new revert feature is it not?

auto_submit/lib/service/validation_service.dart
ricardoamador: I would think it would be fairly simple to integrate this into the autosubmit bot by adding a user membership check. We already are looking for that label. I think this would be the functionality for the revert api. I would like @keyonghan to weight in to though.

keyonghan: Currently we use `warning: land on red to fix tree breakage` only to ignore tree status.
I guess the purpose of this label is different from what we are targeting here: for PRs that intend to unblock the tree may need pre-submit check validation, and then get landed by ignoring the tree status.
Here we may need a new label. With that, it will get approval and auto landed immediately without waiting for any check/status.
I believe the efforts should be straightforwards based on existing validation workflows: 
1) add new label check when auto approving
2) ignore statuses/checks in CI validation

One thing to note is the autosubmit app is listening to `autosubmit` label, there might be some race condition when adding the new label.

keyonghan:
Closing as we plan to support a more general revert.

PR #2010, opened 2022-07-27
[scheduler] Run guranteed policy on release branches and engine PRs

CaseyHillers:
Release branches are currently blocked as they will use batch scheduling for the engine on mac, which isn't expected.

I assume this is similar logic we also want for the engine repo.

### DIFF
--- a/app_dart/lib/src/service/scheduler.dart
+++ b/app_dart/lib/src/service/scheduler.dart
@@ -134,7 +134,11 @@ class Scheduler {
     final List<Tuple<Target, Task, int>> toBeScheduled = <Tuple<Target, Task, int>>[];
     for (Target target in initialTargets) {
       final Task task = tasks.singleWhere((Task task) => task.name == target.value.name);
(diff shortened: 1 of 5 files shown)


app_dart/lib/src/service/scheduler.dart
```diff
@@ -134,7 +134,11 @@ class Scheduler {
     final List<Tuple<Target, Task, int>> toBeScheduled = <Tuple<Target, Task, int>>[];
     for (Target target in initialTargets) {
       final Task task = tasks.singleWhere((Task task) => task.name == target.value.name);
-      final SchedulerPolicy policy = target.schedulerPolicy;
+      SchedulerPolicy policy = target.schedulerPolicy;
```
CaseyHillers, line 137: I would've preferred to add this into target, but we don't pass the branch down to the target

zanderso (commented):
nit: gurantee -> guarantee

I'm missing context for a proper review, but I can confirm that we should be building every engine commit.

keyonghan (approved):
LGTM

PR #2015, opened 2022-07-28
[auto_submit] Run flutter packages upgrade to fix analyzer issues

CaseyHillers:
Autosubmit is failing on tip of tree. This upgrades its analyzer dependencies (unpins a few unnecessary lint packages, and adds some that were transitively depended on that should be required)


ricardoamador:
Why the change from private to public on all these fields? Is this for testing?

Also you just need to run the formatter.

CaseyHillers:
> Why the change from private to public on all these fields? Is this for testing?

These are fields that should never have been private as they're variables only used within a method.

ricardoamador (approved):
LGTM

PR #2025, opened 2022-07-29
add fluttergithubbot to list of allow-listed autoroller accounts

christopherfujino:
Related to https://github.com/flutter/flutter/pull/108635
Part of https://github.com/flutter/flutter/issues/106371

@keyonghan is it ok to add this user to this list? I know this account is also used for the flaky bot PRs.

@CaseyHillers this seems like a silly test I've added, since I'm setting `config.rollerAccountsValue` in the test. Should I refactor this code so I can actually test the prod value?

### DIFF
--- a/app_dart/lib/src/service/config.dart
+++ b/app_dart/lib/src/service/config.dart
@@ -323,6 +323,8 @@ class Config {
         'skia-flutter-autoroll',
         'engine-flutter-autoroll',
         'dependabot',
+        // for pub package autorolls https://github.com/flutter/flutter/issues/106371
+        'fluttergithubbot',
       };
 
   Future<String> generateJsonWebToken() async {
--- a/app_dart/test/request_handlers/check_for_waiting_pull_requests_test.dart
+++ b/app_dart/test/request_handlers/check_for_waiting_pull_requests_test.dart
@@ -371,6 +371,35 @@ void main() {
       );
     });
 
+    test('Merges unapproved pub autoroll PR from fluttergithubbot', () async {
+      // Ensure there is at least one cirrus status.
+      statuses = <dynamic>[
+        <String, String>{'id': '3', 'status': 'COMPLETED', 'name': 'test2'}
+      ];
+      config.rollerAccountsValue = <String>{'fluttergithubbot'};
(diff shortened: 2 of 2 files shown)


keyonghan (changes requested):
> @keyonghan is it ok to add this user to this list? I know this account is also used for the flaky bot PRs.

I am afraid not. This will allow all flake/deflake PRs to auto land without test owners' approval. However, we do need test owners' approval for such PRs, as they may not be suitable for different reasons.

I would suggest use a different bot for this rolling purpose, same as other rolling bots.

christopherfujino:
> This will allow all flake/deflake PRs to auto land without test owners' approval. However, we do need test owners' approval for such PRs, as they may not be suitable for different reasons.

Technically it would require an org member adding the `autosubmit` label, and that org member would also have permissions to approve it, even if they're not the code owner.

keyonghan:
> > This will allow all flake/deflake PRs to auto land without test owners' approval. However, we do need test owners' approval for such PRs, as they may not be suitable for different reasons.
> 
> Technically it would require an org member adding the `autosubmit` label, and that org member would also have permissions to approve it, even if they're not the code owner.

Considering flake bot will not auto add `autosubmit` label, this will not cause flake/deflake PRs to auto land. Another question is: the auto submit bot is using `fluttergithubbot` to auto approve roller PRs, I am not sure if the bot could auto approve its own PRs.

christopherfujino:
> > > This will allow all flake/deflake PRs to auto land without test owners' approval. However, we do need test owners' approval for such PRs, as they may not be suitable for different reasons.
> > 
> > 
> > Technically it would require an org member adding the `autosubmit` label, and that org member would also have permissions to approve it, even if they're not the code owner.
> 
> Considering flake bot will not auto add `autosubmit` label, this will not cause flake/deflake PRs to auto land. Another question is: the auto submit bot is using `fluttergithubbot` to auto approve roller PRs, I am not sure if the bot could auto approve its own PRs.

oh, I hadn't considered that

christopherfujino:
Shucks, I'll have to figure out another way to do this then, thanks @keyonghan!

PR #2035, opened 2022-08-04
Call pub/sub pull API multiple times to return more messages

keyonghan:
Fixes: https://github.com/flutter/flutter/issues/108767

### DIFF
--- a/auto_submit/lib/requests/check_pull_request.dart
+++ b/auto_submit/lib/requests/check_pull_request.dart
@@ -40,16 +40,16 @@ class CheckPullRequest extends AuthenticatedRequestHandler {
   Future<Response> get() async {
     final Set<int> processingLog = <int>{};
     final ApproverService approver = approverProvider(config);
-    final List<pub.ReceivedMessage> messageSet = await pullMessages();
-    if (messageSet.isEmpty) {
+    final List<pub.ReceivedMessage> messageList = await pullMessages();
+    if (messageList.isEmpty) {
       log.info('There are no requests in the queue');
       return Response.ok('No requests in the queue.');
     }
-    log.info('Processing ${messageSet.length} messages');
+    log.info('Processing ${messageList.length} messages');
     ValidationService validationService = ValidationService(config);
     final List<Future<void>> futures = <Future<void>>[];
 
-    for (pub.ReceivedMessage message in messageSet) {
+    for (pub.ReceivedMessage message in messageList) {
       final String messageData = message.message!.data!;
       final rawBody = json.decode(String.fromCharCodes(base64.decode(messageData))) as Map<String, dynamic>;
       final PullRequest pullRequest = PullRequest.fromJson(rawBody);
--- a/auto_submit/lib/requests/check_pull_request.dart
+++ b/auto_submit/lib/requests/check_pull_request.dart
@@ -34,22 +34,32 @@ class CheckPullRequest extends AuthenticatedRequestHandler {
   final ApproverServiceProvider approverProvider;
 
   static const int kPullMesssageBatchSize = 100;
-  static const int kPubsubPullNumber = 5;
 
   @override
   Future<Response> get() async {
+    // Loops [config.kPubsubPullNumber] times to try pulling/processing as many
+    // messages as possible.
+    for (int i = 0; i < config.kPubsubPullNumber; i++) {
+      await checkPullRequest();
+    }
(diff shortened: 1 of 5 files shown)


ricardoamador (approved):
LGTM

auto_submit/lib/requests/check_pull_request.dart
```diff
@@ -34,25 +34,27 @@ class CheckPullRequest extends AuthenticatedRequestHandler {
...
 
   @override
   Future<Response> get() async {
     final Set<int> processingLog = <int>{};
-    final pub.PullResponse pullResponse = await pubsub.pull('auto-submit-queue-sub', kPullMesssageBatchSize);
     final ApproverService approver = approverProvider(config);
-    final List<pub.ReceivedMessage>? receivedMessages = pullResponse.receivedMessages;
-    if (receivedMessages == null) {
+    final List<pub.ReceivedMessage> messageList = await pullMessages();
+    if (messageList.isEmpty) {
```
godofredoc, line 44: I'd recommend removing this check. Sometimes even though there are messages in the queue a call to the api return empty results.

auto_submit/lib/requests/check_pull_request.dart
```diff
@@ -34,25 +34,27 @@ class CheckPullRequest extends AuthenticatedRequestHandler {
   final ApproverServiceProvider approverProvider;
 
   static const int kPullMesssageBatchSize = 100;
+  static const int kPubsubPullNumber = 5;
```
godofredoc, line 37: How many iterations can we run in 60s? I recommend making this # closer to a 1/2 of the max.

auto_submit/lib/requests/check_pull_request.dart
```diff
@@ -69,4 +71,24 @@ class CheckPullRequest extends AuthenticatedRequestHandler {
...
+      final List<pub.ReceivedMessage>? receivedMessages = pullResponse.receivedMessages;
+      if (receivedMessages == null) {
+        continue;
+      }
+      for (pub.ReceivedMessage message in receivedMessages) {
+        final String messageId = message.message!.messageId!;
+        messageMap[messageId] = message;
+      }
+    }
+    return messageMap.values.toList();
```
godofredoc, line 92: Not sure how long it takes for pub/sub api to return but if it takes something more that a couple hundred ms I'd recommend to process the messages as we get the rather than accumulating and processing at the end.

auto_submit/lib/requests/check_pull_request.dart
keyonghan: I need to monitor/adjust this a bit based on the response time. Here 5 is a relatively conservative number to begin with.

auto_submit/lib/requests/check_pull_request.dart
keyonghan: We want to exit early if no message is returned. Updated the message to `No messages are pulled.` to avoid confusion.

auto_submit/lib/requests/check_pull_request.dart
keyonghan: What would be the advantage by doing that?
With consolidating first, we just need to iterate all distinct messages once with same goal. But by processing messages following each call, we are repeatedly processing same message 5 times in each cronjob call.
We can keep close monitoring the execution time in the beginning.

auto_submit/lib/requests/check_pull_request.dart
godofredoc: Please add a TODO to not forget to update this number to an optimal value.

auto_submit/lib/requests/check_pull_request.dart
godofredoc: The main concern is applying synchronous processing logic to asynchronous technologies, ideally in an asynchronous world we get a bunch of messages and delegate the processing to something else without blocking in the processing. There are several subscription configurations to avoid the behavior you mentioned https://cloud.google.com/pubsub/docs/exactly-once-delivery#exactly-once_delivery_guarantees

keyonghan: Updated to loop through each` pull` call and process messages immediately in each loop.

auto_submit/lib/requests/check_pull_request.dart
keyonghan: Added

godofredoc (approved):

PR #2039, opened 2022-08-05
[app_dart] Fix Gerrit CreateBranch request body

CaseyHillers:
https://github.com/flutter/flutter/issues/102325

The Gerrit documentation was out of date, and this should only be the string of the sha.

Next issue is dealing with the 403's coming from Gerrit


keyonghan (approved):
LGTM

PR #1972, opened 2022-07-12
Xilaizhang refactor 1

XilaiZhang:
*Replace this paragraph with a description of what this PR is changing or adding, and why. Consider including before/after screenshots.*

*List which issues are fixed by this PR. You must list at least one issue.*

*If you had to change anything in the [flutter/tests] repo, include a link to the migration guide as per the [breaking change policy].*

## Pre-launch Checklist

- [ ] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [ ] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [ ] I read the [Flutter Style Guide] _recently_, and have followed its advice.
- [ ] I signed the [CLA].
- [ ] I listed at least one issue that this PR fixes in the description above.
- [ ] I updated/added relevant documentation (doc comments with `///`).
- [ ] I added new tests to check the change I am making, or this PR is [test-exempt].
- [ ] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[test-exempt]: https://github.com/flutter/flutter/wiki/Tree-hygiene#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat

PR #1899, opened 2022-06-10
[codesign] skeleton template from dart create

XilaiZhang:
context: this is the first split of https://github.com/flutter/cocoon/pull/1861.

This is an effort to split the original PR into smaller PRs for easier reviews. The current PR is the skeleton code obtained from running `dart create`.

PR #2044, opened 2022-08-10
[codesign] Logic to visit a binary file

XilaiZhang:
context: this is the fourth split (a smaller split) of https://github.com/flutter/cocoon/pull/1861.

prior splits (merged, excluding abandoned) : https://github.com/flutter/cocoon/pull/1972, https://github.com/flutter/cocoon/pull/1899, https://github.com/flutter/cocoon/pull/2003.

This split adds in logic to visit a binary file, and code sign it with or without entitlement, based on the configuration file.

-------
Thanks for reviewing! Incorporated feedbacks from Christopher

### DIFF
--- a/codesign/lib/src/file_codesign_visitor.dart
+++ b/codesign/lib/src/file_codesign_visitor.dart
@@ -90,11 +90,11 @@ class FileCodesignVisitor {
   static const String fixItInstructions = '''
 Codesign test failed.
 
-We compared binary files in engine artifacts with those listed in 
-entilement.txt and withoutEntitlements.txt, and the binary files do not match. 
-*entitlements.txt is the configuartion file encoded in engine artifact zip, 
+We compared binary files in engine artifacts with those listed in
+entilement.txt and withoutEntitlements.txt, and the binary files do not match.
+*entitlements.txt is the configuartion file encoded in engine artifact zip,
 built by BUID.gn and Ninja, to detail the list of entitlement files.
-Either an expected file was not found in *entitlements.txt, or an unexpected 
+Either an expected file was not found in *entitlements.txt, or an unexpected
 file was found in entitlements.txt.
 
 This usually happens either during an engine roll.
@@ -149,8 +149,7 @@ entitlements applied during codesigning.
           zipEntity: entity,
           entitlementParentPath: entitlementParentPath,
         );
-      }
-      else if (childType == FileType.binary){
+      } else if (childType == FileType.binary) {
         await visitBinaryFile(binaryEntity: entity, entitlementParentPath: entitlementParentPath);
       }
       log.info('Child file of direcotry ${directory.basename} is ${entity.basename}\n');
@@ -184,7 +183,6 @@ entitlements applied during codesigning.
       processManager: processManager,
     );
   }
-
 
   /// Visit and codesign a binary with / without entilement.
   ///
@@ -232,5 +230,4 @@ entitlements applied during codesigning.
     }
     fileConsumed.add(entitlementCurrentPath);
   }
-
 }
--- a/codesign/lib/src/file_codesign_visitor.dart
+++ b/codesign/lib/src/file_codesign_visitor.dart
@@ -87,7 +87,24 @@ class FileCodesignVisitor {
 </plist>
 ''';
 
-  void _initialize() {
+  static const String fixItInstructions = '''
+Codesign test failed.
+
+We compared binary files in engine artifacts with those listed in 
+entilement.txt and withoutEntitlements.txt, and the binary files do not match. 
+*entitlements.txt is the configuartion file encoded in engine artifact zip, 
+built by BUID.gn and Ninja, to detail the list of entitlement files.
+Either an expected file was not found in *entitlements.txt, or an unexpected 
+file was found in entitlements.txt.
+
+This usually happens either during an engine roll.
+If this is a valid change, then BUILD.gn needs to be changed.
+Specifically, update [entitlements.txt] or
+[withoutEntitlements.txt] lists, depending on if the file should have macOS
+entitlements applied during codesigning.
+''';
+
+  void initialize() {
     entitlementsFile = tempDir.childFile('Entitlements.plist')..writeAsStringSync(_entitlementsFileContents);
     remoteDownloadsDir = tempDir.childDirectory('downloads')..createSync();
     codesignedZipsDir = tempDir.childDirectory('codesigned_zips')..createSync();
@@ -95,7 +112,7 @@ class FileCodesignVisitor {
 
   /// The entrance point of examining and code signing an engine artifact.
   Future<void> validateAll() async {
-    _initialize();
+    initialize();
 
     await Future<void>.value(null);
     log.info('Codesigned all binaries in ${tempDir.path}');
@@ -133,6 +150,9 @@ class FileCodesignVisitor {
           entitlementParentPath: entitlementParentPath,
         );
       }
+      else if (childType == FileType.binary){
+        await visitBinaryFile(binaryEntity: entity, entitlementParentPath: entitlementParentPath);
+      }
       log.info('Child file of direcotry ${directory.basename} is ${entity.basename}\n');
     }
   }
@@ -164,4 +184,53 @@ class FileCodesignVisitor {
       processManager: processManager,
     );
   }
+
+
+  /// Visit and codesign a binary with / without entilement.
+  ///
+  /// At this stage, the virtual [entitlementCurrentPath] accumulated through the recursive visit, is compared
+  /// against the paths extracted from [fileWithEntitlements], to help determine if this file should be signed
+  /// with entitlements.
+  Future<void> visitBinaryFile({required FileSystemEntity binaryEntity, required String entitlementParentPath}) async {
+    if (await isSymlink(binaryEntity.absolute.path, processManager)) {
+      return;
+    }
+
+    final String currentFileName = binaryEntity.absolute.path.split('/').last;
+    final String entitlementCurrentPath = '$entitlementParentPath/$currentFileName';
+
+    if (!fileWithEntitlements.contains(entitlementCurrentPath) &&
+        !fileWithoutEntitlements.contains(entitlementCurrentPath)) {
+      log.severe('The system has detected a binary file at $entitlementCurrentPath\n'
+          'but it is not in the entitlements configuartion files your provided \n'
+          'if this is a new engine artifact, please add it to one of the entilements.txt files\n');
+      throw CodesignException(fixItInstructions);
+    }
+    log.info('\n signing file at path ${binaryEntity.absolute.path}');
+    log.info('\n the virtual entitlement path associated with file is $entitlementCurrentPath');
+    log.info('\n the decision to sign with entitlement is ${fileWithEntitlements.contains(entitlementCurrentPath)}');
+    final List<String> args = <String>[
+      'codesign',
+      '-f', // force
+      '-s', // use the cert provided by next argument
+      codesignCertName,
+      binaryEntity.absolute.path,
+      '--timestamp', // add a secure timestamp
+      '--options=runtime', // hardened runtime
+      if (fileWithEntitlements.contains(entitlementCurrentPath)) ...<String>[
+        '--entitlements',
+        entitlementsFile.absolute.path
+      ],
+    ];
+    final io.ProcessResult result = await processManager.run(args);
+    if (result.exitCode != 0) {
+      throw CodesignException(
+        'Failed to codesign ${binaryEntity.absolute.path} with args: ${args.join(' ')}\n'
+        'stdout:\n${(result.stdout as String).trim()}\n'
+        'stderr:\n${(result.stderr as String).trim()}',
+      );
+    }
+    fileConsumed.add(entitlementCurrentPath);
+  }
+
 }
--- a/codesign/lib/src/file_codesign_visitor.dart
+++ b/codesign/lib/src/file_codesign_visitor.dart
@@ -103,9 +103,11 @@ file was found in entitlements.txt.
 
 This usually happens during an engine roll.
 If this is a valid change, then BUILD.gn needs to be changed.
-For example, If this is a new binary that need to be code signed with entitlements, add it
+Binaries that will run on a macOS host require entitlements, and
+binaries that run on an iOS device must NOT have entitlements.
+For example, If this is a new binarythat runs on macOS host, add it
 to [entitlements.txt] file inside the zip artifact produced by BUILD.gn.
-If this is a new binary that need to be code signed without entitlements, add it
+If this is a new binary that need to be run on iOS device, add it
 to [withoutEntitlements.txt].
 If there are obsolete binaries in entitlements configuration files, please delete or
 update these file paths accordingly.
--- a/codesign/lib/src/file_codesign_visitor.dart
+++ b/codesign/lib/src/file_codesign_visitor.dart
@@ -135,9 +135,6 @@ update these file paths accordingly.
     final List<FileSystemEntity> entities = await directory.list().toList();
     for (FileSystemEntity entity in entities) {
       if (entity is io.Directory) {
-        if (io.FileSystemEntity.isLinkSync(entity.absolute.path)) {
-          return;
-        }
         await visitDirectory(
           directory: directory.childDirectory(entity.basename),
           entitlementParentPath: entitlementParentPath,
--- a/codesign/lib/src/file_codesign_visitor.dart
+++ b/codesign/lib/src/file_codesign_visitor.dart
@@ -95,7 +95,7 @@ class FileCodesignVisitor {
 Codesign test failed.
 
 We compared binary files in engine artifacts with those listed in
-entilement.txt and withoutEntitlements.txt, and the binary files do not match.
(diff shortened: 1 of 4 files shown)


codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -87,15 +87,32 @@ class FileCodesignVisitor {
...
 ''';
 
-  void _initialize() {
+  static const String fixItInstructions = '''
+Codesign test failed.
+
+We compared binary files in engine artifacts with those listed in
+entilement.txt and withoutEntitlements.txt, and the binary files do not match.
+*entitlements.txt is the configuartion file encoded in engine artifact zip,
+built by BUID.gn and Ninja, to detail the list of entitlement files.
```
christopherfujino, line 96:
```suggestion
built by BUILD.gn and Ninja, to detail the list of entitlement files.
```

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -87,15 +87,32 @@ class FileCodesignVisitor {
...
+Either an expected file was not found in *entitlements.txt, or an unexpected
+file was found in entitlements.txt.
+
+This usually happens either during an engine roll.
+If this is a valid change, then BUILD.gn needs to be changed.
+Specifically, update [entitlements.txt] or
+[withoutEntitlements.txt] lists, depending on if the file should have macOS
```
christopherfujino, line 103: how does one determine the difference?

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -164,4 +183,51 @@ class FileCodesignVisitor {
       processManager: processManager,
     );
   }
+
+  /// Visit and codesign a binary with / without entilement.
```
christopherfujino, line 187:
```suggestion
  /// Visit and codesign a binary with / without entitlement.
```

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -164,4 +183,51 @@ class FileCodesignVisitor {
...
+  ///
+  /// At this stage, the virtual [entitlementCurrentPath] accumulated through the recursive visit, is compared
+  /// against the paths extracted from [fileWithEntitlements], to help determine if this file should be signed
+  /// with entitlements.
+  Future<void> visitBinaryFile({required FileSystemEntity binaryEntity, required String entitlementParentPath}) async {
```
christopherfujino, line 192: I'm thinking `binaryEntity` can be a `File` rather than `FileSystemEntity`, right? That is to say, we should never be passing a directory to this method.

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -164,4 +183,51 @@ class FileCodesignVisitor {
...
+    if (await isSymlink(binaryEntity.absolute.path, processManager)) {
+      return;
+    }
+
+    final String currentFileName = binaryEntity.absolute.path.split('/').last;
```
christopherfujino, line 197: use `.basename` https://pub.dev/documentation/file/latest/file/FileSystemEntity/basename.html

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -164,4 +183,51 @@ class FileCodesignVisitor {
...
+    final String entitlementCurrentPath = '$entitlementParentPath/$currentFileName';
+
+    if (!fileWithEntitlements.contains(entitlementCurrentPath) &&
+        !fileWithoutEntitlements.contains(entitlementCurrentPath)) {
+      log.severe('The system has detected a binary file at $entitlementCurrentPath\n'
+          'but it is not in the entitlements configuartion files your provided \n'
```
christopherfujino, line 203:
```suggestion
          'but it is not in the entitlements configuration files you provided \n'
```

christopherfujino: Actually, I suspect logging just the path is enough, the rest of this message seems redundant with `fixItInstructions`.

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -164,4 +183,51 @@ class FileCodesignVisitor {
...
+          'if this is a new engine artifact, please add it to one of the entilements.txt files\n');
+      throw CodesignException(fixItInstructions);
+    }
+    log.info('\n signing file at path ${binaryEntity.absolute.path}');
```
christopherfujino, line 207: why the leading `'\n '` in all these info logs?

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -164,4 +183,51 @@ class FileCodesignVisitor {
...
+    log.info('\n the virtual entitlement path associated with file is $entitlementCurrentPath');
+    log.info('\n the decision to sign with entitlement is ${fileWithEntitlements.contains(entitlementCurrentPath)}');
```
christopherfujino, line 209: is this meant to be in production?

christopherfujino: nvm, you're probably right, it's good to have this in the logs.

christopherfujino: actually, even better, why don't you move this log until after the next line, and log the value of `args`.

codesign/lib/src/utils.dart
```diff
@@ -29,6 +29,21 @@ enum FileType {
   }
 }
 
+Future<bool> isSymlink(String fileOrFolderPath, ProcessManager processManager) async {
```
christopherfujino, line 32: You should avoid calling out to sub-processes and parsing their output when possible. You can check if it's a link from a package:file entity by checking if it's a `Link` https://pub.dev/documentation/file/latest/file/Link-class.html

christopherfujino: specifically this method: https://api.dart.dev/stable/2.17.5/dart-io/FileSystemEntity/isLinkSync.html

codesign/test/file_codesing_visitor_test.dart
```diff
@@ -37,6 +37,7 @@ void main() {
         tempDir: tempDir,
```
christopherfujino, line 37: please rename this file `file_codesign_visitor_test.dart`

codesign/test/file_codesing_visitor_test.dart
```diff
@@ -37,6 +37,7 @@ void main() {
...
       );
       codesignVisitor.directoriesVisited.clear();
+      codesignVisitor.initialize();
```
christopherfujino, line 40: Is there any reason why the initialization code isn't run in the visitor's constructor? This seems like an easy candidate for a future bug, where someone forgets to call initialize before using the visitor, or it gets called twice.

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: Very true, thanks for correcting!

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: Sure, changed description.

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: thanks for correcting!

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: Yeah `visitBinary` is reached if the file type is `application/x-mach-binary`, but for sure I have changed it to be a `File` type.

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: Good point, thanks for pointing out!

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: Thanks for correcting!

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: True, log already has formatting. Removed all log related \n.

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: umm I thought this would be helpful in tests to check whether we code signed each binary with entitlements or not. Should we find another way to obtain this information in tests?

codesign/lib/src/utils.dart
XilaiZhang: Yeah I tried to use Link previously, but the fileSystemEntity does not return true in the check of `entity is io.Link` or `entity is Link`, when the FileSystemEntity is a link. e.g. listing a folder A which contains child file symlink B, and B would fail the check of `B is Link`. umm maybe there is a better way to check if an entity resolves to Link rather than `is`?

codesign/test/file_codesing_visitor_test.dart
XilaiZhang: This is embarrassing. Very true, renamed.

codesign/test/file_codesing_visitor_test.dart
XilaiZhang: True, I forgot why i had it separately initialized. Moved.

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: Oh sorry i didn't refresh the page, there are more comments from Christopher. Sure I kept the logs and moved the \n out of the way.

codesign/lib/src/utils.dart
christopherfujino: check my previous comment

XilaiZhang: Ahhh there is a new comment every time I refresh. Thanks a lot Christopher!

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -93,27 +97,22 @@ Codesign test failed.
...
 Either an expected file was not found in *entitlements.txt, or an unexpected
 file was found in entitlements.txt.
 
-This usually happens either during an engine roll.
+This usually happens during an engine roll.
 If this is a valid change, then BUILD.gn needs to be changed.
-Specifically, update [entitlements.txt] or
-[withoutEntitlements.txt] lists, depending on if the file should have macOS
-entitlements applied during codesigning.
+For example, If this is a new binary that need to be code signed with entitlements, add it
```
christopherfujino, line 106: Most people won't know which binaries need entitlements, so I think you should clarify that binaries that will run on a macOS host require entitlements--binaries that run on an iOS device must NOT have entitlements.

codesign/lib/src/utils.dart
XilaiZhang: Done.

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: Yes sir, done.

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -87,16 +91,30 @@ class FileCodesignVisitor {
...
+*entitlements.txt is the configuartion file encoded in engine artifact zip,
+built by BUILD.gn and Ninja, to detail the list of entitlement files.
+Either an expected file was not found in *entitlements.txt, or an unexpected
+file was found in entitlements.txt.
+
+This usually happens during an engine roll.
+If this is a valid change, then BUILD.gn needs to be changed.
+Binaries that will run on a macOS host require entitlements, and
+binaries that run on an iOS device must NOT have entitlements.
+For example, If this is a new binarythat runs on macOS host, add it
```
christopherfujino, line 108:
```suggestion
For example, if this is a new binary that runs on a macOS host, add it
```

XilaiZhang: Very true, I really need to improve on writing. Thanks for correcting them Christopher!

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -132,8 +150,10 @@ class FileCodesignVisitor {
           zipEntity: entity,
           entitlementParentPath: entitlementParentPath,
         );
+      } else if (childType == FileType.binary) {
+        await visitBinaryFile(binaryFile: entity as File, entitlementParentPath: entitlementParentPath);
       }
-      log.info('Child file of direcotry ${directory.basename} is ${entity.basename}\n');
+      log.info('Child file of direcotry ${directory.basename} is ${entity.basename}');
```
christopherfujino, line 156:
```suggestion
      log.info('Child file of directory ${directory.basename} is ${entity.basename}');
```

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -142,8 +162,8 @@ class FileCodesignVisitor {
     required FileSystemEntity zipEntity,
     required String entitlementParentPath,
   }) async {
-    log.info('This embedded file is ${zipEntity.path} and entilementParentPath is $entitlementParentPath\n');
-    final String currentFileName = zipEntity.path.split('/').last;
+    log.info('This embedded file is ${zipEntity.path} and entilementParentPath is $entitlementParentPath');
```
christopherfujino, line 165:
```suggestion
    log.info('This embedded file is ${zipEntity.path} and entitlementParentPath is $entitlementParentPath');
```

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -164,4 +184,51 @@ class FileCodesignVisitor {
...
     );
   }
+
+  /// Visit and codesign a binary with / without entitlement.
+  ///
+  /// At this stage, the virtual [entitlementCurrentPath] accumulated through the recursive visit, is compared
+  /// against the paths extracted from [fileWithEntitlements], to help determine if this file should be signed
+  /// with entitlements.
+  Future<void> visitBinaryFile({required File binaryFile, required String entitlementParentPath}) async {
+    if (io.FileSystemEntity.isLinkSync(binaryFile.absolute.path)) {
```
christopherfujino, line 194: I actually don't think this is possible, since we should only get here if `file --mime-type /path/to/entity` returns `application/x-mach-binary`, while a link would return `inode/symlink`. I think you can omit this, or if you want to be safe, instead throw a `StateError('Expected a binary but got a link at ${binaryFile.path}');`

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -164,4 +184,51 @@ class FileCodesignVisitor {
...
+      return;
+    }
+
+    final String currentFileName = binaryFile.absolute.path.split('/').last;
```
christopherfujino, line 198:
```suggestion
    final String currentFileName = binaryFile.basename;
```

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: Yeah i was confused by this previously as well. For some reasons `mime type` check would not pick up the symlink. it returns a directory type for FlutterMacOS.framework.zip/Versions/Current , which should be
 a link. But for some reasons the symlink check would pick it up and output the correct information.
Screen Shot 2022-08-10 at 4 19 49 PM

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: My bad, I did a replace all on entilement (a messed up word). Thanks for catching it!

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: Very true, I should have caught them all.

codesign/lib/src/file_codesign_visitor.dart
christopherfujino: Right, but `FlutterMacOS.framework.zip/Versions/Current` will never his this method, right?

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: my man I really spelled a lot of things incorrectly. Thanks for catching them, it's impressive Christopher noticed all these details!

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: True, I will move this check to the directory level

christopherfujino: Sounds good.

The point is that rather than quietly handling this state, I suspect if we ever hit it there was a bug elsewhere, and it's better to crash than to quietly do the wrong thing.

XilaiZhang: umm previously we implemented a crash on loop (when symlink directs back to parent folder etc.), I thought that covered the crash by symlinks. Umm maybe I am wrong and I am thinking, is there another bug that could be caused by symlink

XilaiZhang: Actually I rethought about it, since symlinks are quite often in engine artifacts, I think Christopher is right, it's a good idea to follow the symlinks, and if they form a cycle, let them crash. early termination on symlink will bypass the loop check which isn't ideal. I have removed this early termination check on symlinks.

christopherfujino (approved):
LGTM

PR #2045, opened 2022-08-10
Check autosubmit label existence before landing a PR

keyonghan:
Fixes: https://github.com/flutter/flutter/issues/109004


auto_submit/lib/validations/autosubmit_label_existence.dart
```diff
@@ -0,0 +1,28 @@
...
+  AutosubmitLabelExistence({
+    required Config config,
+  }) : super(config: config);
+
+  @override
+
+  /// The `autosubmit` label must exist when landing PRs.
+  Future<ValidationResult> validate(QueryResult result, github.PullRequest messagePullRequest) async {
+    // List of labels associated with the pull request.
+    final List<String> labelNames = (messagePullRequest.labels as List<github.IssueLabel>)
```
godofredoc, line 22: Do we need to get the labels directly using the github client? as it can still land the PR if the message was queued before the label was removed.

auto_submit/lib/validations/autosubmit_label_existence.dart
```diff
@@ -0,0 +1,28 @@
...
+        .map<String>((github.IssueLabel labelMap) => labelMap.name)
+        .toList();
+    String message = '- The autosubmit label has been removed. Please add it back when ready.';
```
godofredoc, line 25: Should we just log server side and do not add the comment to the PR? The message may be misleading specially because the label was removed by the author manually.

auto_submit/lib/validations/autosubmit_label_existence.dart
```diff
@@ -0,0 +1,28 @@
...
+/// Validates that the list of checks for the PR is not empty.
+class AutosubmitLabelExistence extends Validation {
+  AutosubmitLabelExistence({
+    required Config config,
+  }) : super(config: config);
+
+  @override
+
+  /// The `autosubmit` label must exist when landing PRs.
+  Future<ValidationResult> validate(QueryResult result, github.PullRequest messagePullRequest) async {
```
ricardoamador, line 20: I'm not sure I understand. If we are validating that the label exists why do we remove the label. shouldn't this just be a simple existence check?

auto_submit/lib/validations/autosubmit_label_existence.dart
keyonghan: Switched the logic to `shouldProcess`, which handles this naturally.

auto_submit/lib/validations/autosubmit_label_existence.dart
keyonghan: Rethinking this, I thing `shouldProcess` is a better place for this label check. Updated.

auto_submit/lib/validations/autosubmit_label_existence.dart
keyonghan: I guess the confusing part is the no-op action.
Now we switched the logic to `shouldProcess`, this should be clear.

auto_submit/lib/service/validation_service.dart
```diff
@@ -49,13 +49,16 @@ class ValidationService {
   final Config config;
   final Set<Validation> validations = <Validation>{};
 
-  /// Checks if a pullRequest is still open before trying to process it.
+  /// Checks if a pullRequest is still open and with autosubmit label before trying to process it.
   Future<bool> shouldProcess(github.PullRequest pullRequest) async {
-    github.RepositorySlug slug = pullRequest.base!.repo!.slug();
-    github.GitHub gitHub = await config.createGithubClient(pullRequest.base!.repo!.slug());
-    github.PullRequest currentPullRequest = await gitHub.pullRequests.get(slug, pullRequest.number!);
+    final github.RepositorySlug slug = pullRequest.base!.repo!.slug();
```
ricardoamador, line 54: Thanks this is much more readable.

ricardoamador (approved):
LGTM

auto_submit/test/requests/github_webhook_test_data.dart
```diff
@@ -71,7 +71,7 @@ String generateWebhookEvent(
 
 PullRequest generatePullRequest(
     {String? labelName,
-    String? autosubmitLabel,
+    String? autosubmitLabel = 'autosubmit',
```
godofredoc, line 74: Should we use the config constant here?

godofredoc (approved):

auto_submit/test/requests/github_webhook_test_data.dart
keyonghan: Updated.

PR #2056, opened 2022-08-16
Update CI_YAML docs with platform_properties

CaseyHillers:


CI_YAML.md
```diff
@@ -22,10 +22,19 @@ enabled_branches:
...
+    properties:
+      # os will be inherited by all Linux targets, but it can be overrided at the target level
+      os: Linux
+
 targets:
 # A Target is an individual unit of work that is scheduled by Flutter infra
 # Target's are composed of the following properties:
 # name: A human readable string to uniquely identify this target.
+#       The first word indicates the platform this test will be run on. This should match
+#       to an existing platform under platform_properties.
```
yaakovschectman, line 37: Would it be helpful to also specify here how the target's first word can indicate both a building machine and a target machine (e.g. `Linux_android`)?

yaakovschectman (approved):
LGTM

CI_YAML.md
CaseyHillers: I'm only aware of this being an edge case for the engine's build rules, which is separate from the infra configs

keyonghan (approved):
LGTM

PR #2068, opened 2022-08-22
[app_dart] Push GitHub webhooks to PubSub

CaseyHillers:
## Overview

* Create /api/github/webhook-subscription
* Migrate most GitHub logic to the subscription
* Create GithubWebhookMessage for passing data to the subscription
* Cleanup tests
  * Use defined RepositorySlug from Config
  * Don't pass default values

## Why

Generating the recipe branches is failing as GitHub is taking too long to replicate 75% of the time. This allows us to exponentially retry webhook logic, and is something that would be valuable for all of Cocoon to have.

## Issues

Fixes https://github.com/flutter/flutter/issues/81579

## Future work

* Split github webhook into smaller subscriptions to mitigate breakages
* Generate pubsubs via code instead of the GCP UI


app_dart/lib/src/request_handlers/github/webhook_subscription.dart
```diff
@@ -0,0 +1,659 @@
...
+  'flutter/packages',
+  'flutter/plugins',
+};
+
+final RegExp kEngineTestRegExp = RegExp(r'(tests?|benchmarks?)\.(dart|java|mm|m|cc)$');
+final List<String> kNeedsTestsLabels = <String>['needs tests'];
+
+/// Subscription for processing GitHub webhooks.
+///
+// TODO(chillers): Change this to an actual subscription.
```
ricardoamador, line 43: Do you need to do these?

app_dart/lib/src/request_handlers/github/webhook_subscription.dart
```diff
@@ -0,0 +1,659 @@
...
+          topicName: 'github-webhooks',
+        );
+
+  /// Cocoon scheduler to trigger tasks against changes from GitHub.
+  final Scheduler scheduler;
+
+  /// To provide build status updates to GitHub pull requests.
+  final GithubChecksService? githubChecksService;
+
+  ///
```
ricardoamador, line 77: Remove or add comment.

app_dart/lib/src/request_handlers/github/webhook_subscription.dart
```diff
@@ -0,0 +1,659 @@
...
+  @override
+  Future<Body> post() async {
+    if (message.data == null || message.data!.isEmpty) {
+      log.warning('GitHub webhook message was empty. No-oping');
+      return Body.empty;
+    }
+
+    final pb.GithubWebhookMessage webhook = pb.GithubWebhookMessage.fromJson(message.data!);
+
+    log.fine('Processing ${webhook.event}');
```
ricardoamador, line 90: What is this log level? I am not familiar with it and when do we want to use it?

app_dart/lib/src/request_handlers/github/webhook_subscription.dart
```diff
@@ -0,0 +1,659 @@
...
+    await gitHubClient.pullRequests.createReview(slug, review);
+  }
+
+  Future<void> _checkForLabelsAndTests(PullRequestEvent pullRequestEvent) async {
+    final PullRequest pr = pullRequestEvent.pullRequest!;
+    final String? eventAction = pullRequestEvent.action;
+    final String repo = pr.base!.repo!.fullName.toLowerCase();
+    if (kNeedsCheckLabelsAndTests.contains(repo)) {
+      final GitHub gitHubClient = await config.createGitHubClient(pullRequest: pr);
+      try {
```
ricardoamador, line 231: I am assuming that we do not catch the exception because it trickles back to the call of the REST API. Is that correct?

app_dart/lib/src/service/config.dart
```diff
@@ -29,9 +29,11 @@ import 'logging.dart';
 const String kDefaultBranchName = 'master';
 
 /// Name of an example release base branch name.
+// TODO(chillers): Delete this as it's only used for tests.
```
ricardoamador, line 32: Delete?

app_dart/test/request_handlers/github/webhook_subscription_test.dart
```diff
@@ -0,0 +1,2382 @@
...
+import '../../src/request_handling/fake_http.dart';
+import '../../src/request_handling/subscription_tester.dart';
+import '../../src/service/fake_buildbucket.dart';
+import '../../src/service/fake_github_service.dart';
+import '../../src/service/fake_scheduler.dart';
+import '../../src/utilities/entity_generators.dart';
+import '../../src/utilities/mocks.dart';
+import '../../src/utilities/webhook_generators.dart';
+
+void main() {
```
ricardoamador, line 28: Nice thorough testing!

ricardoamador (approved):
LGTM, I just have some questions that don't require code changes.

app_dart/lib/src/request_handlers/github/webhook_subscription.dart
CaseyHillers: See https://pub.dev/packages/logging for the full details, but tl;dr is it's similar to what other loggers would have as debug. For our uses, we log every level so it doesn't matter beyond filtering purposes.

app_dart/lib/src/service/config.dart
CaseyHillers: I'd prefer to leave it in. Some of these TODOs I didn't want to distract as it's already a large diff. I can follow up in a separate PR as this should be moved directly into the tests.

keyonghan:
> Generating the recipe branches is failing as GitHub is taking too long to replicate 75% of the time. This allows us to exponentially retry webhook logic

If webhook API fails due to timing out, what does the pub/sub bring to us considering the timing point of view? If it fails in original API, I would expect it fails in the new pub/sub processing API as well, as they take the same time. Instead, shall we separate the logic of creating recipes from the webhook logic? Say just pub/sub the recipes branch logic.

app_dart/lib/src/request_handlers/github/webhook_subscription.dart
CaseyHillers: Correct!

CaseyHillers:
> If webhook API fails due to timing out, what does the pub/sub bring to us considering the timing point of view? If it fails in original API, I would expect it fails in the new pub/sub processing API as well, as they take the same time. Instead, shall we separate the logic of creating recipes from the webhook logic? Say just pub/sub the recipes branch logic.

I agree, this is a bit of a yak shave to the original issue.

As you pointed out, I think we want to separate the branching flutter recipes logic from the GitHub webhook. I think making this into a pubsub that it can subscribe to, and pull every 3 minutes would work well.

app_dart/lib/src/request_handlers/github/webhook_subscription.dart
CaseyHillers: This is done.

keyonghan:
> As you pointed out, I think we want to separate the branching flutter recipes logic from the GitHub webhook. I think making this into a pubsub that it can subscribe to, and pull every 3 minutes would work well.

Instead of a pull logic, a push sounds a better fit. From the webhook API, we just need to save message to the pub/sub and leave the processing async on the new API. This way we almost have no effect of existing workflow but a new API handling the recipes branching.

CaseyHillers:
> Instead of a pull logic, a push sounds a better fit. From the webhook API, we just need to save message to the pub/sub and leave the processing async on the new API. This way we almost have no effect of existing workflow but a new API handling the recipes branching.

I think we're in alignment. This PR enables GitHub webhooks to be processed via pubsub so we can create separate APIs for the various logic in Cocoon.

In future PRs, we can fork out the logic in github/webhook_subscription.dart to be various feature oriented subscriptions.

keyonghan (approved):
Let's monitor the tree a bit when it lands. Hopefully the swap is smooth.

CaseyHillers:
I'll land this tomorrow, and monitor the Monday rollout. I already have a few changes in the pipeline that will need to be monitored tomorrow.

PR #2075, opened 2022-08-24
[skia + cocoon] deduplicate branch name alias

XilaiZhang:
fixes https://github.com/flutter/flutter/issues/109255. The root cause was that listing branch would give several aliases for the same branch. For example, "beta", "flutter-1.0-candidate-1" could be two branch instances (as returned by the list branch api), but they point to the same branch. This commit deduplicates the branch aliases.


app_dart/lib/src/service/branch_service.dart
```diff
@@ -125,8 +125,10 @@ class BranchService {
 
     List<gh.Branch> branches = await github.repositories.listBranches(slug).toList();
     final String devName = await _getDevBranch(github: github, slug: slug, branches: branches);
-    final String betaName = branches.where(((gh.Branch b) => b.commit!.sha == betaSha)).single.name!;
-    final String stableName = branches.where(((gh.Branch b) => b.commit!.sha == stableSha)).single.name!;
+    final String betaName =
+        branches.where(((gh.Branch b) => b.commit!.sha == betaSha && b.name != 'beta' && b.name != 'dev')).single.name!;
```
godofredoc, line 129: Would it be possible to get two release candidate branches with the same commit as the beta branch?

godofredoc: Also if this logic is handling beta should it include the dev check?

XilaiZhang: Right, yes so if two beta branches have the same commit they will both pass the `b.commit!.sha == betaSha` check
And the alias with the name `beta` will be filtered out by the check `b.name != 'beta'`, and the dev branch will be filtered out by the check `b.name != 'dev'`. So I put down `b.commit!.sha == betaSha && b.name != 'beta' && b.name != 'dev'`

Using the test as an example, the following three branch instances
```
      gh.Branch betaBranch = generateBranch(3, name: 'flutter-3.4-candidate.5', sha: '789beta');
      gh.Branch devBranch = generateBranch(4, name: 'dev', sha: '789beta');
      gh.Branch betaAliasBranch = generateBranch(5, name: 'beta', sha: '789beta');
```
will be deduplicated to flutter-3.4-candidate-5

CaseyHillers: I think this is a good example of why we create [release-candidate-branch.version](https://cs.opensource.google/flutter/flutter/+/beta:bin/internal/release-candidate-branch.version).

Can we refactor the logic to use the GitHub files API for finding the beta and stable branches using this version file?

app_dart/lib/src/service/branch_service.dart
```diff
@@ -119,15 +119,15 @@ class BranchService {
...
-    final String devName = await _getDevBranch(github: github, slug: slug, branches: branches);
-    final String betaName = branches.where(((gh.Branch b) => b.commit!.sha == betaSha)).single.name!;
-    final String stableName = branches.where(((gh.Branch b) => b.commit!.sha == stableSha)).single.name!;
-
+      {required GithubService githubService, required gh.RepositorySlug slug}) async {
+    List<gh.Branch> branches = await githubService.github.repositories.listBranches(slug).toList();
+    final String devName = await _getDevBranch(github: githubService.github, slug: slug, branches: branches);
+
+    final String betaName =
+        (await githubService.getFileContent(slug, 'bin/internal/release-candidate-branch.version', ref: 'beta')).trim();
```
XilaiZhang, line 127: I added a trim() after the `replaceAll` and `decode` operations. Based on experiments with fetch file on [stable through github api](https://api.github.com/repos/flutter/flutter/contents/bin/internal/release-candidate-branch.version?ref=stable) and [beta through github api](https://api.github.com/repos/flutter/flutter/contents/bin/internal/release-candidate-branch.version?ref=beta), the decoded release branch version files would for some reasons have trailing '\n' attached to it, and the trailing '\n' could not be removed by `replaceAll`.  I tested this with my personal repo as well, here is the [api response for my personal repo](https://api.github.com/repos/XilaiZhang/miscellaneous-side-project/contents/push_box/README.md). 

So I ended up adding a trim() at the end.

app_dart/lib/src/service/branch_service.dart
XilaiZhang: Sure, the new commit is now based on file content.

app_dart/lib/src/service/branch_service.dart
CaseyHillers: All version files (and part of the Flutter style guide, and generally a widely accepted practice) is to have a trailing newline before EOF

app_dart/lib/src/service/branch_service.dart
```diff
@@ -119,15 +119,15 @@ class BranchService {
...
-      {required gh.GitHub github, required gh.RepositorySlug slug}) async {
-    final String betaSha = (await github.repositories.getBranch(slug, 'beta')).commit!.sha!;
-    final String stableSha = (await github.repositories.getBranch(slug, 'stable')).commit!.sha!;
-
-    List<gh.Branch> branches = await github.repositories.listBranches(slug).toList();
-    final String devName = await _getDevBranch(github: github, slug: slug, branches: branches);
-    final String betaName = branches.where(((gh.Branch b) => b.commit!.sha == betaSha)).single.name!;
-    final String stableName = branches.where(((gh.Branch b) => b.commit!.sha == stableSha)).single.name!;
-
+      {required GithubService githubService, required gh.RepositorySlug slug}) async {
```
CaseyHillers, line 122: nit: trailing comma

app_dart/lib/src/service/branch_service.dart
```diff
@@ -119,15 +119,15 @@ class BranchService {
...
+    List<gh.Branch> branches = await githubService.github.repositories.listBranches(slug).toList();
```
CaseyHillers, line 123: Does this need to be a List instead of Iterable? If not, consider simplifying to an Iterable

app_dart/lib/src/service/branch_service.dart
```diff
@@ -119,15 +119,15 @@ class BranchService {
...
+    final String devName = await _getDevBranch(github: githubService.github, slug: slug, branches: branches);
```
CaseyHillers, line 124: Instead of dev, we should call this the latestCandidateBranch (to avoid confusion with the deprecated dev branch)

app_dart/lib/src/service/branch_service.dart
CaseyHillers, line 127: Since beta and stable both perform this logic, we should move it into a helper function

app_dart/lib/src/service/github_service.dart
```diff
@@ -238,10 +238,12 @@ class GithubService {
 
   /// Gets the file content as UTF8 string of the file specified by the `path`
   /// in the repository.
-  Future<String> getFileContent(RepositorySlug slug, String path) async {
+  Future<String> getFileContent(RepositorySlug slug, String path, {String? ref}) async {
     ArgumentError.checkNotNull(slug);
     ArgumentError.checkNotNull(path);
-    final RepositoryContents contents = await github.repositories.getContents(slug, path);
+    final RepositoryContents contents = ref == null
```
CaseyHillers, line 244: Why does this need a ternary? Ideally, this would look like

```dart
final RepositoryContents contents = await github.repositories.getContents(slug, path, ref: ref);
```

app_dart/test/service/branch_service_test.dart
```diff
@@ -123,12 +121,6 @@ void main() {
       gh.Branch candidateBranchTwo = generateBranch(5, name: 'flutter-2.15-candidate.99', sha: 'superLargeYZvalue');
       gh.Branch candidateBranchThree = generateBranch(6, name: 'flutter-0.5-candidate.0', sha: 'someZeroValues');
 
-      when(mockRepositoriesService.getBranch(any, 'stable')).thenAnswer((Invocation invocation) {
```
CaseyHillers, line 126: Thank you for the cleanup!

app_dart/test/src/service/fake_github_service.dart
```diff
@@ -62,8 +64,19 @@ class FakeGithubService implements GithubService {
   }
 
   @override
-  Future<String> getFileContent(RepositorySlug slug, String path) async {
-    return '';
+  Future<String> getFileContent(RepositorySlug slug, String path, {String? ref}) async {
+    if (path == 'bin/internal/release-candidate-branch.version') {
+      String encodedResponse = '';
+      if (ref == 'beta') {
+        encodedResponse = 'Zmx1dHRlci0zLjItY2FuZGlkYXRlLjUK\n';
```
CaseyHillers, line 71: 1. Avoid making test values unreadable to humans.

2. Do we need to handle encoding in the fake? It seems like its encoded then decoded

```dart
final String encodedResponse = base64.encode(utf8.encode('flutter-3.2-candidate.5\n'));
```

Or simpler:

```dart
// Adding async to the method doesn't inherently makes all returned values a Future
if (ref == 'beta') {
  return 'flutter-3.2-candidate.5\n';
}
```

app_dart/lib/src/service/branch_service.dart
XilaiZhang: Yeah the reason i added trim() following a replaceAll('\n') is related to decoding. I explained in more details in https://github.com/flutter/cocoon/pull/2075#discussion_r960976435.

app_dart/lib/src/service/branch_service.dart
XilaiZhang: True, done

app_dart/lib/src/service/branch_service.dart
XilaiZhang: umm i read [the pros and cons of iterable and list](https://stackoverflow.com/questions/52345694/dart-iterablee-vs-liste-always-use-iterable) and it looks like in our case we would want a list so the sort operation would be efficient.

app_dart/lib/src/service/branch_service.dart
XilaiZhang: Sure, renamed the functions and tests.

app_dart/lib/src/service/branch_service.dart
XilaiZhang: Sure, refactored. Yeah previously i didn't create a helper function cause I thought they are one liners.

app_dart/lib/src/service/github_service.dart
XilaiZhang: Yeah this would involve many changes on flaky builders/issues handlers and their tests. I have updated those files and their tests to use the default branch as ref. Yeah previously i thought I could avoid these changes by supplying a tenary.

CaseyHillers (approved):
LGTM pending the nullable ref comment

flutter-lgtm

app_dart/lib/src/request_handlers/check_flaky_builders.dart
```diff
@@ -55,7 +55,11 @@ class CheckFlakyBuilders extends ApiRequestHandler<Body> {
     final RepositorySlug slug = Config.flutterSlug;
     final GithubService gitHub = config.createGithubServiceWithToken(await config.githubOAuthToken);
     final BigqueryService bigquery = await config.createBigQueryService();
-    final String ciContent = await gitHub.getFileContent(slug, kCiYamlPath);
+    final String ciContent = await gitHub.getFileContent(
+      slug,
+      kCiYamlPath,
+      ref: Config.defaultBranch(slug),
```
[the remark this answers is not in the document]
XilaiZhang: Very True, switched to nullable. Thanks for the advice!

PR #2079, opened 2022-08-24
Autosubmit bot does not remove label on failure label.

ricardoamador:
## Description

If merging has failed the autosubmit cannot remove the label in an exceptional case. 
See https://github.com/flutter/flutter/issues/109754.

When an exception occurs we return from the call to the API caller with an error code and are unable to remove the label allowing this failure loop to continue.

## Related issues

fixes https://github.com/flutter/flutter/issues/109754

## Pre-launch Checklist

- [x] I signed the [CLA].


auto_submit/lib/service/validation_service.dart
```diff
@@ -146,9 +146,26 @@ class ValidationService {
       }
     }
     // If we got to this point it means we are ready to submit the PR.
-    bool processed = await processMerge(config, result, messagePullRequest);
-    if (processed) await pubsub.acknowledge('auto-submit-queue-sub', ackId);
-    log.info('Ack the processed message : $ackId.');
+    try {
+      bool processed = await processMerge(config, result, messagePullRequest);
+      if (processed) {
+        await pubsub.acknowledge('auto-submit-queue-sub', ackId);
```
keyonghan, line 152: We should ack the message for any case till this point.

auto_submit/lib/service/validation_service.dart
```diff
@@ -146,9 +146,26 @@ class ValidationService {
       }
     }
     // If we got to this point it means we are ready to submit the PR.
-    bool processed = await processMerge(config, result, messagePullRequest);
-    if (processed) await pubsub.acknowledge('auto-submit-queue-sub', ackId);
-    log.info('Ack the processed message : $ackId.');
+    try {
```
keyonghan, line 149: Nit: move this block logic to a separate sub function for better readibility.

auto_submit/lib/service/validation_service.dart
```diff
@@ -146,9 +146,26 @@ class ValidationService {
...
+      bool processed = await processMerge(config, result, messagePullRequest);
+      if (processed) {
+        await pubsub.acknowledge('auto-submit-queue-sub', ackId);
+        log.info('Ack the processed message : $ackId.');
+      } else {
+        String message = 'Unable to merge pull request $prNumber.';
+        log.warning(message);
+        await gitHubService.removeLabel(slug, prNumber, Config.kAutosubmitLabel);
+        await gitHubService.createComment(slug, prNumber, message);
```
keyonghan, line 158: Nit: encapsulate these repeated logics in a sub-function.

auto_submit/lib/service/validation_service.dart
keyonghan, line 149: Could we add unit tests to cover?

ricardoamador: Yes I can add those.

ricardoamador: Done.

auto_submit/lib/service/validation_service.dart
ricardoamador: Done.

auto_submit/lib/service/validation_service.dart
ricardoamador: So even when we temporarily ignore in the result action we need to ack the message? Because that is the only case we are in this path when processing the pull request.

keyonghan: For the case of `ignore`, we should keep.
For for any place where we remove the label, we should ack.

ricardoamador: Okay. I added the acknowledge everywhere we remove the label. It seems that in any case but ignore we then remove the label.

auto_submit/lib/service/validation_service.dart
```diff
@@ -146,20 +145,112 @@ class ValidationService {
...
-    bool processed = await processMerge(config, result, messagePullRequest);
-    if (processed) await pubsub.acknowledge('auto-submit-queue-sub', ackId);
-    log.info('Ack the processed message : $ackId.');
+    await processMergeSafely(
+        config, gitHubService, messagePullRequest, result, pubsub, ackId, slug, prNumber, Config.kAutosubmitLabel);
+  }
+
+  /// The logic for processing a revert request and opening the follow up
+  /// review issue in github.
+  Future<void> processRevertRequest(
```
CaseyHillers, line 154: For large methods, the Dart way is to make everything a named parameter. Generally, after 2 parameters is considered large.

auto_submit/lib/service/validation_service.dart
```diff
@@ -146,20 +145,112 @@ class ValidationService {
...
+    gh.RepositorySlug slug = messagePullRequest.base!.repo!.slug();
+    final int prNumber = messagePullRequest.number!;
+    final GithubService githubService = await config.createGithubService(slug);
+
+    if (revertValidationResult.result) {
+      try {
+        bool processed = await processMerge(config, result, messagePullRequest);
+        if (processed) {
+          gh.Issue issue = await githubService.createIssue(
+            slug,
```
CaseyHillers, line 172: Flutter only creates issues on flutter/flutter. This will need to be hard coded instead of dynamic to the repo-under-test

auto_submit/lib/service/validation_service.dart
```diff
@@ -146,20 +145,112 @@ class ValidationService {
...
+            'Follow up review for revert pull request $prNumber',
+            'Revert request by author ${result.repository!.pullRequest!.author}',
```
CaseyHillers, line 174: In it's current state, there is little info. Could we file a P1 to ensure it is triaged, and action steps (such as having someone other than the author close the issue to grant LGTM)

auto_submit/lib/service/validation_service.dart
```diff
@@ -146,20 +145,112 @@ class ValidationService {
...
+          );
+          log.info('Issue #${issue.id} was created to track the review for $prNumber in ${slug.fullName}');
+        } else {
+          String message = 'Unable to merge pull request $prNumber.';
+          log.warning(message);
+          await removeLabelAndComment(githubService, slug, prNumber, Config.kRevertLabel, message);
+        }
+      } catch (exception) {
```
CaseyHillers, line 182: Prefer specifying the type to catch, instead of a generic catch all.

Would this be a GitHubError exception?

ricardoamador: Yeah that is one of the cons of dart, the fact that you need to guess or go dig through code when a throws declaration would could be immensely helpful. /done ranting. I will look for the explicit exception.

auto_submit/lib/service/validation_service.dart
ricardoamador: Odd why the minute the parameters become named they need to be explicitly defined as nullable...

auto_submit/lib/service/validation_service.dart
ricardoamador: OOOh good catch.

auto_submit/lib/service/validation_service.dart
ricardoamador: Good point. Let me look into this.

ricardoamador: Okay so there isn't too much you can add to the issue at creation time.
But I did add the following:

required RepositorySlug repositorySlug,
    required String title,
    required String body,
    List<String>? labels,
    String? assignee,
    List<String>? assignees,
    String? state,

I assume the p1 is just a labelbut not sure we can add an action step. It doesn't look like we can add that.

auto_submit/test/service/validation_service_test.dart
```diff
@@ -99,14 +105,103 @@ void main() {
...
+  bool threwException = false;
+
+  @override
+  Future<bool> processMerge(Config config, qr.QueryResult queryResult, gh.PullRequest messagePullRequest) async {
+    if (returnValue) {
+      return processMergeReturn;
+    }
+    threwException = true;
+    throw Exception();
+  }
```
keyonghan, line 206: Instead of faking this service, we should validate the logics. Other tests in this file may give you some hint.

auto_submit/lib/service/validation_service.dart
CaseyHillers: You can do `required String myParam`

auto_submit/lib/service/validation_service.dart
ricardoamador: Added catch for explicit GithubError and Exception as they are both thrown.

CaseyHillers: Everything thrown in Dart should be a subtype of Exception. Is there a specific exception you're trying to catch?

ricardoamador: Yes but Exception is thrown when we are calling createGitHubGraphQLClient(...) and if we do not have permissions our method in config.dart for _generateGithubToken will throw an exception.
 https://github.com/flutter/cocoon/blob/main/auto_submit/lib/service/config.dart#L128
But I see a way to rework this now from this conversation.

ricardoamador:
I have made a mess of this. There is a much simpler fix to be made here. Closing.

PR #2082, opened 2022-08-25
[dashboard] Remove unused url variable

CaseyHillers:


dashboard/lib/build_dashboard_page.dart
```diff
@@ -126,7 +126,7 @@ class BuildDashboardPageState extends State<BuildDashboardPage> {
               height: 360,
               child: ListView(
                 children: <Widget>[
-                  if (_smallScreen) ..._slugSelection(context, buildState),
+                  if (_smallScreen) ..._repositorySelection(context, buildState),
```
CaseyHillers, line 129: I also renamed this function nearby as it was incorrect (it does slug and branch selection)

yusuf-goog: Can we rename to more descriptive name

yusuf-goog (approved):
LGTM with one comment.

PR #2086, opened 2022-08-25
Create cocoon analysis_options.yaml

CaseyHillers:
After migrating app_dart to use super params, I realize we should have a global analysis options.

Currently, Cocoon's subprojects copy their own analysis which creates microcosms within the repo.

This starts creating some shared options, but leaves a few overrides for some of the projects (like app_dart needing prints)


christopherfujino (approved):
LGTM

keyonghan (approved):
LGTM. Thanks for the cleanup!

PR #2090, opened 2022-08-26
[dashboard] Create fake for FlutterAppIcon in tests

CaseyHillers:
* Uprev flutter_app_icons to 0.0.8
* Fix tests failing as there was no fake plugin created for it as it was originally a no-op

### DIFF
--- a/dashboard/test/state/build_test.dart
+++ b/dashboard/test/state/build_test.dart
@@ -4,6 +4,7 @@
 
 import 'dart:async';
 
+import 'package:flutter_app_icons/flutter_app_icons_platform_interface.dart';
 import 'package:flutter_dashboard/model/branch.pb.dart';
 import 'package:flutter_dashboard/model/build_status_response.pb.dart';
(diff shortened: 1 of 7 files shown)


ricardoamador (approved):
LGTM

dashboard/test/state/build_test.dart
```diff
@@ -45,6 +47,8 @@ void main() {
             ..repository = 'flutter'
         ]),
       );
+
+      FlutterAppIconsPlatform.instance = FakeFlutterAppIcons();
```
stuartmorgan, line 51: You could just do this at the start of `main` instead of `setUp`, since it doesn't need to run per test, but it's also harmless to recreate the fake.

stuartmorgan (approved):
LGTM

dashboard/test/state/build_test.dart
CaseyHillers: Thanks, but I'll keep it in setUp. In the off chance this stops being hermetic, it won't start to cause flakes

PR #2099, opened 2022-08-30
[app_dart] Fix encoding GithubWebhookMessage for pubsub

CaseyHillers:
Fixes https://github.com/flutter/flutter/issues/110487

This is somewhat difficult to test as the issue in the differences between the prod pubsub and the fake pubsub.

### DIFF
--- a/app_dart/lib/src/request_handlers/github_webhook.dart
+++ b/app_dart/lib/src/request_handlers/github_webhook.dart
@@ -37,10 +37,10 @@ class GithubWebhook extends RequestHandler<Body> {
 
     final String requestString = utf8.decode(requestBytes);
 
-    // TODO(chillers): This won't work until using proto.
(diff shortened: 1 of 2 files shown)


ricardoamador (approved):

app_dart/lib/src/request_handlers/github_webhook.dart
```diff
@@ -41,7 +41,7 @@ class GithubWebhook extends RequestHandler<Body> {
     final pb.GithubWebhookMessage message = pb.GithubWebhookMessage.create()
       ..event = event
       ..payload = requestString;
-    await pubsub.publish('github-wehbooks', message);
+    await pubsub.publish('github-wehbooks', message.writeToJsonMap());
```
keyonghan, line 44: Does a log info help here?

keyonghan (approved):
LGTM

PR #2107, opened 2022-09-01
[codesign] add function to check notarization status

XilaiZhang:
Context: Boost codesign priority and add more commits before the Christopher meeting next Tuesday.

This commit adds the function and tests to check notarization status of an uploaded zip artifact.

prior prs: https://github.com/flutter/cocoon/pull/1972, https://github.com/flutter/cocoon/pull/1899, https://github.com/flutter/cocoon/pull/2003, https://github.com/flutter/cocoon/pull/2044, https://github.com/flutter/cocoon/pull/1861, https://github.com/flutter/cocoon/pull/2059

PR #2059, opened 2022-08-17
[codesign] add function to parse entitlement configs

XilaiZhang:
This the fifth split (and a shorter split) of https://github.com/flutter/cocoon/pull/1861.

This PR adds function to parse and store entitlement files that were generated through GN.

prior splits: https://github.com/flutter/cocoon/pull/1972, https://github.com/flutter/cocoon/pull/1899, https://github.com/flutter/cocoon/pull/2003, https://github.com/flutter/cocoon/pull/2044.

PR #2111, opened 2022-09-02
[codesign] function to Upload to notary server

XilaiZhang:
Context: Add more commits before the Christopher meeting next Tuesday.

This commit adds the function and tests to upload artifact to notary server.

prior prs: https://github.com/flutter/cocoon/pull/1972, https://github.com/flutter/cocoon/pull/1899, https://github.com/flutter/cocoon/pull/2003, https://github.com/flutter/cocoon/pull/2044, https://github.com/flutter/cocoon/pull/1861, https://github.com/flutter/cocoon/pull/2059, https://github.com/flutter/cocoon/pull/2107

### DIFF
--- a/codesign/lib/src/file_codesign_visitor.dart
+++ b/codesign/lib/src/file_codesign_visitor.dart
@@ -92,6 +92,7 @@ class FileCodesignVisitor {
 ''';
   static const Duration _notarizationTimerDuration = Duration(seconds: 45);
   static final RegExp _notarytoolStatusCheckPattern = RegExp(r'[ ]*status: ([a-zA-z ]+)');
+  static final RegExp _notarytoolRequestPattern = RegExp(r'id: ([a-z0-9-]+)');
 
   static const String fixItInstructions = '''
 Codesign test failed.
@@ -273,10 +274,6 @@ update these file paths accordingly.
     await completer.future;
   }
 
-  String uploadZipToNotary(File localFile, [int retryCount = 3]) {
-    throw UnimplementedError('will implement later');
-  }
-
   /// Make a request to the notary service to see if the notary job is finished.
   ///
   /// A return value of true means that notarization finished successfully,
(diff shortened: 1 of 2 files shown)


codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -319,4 +316,44 @@ update these file paths accordingly.
...
+
+      log.info('uploading ${args.join(' ')}');
+      final io.ProcessResult result = processManager.runSync(args);
+
+      final String combinedOutput = (result.stdout as String) + (result.stderr as String);
+      final RegExpMatch? match;
+      match = _notarytoolRequestPattern.firstMatch(combinedOutput);
+
+      if (match == null) {
+        log.warning('Failed to upload to the notary service with args: ${args.join(' ')}');
```
christopherfujino, line 344: Technically this means that we don't understand the output, right? In other words, it could have succeeded but maybe the text of the output changed and now doesn't match the regex.

I wonder, before checking the output, should we first check the exit code of the ProcessResult?

XilaiZhang: Yep good point, added a block and test to check for exit code

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -341,6 +341,12 @@ update these file paths accordingly.
       match = _notarytoolRequestPattern.firstMatch(combinedOutput);
 
       if (match == null) {
+        if (result.exitCode >= 0 && result.exitCode <= 255) {
```
christopherfujino, line 344: I think actually before you even use the regex, after line 337, you could do:

```dart
if (result.exitCode != 0) {
  throw CodesignException('Command "${args.join(' ')}" failed with exit code ${result.exitCode}\nStdout: ${result.stdout}\nStderr: ${result.stderr}');
}
```

There's no point in continuing to run if you have a non-zero exit code, and this requires manual intervention.

XilaiZhang: Sure, added a check for bad exit code <0 or exit code >255 on Mac.

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -335,18 +335,20 @@ update these file paths accordingly.
 
       log.info('uploading ${args.join(' ')}');
       final io.ProcessResult result = processManager.runSync(args);
+      if (result.exitCode < 0 || result.exitCode > 255) {
```
christopherfujino, line 338: Don't we want `if (result.exitCode != 0) {`? For example, exit code of 1 would be not good, right?

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -335,18 +335,20 @@ update these file paths accordingly.
...
       match = _notarytoolRequestPattern.firstMatch(combinedOutput);
 
       if (match == null) {
-        if (result.exitCode >= 0 && result.exitCode <= 255) {
-          log.warning('The upload to notary service completed normally, but'
-              ' the output format does not match the current notary tool version.'
-              ' If after inspecting the output below, you believe the process finished '
-              'successfully but was not detected, please contact fujino@');
-        }
+        log.warning('The upload to notary service completed normally, but'
```
christopherfujino, line 348: I think we want to fail the build here. I've learned from experience, if the build succeeds, nobody reads the logs.

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: umm according to process.dart a value in the range of 1 and 255 would be good. An excerpt from dart looks like
```
  /// On Linux and OS X a normal exit code will be a positive value in
  /// the range `[0..255]`. If the process was terminated due to a signal
  /// the exit code will be a negative value in the range `[-255..-1]`,
  /// where the absolute value of the exit code is the signal
  /// number. For example, if a process crashes due to a segmentation
  /// violation the exit code will be -11, as the signal SIGSEGV has the
  /// number 11.
```

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: Thanks for the suggestion! I moved this warning to the place where it is after using up all retry counts

XilaiZhang: umm I think if the output does not match because of notary tool version, it will fail consistently and eventually throw an exception. but yeah i moved it to right before we throw the exception so it is more obvious.

codesign/lib/src/file_codesign_visitor.dart
christopherfujino: Right, but if it's not zero, that usually means the process exited because of an error: https://www.gnu.org/software/bash/manual/html_node/Exit-Status.html#:~:text=A%20non%2Dzero%20exit%20status,N%20as%20the%20exit%20status.

XilaiZhang: Ahh thanks for explaining! Update to be nonzero check

christopherfujino (approved):
LGTM

PR #2125, opened 2022-09-13
[codesign] the semi entry point of visiting files

XilaiZhang:
Adds visitRemoteZip function, which triggers a recursive visit of an engine artifact's content. It downloads the artifact from g cloud, unzip it, parse entitlements in it, have visitors recursively visit its contents, triggers notraization upload and check for notarization status. The 'recursively visit its contents' part is delegated to sub apis, but it includes logic for other parts.


codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -124,6 +125,66 @@ update these file paths accordingly.
     await tempDir.delete(recursive: true);
   }
 
+  /// Retrieve eninge artifact from google cloud storage and kick start a
```
CaseyHillers, line 128:
```suggestion
  /// Retrieve engine artifact from google cloud storage and start a
```

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -124,6 +125,66 @@ update these file paths accordingly.
...
+  /// recursive visit of its contents.
+  ///
+  /// Invokes [visitDirectory] to recursively visit the contents of the remote
+  /// zip. Also downloads, notarizes and uploads the engine artifact.
+  Future<void> processRemoteZip({
+    required String artifactFilePath,
+    required Directory parentDirectory,
+  }) async {
+    final FileSystem fs = tempDir.fileSystem;
```
CaseyHillers, line 137: Why is tempDir a class field? You can do `fileSystem.systemTempDirectory.createTempSync()`

https://api.flutter.dev/flutter/package-file_file/Directory/createTempSync.html

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -268,7 +332,7 @@ update these file paths accordingly.
 
     // check on results
     Timer.periodic(
-      _notarizationTimerDuration,
+      notarizationTimerDuration,
```
CaseyHillers, line 335: For polling, 45 seconds seems like a long interval. What about 5 seconds?

codesign/lib/src/utils.dart
```diff
@@ -5,10 +5,45 @@
...
 
 import 'package:args/args.dart';
+import 'package:file/file.dart';
 import 'package:process/process.dart';
 
 import 'log.dart';
 
+const String gsCloudBaseUrl = r'gs://flutter_infra_release';
+
+Future<void> upload({
```
CaseyHillers, line 15: These should be moved into a dedicated gsutil class to ensure they have test coverage, and can be easily faked

codesign/lib/src/utils.dart
```diff
@@ -5,10 +5,45 @@
...
+  final String fullRemotePath = '$gsCloudBaseUrl/flutter/$commitHash/$remotePath';
+  final ProcessResult result = await processManager.run(
+    <String>['gsutil', 'cp', localPath, fullRemotePath],
+  );
+  if (result.exitCode != 0) {
+    throw Exception('Failed to upload $localPath to $fullRemotePath');
+  }
+}
+
+Future<File> download({
```
CaseyHillers, line 30: dartdocs?

codesign/lib/src/utils.dart
```diff
@@ -5,10 +5,45 @@
...
+  required String remotePath,
+  required String localPath,
+  required String commitHash,
```
CaseyHillers, line 33: Consider adding a wrapper function, like `downloadEngineArtifacts()` as a generic `download` function will only have a `uri` parameter and the `destination` of the file

codesign/lib/src/utils.dart
```diff
@@ -5,10 +5,45 @@
...
+  required ProcessManager processManager,
+  required Directory tempDir,
+}) async {
+  final String source = '$gsCloudBaseUrl/flutter/$commitHash/$remotePath';
+  final ProcessResult result = await processManager.run(
+    <String>['gsutil', 'cp', source, localPath],
+  );
+  if (result.exitCode != 0) {
```
CaseyHillers, line 41: Should this be updated to also verify the checksum matches what GCS has?

codesign/test/file_codesign_visitor_test.dart
```diff
@@ -246,6 +247,124 @@ void main() {
       expect(messages, contains('Visiting directory ${tempDir.absolute.path}/embedded_zip_${zipFileName.hashCode}'));
     });
 
+    test('procesRemotezip triggers correct workflow', () async {
```
CaseyHillers, line 250: What is gsutil fails on download? What if it fails on upload? What happens when notarization fails

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -124,6 +125,66 @@ update these file paths accordingly.
...
+      inputDir: parentDirectory,
+      outputZip: codesignedFile,
+      processManager: processManager,
+    );
+
+    // notarize
+    await notarize(codesignedFile);
+
+    // we will only upload for production
+    if (production) {
```
CaseyHillers, line 178: The production flag is an antipattern. See other comment about having a Gsutil class that can be injected

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: True, corrected naming.

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: Right yeah the naming is probably bad, I renamed tempDir to rootDirectory. I treated tempDir as the root directory for file processing. so that api calls such as `visitZip` `processZip` that need to create folder relative to the root directory, could access the root directory through class field. Otherwise we would need to pass root directory information (formerly tempDir) as a parameter to the api calls.

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: Sure, changed to 5 seconds to see if it works out (in case it doesn't i can switch back to 45). I could be wrong but I think previous polls might take more than 3 queries (>135 sec) before the notarization completes.

codesign/lib/src/utils.dart
XilaiZhang: Sure, I grouped the four functions together and refactored them into a class called `engineArtifactTransfer`. It would take in gcloud base url as the parameter for the class. The fake instances of upload and download classes remove the process commands, and only return based on exit code.

codesign/lib/src/utils.dart
XilaiZhang: Sure, added wrapper for both download and upload.

codesign/lib/src/utils.dart
XilaiZhang: umm seems like "gsutil automatically performs integrity checks on all uploads and downloads" based on the doc at https://cloud.google.com/storage/docs/gsutil/addlhelp/CRC32CandInstallingcrcmod. looks like hash verification is implemented as  https://cloud.google.com/storage/docs/gsutil/commands/cp#checksum-validation. looks like automatic retry has been implemented with https://cloud.google.com/storage/docs/gsutil/commands/cp#retry-handling. 

Looks like if retry of hash mismatch fail, gsutil will eventually throw a bad exit code, and we are catching the exit code in our code.

codesign/test/file_codesign_visitor_test.dart
XilaiZhang: Sure, added tests for download fail upload succeed, download succeed upload fail, and both download and upload succeed. The notarization related tests were added in previous PRs under the test group('notarization tests: ')

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: Sure, changed to an injected class. the class will do no-op if not in production mode.

codesign/lib/src/utils.dart
XilaiZhang: Sure, added docs.

codesign/lib/src/file_codesign_visitor.dart
CaseyHillers: I'm still confused. Generally, you should only need the `Directory` or `FileSystem`. Why do you need both?

codesign/lib/src/file_codesign_visitor.dart
CaseyHillers: Yes, but that'll take longer for the overall script. If it takes 95 seconds, the script waits an extra 40 seconds. Over 20 artifacts, that adds up to taking an extra 13 minutes.

codesign/lib/src/file_codesign_visitor.dart
CaseyHillers: Seems like the class wasn't uploaded.

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: Right, yeah I am using `Directory` for folders related creation in the api calls, like rootDirectory.childDirectory(something). So api calls can share the instance of the rootDirectory, and create/delete directories relative to the root directory.

umm my understanding is that `FileSystem` is an abstract class, and to use utilities like fileSystem.path.join and others, we need to invoke based on an instance of the fileSystem. I created an instance of the `FileSystem` as a class field, so api calls can share this instance and invoke the filesystem related utility functions.

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: umm I invoked the upload function with the new class like
```    
await engineArtifactTransfer.uploadEngineArtifact(
      localPath: codesignedFile.path,
      remotePath: artifactFilePath,
      commitHash: commitHash,
    );
```
umm I think the engine artifact will be uploaded. umm would be great if Casey could help elaborate a little bit on the class wasn't uploaded part

CaseyHillers: It would be easier to find if it was in it's own file, and not hidden in utils.dart.

Also, the global functions in utils.dart are still concerning.

XilaiZhang: Sure, I refactored it into its own file transfer_utils.dart.

And sure, I shifted the two global functions (upload and download) into within the EngineArtifactTransfer class, and provided default implementations of them within the class. The class will use default implementations of upload/download if nothing is passed in, and will use user defined download/upload functions if they are passed in.

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -66,6 +67,7 @@ class FileCodesignVisitor {
   late final File entitlementsFile;
   late final Directory remoteDownloadsDir;
   late final Directory codesignedZipsDir;
+  late final Duration notarizationTimerDuration;
```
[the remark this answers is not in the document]
XilaiZhang: very true, thanks for the suggestion Christopher!

codesign/lib/src/transfer_utils.dart
```diff
@@ -0,0 +1,92 @@
...
+
+import 'dart:io';
+
+import 'package:file/file.dart';
+import 'package:process/process.dart';
+
+import './utils.dart';
+
+/// Utility function class to handle upload/download of files from/to google cloud.
+class EngineArtifactTransfer {
```
CaseyHillers, line 13: Best practice for file names is to be named after the class they're in. Utils is generally a bad name as it's very vague

`engine_artifact_transfer.dart`

codesign/lib/src/transfer_utils.dart
```diff
@@ -0,0 +1,92 @@
...
+  EngineArtifactTransfer({
+    required this.gsCloudBaseUrl,
```
CaseyHillers, line 15: You could simplify this by only passing the gcs bucket.

Then, internally you'd need to rewrite some of the urls to be `gs://$bucket/$path`

codesign/lib/src/transfer_utils.dart
```diff
@@ -0,0 +1,92 @@
...
+    this.uploadFunction,
+    this.downloadFunction,
+  }) {
+    uploadFunction ??= defaultUploadFunction;
```
CaseyHillers, line 19: This is a bad practice. It's hard to debug and be sure what's used in prod vs tests.

Prefer injecting the services needed instead of the functions being used. In this case, you only need to pass `ProcessManager` and `FileSystem`

codesign/lib/src/transfer_utils.dart
CaseyHillers, line 13: Based on the dart doc, a clearer name would be `GoogleCloudStorage`

codesign/lib/src/transfer_utils.dart
```diff
@@ -0,0 +1,92 @@
...
+    final ProcessResult result = await processManager.run(
+      <String>['gsutil', 'cp', localPath, destinationUrl],
+    );
+    if (result.exitCode != 0) {
+      throw Exception('Failed to upload $localPath to $destinationUrl');
+    }
+  }
+
+  /// Utility function to download a file from google cloud.
+  Future<File> defaultDownloadFunction({
```
CaseyHillers, line 78: Can we add tests for this class?

codesign/lib/src/transfer_utils.dart
XilaiZhang: Sure, renamed to GoogleCloudStorage.dart based on https://github.com/flutter/cocoon/pull/2125#discussion_r975895105

codesign/lib/src/transfer_utils.dart
XilaiZhang: Sure, simplified constructor parameters.

codesign/lib/src/transfer_utils.dart
XilaiZhang: Sure, renamed.

codesign/lib/src/transfer_utils.dart
XilaiZhang: umm the tests are already there but maybe the naming wasn't obvious. I changed the test group name to group('test google cloud storage and processRemoteZip workflow') so it is more obvious.

codesign/lib/src/google_cloud_storage.dart
```diff
@@ -10,19 +10,10 @@ import 'package:process/process.dart';
...
-    uploadFunction ??= defaultUploadFunction;
-    downloadFunction ??= defaultDownloadFunction;
-  }
+class GoogleCloudStorage {
+  GoogleCloudStorage();
 
-  String gsCloudBaseUrl;
-  Function? uploadFunction;
-  Function? downloadFunction;
+  String gsCloudBaseUrl = r'gs://flutter_infra_release';
```
CaseyHillers, line 16: This doesn't need a raw string

codesign/lib/src/google_cloud_storage.dart
```diff
@@ -51,17 +41,16 @@ class EngineArtifactTransfer {
     int exitCode = 0,
   }) async {
     final String sourceUrl = '$gsCloudBaseUrl/flutter/$commitHash/$remotePath';
-    return await downloadFunction!(
```
CaseyHillers: What's the value in having downloadFunction and downloadEngineArtifact?

nit: You don't need to include "Function" in the name of your methods

codesign/lib/src/google_cloud_storage.dart
```diff
@@ -75,7 +64,7 @@ class EngineArtifactTransfer {
   }
 
   /// Utility function to download a file from google cloud.
-  Future<File> defaultDownloadFunction({
+  Future<File> downloadFunction({
     required String sourceUrl,
     required String localPath,
     required ProcessManager processManager,
```
CaseyHillers, line 70: Instead of passing ProcessManager and FileSystem on every function call, it can be passed when `GoogleCloudStorage` is initialized

codesign/lib/src/google_cloud_storage.dart
```diff
@@ -10,19 +10,10 @@ import 'package:process/process.dart';
 import './utils.dart';
 
 /// Utility function class to handle upload/download of files from/to google cloud.
```
CaseyHillers, line 12: This is a bit more than a utility. It's a service for interacting with the local `gsutil`

codesign/test/src/fake_google_cloud_storage.dart
```diff
@@ -0,0 +1,75 @@
...
+    return await fakeDownloadFunction(
+      sourceUrl: sourceUrl,
+      localPath: localPath,
+      processManager: processManager,
+      rootDirectory: rootDirectory,
+      exitCode: exitCode,
+    );
+  }
+
+  Future<void> fakeUploadFunction({
```
CaseyHillers, line 51: These seem simple enough to inline in the original methods

codesign/lib/src/google_cloud_storage.dart
XilaiZhang: Sure, removed.

codesign/lib/src/google_cloud_storage.dart
XilaiZhang: umm right they come from an earlier review https://github.com/flutter/cocoon/pull/2125#discussion_r971240252 to have a wrapper class around the generic download function. I have reverted it back to its original state (without generic-wrapper).

codesign/lib/src/google_cloud_storage.dart
XilaiZhang: Good advice! I refactored this class

codesign/lib/src/google_cloud_storage.dart
XilaiZhang: Sure, updated description.

codesign/test/src/fake_google_cloud_storage.dart
XilaiZhang: umm I removed this fake class, and reverted back to the fake processManager topology. 

umm I might be wrong, when inlining fake class into real class, I found it a bit messy to `if` check which class is real and provide exit code accordingly. So I removed the fake class.

codesign/lib/src/google_cloud_storage.dart
```diff
@@ -9,67 +9,41 @@ import 'package:process/process.dart';
...
-  String gsCloudBaseUrl = r'gs://flutter_infra_release';
-
-  /// Wrapper function to upload code signed flutter engine artifact to google cloud bucket.
+  GoogleCloudStorage({
+    required this.processManager,
+    required this.rootDirectory,
+    required this.commitHash,
+  });
+
+  ProcessManager processManager;
```
CaseyHillers, line 20: These fields should be final. It looks like the analyzer failed to catch this. Can you try rebasing your branch with main?

https://dart-lang.github.io/linter/lints/prefer_final_fields.html

codesign/lib/src/google_cloud_storage.dart
```diff
@@ -9,67 +9,41 @@ import 'package:process/process.dart';
...
     if (result.exitCode != 0) {
-      throw Exception('Failed to upload $localPath to $destinationUrl');
+      throw CodesignException('Failed to upload $localPath to $destinationUrl');
     }
   }
 
-  /// Utility function to download a file from google cloud.
-  Future<File> downloadFunction({
-    required String sourceUrl,
+  /// Function to download flutter engine artifact from google cloud bucket.
```
CaseyHillers, line 41: nitty nit: Since this is part of a class, it's a method

codesign/lib/src/google_cloud_storage.dart
```diff
@@ -9,67 +9,41 @@ import 'package:process/process.dart';
...
+  Future<File> downloadEngineArtifact({
+    required String remotePath,
```
CaseyHillers, line 43: A common way of phrasing this is `from` and `destination`

CaseyHillers (approved):

codesign/lib/src/google_cloud_storage.dart
XilaiZhang: Sure, rebased branch with main, also added finals keywords.

codesign/lib/src/google_cloud_storage.dart
XilaiZhang: Good point! updated description

codesign/lib/src/google_cloud_storage.dart
XilaiZhang: Sure, updated all the names in src and tests to be `from` and `destination`

PR #2142, opened 2022-09-15
Add devicelab example to CI_YAML doc

keyonghan:
We are lacking doc example for devicelab targets.

This will help reduce tractions when team members add such targets: https://github.com/flutter/flutter/pull/111658

### DIFF
--- a/CI_YAML.md
+++ b/CI_YAML.md
@@ -69,6 +69,23 @@ targets:
        ["framework", "hostonly"]
    postsubmit_properties:
      - upload_metrics: "true"
+
+#
+# Devicelab example:
+# For tests that are located https://github.com/flutter/flutter/tree/master/dev/devicelab/bin/tasks:
+# 1) a `taskname` property is required, which should match the task name
+# 2) target name follows format of `<platform> <taskname>`
+# 3) update `tags` accordingly
+#
+# Here is the target config for a task named: `analyzer_benchmark.dart`.
+ - name: Linux_android analyzer_benchmark
+   recipe: devicelab/devicelab_drone
+   presubmit: false
+   timeout: 60
+   properties:
+     tags: >
+       ["devicelab", "android", "linux"]
+     task_name: analyzer_benchmark
 ```
 
 ### Adding new targets
--- a/CI_YAML.md
+++ b/CI_YAML.md
@@ -35,6 +35,8 @@ targets:
 # name: A human readable string to uniquely identify this target.
 #       The first word indicates the platform this test will be run on. This should match
 #       to an existing platform under platform_properties.
+# recipes: LUCI recipes the target follows to run tests
+#          https://flutter.googlesource.com/recipes/+/refs/heads/main/recipes/
 # bringup: Whether this target is under active development and should not block the tree.
 #          If true, will not run in presubmit and will not block postsubmit.
 # presubmit: Whether to run this target on presubmit (defaults to true).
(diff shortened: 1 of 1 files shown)


CI_YAML.md
```diff
@@ -69,6 +69,23 @@ targets:
...
+# For tests that are located https://github.com/flutter/flutter/tree/master/dev/devicelab/bin/tasks:
+# 1) a `taskname` property is required, which should match the task name
+# 2) target name follows format of `<platform> <taskname>`
+# 3) update `tags` accordingly
+#
+# Here is the target config for a task named: `analyzer_benchmark.dart`.
+ - name: Linux_android analyzer_benchmark
+   recipe: devicelab/devicelab_drone
+   presubmit: false
+   timeout: 60
```
CaseyHillers, line 84: Consider removing timeout. I think we want to start lowering these as most finish within a few minutes.

CaseyHillers (approved):

CI_YAML.md
```diff
@@ -69,6 +69,23 @@ targets:
        ["framework", "hostonly"]
    postsubmit_properties:
      - upload_metrics: "true"
+
+#
+# Devicelab example:
+# For tests that are located https://github.com/flutter/flutter/tree/master/dev/devicelab/bin/tasks:
+# 1) a `taskname` property is required, which should match the task name
+# 2) target name follows format of `<platform> <taskname>`
+# 3) update `tags` accordingly
```
jensjoha, line 78: What does this part mean? What is "accordingly"?

CI_YAML.md
```diff
@@ -69,6 +69,23 @@ targets:
...
+#
+# Here is the target config for a task named: `analyzer_benchmark.dart`.
+ - name: Linux_android analyzer_benchmark
+   recipe: devicelab/devicelab_drone
```
jensjoha, line 82: What does these actually mean? (e.g. presubmit: false --- I assume it means don't run on presubmit, but I'm guessing --- and "timeout" -- from @CaseyHillers comment I'm guessing it's in minutes, but again I don't know... And if there is none (as suggested) what does that mean?

CI_YAML.md
```diff
@@ -69,6 +69,23 @@ targets:
        ["framework", "hostonly"]
    postsubmit_properties:
      - upload_metrics: "true"
+
+#
+# Devicelab example:
+# For tests that are located https://github.com/flutter/flutter/tree/master/dev/devicelab/bin/tasks:
+# 1) a `taskname` property is required, which should match the task name
+# 2) target name follows format of `<platform> <taskname>`
```
jensjoha, line 77: The numbering here is not logical to me. "1" refers to the last line of the segment, "2" to the first.

CI_YAML.md
keyonghan: Removed.

CI_YAML.md
keyonghan: Updated to make it clear.

CI_YAML.md
keyonghan: Added a comment to `targets` section to explain recipes.

Removing the `timeout` to avoid confusion. We have default timeout setup from recipes side to control test timeout if hanging.

CI_YAML.md
keyonghan: Recordered.

jensjoha:
LGTM. (I'll see if I can find a button to click in a moment).

I might add, though, that I had never found out that I had to add stuff in `.ci.yaml` though (and thus not been able to find the link for this file either) if it hadn't been for the explanation in https://github.com/flutter/flutter/pull/111461#issuecomment-1247113905.

jensjoha (approved):

keyonghan:
> I might add, though, that I had never found out that I had to add stuff in .ci.yaml though (and thus not been able to find the link for this file either) if it hadn't been for the explanation in https://github.com/flutter/flutter/pull/111461#issuecomment-1247113905.

F.Y.I. here is the Flutter wiki about how to add a new test shard: https://github.com/flutter/flutter/wiki/Adding-a-new-Test-Shard

jensjoha:
> > I might add, though, that I had never found out that I had to add stuff in .ci.yaml though (and thus not been able to find the link for this file either) if it hadn't been for the explanation in [flutter/flutter#111461 (comment)](https://github.com/flutter/flutter/pull/111461#issuecomment-1247113905).
> 
> F.Y.I. here is the Flutter wiki about how to add a new test shard: https://github.com/flutter/flutter/wiki/Adding-a-new-Test-Shard

Interesting. I was looking for "benchmark" which is probably why I didn't find it.

PR #2148, opened 2022-09-19
Schedule builds after db updates in batch-backfiller

keyonghan:
Fixes: https://github.com/flutter/flutter/issues/111899

### DIFF
--- a/app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
+++ b/app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
@@ -97,7 +97,7 @@ class BatchBackfiller extends RequestHandler {
     }
   }
 
-  // Create list of backfill requests.
+  /// Creates a list of backfill requests.
   List<Future<void>> backfillRequestList(List<Tuple<Target, FullTask, int>> backfill) {
     final List<Future<void>> futures = <Future<void>>[];
     for (Tuple<Target, FullTask, int> tuple in backfill) {
--- a/app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
+++ b/app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
@@ -91,14 +91,14 @@ class BatchBackfiller extends RequestHandler {
       });
       // Schedule all builds asynchronously.
       // Schedule after db updates to avoid duplicate scheduling when db update fails.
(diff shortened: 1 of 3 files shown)


app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
```diff
@@ -80,7 +80,25 @@ class BatchBackfiller extends RequestHandler {
...
+      });
+      // Schedule all builds asynchronously.
+      // Schedule after db updates to avoid duplicate scheduling when db update fails.
+      await Future.wait<void>(backfillRequestList(backfill));
+    } catch (error) {
+      log.severe('Failed to update tasks when backfilling: $error');
+    }
+  }
+
+  // Create list of backfill requests.
```
ricardoamador, line 100: nit: should this be '///' instead since it is a doc string?

app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
```diff
@@ -80,7 +80,25 @@ class BatchBackfiller extends RequestHandler {
...
+    try {
+      await datastore.withTransaction<void>((Transaction transaction) async {
+        transaction.queueMutations(inserts: backfillTasks);
+        await transaction.commit();
+        log.fine(
+            'Updated ${backfillTasks.length} tasks: ${backfillTasks.map((e) => e.name).toList()} when backfilling.');
+      });
+      // Schedule all builds asynchronously.
+      // Schedule after db updates to avoid duplicate scheduling when db update fails.
+      await Future.wait<void>(backfillRequestList(backfill));
```
ricardoamador, line 94: Should we do this in a Finally block instead in case there is an exception?

ricardoamador (approved):
LGTM

app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
keyonghan: Yeah, you are right (copied form earlier comment). Updated.

app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
keyonghan: No. We do not want to schedule when exception happens to avoid duplicate scheduling. If exception happens, we should not update task status, and should not schedule either to be consistent.

PR #2156, opened 2022-09-22
[scheduler] Ignore bringup on release branches

CaseyHillers:
Remove bringup targets from being scheduled on release branches. The release team doesn't look at these, they create noise, and use extra capacity.


itsjustkevin (approved):

ricardoamador (approved):
LGTM

PR #2157, opened 2022-09-22
[app_dart] Remove unused LuciSerivce and LuciBuilder

CaseyHillers:
With the migration to the Cocoon scheduler, these appear to finally be unused. Cleaning up all references from the codebase.


keyonghan (approved):
LGTM

PR #2172, opened 2022-09-28
[dashboard] Truncate branch names

CaseyHillers:
This trims branches to not be as long in the build dashboard. It makes the max length able to show full release candidate branches, but any longer is trimmed.

Adds `package:truncate` to `dashboard`. This is already used in `app_dart`

### DIFF
--- a/dashboard/lib/build_dashboard_page.dart
+++ b/dashboard/lib/build_dashboard_page.dart
@@ -53,6 +53,11 @@ class BuildDashboardPageState extends State<BuildDashboardPage> {
   /// master on one repo, to main for a different repo.
   String? branch;
 
+  /// Example branch for [truncate].
+  ///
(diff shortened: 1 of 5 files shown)


dashboard/lib/build_dashboard_page.dart
```diff
@@ -248,7 +249,7 @@ class BuildDashboardPageState extends State<BuildDashboardPage> {
               padding: const EdgeInsets.only(top: 9.0),
               child: Center(
                 child: Text(
-                  buildState.currentBranch,
+                  truncate(buildState.currentBranch, 28),
```
yusuf-goog, line 252: nit: should this number be a constant?

yusuf-goog (approved):
LGTM with 1 nit.

ricardoamador (approved):
LGTM
I agree with Yusuf on the magic number.

PR #2179, opened 2022-09-29
[codesign] optional switch and dryrun

XilaiZhang:
Add support of ios usb dependency artifacts inherited from https://github.com/christopherfujino/codesign.py/blob/master/codesign.py#L635. This PR adds path and arg support to these artifacts. Dryrun is set to be the default mode that does not upload the signed artifacts, as abiding https://github.com/flutter/flutter/issues/104450#event-6664083158. 

update 10/17: Tested ios usb artifacts by signing the ios usb led artifacts. Verified locally that the app signs them with/without entitlements correctly.

update 10/04: refactored the app to take in 'file path patterns', which are regex-ish file paths to support a generic signing of different types of artifacts


codesign/bin/codesign.dart
```diff
@@ -71,14 +73,15 @@ Future<void> main(List<String> args) async {
       help: 'the Flutter engine commit revision for which Google Cloud Storage binary artifacts should be codesigned.',
     )
     ..addMultiOption(
-      kCodesignFilepathOption,
-      help: 'The zip file paths to be codesigned. Pass this option multiple'
-          'times to codesign multiple zip files',
-      valueHelp: 'darwin-x64/font-subset.zip',
+      kOptionalSwitch,
```
CaseyHillers, line 76: This approach isn't scalable. Someone should be able to give a generic zip with the list of paths they want codesigned.

codesign/bin/codesign.dart
```diff
@@ -14,31 +14,33 @@ import 'package:process/process.dart';
 /// Definitions of variables are included in help texts below.
 const String kHelpFlag = 'help';
 const String kCommitOption = 'commit';
-const String kProductionFlag = 'production';
+const String kDryrunFlag = 'dryrun';
```
CaseyHillers, line 17: nit: It seems like this dryrun change could've been a separate PR. This PR is already extra large, so avoid adding in features that are unrelated.

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -120,9 +118,55 @@ If there are obsolete binaries in entitlements configuration files, please delet
 update these file paths accordingly.
 ''';
 
+  List<String> engineArtifactFilePaths = [
```
CaseyHillers, line 121: I thought the goal with entitlements.txt was to avoid needing this hardcoded list?

codesign/bin/codesign.dart
XilaiZhang: umm right the current implementation signs through gcloud buckets. it takes in an artifact path (that exists on gcloud buckets), and signs all the binaries inside the zip. I might be misunderstanding, umm for the requirement of `Someone give a generic zip with the list of paths they want codesigned`, would the action item be taking in extra parameters for custom gcloud upload/download bucket paths? (or I can add a parameter to take in a custom local zip path, where the zip is located by local file system) 

Right, the current implementation accommodates the various gcloud paths of ios usb dependencies, like the unsigned bucket path '$bucketPrefix/ios-usb-dependencies/unsigned/$rawName/$commitHash/$from', or the artifact bucket path`ios-usb-dependencies/$rawName/$commitHash/$destination`. umm in terms of customized support, would it make sense for us to take the gcloud bucket paths as input parameters instead?

I might be wrong, based on Casey's idea, I am thinking if we take the gcloud bucket paths instead, we would still support the engine paths by default (cause there are like 32 such cloud paths prefix and would be hard to provide through command line). umm I am not sure about scalability, I thought we are signing engine artifacts and ios usb artifacts, or maybe we would like to sign others as well?

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: umm Right yeah it removes the hard coded list of the files within each artifact. my understanding is that the artifact list might still be needed, so we know where to fetch them from/ upload to with respect to the gcloud buckets. (like we wouldn't need to know about darwin-x64/artifacts.zip/FlutterMacOs/framework/blah/blah which is a binary, but we still need the information on the artifact darwin-x64/artifacts.zip so we could fetch it from google cloud)

codesign/bin/codesign.dart
XilaiZhang: umm yeah i agree. umm `dryrun` is actually a rename from the old parameter `production` to make its usage more obvious

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -120,9 +118,54 @@ If there are obsolete binaries in entitlements configuration files, please delet
...
+    });
+    await Future.wait(
+      futures,
+      eagerError: true,
+    );
+
+    if (dryrun) {
+      log.info('code signing dry run has completed, If you intend to upload the artifacts back to'
+          ' google cloud storage, please use the --dryrun=false flag to run code signing script.'
+          ' thanks for understanding the security concerns!');
```
jmagman, line 167: I don't think we need this line 🙂

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -161,7 +204,7 @@ update these file paths accordingly.
     log.info('parsed binaries without entitlements are $fileWithEntitlements');
 
     // recursively visit extracted files
-    await visitDirectory(directory: parentDirectory, entitlementParentPath: artifactFilePath);
+    await visitDirectory(directory: parentDirectory, entitlementParentPath: "");
```
jmagman, line 207: What does an empty string mean?  Should `entitlementParentPath` be nullable instead?

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -196,7 +243,7 @@ update these file paths accordingly.
       if (entity is io.Directory) {
         await visitDirectory(
           directory: directory.childDirectory(entity.basename),
-          entitlementParentPath: entitlementParentPath,
+          entitlementParentPath: joinEntitlementPaths(entitlementParentPath, entity.basename),
```
jmagman, line 246: When would `entitlementParentPath` not be the parent, but instead the entity itself?  This is confusing...

codesign/lib/src/google_cloud_storage.dart
```diff
@@ -15,19 +15,27 @@ class GoogleCloudStorage {
...
   Future<void> uploadEngineArtifact({
     required String from,
     required String destination,
   }) async {
-    final String destinationUrl = '$gsCloudBaseUrl/flutter/$commitHash/$destination';
+    final String destinationUrl;
+    if (optionalSwitch.isEmpty) {
+      destinationUrl = '$bucketPrefix/flutter/$commitHash/$destination';
+    } else {
+      final String rawName = destination.replaceAll(".zip", "");
```
jmagman, line 36: Why does this not need the zip but the other one does?  Can you add a comment?

codesign/lib/src/google_cloud_storage.dart
```diff
@@ -43,7 +51,13 @@ class GoogleCloudStorage {
     required String from,
     required String destination,
   }) async {
-    final String sourceUrl = '$gsCloudBaseUrl/flutter/$commitHash/$from';
+    final String sourceUrl;
+    if (optionalSwitch.isEmpty) {
+      sourceUrl = '$bucketPrefix/flutter/$commitHash/$from';
+    } else {
+      final String rawName = from.replaceAll(".zip", "");
```
jmagman, line 58: Same, why doesn't this need a zip extension but the other one does?

codesign/lib/src/utils.dart
```diff
@@ -124,3 +124,11 @@ String? getValueFromEnvOrArgs(
   throw CodesignException('Expected either the CLI arg --$name or the environment variable $envName '
       'to be provided!');
 }
+
+String joinEntitlementPaths(String entitlementParentPath, String pathToJoin) {
+  if (entitlementParentPath == "") {
+    return pathToJoin;
```
jmagman, line 130: The magic empty string is confusing, I'd rather the caller be explicit about the path than abstracting it into this method.

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: Right good question. Yeah it is kind of confusing, i have renamed `entitlementParentPath` to be `parentVirtualPath`.

Right, the purpose of this variable is to track the theoretical logical path of the file in the file system, assuming that none of the files have been moved or shuffled (either due to extraction or copying around). Previously I wrote a little example in go/code-signing-standalone: 
```
As we move along the file system, we sometimes deviate from the file tree and go to different folders. 
For example, we might extract current zip to another folder for processing, and switch into the other folder. 
This deviation makes it hard to compare our absolute file path with the file path listed in entitlements.txt, because our absolute file path is not accurate.
Instead, we maintain a `virtual file path` along the way, denoting the theoretical position in the filesystem had we not switched folders. 
This virtual path can be easily compared with the paths listed in entitlements.txt.

For example, in the picture above, even though we extract parent_1/parent_2/ios.zip to folder parent_1/folder_ios for processing, in the recursive function call, we will note the virtual file path as parent_1/parent_2/ios.zip. 
```

in other words, if we have a file a/b/c/d/e/f/g/file_binary, and assume while traversing the file system we extracted it and shuffled files around, so by the time we see file_binary it is now located at t/r/y/u/v/s/t/e/file_binary. This virtual path variable would serve as the source of truth that dictates the theoretical path. The virtual path variable would have been passed along and updated during the file system traversal and tells us the theorectical path of file_binary is a/b/c/d/e/f/g/file_binary.

And yeah I set the start path of virtual path to be an empty string. And it adds on filenames as it goes deeper into the file structure. For example, assume we have a file artifacts.zip/impellerc,  at the point of artifacts.zip the virtual path is empty string "", but at the point of impellerc the virtual path is updated to "" + "impellerc" which becomes "impellerc". As another example, assume the file is now artifacts.zip/Flutter/impellerc,  (1) at the point of artifacts.zip the virtual path is empty string "",  (2) at the point of  artifacts.zip/Flutter the virtual path is updated to "" + "Flutter" which becomes "Flutter", (3) at the point of artifacts.zip/Flutter/impellerc the virtual path is updated to "Flutter" + "impellerc" which becomes "Flutter/impellerc". 

So the concatenation of file paths is different based on whether we are starting a virtual path, or we are updating an existing virtual path. If we are starting with a virtual path, we don't want a front slash, so the virtual path of artifacts.zip/impellerc would be "impellerc", as opposed to "/impellerc".  If we are updating a virtual path, we don't want a slash in between, so the virtual path of artifacts.zip/Flutter/impellerc would be "Flutter/impellerc", as opposed to "Flutterimpellerc".

codesign/lib/src/google_cloud_storage.dart
XilaiZhang: Sure, I added a comment. The trimming of .zip is mostly used to create the full gcloud bucket path for ios usb artifacts. For example, assume we have destination as ios-deploy.zip. If this is an engine artifact, the bucket path would be something like ...blah/ios-deploy.zip, where the name is only used once. If this is an ios usb artifact, we process the bucket path to be something like ...blah/ios-deploy/...blah/ios-deploy.zip, which is sort of repeated in the gcloud bucket path.

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: umm I added it here because of https://github.com/flutter/flutter/issues/104387 and https://github.com/flutter/flutter/issues/104450#event-6664083158. (caused by me uploading corrupted artifacts back to gcloud, and created havoc for people). So by design codesign app won't upload artifacts back to gcloud. Since users might want to upload artifacts back instead (to complete the cycle of code signing), I added a prompt saying users could use --dryrun=false to state that they are ready to upload the artifacts back to gcloud (and inform users they would be responsible for it)

CaseyHillers: The "thanks for understanding the security concerns!" is not needed

codesign/bin/codesign.dart
```diff
@@ -14,31 +14,33 @@ import 'package:process/process.dart';
...
+/// For [kOptionalSwitch], it is an optional parameter that you can supply to code sign binaries other than flutter engine binaries.
+/// The five options are: ios-deploy.zip, libimobiledevice.zip, libplist.zip, libusbmuxd.zip, openssl.zip.
+///
 /// Usage:
 /// ```shell
 /// dart run bin/main.dart --commit=$commitSha
-/// --filepath=darwin-x64/FlutterMacOS.framework.zip --filepath=ios/artifacts.zip
-/// --filepath=dart-sdk-darwin-arm64.zip
-/// ( add `--production` if this is intended for production)
+/// ( add `--dryrun` if this is intended for dryrun)
```
CaseyHillers, line 43: nit: Use unix format

```
dart run bin/main.dart --commit=$commitSha [--dryrun]
```

codesign/bin/codesign.dart
```diff
@@ -71,14 +73,15 @@ Future<void> main(List<String> args) async {
       help: 'the Flutter engine commit revision for which Google Cloud Storage binary artifacts should be codesigned.',
     )
     ..addMultiOption(
-      kCodesignFilepathOption,
-      help: 'The zip file paths to be codesigned. Pass this option multiple'
-          'times to codesign multiple zip files',
-      valueHelp: 'darwin-x64/font-subset.zip',
+      kOptionalSwitch,
+      help:
+          'the list of binaries that are supported besides flutter engine binaries. e.g. ios-deploy.zip, libimobiledevice.zip, libplist.zip, libusbmuxd.zip, openssl.zip',
```
CaseyHillers, line 78: nit: You can use `allowedHelp` to make the error message more user friendly

codesign/bin/codesign.dart
```diff
@@ -71,14 +73,15 @@ Future<void> main(List<String> args) async {
...
+      allowed: ["ios-deploy.zip", "libimobiledevice.zip", "libplist.zip", "libusbmuxd.zip", "openssl.zip"],
+      defaultsTo: <String>[],
     )
     ..addFlag(
-      kProductionFlag,
-      help: 'whether we are going to upload the artifacts back to GCS for production',
+      kDryrunFlag,
+      help: 'whether we are going to upload the artifacts back to GCS for dryrun',
```
CaseyHillers, line 84: Dryrun seems to do a few things here. I would also expect it to skip the codesigning process. A short description may be "If passed, production services will not be used"

codesign/lib/src/file_codesign_visitor.dart
CaseyHillers: This is something that should be in the v2 engine recipes for each codesign sub build.

The codesign script should generalize the `switch` option to be an `path` option. Then callers will be able to codesign any generic artifact without needing to update this script.

codesign/bin/codesign.dart
XilaiZhang: Good point, thanks Casey!

codesign/bin/codesign.dart
XilaiZhang: Sure, I have removed kOptionalSwitch based on Casey's advice but this is good to learn!

codesign/bin/codesign.dart
XilaiZhang: umm yeah right currently `dryrun` signs the artifacts, notarizes them, but skips uploading them back to gcs. I guess the benefit of having the codesigning process is that notarization could potentially catch some extra exceptions, and the advantage of skipping codesigning is that we would have a faster dry run. Shall I change the code to make `dryrun` to skip codesigning as well? (If so I will probably do it in a follow up commit, planning to log off now to get dinner)

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: Sure, I removed the list from codesign app. Will remember to add them in the engine recipe.

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: Oh I see, sure, removed!

codesign/bin/codesign.dart
XilaiZhang: Right in the newest commit I have changed `dryrun` to skip codesign and notarization.

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -181,12 +181,12 @@ update these file paths accordingly.
       processManager: processManager,
     );
 
-    // notarize
-    await notarize(codesignedFile);
-
     // `dryrun` flag defaults to true to prevent uploading artifacts back to google cloud.
     // This would help prevent https://github.com/flutter/flutter/issues/104387
     if (!dryrun) {
+      // notarize
```
CaseyHillers, line 187: This comment isn't providing any value to the following line

codesign/test/file_codesign_visitor_test.dart
```diff
@@ -591,6 +591,20 @@ void main() {
...
+        appSpecificPassword: randomString,
+        codesignAppstoreId: randomString,
+        codesignTeamId: randomString,
+        googleCloudStorage: googleCloudStorage,
+        fileSystem: fileSystem,
+        processManager: processManager,
+        rootDirectory: rootDirectory,
+        filePaths: [randomString],
+        dryrun: false,
+        notarizationTimerDuration: const Duration(seconds: 0),
```
CaseyHillers, line 606:
```suggestion
        notarizationTimerDuration: Duration.zero,
```

codesign/bin/codesign.dart
```diff
@@ -69,14 +69,32 @@ Future<void> main(List<String> args) async {
       help: 'Team-id is used by notary service for xcode version 13+.',
     )
     ..addOption(
-      kCommitOption,
-      help: 'the Flutter engine commit revision for which Google Cloud Storage binary artifacts should be codesigned.',
+      kGCloudDownloadPattern,
+      help: 'A required pattern to specify the google cloud bucket prefix of the artifacts stored in google cloud. '
+          'In a pattern, the word ARTIFACTRAWNAME will get replaced to the raw name of artifact without suffix or prefix, the word FILEPATH will get replaced to the artifact path\n'
+          'For example, to code sign ios usb dependency artifacts ios-deploy.zip and libplist.zip, supply the pattern\n'
+          '`--google-cloud-download-pattern = ios-usb-dependencies/unsigned/ARTIFACTRAWNAME/<commitHash>/FILEPATH`\n,'
```
CaseyHillers, line 76: Few notes here,
1. I think this makes more sense in the README.md
2. nit: There's a bash syntax issue with space around the "="
3. I'm not sure what I'm supposed to add here. Is ARTIFACTRAWNAME expected? Is commit hash inputted? For user variables in bash, it helps to denote them as environment variables (like $ARTIFACT_RAW_NAME). Then users can copy and paste the exact command, but before it set up the inputs.
4. File path seems duplicate, and could be combined with this variable to just be `gcsPath`

XilaiZhang: umm right i think the main confusion comes from the fact that each type of artifact requires a different type of pattern to be used. for example, ios-usb artifact might want a gcs path upload pattern of a/b/ios-usb/c/ios-usb.zip, and a download path of a/b/ios-usb/c/unsigned/ios-usb.zip, whereas an engine artifact needs a gcs path with the pattern flutter/commit/engine.zip. I think there are 3 options:

1. let user input the pattern
pros: scalable
cons: it becomes confusing for the user to craft the pattern themself.

2. assume the pattern ourselves, and let user specify the type of artifact.
pros: clear, env variables and parameters easy to use
cons: not scalable, need to add a new pattern for each new type of artifact introduced.

3. use a single gcs path parameter. I think the problem with only taking in a gcsPath parameter is that, gcs upload paths are different from gcs download paths. and when signing multiple artifacts, users might supply flags such as --upload_1 --upload_2 --download_2 --download_1, and it would be hard for code sign app to match and pair the upload and download paths.

With option 2 we can assume the pattern for users and reduce the confusions that would be caused by large blob of texts instructing users how to craft their own patterns. The downside of option 2 is that it is not as scalable as Casey pointed out, because we have to introduce a new pattern whenever a new tip of artifact is introduced. option 1 is scalable but like Casey pointed out it is confusing to read, and option 3 does not quite work because codesign app can't auto pair the download of upload paths.

umm I am thinking if there is another way of implementing it so that it is clear.

CaseyHillers: > an engine artifact needs a gcs path with the pattern flutter/commit/engine.zip

I think you answer that in 3. For the download path, users would input gs://flutter_infra_releases/flutter/commitabc123/mac_arm/...

Can you expand on why users would need to input multiple downloads and uploads instead of rerunning the app? My understanding is the signing sub builds will only execute one download/upload.

For the issue with mapping downloads/uploads, using package:args you can have a single flag that takes in multiple values.

XilaiZhang: Synced with Casey yesterday and the plan was to have codesign app only support signing a single artifact each time, so that we can streamline the process of generic signing, and pass in parameters without user crafting. I went ahead and removed multi artifact signing logic but there is a chance that we still need to sign multiple artifacts at the same time:

As the way it is implemented in https://flutter-review.googlesource.com/c/recipes/+/34501,
1. for global generators, they will produce multiple artifacts in the multiple out directories, and this might require a multi artifacts parameter sign.
2. for each archive, it produces multiple artifacts as shown in `included_paths`, which might require a multi parameters sign. 

To be able to sign single parameter, and align with what we would like in the sub builder design in go/code-signing-bot, my understanding is that we would need a triggering of _schedule_with_led for code signing job from within the _schedule_with_led of the sub build of the artifacts. But my understanding could be off. Would be great if sir @godofredoc could provide some insights! Thank you!

godofredoc: Simpler implementations are better. Ideally we can sign a single artifact at a time and the recipe logic can take care of running the tool for as many artifacts as the build is generating.

godofredoc: We can split the work here, if you tell what the interface for calling the tool is and we manually generate a cipd package I can help with the signer recipe implementation.

XilaiZhang: Synced with sir Godofredo and my current plan is to 
1. let codesign support single parameter file path (meaning signing one artifact at most each time) 
2. parse the "parameters" field or add "included paths" in generator configs, so that we can later retrieve the total file paths to be codesigned for global generators. 
4. for sub build such as 'host_debug' which produces multiple 'included paths', or global generators, wait until the group of 'included paths' have been built, and feed them one by one to the schedule with led for codesign.

jmagman:
Any update on getting this merged?  Can we do a one-off code signing until the details are worked out?  We really need to update these binaries before the next beta to unblock go/flutter-macos-binary-verification

codesign/test/file_codesign_visitor_test.dart
XilaiZhang: Neat, this is very helpful! Thanks for pointing this out

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: True, deleted.

XilaiZhang:
>Any update on getting this merged? Can we do a one-off code signing until the details are worked out? We really need to update these binaries before the next beta to unblock go/flutter-macos-binary-verification

Yes for sure, in the newest commit I addressed the previous feedbacks, and changed the path argument to be a single parameter. In its current state, this app can code sign ios usb artifacts with commands similar to `dart run bin/codesign.dart [--dryrun] --google-cloud-download-upload-path=flutter/<commit>/android-arm-profile/artifacts.zip#flutter/<commit>/android-arm-profile/artifacts.zip`. I tested it with the led artifact (for example, `ios-usb-dependencies/unsigned/libimobiledevice/led/libimobiledevice.zip`) and it signs correctly with/without entitlements. 

As for the next step based on urgency, since https://flutter-review.googlesource.com/c/recipes/+/34300 is merged, we should be expecting entitlement configs well exist in the artifacts. Therefore if this PR gets merged, we can use this app to sign ios usb artifacts. If this PR takes extra time to review, feel free to download the app at the commit [805a295](https://github.com/flutter/cocoon/pull/2179/commits/805a295aa4792de56fdd87f336ca0af1b3746270) to do a one-off signing. (takes each file path parameter separately) [a075c1d](https://github.com/flutter/cocoon/pull/2179/commits/a075c1de72476550cb11aed555c6db9c0b8327a7) could be used for one-off signing as well. (signs 5 ios usb dependency in one go). 

I think it is likely that we will just get this PR merged. But yeah feel free to do one-off signing based on urgency.

codesign/bin/codesign.dart
```diff
@@ -30,17 +28,16 @@ const String kGCloudUploadPattern = 'google-cloud-upload-pattern';
...
 /// true.
 ///
-/// For [kCodesignFilepathOption], supply the filepaths of binaries to be codesigned.
-/// e.g. supply `--filepath=darwin-x86/artifacts.zip --ios-release/artifacts.zip` if you would like to codesign darwin-x86/artifacts.zip and ios-release/artifacts.zip
-///
-/// For [kGCloudDownloadPattern], it is a required parameter to specify the google cloud bucket prefix of the artifacts stored in google cloud.
-/// Artifacts stored with this bucket prefix path are downloaded from google cloud and codesigned.
-/// For [kGCloudUploadPattern], it is a required parameter to specify the google cloud bucket prefix for the code signed artifacts to be uploaded to.
-/// Code signed artifacts are uploaded to google cloud under this bucket prefix.
+/// For [kGCloudDownloadUploadPathOption], it is a required parameter to specify the google cloud bucket paths.
```
CaseyHillers, line 31: This should be two separate options (a download path and an upload path)

codesign/lib/src/google_cloud_storage.dart
```diff
@@ -14,27 +14,20 @@ class GoogleCloudStorage {
   GoogleCloudStorage({
     required this.processManager,
     required this.rootDirectory,
-    required this.gCloudDownloadPattern,
-    required this.gCloudUploadPattern,
+    required this.gCloudDownloadUploadPath,
```
CaseyHillers, line 17: This should remain in the methods instead of the constructor.

If we ever need to support multiple downloads/uploads, it would have to be changed.

codesign/bin/codesign.dart
XilaiZhang: Sure, done.

codesign/lib/src/google_cloud_storage.dart
XilaiZhang: Sure, refactored.

codesign/bin/codesign.dart
XilaiZhang: (sneak in one extra comment: Casey looks very strong in the new github profile picture)

codesign/bin/codesign.dart
```diff
@@ -19,7 +19,8 @@ const String kCodesignUserNameOption = 'codesign-username';
 const String kAppSpecificPasswordOption = 'app-specific-password';
 const String kCodesignAppStoreIdOption = 'codesign-appstore-id';
 const String kCodesignTeamIdOption = 'codesign-team-id';
-const String kGCloudDownloadUploadPathOption = 'google-cloud-download-upload-path';
+const String kGCloudDownloadPathOption = 'google-cloud-download-path';
```
CaseyHillers, line 22: nit: You can make these shorter by saying `gcs-download-path`

Similarly, instead of saying `kGCloudDownloadPathOption`, it would be more specific to say `kGcsDownloadPathOption` as there's a few ways to download from GCP

codesign/bin/codesign.dart
```diff
@@ -28,16 +29,16 @@ const String kGCloudDownloadUploadPathOption = 'google-cloud-download-upload-pat
...
+/// For [kGCloudDownloadPathOption] and [kGCloudUploadPathOption], they are required parameter to specify the google cloud bucket paths.
+/// [kGCloudDownloadPathOption] is the google cloud bucket prefix to download the remote artifacts,
+/// [kGCloudUploadPathOption] is the cloud bucket prefix to upload codesigned artifact to.
+/// For example, supply '--google-cloud-download-path=ios-usb-dependencies/unsigned/libimobiledevice/<commit>/libimobiledevice.zip',
+/// and code sign app will download the artifact at 'ios-usb-dependencies/unsigned/libimobiledevice/<commit>/libimobiledevice.zip'
 ///
 /// Usage:
 /// ```shell
-/// dart run bin/codesign.dart [--dryrun] --google-cloud-download-upload-path=flutter/<commit>/android-arm-profile/artifacts.zip#flutter/<commit>/android-arm-profile/artifacts.zip
+/// dart run bin/codesign.dart [--dryrun] --google-cloud-download-path=flutter/<commit>/android-arm-profile/artifacts.zip
```
CaseyHillers, line 40: Should this be gs://flutter_infra_release/flutter/<commit>/android-arm-profile/artifacts.zip?

codesign/lib/src/google_cloud_storage.dart
```diff
@@ -14,19 +14,17 @@ class GoogleCloudStorage {
...
 
   final ProcessManager processManager;
   final Directory rootDirectory;
-  final String gCloudDownloadUploadPath;
   final String bucketPrefix = 'gs://flutter_infra_release';
 
   /// Method to upload code signed flutter engine artifact to google cloud bucket.
   Future<void> uploadEngineArtifact({
     required String from,
+    required String gCloudUploadPath,
```
CaseyHillers, line 26: Since this is already in a GCS service class, you can simplify it back to `destination`

CaseyHillers (approved):
I have a few comments left on the ergonomics, once addressed LGTM!

codesign/bin/codesign.dart
XilaiZhang: Sure, renamed.

codesign/lib/src/google_cloud_storage.dart
XilaiZhang: Sure, renamed.

codesign/bin/codesign.dart
XilaiZhang: Good point, yeah I thought about it and right it's better for user to provide the full path including gs:// . refactored.

PR #2183, opened 2022-10-03
[codesign + yaml] publish codesign cipd

XilaiZhang:
remove bring up to have codesign cipd published..

context: based on the led results on https://ci.chromium.org/raw/build/logs.chromium.org/flutter/led/xilaizhang_google.com/9443be22646c25c8114585337a18218e2b3a99cc6d2029a4191d41703726da80/+/build.proto?server=chromium-swarm.appspot.com, and inspecting device doctor and codesign.


XilaiZhang:
getting a write permission at [go/bug-luci-admin](https://goto.google.com/bug-luci-admin) based on https://ci.chromium.org/ui/p/flutter/builders/try/Mac%20codesign/1/overview. But the write permission server returns a 500 error as of now.

.ci.yaml
```diff
@@ -60,8 +60,7 @@ targets:
       - .ci.yaml
 
   - name: Mac codesign
-    bringup: true
-    postsubmit: false
+    presubmit: false
```
CaseyHillers, line 63: Should this run on presubmit?

XilaiZhang: umm my understanding is that like Casey said, presubmit would use a try builder to write to cipd, so we might want to avoid this and not building and publishing on presubmit

CaseyHillers: The recipe should be updated to only upload in prod

XilaiZhang: umm device doctor and codesign shares the same recipe. Like Casey pointed out, device doctor needs the presubmit try builders to upload the cipd packages. If updating to only upload in prod it would also affect device doctor. umm I probably don't have access, shall we modify ACL for device doctor?

CaseyHillers: I sent https://flutter-review.googlesource.com/c/recipes/+/34580

CaseyHillers: Recipes is better for things that will have ACL issues (like whether to upload artifacts), and ci.yaml is better for build/test configurations. In this case, you can configure in ci.yaml what cipd packages should be built, but recipes should have the upload logic.

XilaiZhang: Thanks for explaining Casey!

CaseyHillers (approved):

PR #2188, opened 2022-10-05
Enable postsubmit run for device_doctor builders

keyonghan:
With https://flutter-review.googlesource.com/c/recipes/+/34580, packages will not be uploaded to cipd for try runs.

This PR enable post-submit runs, so that we can upload packages, which happens after reviews.


CaseyHillers (approved):

ricardoamador (approved):
LGTM

PR #2199, opened 2022-10-07
Revert will wait for required checks to complete.

ricardoamador:
## Description

When creating a revert request, a user can add a label to the request while opening the revert. This means the checks will be made immediately and when this occurs we can sometimes get in the situation that he ci.yaml a required check has not yet completed. This means the label will be removed and the user will be notified we cannot merge the request. 

This fix is to wait only for the ci.yaml check to complete before merging as it is the only required check. 

image

## Issues Addressed

Fixes https://github.com/flutter/flutter/issues/113070

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/auto_submit/lib/service/config.dart
+++ b/auto_submit/lib/service/config.dart
@@ -37,8 +37,12 @@ class Config {
   static const String kRevertLabel = 'revert';
 
   static const int backOfMultiplier = 200;
-  static const int backoffAttempts = 3;
+  static const int backoffAttempts = 5;
   static const int maxDelaySeconds = 1;
+
+  static const int backOfMultiplierCheckWait = 500;
+  static const int backoffAttemptsCheckWait = 5;
+  static const int maxDelaySecondsCheckWait = 5;
 
   static const String flutter = 'flutter';
   static const String flutterGcpProjectId = 'flutter-dashboard';
--- a/auto_submit/lib/validations/required_check_runs.dart
+++ b/auto_submit/lib/validations/required_check_runs.dart
@@ -1,0 +1,5 @@
+// Copyright 2022 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+const List<String> requiredCheckRuns = ['ci.yaml validation'];
--- a/auto_submit/lib/validations/revert.dart
+++ b/auto_submit/lib/validations/revert.dart
@@ -2,25 +2,38 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'package:auto_submit/exception/retryable_checkrun_exception.dart';
 import 'package:auto_submit/model/auto_submit_query_result.dart' as auto;
+import 'package:auto_submit/service/config.dart';
 import 'package:auto_submit/service/github_service.dart';
+import 'package:auto_submit/validations/required_check_runs.dart';
 import 'package:auto_submit/validations/validation.dart';
 import 'package:github/github.dart' as github;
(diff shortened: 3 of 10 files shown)


auto_submit/lib/service/config.dart
```diff
@@ -37,9 +37,13 @@ class Config {
   static const String kRevertLabel = 'revert';
 
   static const int backOfMultiplier = 200;
-  static const int backoffAttempts = 3;
+  static const int backoffAttempts = 5;
```
CaseyHillers, line 40: It would be cleaner to consolidate these into a [RetryOption](https://pub.dev/documentation/retry/latest/retry/RetryOptions-class.html)

auto_submit/lib/validations/required_check_runs.dart
```diff
@@ -0,0 +1,5 @@
+// Copyright 2022 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+const List<String> requiredCheckRuns = ['ci.yaml validation'];
```
CaseyHillers, line 5: It's weird that the GitHub API doesn't surface what is considered a required check run. You may want to leave a comment here as I went digging in the API to confirm this.

auto_submit/lib/validations/revert.dart
```diff
@@ -100,4 +129,73 @@ class Revert extends Validation {
...
+
+    return checksCompleted;
+  }
+}
+
+/// Function signature that will be executed with retries.
+typedef RetryHandler = Function();
+
+/// Runs the internal verify checks runs function with retries.
+Future<void> _waitForRequiredCheckRunsWithRetries(RetryHandler retryHandler, RetryOptions retryOptions) {
```
CaseyHillers, line 179: This is simple enough that you may want to inline it

auto_submit/lib/validations/revert.dart
```diff
@@ -100,4 +129,73 @@ class Revert extends Validation {
...
+              slug,
+              githubService,
+              checkRun,
+            );
+          },
+          retryOptions,
+        );
+      }
+    } catch (e) {
+      log.warning('Required check has not completed in time. ${e.toString()}');
```
CaseyHillers, line 167: What happens if we enable autosubmit on non-flutter repos that would not have the ci.yaml validation check?

auto_submit/lib/validations/required_check_runs.dart
ricardoamador: Haha yeah I spent a bit of time trying to see if I could get that information as well. I will definitely add a note.

auto_submit/lib/validations/revert.dart
ricardoamador: That is a good question. Currently this will be an issue. We need to map the required tests to repositories since github does not expose required checks.

auto_submit/lib/service/config.dart
ricardoamador: That is a good idea.

auto_submit/lib/validations/revert.dart
ricardoamador: After changing to a mapping with repository name, this will work fine. If a repository does not have any required checks it would not run the retry function. The drawback to this is that since github does not expose required checks we have to keep track of this manually within the source.

auto_submit/lib/validations/revert.dart
ricardoamador: Done.

CaseyHillers (approved):
flutter-lgtm

PR #2201, opened 2022-10-07
Add delay between deflake github requests

keyonghan:
When multiple github requests are called in a short period, it hits 
```
You have exceeded a secondary rate limit 
```
This PR adds a 1s delay between calls following: https://docs.github.com/en/rest/guides/best-practices-for-integrators#dealing-with-secondary-rate-limits

More context: https://github.com/flutter/flutter/issues/102402

### DIFF
--- a/app_dart/lib/src/request_handlers/check_flaky_builders.dart
+++ b/app_dart/lib/src/request_handlers/check_flaky_builders.dart
@@ -81,6 +81,9 @@ class CheckFlakyBuilders extends ApiRequestHandler<Body> {
           await bigquery.listRecentBuildRecordsForBuilder(kBigQueryProjectId, builder: info.name, limit: kRecordNumber);
       if (_shouldDeflake(builderRecords)) {
         await _deflakyPullRequest(gitHub, slug, info: info, ciContent: ciContent, testOwnership: testOwnership);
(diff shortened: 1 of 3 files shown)


app_dart/lib/src/request_handlers/check_flaky_builders.dart
```diff
@@ -81,6 +81,9 @@ class CheckFlakyBuilders extends ApiRequestHandler<Body> {
           await bigquery.listRecentBuildRecordsForBuilder(kBigQueryProjectId, builder: info.name, limit: kRecordNumber);
       if (_shouldDeflake(builderRecords)) {
         await _deflakyPullRequest(gitHub, slug, info: info, ciContent: ciContent, testOwnership: testOwnership);
+        // Manually add a 1s delay between consecutive GitHub requests to deal with secondary rate limit error.
+        // https://docs.github.com/en/rest/guides/best-practices-for-integrators#dealing-with-secondary-rate-limits
+        await Future.delayed(const Duration(seconds: 1));
```
CaseyHillers, line 86: This should be injectable so tests can use `Duration.zero`

keyonghan: Defined in config so that tests could use default zero from fake.

CaseyHillers (approved):

ricardoamador (approved):
LGTM

PR #2206, opened 2022-10-10
Using protobuf instead of ciyaml directly.

ricardoamador:
## Description

Update to flaky issues handler code so that we use the protobuf and wrappers instead of using the ciYaml directly. This will avoid issues with deprecated features as we will only receive changes via the protobuf definitions.

## Issue(s)

https://github.com/flutter/flutter/issues/87529

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read the [Flutter Style Guide] _recently_, and have followed its advice.
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [ ] I updated/added relevant documentation (doc comments with `///`).
- [ ] I added new tests to check the change I am making, or this PR is [test-exempt].
- [ ] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[test-exempt]: https://github.com/flutter/flutter/wiki/Tree-hygiene#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat

PR #2220, opened 2022-10-17
Revert "Using protobuf instead of ciyaml directly."

keyonghan:
Reverts flutter/cocoon#2206

This is causing tree breakage: https://github.com/flutter/flutter/issues/113473


fluttergithubbot (approved):

ricardoamador:
Are the test owners in the ciyaml? I am not sure how that was caused by these changes. I never touched anything in any test ownership related file.

ricardoamador:
Found the test owners file now I am confused as to how this would affect what is being retrieved from that file. Is that test a presubmit test?

CaseyHillers:
@ricardoamador it sounds like we need to add an integration test here so Cocoon can run the same test. There's something similar for validating ci.yamls that can be reused.

ricardoamador:
The issue with changing to use the protobuf file is that the current line: 
```
# verify_binaries_codesigned @christopherfujino @flutter/releases
```
uses @flutter/releases which is invalid. The test will work if it is changed to something like flutter/tool.

PR #2250, opened 2022-10-21
flutter upgrade to mockito 5.3.2

CaseyHillers:
Upgrades autosubmit and app_dart to the latest versions.

Inline custom_mocks for app_dart to get past analyzer issue

Fix latest app_dart lints

Regenerate codegen


auto_submit/pubspec.yaml
```diff
@@ -17,7 +17,7 @@ dependencies:
   googleapis: ^9.2.0
   googleapis_auth: ^1.3.1
   graphql: ^5.1.1
-  gql: ^0.13.1
+  gql: ^0.14.0
```
keyonghan, line 20: (mentioned for line 11)
Do you intend to update sdk to `'>=2.19.0-0 <3.0.0'`?

CaseyHillers: Yep! app_dart is intended to be built against the tip of tree, so this is an uprev to the latest version

ricardoamador (approved):
LGTM

PR #2253, opened 2022-10-21
remove test dependency on int hashcode

jonahwilliams:
Allows Dart to change int hashcode implementation without breaking cocoon test


dashboard/lib/service/dev_cocoon.dart
```diff
@@ -178,8 +178,92 @@ class DevelopmentCocoonService implements CocoonService {
   final List<String> _authors = <String>['alice', 'bob', 'charlie', 'dobb', 'eli', 'fred'];
   final List<int> _messagePrimes = <int>[3, 11, 17, 23, 31, 41, 47, 67, 79];
   final List<String> _words = <String>['fixes', 'issue', 'crash', 'developer', 'blocker', 'intermittent', 'format'];
+  final List<String> _commits = <String>[
```
jonahwilliams, line 181: git log --pretty=online from the cocoon repo, I grabbed some extras. just in case.

zanderso (approved):

CaseyHillers (approved):

PR #2276, opened 2022-10-29
[app_dart] Reduce create branch delays

CaseyHillers:
An attempted fix at getting the pubsub retries to kick in by reducing max runtime to less than 60 seconds.

https://github.com/flutter/flutter/issues/112036


keyonghan (approved):
Rebase is needed to pass the CI.

PR #2289, opened 2022-11-03
[codesign + recipe] refactor gsutil and zip utilities 

XilaiZhang:
Refactors `Gsutil` utility to be delegated to the recipe, because triggering gsutil from codesign app will need authentication and other complexities (as according to sir Godofredo). Luci will now handle xcode isntallation, gsutil download and upload. The codesign app will now take in arguments `input-zip-file-path` and `output-zip-file-path`, code sign the artifacts located at `input-zip-file-path`, and output the file to `output-zip-file-path`.

Note that a corrupted output zip file (not of zip type but with same name) at the output-zip-file-path location would cause failures. 

For Casey's context, the motivation of this PR comes from the led run of codesign recipe.


codesign/bin/codesign.dart
```diff
@@ -81,16 +81,14 @@ Future<void> main(List<String> args) async {
           'the name of the certificate for flutter, for example, is: FLUTTER.IO LLC',
     )
     ..addOption(
-      kGcsDownloadPathOption,
-      help: 'The google cloud bucket path to download the artifact from\n'
-          'e.g. supply `--gcs-download-path=gs://flutter_infra_release/ios-usb-dependencies/unsigned/ios-deploy/<commit>/ios-deploy.zip`'
-          ' if you would like to codesign ios-deploy.zip, which has a google cloud bucket path of flutter_infra_release/ios-usb-dependencies/unsigned/ios-deploy/<commit>/ios-deploy.zip to be downloaded from \n',
+      kInputZipPathOption,
+      help:
+          'File path to the unsigned artifact zip file. The artifact should have been pre-fetched from gcs bucket through recipe \n',
```
christopherfujino, line 86:
```suggestion
          'File path to the unsigned artifact zip file. The artifact should have been pre-fetched from gcs bucket through recipe.',
```

codesign/bin/codesign.dart
```diff
@@ -81,16 +81,14 @@ Future<void> main(List<String> args) async {
...
     )
     ..addOption(
-      kGcsUploadPathOption,
-      help: 'The google cloud bucket path to upload the artifact to. \n'
-          'e.g. supply `--gcs-upload-path=gs://flutter_infra_release/ios-usb-dependencies/ios-deploy/<commit>/ios-deploy.zip`'
-          ' if you would like to codesign ios-deploy.zip, which has a google cloud bucket path of flutter_infra_release/ios-usb-dependencies/ios-deploy/<commit>/ios-deploy.zip to be uploaded to',
+      kOutputZipPathOption,
+      help: 'File path to store codesigned artifact zip file.'
+          ' The recipe will take codesigned zip from this location and upload back to gcs bucket \n',
```
christopherfujino, line 91:
```suggestion
          ' The recipe will take codesigned zip from this location and upload back to gcs bucket.',
```

codesign/test/file_codesign_visitor_test.dart
```diff
@@ -20,33 +19,29 @@ void main() {
...
-        googleCloudStorage: googleCloudStorage,
         fileSystem: fileSystem,
         appSpecificPasswordFilePath: appSpecificPasswordFilePath,
         codesignAppstoreIDFilePath: codesignAppstoreIDFilePath,
         codesignTeamIDFilePath: codesignTeamIDFilePath,
         processManager: processManager,
         rootDirectory: rootDirectory,
-        gcsDownloadPath: 'gs://flutter/$randomString/$randomString',
-        gcsUploadPath: 'gs://flutter/$randomString/$randomString',
+        inputZipPath: randomString,
```
christopherfujino, line 43: do we really want the input and output path to be the same? also, shouldn't these files have a .zip extension?

codesign/bin/codesign.dart
XilaiZhang: Sure, removed ending newline.

codesign/bin/codesign.dart
XilaiZhang: Indeed, tested and it looks like the ending newline had no effect. Thanks for the suggestion Christopher!

codesign/test/file_codesign_visitor_test.dart
XilaiZhang: Very true I meant to set them to inputZipPath and outputZipPath. Good eye Christopher!

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -173,32 +166,29 @@ update these file paths accordingly.
...
 
-    if (dryrun) {
-      log.info('code signing dry run has completed, If you intend to upload the artifacts back to'
-          ' google cloud storage, please use the --dryrun=false flag to run code signing script.');
+    if (codesignedFilePath != null) {
+      log.info('code signing dry run has completed, If you have uploaded the artifacts back to'
+          ' google cloud storage, please delete the folder $codesignedFilePath and $inputZipPath');
     }
-    log.info('Codesigned all binaries in ${rootDirectory.path}');
+    log.info('Codesign completed. Codesigned zip is located at ${rootDirectory.path}');
```
christopherfujino, line 175: `rootDirectory` is a directory, not a zip file, right?

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -173,32 +166,29 @@ update these file paths accordingly.
...
 
-    await rootDirectory.delete(recursive: true);
+    rootDirectory.list(recursive: true).listen((FileSystemEntity file) async {
+      if (file is File && file.path != codesignedFilePath) {
```
godofredoc, line 178: Is this for extracted paths? if it is I'd recommend to unzip, read, sign and delete temp files from a single try/finally

codesign/lib/src/google_cloud_storage.dart
```diff
@@ -1,49 +0,0 @@
-// Copyright 2019 The Flutter Authors. All rights reserved.
```
godofredoc, line 1: Should this file be removed rather than just deleting the content of it?

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: Very true, corrected naming.

codesign/bin/codesign.dart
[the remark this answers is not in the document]
XilaiZhang: Sure, done.

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -173,32 +166,29 @@ update these file paths accordingly.
     codesignTeamId = availablePasswords['CODESIGN_TEAM_ID']!;
     appSpecificPassword = availablePasswords['APP_SPECIFIC_PASSWORD']!;
 
-    await processRemoteZip();
+    final String? codesignedFilePath = await processRemoteZip();
 
-    if (dryrun) {
-      log.info('code signing dry run has completed, If you intend to upload the artifacts back to'
-          ' google cloud storage, please use the --dryrun=false flag to run code signing script.');
+    if (codesignedFilePath != null) {
```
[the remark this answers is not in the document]
XilaiZhang: Right the exceptions will be thrown as exceptions. Right the nullable value is for the dryrun mode which skips the notarzaiton process and prompts the user with the hint that 'code signing dry run has completed, ... blah'. I changed the check to `if dryrun` to be more explicit. This value is only shared to the cleanup process to determine which files to remove.

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: Yep its for cleanup. Sure thanks for the suggestion! I reorganized it in a try finally block.

codesign/lib/src/google_cloud_storage.dart
XilaiZhang: umm for some reasons this file is shown as removed on github ui on my end. 
Screen Shot 2022-11-03 at 2 41 11 PM

XilaiZhang:
Thanks a lot for all the feedbacks! 
(I re-read through my code a few times and hopefully there are at least no more spelling mistakes :)

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -158,47 +151,49 @@ update these file paths accordingly.
...
-    appSpecificPassword = availablePasswords['APP_SPECIFIC_PASSWORD']!;
-
-    await processRemoteZip();
-
-    if (dryrun) {
-      log.info('code signing dry run has completed, If you intend to upload the artifacts back to'
-          ' google cloud storage, please use the --dryrun=false flag to run code signing script.');
+    // The value of codesignedFilePath is used in the cleanup process in the finally block.
+    String? codesignedFilePath;
+    try {
```
godofredoc, line 156: The code will be more readable if you cleanup the temp file as part of the iteration.

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -158,47 +151,49 @@ update these file paths accordingly.
...
+        await readPassword(passwordFilePath);
+      }
+      if (availablePasswords.containsValue('')) {
+        throw CodesignException('certian passwords are missing. \n'
+            'make sure you have provided <CODESIGN_APPSTORE_ID>, <CODESIGN_TEAM_ID>, and <APP_SPECIFIC_PASSWORD>');
+      }
+      codesignAppstoreId = availablePasswords['CODESIGN_APPSTORE_ID']!;
+      codesignTeamId = availablePasswords['CODESIGN_TEAM_ID']!;
+      appSpecificPassword = availablePasswords['APP_SPECIFIC_PASSWORD']!;
+      codesignedFilePath = await processRemoteZip();
```
godofredoc, line 171: if this line is the one generating the temp files you can probably move the try - finally  block here.

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -151,38 +151,39 @@ update these file paths accordingly.
 
   /// The entrance point of examining and code signing an engine artifact.
   Future<void> validateAll() async {
+    for (String passwordFilePath in [
+      codesignAppstoreIDFilePath,
+      codesignTeamIDFilePath,
+      appSpecificPasswordFilePath,
+    ]) {
+      await readPassword(passwordFilePath);
```
godofredoc, line 159: Is the output of this method supposed to be saved somewhere?  I'd suggest to refactor readPassword to return the value from the file and remove the validations. If the user is passing a file with wrong values that is the responsibility of the caller.

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: Right I was thinking about the clean up process. Currently I told the code to "remove files in root directory except the output path". I was considering the different cases that would trigger a remove if clean up is delegated to the visitors during the iteration, but I think there might be corner cases. I think the current way might be less error prone but I might be wrong, what do you think? If a delegated cleanup is more ideal I can submit it in a followup PR since there would probably be many changes.

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: Sure, moved.

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -151,38 +151,39 @@ update these file paths accordingly.
...
+    }
+    if (availablePasswords.containsValue('')) {
+      throw CodesignException('certian passwords are missing. \n'
+          'make sure you have provided <CODESIGN_APPSTORE_ID>, <CODESIGN_TEAM_ID>, and <APP_SPECIFIC_PASSWORD>');
+    }
```
godofredoc, line 164: maybe something like:

`codesignAppstoreId = readPassword(codesignAppstoreIDFilePath);`

 and remove  all the other parts.

codesign/lib/src/file_codesign_visitor.dart
godofredoc: From what I understand the processing of a single file is a hermetic process. E.g. get the path to the file, unzips it to a temp folder, read the metadata file, signs the relevant files, creates a new zip in the output location and removes the temporary files.

Not sure what corner cases are we expecting to get if we delete the files after processing each file but I see multiple benefits, the code will be cleaner, temp files will be removed right after they are not needed anymore(this is even more important for the bot with the limited disk space), reduce the risk of potentially leaving temp files behind.

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: Sure, refactored.

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: Sure, thanks for the insights! Todo for myself is to update encrypted password files and update recipe.

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: Oh maybe I have misunderstood "as part of iteration", I thought iteration meant the traversal of the file tree. But now I re-read it and I think maybe it meant delegating cleanup from `validateAll()` to `processRemoteZip()`?

godofredoc (approved):
Thanks for the changes.

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: I retraced the tmp file generation and changed the cleanup process to be two delete statements, both of which like sir Godofredo said removes the unzipped folders. I think these two places should cover all the cleanup processes. Yeah in deed this looks cleaner than it used to be. Thanks a lot for providing the insights!

XilaiZhang:
> Thanks for the changes.

Yeah I wasn't as big brain as sir Godofredo and Christopher. Now that I revisit and savor over these improvements I start to appreciate the beauty of them.

PR #2297, opened 2022-11-10
[codesign] remove password files after runs

XilaiZhang:
fixes https://github.com/flutter/flutter/issues/115117.

Ensure the password files are cleaned up after runs.

### DIFF
--- a/codesign/test/file_codesign_visitor_test.dart
+++ b/codesign/test/file_codesign_visitor_test.dart
@@ -1107,6 +1107,10 @@ status: Invalid''',
             'going through the notary service. To run the full codesign process, use --no-dryrun flag.'),
       );
       rootDirectory = fileSystem.systemTempDirectory.createTempSync('conductor_codesign');
+
+      await fileSystem.file(appSpecificPasswordFilePath).delete();
+      await fileSystem.file(codesignAppstoreIDFilePath).delete();
+      await fileSystem.file(codesignTeamIDFilePath).delete();
     });
 
     test('upload optional switch artifacts when dryrun is false', () async {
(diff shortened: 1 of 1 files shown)


codesign/test/file_codesign_visitor_test.dart
```diff
@@ -1200,6 +1204,10 @@ status: Invalid''',
         ),
       );
       rootDirectory = fileSystem.systemTempDirectory.createTempSync('conductor_codesign');
+
+      await fileSystem.file(appSpecificPasswordFilePath).delete();
```
CaseyHillers, line 1208: L24 should be made:

```
  late file FileSystem fileSystem;
```

Then the `setUp` should instantiate the filesystem.

XilaiZhang: Done. Yeah this is cleaner. By the way would recreating filesystems be cheaper than recreating files?

codesign/test/file_codesign_visitor_test.dart
```diff
@@ -21,16 +21,17 @@ void main() {
...
+  late MemoryFileSystem fileSystem;
   late FakeProcessManager processManager;
   late cs.FileCodesignVisitor codesignVisitor;
-
-  Directory rootDirectory = fileSystem.systemTempDirectory.createTempSync('conductor_codesign');
+  late Directory rootDirectory;
 
   group('test reading in passwords: ', () {
     setUp(() {
+      fileSystem = MemoryFileSystem.test();
```
CaseyHillers, line 33: This doesn't block your PR, but in the future you can remove a lot of this duplicate work in a refactor by nesting your setUps.

It's common to have a group for the entire file, then groups for different parts of logic

CaseyHillers (approved):

codesign/test/file_codesign_visitor_test.dart
XilaiZhang: Thanks a lot for the suggestion! Refactored

PR #2316, opened 2022-11-21
[codesign] rename codesign app name to avoid collision

XilaiZhang:
The codesign app is renamed from "codesign" to "flutter_codesign".

When the app name is "codesign", and when executing recipe in led run, the system recipe would automatically look for system codesign. Also, the codesign app itself calls upon the system codesign, so it would be tricky to get away the naming collision with `with api.context`.


XilaiZhang:
actually this might be the wrong place to change it, should be a change on the cipd recipe

godofredoc (commented):
Any reason why you thought this was the wrong place to change the name of the binary?

XilaiZhang:
> Any reason why you thought this was the wrong place to change the name of the binary?

Thanks for the hint! I have updated the PR. Yeah I eventually realized that the key is the -o of the binary name. For example, `-o build/flutter_codesign` as opposed to `-o build/codesign`, when using the build script`tool/dart-sdk/bin/dart compile exe bin/flutter_codesign.dart -o build/flutter_codesign`

christopherfujino (commented):
Drive by comment, when invoking this from the recipe, we should use the absolute path, rather than depend on the name being present and resolving to the right binary on the path

XilaiZhang:
> Drive by comment, when invoking this from the recipe, we should use the absolute path, rather than depend on the name being present and resolving to the right binary on the path

Umm my understanding is that an absolute path invocation could be challenging, since our cipd pkg was installed to a `self.m.path.mkdtemp()` folder in recipe. Maybe it would be hard to retreive this absolute path.

CaseyHillers:
> self.m.path.mkdtemp() folder in recipe

`abspath` can be used on the returned `Path` object in recipes.

XilaiZhang:
> > self.m.path.mkdtemp() folder in recipe
> 
> `abspath` can be used on the returned `Path` object in recipes.

True, umm currently the path isn't returned from flutter deps. Should we change flutter deps to return the path?

XilaiZhang:
> > > self.m.path.mkdtemp() folder in recipe
> > 
> > 
> > `abspath` can be used on the returned `Path` object in recipes.
> 
> True, umm currently the path isn't returned from flutter deps(https://flutter.googlesource.com/recipes/+/refs/heads/main/recipe_modules/flutter_deps/api.py#424). Should we change flutter deps to return the path?

christopherfujino:
> > > > self.m.path.mkdtemp() folder in recipe
> > > 
> > > 
> > > `abspath` can be used on the returned `Path` object in recipes.
> > 
> > 
> > True, umm currently the path isn't returned from flutter deps(https://flutter.googlesource.com/recipes/+/refs/heads/main/recipe_modules/flutter_deps/api.py#424). Should we change flutter deps to return the path?

That should be a safe change, since currently it returns nothing (and thus you won't break existing callers).

godofredoc:
> urned `Path` object in recipes.

> True, umm currently the path isn't returned from flutter deps. Should we change flutter deps to return the path?

The generic way of managing dependencies in recipes is to provide the `dependencies` property with a dictionary per dependency and getting all the dependencies installed with a single call to https://flutter.googlesource.com/recipes/+/refs/heads/main/recipe_modules/flutter_deps/api.py#45. Calling  https://flutter.googlesource.com/recipes/+/refs/heads/main/recipe_modules/flutter_deps/api.py#424 directly is an option but we would need to pass an adhoc property for the tool version.

XilaiZhang:
Right, correct me if I am wrong, I think Christopher and Casey's suggestion would be something like https://flutter-review.git.corp.google.com/c/recipes/+/35322/10..12 . My understanding is that both the absolute path approach and the renaming binary approach would be able to resolve name collision.

godofredoc:
> 



> Right, correct me if I am wrong, I think Christopher and Casey's suggestion would be something like https://flutter-review.git.corp.google.com/c/recipes/+/35322/10..12 . My understanding is that both the absolute path approach and the renaming binary approach would be able to resolve name collision.

I have no strong opinions about either solution, will let @christopherfujino @CaseyHillers LGTM this change

CaseyHillers:
> I have no strong opinions about either solution, will let @christopherfujino @CaseyHillers LGTM this change

@XilaiZhang this sounds like a recipe only fix. Closing this PR SGTM, and patching recipes to use the abspath is what I would expect. Generally, you want to avoid relying on binaries in PATH as it makes it harder to debug.

XilaiZhang:
Sure, thanks for sharing the insights! Closing now

PR #2326, opened 2022-11-29
[codesign] absolute path for system codesign

XilaiZhang:
Change system codesign to use absolute path.

For our mac bots, system codesign are located at /usr/bin/codesign. Based on discussion in https://github.com/flutter/cocoon/pull/2316 , we would like to use the absolute path to identify the different callable `codesign`s. Without specifying an absolute path, the name resolution would fail during the led recipe run.


christopherfujino (approved):
lgtm

PR #2329, opened 2022-12-02
[codesign + open to suggestions] log warning instead of exception when metadata isn't present

XilaiZhang:
open to suggestions, we have two options here:
1. enforce every artifact to have both `entitlements.txt` and `without_entitlements.txt` as metadata, and throw exception when any of the metadata is missing.
2. allow artifact to have a single or no metadata file, to imply that the artifact does not have any binary files to be codesigned (with/without entitlements). In this case we will convert exception to warning.

### DIFF
--- a/codesign/lib/src/file_codesign_visitor.dart
+++ b/codesign/lib/src/file_codesign_visitor.dart
@@ -317,8 +317,10 @@ update these file paths accordingly.
         ? fileSystem.path.join(parent.path, 'entitlements.txt')
         : fileSystem.path.join(parent.path, 'without_entitlements.txt');
     if (!(await fileSystem.file(entitlementFilePath).exists())) {
-      throw CodesignException('$entitlementFilePath not found \n'
-          'make sure you have provided them along with the engine artifacts \n');
+      log.warning('$entitlementFilePath not found. '
+          'by default, system will assume there is no ${entitlements ? '' : 'without_'}entitlements file.'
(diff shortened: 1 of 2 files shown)


XilaiZhang:
maybe @christopherfujino and @godofredoc would have some insights on which approach would be better. Thank you!

godofredoc:
@XilaiZhang should we close or land this PR? is still relevant?

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -317,8 +317,10 @@ update these file paths accordingly.
         ? fileSystem.path.join(parent.path, 'entitlements.txt')
         : fileSystem.path.join(parent.path, 'without_entitlements.txt');
     if (!(await fileSystem.file(entitlementFilePath).exists())) {
-      throw CodesignException('$entitlementFilePath not found \n'
-          'make sure you have provided them along with the engine artifacts \n');
+      log.warning('$entitlementFilePath not found. '
```
christopherfujino, line 320: The result is that no binaries will be signed, right? I think we should make that clear in this warning message (it was not clear to me until I read the PR description).

XilaiZhang: Good advice! Added a line informing user no binaries will be signed.

godofredoc (approved):

PR #2343, opened 2022-12-09
Try to find existing check run instead of creating new one when re-running a postsubmit build from the dashboard

nehalvpatel:
When you currently hit RERUN on a postsubmit build from the dashboard, it creates a new check run on the GitHub PR. This PR tries to find and use the existing check run from the previous attempt.

Addresses https://github.com/flutter/flutter/issues/116739, which is tracking https://github.com/flutter/flutter/issues/116226

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/app_dart/lib/src/foundation/github_checks_util.dart
+++ b/app_dart/lib/src/foundation/github_checks_util.dart
@@ -140,4 +140,55 @@ class GithubChecksUtil {
       output: output,
     );
   }
+
+  /// Retrieves all check runs that match the Git [ref].
+  ///
+  /// Optionally, you can supply [checkName], [status], and/or [filter] to refine the query.
+  Future<Stream<github.CheckRun>> listCheckRunsForRef(
+    Config cocoonConfig,
+    github.RepositorySlug slug, {
+    required String ref,
+    String? checkName,
+    github.CheckRunStatus? status,
+    github.CheckRunFilter? filter,
+  }) async {
+    const RetryOptions r = RetryOptions(
+      maxAttempts: 3,
+      delayFactor: Duration(seconds: 2),
+    );
+    return r.retry(
+      () async {
+        return _listCheckRunsForRef(
+          cocoonConfig,
+          slug,
+          ref: ref,
+          checkName: checkName,
+          status: status,
+          filter: filter,
+        );
+      },
+      retryIf: (Exception e) => true,
+      onRetry: (Exception e) => log.warning(
+        'listCheckRunsForRef fails for slug: ${slug.fullName}, ref: $ref, checkName: $checkName, status: $status, filter: $filter. Exception: ${e.toString()}',
+      ),
+    );
+  }
+
+  Future<Stream<github.CheckRun>> _listCheckRunsForRef(
+    Config cocoonConfig,
+    github.RepositorySlug slug, {
+    required String ref,
+    String? checkName,
+    github.CheckRunStatus? status,
+    github.CheckRunFilter? filter,
+  }) async {
+    final github.GitHub gitHubClient = await cocoonConfig.createGitHubClient(slug: slug);
+    return gitHubClient.checks.checkRuns.listCheckRunsForRef(
+      slug,
+      ref: ref,
+      checkName: checkName,
+      status: status,
+      filter: filter,
+    );
+  }
 }
--- a/app_dart/lib/src/service/luci_build_service.dart
+++ b/app_dart/lib/src/service/luci_build_service.dart
@@ -523,6 +523,7 @@ class LuciBuildService {
     Map<String, Object>? properties,
     Map<String, List<String>>? tags,
     int priority = kDefaultPriority,
+    github.CheckRun? existingCheckRun,
   }) async {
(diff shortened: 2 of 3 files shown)


app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -608,14 +612,15 @@ class LuciBuildService {
   ///   1. It has been tried below the max retry limit
   ///   2. It is for the tip of tree
   ///   3. The last known status is not green
-  ///   4. [ignoreChecks] is false. This allows manual reruns to bypass the Cocoon state.
+  ///   4. [ignoreChecks] is true. This allows manual reruns to bypass the Cocoon state.
```
nehalvpatel, line 615: I think this was a typo in the documentation.

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -580,26 +591,19 @@ class LuciBuildService {
...
     Commit commit,
     Target target,
-    Map<String, dynamic> rawUserData,
   ) async {
     if (!config.githubPostsubmitSupportedRepo(commit.slug) || target.value.bringup) {
-      return;
+      return null;
     }
-    final github.CheckRun checkRun = await githubChecksUtil.createCheckRun(
+    return await githubChecksUtil.createCheckRun(
```
ricardoamador, line 601: Do we need to recreate a check run or can we rerequest an existing checkrun: https://g3doc.corp.google.com/company/teams/eip-cloud/overground/user-guide/quickstart.md?cl=head#onboard-project

app_dart/lib/src/foundation/github_checks_util.dart
```diff
@@ -140,4 +140,55 @@ class GithubChecksUtil {
...
+  Future<Stream<github.CheckRun>> _listCheckRunsForRef(
+    Config cocoonConfig,
+    github.RepositorySlug slug, {
+    required String ref,
+    String? checkName,
+    github.CheckRunStatus? status,
+    github.CheckRunFilter? filter,
+  }) async {
+    final github.GitHub gitHubClient = await cocoonConfig.createGitHubClient(slug: slug);
+    return gitHubClient.checks.checkRuns.listCheckRunsForRef(
```
ricardoamador, line 186: Also just a heads up you might want to look at what this code is actually calling since this is basically a wrapper to a REST call. There might be ways you can get what you need by modifying the library. Just an fyi.

app_dart/lib/src/service/luci_build_service.dart
ricardoamador: Though I am not sure how it associates the new rerequested check run with the particular pr.

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -538,12 +539,22 @@ class LuciBuildService {
...
     log.info('Task task_key: $taskKey for task name: ${task.name}');
 
+    // Received by [PostsubmitLuciSubscription] when LUCI phones back with build results.
     final Map<String, dynamic> rawUserData = <String, dynamic>{
       'commit_key': commitKey,
       'task_key': taskKey,
     };
 
-    await createPostsubmitCheckRun(commit, target, rawUserData);
+    final checkRun = existingCheckRun ?? await createPostsubmitCheckRun(commit, target);
```
ricardoamador, line 548: Maybe the wording is getting me confused but it looks like you are finding a checkrun but ultimately creating a new one again?

app_dart/lib/src/foundation/github_checks_util.dart
nehalvpatel: Oh yea, I was glad to find this function since I was prototyping this implementation by looking at the API. https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28#list-check-runs-for-a-git-reference

This is the command that I was using locally:
```bash
gh api -H "Accept: application/vnd.github+json" "/repos/flutter/plugins/commits/614d7c0df1eb03d2579118b4a0c3032dbedfa364/check-runs?check=Mac_x64%20ios_platform_tests_4_of_4%20master"
```

app_dart/lib/src/service/luci_build_service.dart
nehalvpatel: So, my goal here was to use the `??` null operator to create a new check run if `existingCheckRun` was not passed or if it was explicitly passed as null. This is because this `_createPostsubmitScheduleBuild()` hosting method is also used for brand new builds.

Maybe I could think of a way to clarify it.

---

There is an edge case where maybe the `existingCheckRun` is not found way up in the `ResetProdTask` request handler:
https://github.com/flutter/cocoon/pull/2343/files#diff-8b18660d0517d62995b30d9277a6cd6601f428ccd5def5d4535f668cd6ba8cb3R90

But to make that less likely, I have `listCheckRunsForRef` doing retries so we can at least mostly factor out HTTP connection errors:
https://github.com/flutter/cocoon/pull/2343/files#diff-fd1f79cec05184daf8caf11256548ca4b3ff25cc840d6cf14f9f0cfc6c31d0c6R147

PR #2381, opened 2023-01-10
[cron] Increase vacuum frequency to 10 minutes

CaseyHillers:
Reduce outage latency.

With the removal of the release branch logic from this handler, there should be spare quota to running this more often.


nehalvpatel (approved):
LGTM!

PR #2389, opened 2023-01-13
Cache locking in Cocoon for access tokens

ricardoamador:
Adding locking to the cache around reading and writing values hoping to eliminate the issues with writing and creating tokens at the same.

Fixes https://github.com/flutter/flutter/issues/118358

## Pre-launch Checklist

- [x] I signed the [CLA].


app_dart/lib/src/service/cache_service.dart
```diff
@@ -56,42 +56,40 @@ class CacheService {
...
 
-      // If given createFn, update the cache value if the value returned was null.
-      if (createFn != null && value == null) {
-        // Try creating the value
-        value = await createFn();
-        print("Acquiring getOrCreate write lock.");
-        await set(subcacheName, key, value, ttl: ttl);
+    await m.acquireRead();
+    try {
+      print("Acquiring getOrCreate read lock.");
```
godofredoc, line 64: Should this be logging instead?

app_dart/lib/src/service/cache_service.dart
```diff
@@ -56,42 +56,40 @@ class CacheService {
...
+      value = await subcache[key].get();
+    } catch (e) {
+      if (attempt < maxCacheGetAttempts) {
```
godofredoc, line 67: Can we use the retry module instead?

app_dart/lib/src/service/cache_service.dart
```diff
@@ -56,42 +56,40 @@ class CacheService {
...
+          attempt: ++attempt,
+          createFn: createFn,
+          ttl: ttl,
+        );
+      } else {
+        // Give up on trying to get the value from the cache.
+        value = null;
       }
+    } finally {
+      print("Releasing getOrCreate read lock.");
```
godofredoc, line 80: s/print/logging?

app_dart/lib/src/service/cache_service.dart
```diff
@@ -56,42 +56,40 @@ class CacheService {
...
+      m.release();
+    }
+
+    // If given createFn, update the cache value if the value returned was null.
+    if (createFn != null && value == null) {
+      // Try creating the value
+      value = await createFn();
+      print("Acquiring getOrCreate write lock.");
+      await set(subcacheName, key, value, ttl: ttl);
```
godofredoc, line 89: do we need a write lock for this?

app_dart/lib/src/service/cache_service.dart
```diff
@@ -56,42 +56,40 @@ class CacheService {
...
       }
+    } finally {
+      print("Releasing getOrCreate read lock.");
+      m.release();
+    }
+
+    // If given createFn, update the cache value if the value returned was null.
+    if (createFn != null && value == null) {
+      // Try creating the value
+      value = await createFn();
```
godofredoc, line 87: Should createFn have a default value? with this logic if it is null the returned value will be never recreated.

CaseyHillers: +1 - This code was written pre-null safety, so marking it required should be a quick switch

app_dart/test/service/cache_service_test.dart
```diff
@@ -118,6 +119,28 @@ void main() {
       expect(valueAfterPurge, isNull);
     });
 
+    test('set blocks read attempt', () async {
+      const String testKey = 'abc';
+      final Uint8List expectedValue = Uint8List.fromList('123'.codeUnits);
+
+      unawaited(cache.set(testSubcacheName, testKey, expectedValue));
```
CaseyHillers, line 126: Instead, doing `final cacheWrite = cache.set(testSubcacheName, testKey, expectedValue);` will prevent race conditions with timers left pending.

```dart
final cacheWrite = cache.set(testSubcacheName, testKey, expectedValue);
// Do the tests

await cacheWrite; // Finish pending futures.
```

app_dart/lib/src/service/cache_service.dart
```diff
@@ -20,6 +21,8 @@ class CacheService {
   }) : _provider =
             inMemory ? Cache.inMemoryCacheProvider(inMemoryMaxNumberEntries) : Cache.redisCacheProvider(memorystoreUri);
 
+  final ReadWriteMutex m = ReadWriteMutex();
```
CaseyHillers, line 24: 1. package:mutex will only work locally. During peak hours, Cocoon can have 5 instances. I think this is worth landing as it's simpler than implementing a datastore lock, but keep in mind we may need to continue working on this.
2. Is this lock universal to all cache entries? If so, there is bound to be a lot of contention in it's current state

app_dart/lib/src/service/cache_service.dart
ricardoamador: Yeah sorry I had this for debugging. I can remove it.

app_dart/lib/src/service/cache_service.dart
ricardoamador: You can disregard these. They will get removed.

app_dart/lib/src/service/cache_service.dart
ricardoamador: Yeah no problem. I can fix this.

app_dart/lib/src/service/cache_service.dart
ricardoamador: Hmmm, I see what you are saying. 

This guards against the multiple calls locally but yeah we will still have possible issues when we are attempting to store the new tokens in the app installation access tokens storage.  

We should create a token service.

Yes the lock blocks the cache for all entries, not just tokens. Perhaps we need to keep multiple caches for tokens and other values to keep them separate.

app_dart/lib/src/service/cache_service.dart
```diff
@@ -56,7 +59,9 @@ class CacheService {
     final Cache<Uint8List> subcache = cache.withPrefix(subcacheName);
     Uint8List? value;
 
+    await m.acquireRead();
```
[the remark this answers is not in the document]
ricardoamador: Yeah based on your other comment this might be a good thing to do!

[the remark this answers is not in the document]
ricardoamador: After thinking about this some more I think it would be better to create a different class that specifically has locking in it that the calls for tokens will use. I think this is the safer method.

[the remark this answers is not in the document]
ricardoamador: Hmm okay after looking through the code I will go ahead with the locking methods.

app_dart/test/service/cache_service_test.dart
ricardoamador: Ah okay. Cool. Thank you.

app_dart/lib/src/service/cache_service.dart
ricardoamador: Opening an issue for further work. I will link it here.

app_dart/lib/src/service/cache_service.dart
ricardoamador: Yeah I will change this to use the retry module.

app_dart/lib/src/service/cache_service.dart
```diff
@@ -49,40 +53,70 @@ class CacheService {
...
+      );
     }
 
     return value;
   }
 
+  /// This method is the same as the [getOrCreate] method above except that it
+  /// enforces locking access.
+  ///
+  /// Note: these methods are intended to prevent issues around race conditions
```
CaseyHillers, line 97: It would be good call out this only locks on this instance, not the entire fleet.

app_dart/lib/src/service/cache_service.dart
```diff
@@ -95,10 +129,46 @@ class CacheService {
...
-  Future<void> purge(String subcacheName, String key) {
-    final Cache<Uint8List> subcache = cache.withPrefix(subcacheName);
-    return subcache[key].purge(retries: maxCacheGetAttempts);
+  ///
+  /// Note: these methods are intended to prevent issues around race conditions
+  /// when storing and retrieving github tokens. Care should be taken to use the
+  /// locking methods together when accessing data from an entity using the
+  /// cache.
+  Future<void> purge(String subcacheName, String key) async {
+    await m.acquireWrite();
```
CaseyHillers, line 165: Why does purge need a write lock?

app_dart/lib/src/service/cache_service.dart
```diff
@@ -49,40 +53,70 @@ class CacheService {
...
+  /// when storing and retrieving github tokens. Care should be taken to use the
+  /// locking methods together when accessing data from an entity using the
+  /// cache.
+  Future<Uint8List?> getOrCreateWithLocking(
+    String subcacheName,
+    String key, {
+    required Future<Uint8List> Function()? createFn,
+    Duration ttl = const Duration(minutes: 1),
+  }) async {
+    await m.acquireRead();
```
CaseyHillers, line 107: Do we want to lock the read as well? I would think we'd only lock the write.

app_dart/lib/src/service/cache_service.dart
ricardoamador: Done.

app_dart/lib/src/service/cache_service.dart
ricardoamador: Purge is a write so it is good in the case that something is writing through locking that we prevent that action while attempting to purge the cache. That way we don't immediately write then wipe the value.

app_dart/lib/src/service/cache_service.dart
ricardoamador: You know what, I don't think this will work as it. Let me make a quick change.

app_dart/lib/src/service/cache_service.dart
ricardoamador: Up to you if you don't think we need it.

CaseyHillers (approved):

PR #2402, opened 2023-01-18
Use search API to find matching PR for ci.yaml validation reruns

nehalvpatel:
According to https://github.com/flutter/flutter/issues/100081, this should fix ci.yaml validation errors.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/app_dart/lib/src/foundation/github_checks_util.dart
+++ b/app_dart/lib/src/foundation/github_checks_util.dart
@@ -41,7 +41,7 @@ class GithubChecksUtil {
   }
 
   /// Finds all check suites that are associated with a given git [ref].
-  Future<Stream<github.CheckSuite>> listCheckSuitesForRef(
+  Future<List<github.CheckSuite>> listCheckSuitesForRef(
     github.GitHub gitHubClient,
     github.RepositorySlug slug, {
     required String ref,
@@ -55,7 +55,13 @@ class GithubChecksUtil {
     return r.retry(
       () async {
         return gitHubClient.checks.checkSuites
-            .listCheckSuitesForRef(slug, ref: ref, appId: appId, checkName: checkName);
+            .listCheckSuitesForRef(
+              slug,
+              ref: ref,
+              appId: appId,
+              checkName: checkName,
+            )
+            .toList();
       },
       retryIf: (Exception e) => e is github.GitHubError || e is SocketException,
     );
--- a/app_dart/lib/src/foundation/github_checks_util.dart
+++ b/app_dart/lib/src/foundation/github_checks_util.dart
@@ -40,6 +40,27 @@ class GithubChecksUtil {
     );
   }
 
+  /// Finds all check suites that are associated with a given git [ref].
+  Future<Stream<github.CheckSuite>> listCheckSuitesForRef(
+    github.GitHub gitHubClient,
+    github.RepositorySlug slug, {
+    required String ref,
+    int? appId,
+    String? checkName,
+  }) async {
+    const RetryOptions r = RetryOptions(
+      maxAttempts: 3,
+      delayFactor: Duration(seconds: 2),
+    );
+    return r.retry(
+      () async {
+        return gitHubClient.checks.checkSuites
+            .listCheckSuitesForRef(slug, ref: ref, appId: appId, checkName: checkName);
+      },
+      retryIf: (Exception e) => e is github.GitHubError || e is SocketException,
+    );
+  }
+
   /// Sends a request to github checks api to update a [CheckRun] with a given
   /// [status] and [conclusion].
   Future<void> updateCheckRun(
--- a/app_dart/lib/src/service/github_service.dart
+++ b/app_dart/lib/src/service/github_service.dart
@@ -287,9 +287,12 @@ class GithubService {
 
   /// Returns a [Stream] of [Issue]s that match the given [query].
   ///
+  /// The GitHub package uses the [Issue] object for both issue results and PRs.
+  ///
   /// Reference:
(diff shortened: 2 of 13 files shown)


app_dart/lib/src/service/github_service.dart
```diff
@@ -284,4 +284,20 @@ class GithubService {
     final Random rnd = Random();
     return String.fromCharCodes(Iterable<int>.generate(10, (_) => chars.codeUnitAt(rnd.nextInt(chars.length))));
   }
+
+  /// Returns a [Stream] of [Issue]s that match the given [query].
+  ///
+  /// Reference:
+  ///   * https://docs.github.com/en/rest/search?apiVersion=2022-11-28#search-issues-and-pull-requests
+  Stream<Issue> searchIssues(RepositorySlug slug, String query, {String? sort, int pages = 2}) {
```
godofredoc, line 292: searchIssuesAndPrs?

Maybe also and the format of the query parameter.

app_dart/lib/src/service/github_service.dart
```diff
@@ -284,4 +284,20 @@ class GithubService {
     final Random rnd = Random();
     return String.fromCharCodes(Iterable<int>.generate(10, (_) => chars.codeUnitAt(rnd.nextInt(chars.length))));
   }
+
+  /// Returns a [Stream] of [Issue]s that match the given [query].
```
godofredoc, line 288: Are Issues and PRs returned in the same object?

app_dart/lib/src/service/scheduler.dart
```diff
@@ -466,6 +469,33 @@ class Scheduler {
     return true;
   }
 
+  /// Given a [checkRunEvent], finds the [PullRequest] that the check run must belong to.
+  Future<PullRequest?> _findMatchingPullRequest(cocoon_checks.CheckRunEvent checkRunEvent) async {
```
godofredoc, line 473: Does this belong to the github API? I see potential use cases for it in the future and it won't be intuitive to use from here.

app_dart/lib/src/foundation/github_checks_util.dart
```diff
@@ -40,6 +40,27 @@ class GithubChecksUtil {
     );
   }
 
+  /// Finds all check suites that are associated with a given git [ref].
+  Future<Stream<github.CheckSuite>> listCheckSuitesForRef(
```
godofredoc, line 44: Tests for this method?

app_dart/lib/src/service/github_service.dart
godofredoc, line 288: Tests for the new methods?

app_dart/lib/src/service/scheduler.dart
```diff
@@ -444,11 +444,14 @@ class Scheduler {
         final String? name = checkRunEvent.checkRun!.name;
         bool success = false;
         if (name == kCiYamlCheckName) {
-          // TODO(chillers): This is not rerunning the ci.yaml validation check. https://github.com/flutter/flutter/issues/100081
-          final List<github.PullRequest> pullRequests = checkRunEvent.checkRun!.pullRequests ?? <github.PullRequest>[];
-          for (github.PullRequest pullRequest in pullRequests) {
+          // The [CheckRunEvent.checkRun.pullRequests] array is empty for this
+          // event, so we need to find the matching pull request.
+          final pullRequest = await _findMatchingPullRequest(checkRunEvent);
```
godofredoc, line 449: Type?

app_dart/lib/src/service/github_service.dart
nehalvpatel: Yes, the github package uses the Issue object for both issues and PRs.

app_dart/lib/src/service/scheduler.dart
nehalvpatel: That sounds good. I'll shift it over.

app_dart/lib/src/foundation/github_checks_util.dart
nehalvpatel: Can you suggest how I should test this? I added a test for `searchIssuesAndPRs` to validate that it encodes the query properly, but for `listCheckSuitesForRef` and `getPullRequest` I can't think of anything since they just forward the call unconditionally.

app_dart/lib/src/foundation/github_checks_util.dart
```diff
@@ -40,6 +40,27 @@ class GithubChecksUtil {
     );
   }
 
+  /// Finds all check suites that are associated with a given git [ref].
```
CaseyHillers, line 43: Streams add a lot of complexity in Dart. Simplifying this to an iterable will make it easier to use.

nehalvpatel:
Setting to draft while I fix the scheduler test.

app_dart/lib/src/foundation/github_checks_util.dart
nehalvpatel: Sounds good, updated

godofredoc (approved):
Can we manually release a version to test this change?

nehalvpatel:
> Can we manually release a version to test this change?

I was able to deploy it and successfully perform a ci.yaml validation rerun on https://github.com/flutter/flutter/pull/118690. 😄

The only hitch is that previously failed webhook events which requested to rerun ci.yaml keep being retried. The problem is that the PRs associated are already closed or have had a new commit, which invalidates the check suite. This causes the event processing to fail and is causing looping retries. I think we might have to drain those events.

godofredoc:
> I think we might have to drain those events.

You can take a look at the subscription, it may already be configured to auto drop messages older than 24h

nehalvpatel:
Looks like it is set to 7 days.

Screenshot 2023-01-20 at 11 32 24 AM

PR #2421, opened 2023-01-25
make pattern more general

christopherfujino:
Fixes https://github.com/flutter/flutter/issues/119197
Fixes https://github.com/flutter/flutter/issues/119151
Fixes https://github.com/flutter/flutter/issues/119147

### DIFF
--- a/device_doctor/lib/src/ios_debug_symbol_doctor.dart
+++ b/device_doctor/lib/src/ios_debug_symbol_doctor.dart
@@ -160,7 +160,7 @@ class XCDevice {
     required this.name,
   });
 
-  static const String _debugSymbolDescriptionPattern = 'iPhone is busy: Fetching debug symbols for iPhone';
+  static const String _debugSymbolDescriptionPattern = r' is busy: Fetching debug symbols for ';
 
   /// Parse subset of JSON from `parseJson` associated with a particular XCDevice.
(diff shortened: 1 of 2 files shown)


jmagman (approved):
LGTM, thanks!

yusuf-goog:
@christopherfujino looks like there were a few failures that should have been fixed by your change, but still happened? https://ci.chromium.org/p/flutter/builders/prod/Mac_ios%20route_test_ios

jmagman:
The fix did work, the symbol doctor is now correctly detecting that it needs to open Xcode:
Screenshot 2023-01-26 at 10 33 32 AM

However just opening Xcode doesn't seem to be enough for the first time case https://github.com/flutter/flutter/issues/118924

christopherfujino:
Yeah, I believe https://github.com/flutter/flutter/issues/118924 is tracking the remaining work to be done (i.e. running `xcrun xcodebuild -runFirstLaunch` to dismiss the new dialog). It's confusing as the two issues surfaced at the same time.

jmagman:
> It's confusing as the two issues surfaced at the same time.

I don't think it's a coincidence, because the ios_debug_symbol_doctor wasn't previously infra failing, it was getting past that point for the tests to run, and they flaked but often eventually sometimes passed.

PR #2391, opened 2023-01-13
[scheduler] Add build+test scheduling

CaseyHillers:
This is a draft to estimate the work needed if we wanted to implement a build+test scheduler.

- [x] Postsubmit processor
- [x] Presubmit processor
- [x] Nice to have: Refactor the common code into a library

Fixes https://github.com/flutter/flutter/issues/119036

PR #2455, opened 2023-02-06
[scheduler] Remove _targetsToSchedule filter

CaseyHillers:
* This filters so presubmits are not double scheduled, but it's never filtering anything
  * It adds extra RPCs as it verifies every target has a scheduled build
* Unblocks the build+test PR in https://github.com/flutter/cocoon/pull/2391


app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -178,52 +177,10 @@ class LuciBuildService {
...
-
-    return targets;
-  }
-
-  /// List of targets that should be scheduled.
-  ///
-  /// If a [Target] has a [Build] that is already scheduled or was successful, it shouldn't be scheduled again.
-  ///
-  /// Throws [InternalServerError] is [targets] is empty.
-  Future<List<Target>> _targetsToSchedule(List<Target> targets, github.PullRequest pullRequest) async {
```
CaseyHillers, line 199: This is the function this PR removes, and updates the remaining tests.

ricardoamador (approved):
LGTM

PR #2471, opened 2023-02-09
[scheduler] Add vacuum stale tasks

CaseyHillers:
The build dashboard is getting stuck with stale tasks. Add a vacuum handler to get them to eventually run.

### DIFF
--- a/app_dart/lib/src/request_handlers/scheduler/vacuum_stale_tasks.dart
+++ b/app_dart/lib/src/request_handlers/scheduler/vacuum_stale_tasks.dart
@@ -1,0 +1,85 @@
+// Copyright 2019 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:cocoon_service/cocoon_service.dart';
+import 'package:github/github.dart' as gh;
+import 'package:meta/meta.dart';
+
+import '../../model/appengine/task.dart';
+import '../../service/datastore.dart';
+import '../../service/logging.dart';
+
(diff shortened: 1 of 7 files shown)


app_dart/lib/src/request_handlers/scheduler/vacuum_stale_tasks.dart
```diff
@@ -0,0 +1,85 @@
...
+    @visibleForTesting this.datastoreProvider = DatastoreService.defaultProvider,
+    @visibleForTesting this.nowValue,
+  });
+
+  final DatastoreServiceProvider datastoreProvider;
+
+  /// For testing, can be used to inject a deterministic time.
+  final DateTime? nowValue;
+
+  static const int kCommitLimit = 25;
```
nehalvpatel, line 33: It looks like `kCommitLimit` is not being used anywhere yet?

CaseyHillers: Removed! This will instead rely on the datastore.dart limits

nehalvpatel: nit: I think its hard to grok that `queryRecentTasks` has a particular limit but otherwise LGTM!

nehalvpatel (approved):

app_dart/lib/src/request_handlers/scheduler/vacuum_stale_tasks.dart
CaseyHillers: From here, yes. However, we want all datastore logic contained there so it's easier make changes. It's difficult to trace datastore issues when they're scattered throughout the codebase.

PR #2472, opened 2023-02-09
[scheduler] Do not update postsubmit bringup targets in check runs

CaseyHillers:
https://github.com/flutter/flutter/issues/120331

### DIFF
--- a/app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
+++ b/app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
@@ -4,7 +4,6 @@
 
 import 'package:cocoon_service/ci_yaml.dart';
 import 'package:gcloud/db.dart';
-import 'package:github/github.dart';
 import 'package:meta/meta.dart';
 
 import '../model/appengine/commit.dart';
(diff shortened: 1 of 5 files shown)


app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
```diff
@@ -52,22 +51,6 @@ class PostsubmitLuciSubscription extends SubscriptionHandler {
       return Body.empty;
     }
 
-    if (buildPushMessage.userData.containsKey('repo_owner') && buildPushMessage.userData.containsKey('repo_name')) {
```
drewroengoogle, line 55: For my understanding, could you explain what the removal of this section does?

CaseyHillers: This seems to be a duplicate check of below. We already have the commit slug below via commit_key.

drewroengoogle:
Besides the one question, LGTM!

drewroengoogle (approved):

ricardoamador (approved):
LGTM

PR #2483, opened 2023-02-13
[app_dart] Fix stale tasks timestamp encoding

CaseyHillers:
Follow up to https://github.com/flutter/cocoon/pull/2471

This column can't be set to null.

```
Error while encoding entity (Bad state: Property validation failed for property createTimestamp while trying to serialize entity of kind Task. , #0 
```


ricardoamador (approved):
LGTM

PR #2484, opened 2023-02-14
Log dialog buttons when unexpected one found

jmagman:
When dialog box is showing and the buttons aren't handled by the dialog dismisser, print the app tree (text and buttons).


device_doctor/tool/infra-dialog/infra-dialog.xcodeproj/project.pbxproj
```diff
@@ -249,6 +249,7 @@
 				CODE_SIGN_STYLE = Automatic;
 				DEVELOPMENT_TEAM = S8QB4VV633;
 				INFOPLIST_FILE = "infra-dialogUITests/Info.plist";
+				IPHONEOS_DEPLOYMENT_TARGET = 14.0;
```
jmagman, line 252: `Logger` API is on iOS 14 and higher, we are running this on iOS 16 so just bump the minimum.

ricardoamador (approved):
LGTM

Sorry did not see this earlier. It somehow went to my promotions tabs in gmail.

device_doctor/tool/infra-dialog/infra-dialogUITests/infra_dialogUITests.swift
```diff
@@ -31,9 +33,15 @@ class infra_dialogUITests: XCTestCase {
             for buttonText in buttonTexts {
                 let button = springboard.buttons[buttonText]
                 if button.exists {
+                    defaultLog.log("Found button \(buttonText, privacy: .public)")
+                    defaultLog.log("\(springboard.debugDescription, privacy: .public)")
```
jmagman, line 37: Just added this as well, let's log if we actually find a button.

PR #2505, opened 2023-02-27
[app_dart] Skip postsubmit checks on framework

CaseyHillers:
Mitigation for https://github.com/flutter/flutter/issues/120395

Long term, we should migrate this to graphql to handle the framework load.

### DIFF
--- a/app_dart/lib/src/service/luci_build_service.dart
+++ b/app_dart/lib/src/service/luci_build_service.dart
@@ -574,6 +574,10 @@ class LuciBuildService {
     Target target,
     Map<String, dynamic> rawUserData,
   ) async {
+    if (commit.slug == Config.flutterSlug) {
+      // Framework has too many targets, and will get throttled.
+      return;
+    }
     final github.CheckRun checkRun = await githubChecksUtil.createCheckRun(
       config,
       target.slug,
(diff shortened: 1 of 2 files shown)


CaseyHillers:
I suspect https://cs.opensource.google/flutter/cocoon/+/main:app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart will still maintain these checks. If that works, we can probably remove this logic entirely and instead rely on the build message processors.

ricardoamador:
Just fyi that you get far less requests per hour with graphql so long term that won't really help here. You get 1/3 the request rate. One thing that we should try first is sending the client_id of the autosubmit_app.

CaseyHillers:
> Just fyi that you get far less requests per hour with graphql so long term that won't really help here. You get 1/3 the request rate. One thing that we should try first is sending the client_id of the autosubmit_app.

That won't work as then the check runs will be distributed across different apps (they're prefixed with the GitHub app name, which is why Cirrus and flutter dashboard can have the same check names).

A proper fix would be to move the GitHub check logic to a pull subscription so we can throttle ourselves.

ricardoamador:
> That won't work as then the check runs will be distributed across different apps

I am not sure I follow. Why would adding the client id not work/help? It is fixed to the autosubmit app.

Another option could be to throttle off of the rate limit headers. It would be good to combine these two ideas. Read the header and pull based on that information. We did in fact discuss this approach in the morning infra meeting. :)

With this change are you removing checks from post submit? Does that mean no checks will run after the PR is merged?

ricardoamador (approved):
LGTM

CaseyHillers:
> We did in fact discuss this approach in the morning infra meeting. :)

Please keep the GitHub issues up to date. Meetings are great, but we're an open source project.

I'll close this out since you already have a proper fix in mind.

PR #2514, opened 2023-03-06
Guarantee schedule packages repo and optimize guaranteed policy logic

keyonghan:
Old logic handles the only guaranteed repo `engine` as a special case. This PR standardize the logic:
1) adds a guaranteed schedule repo set
2) guarantee schedule `packages` repo


app_dart/lib/src/service/config.dart
```diff
@@ -49,6 +49,12 @@ class Config {
         pluginsSlug,
       };
 
+  /// List of guaranteed scheduling Github repos.
+  static Set<gh.RepositorySlug> get guaranteedSchedulingRepos => <gh.RepositorySlug>{
```
CaseyHillers, line 53: If we add more repos, we should consider making this a list of batch scheduling repos. I suspect most repos will want guaranteed scheduling over batch scheduling.

CaseyHillers (approved):

PR #2515, opened 2023-03-06
Disable vaccum task job

keyonghan:
Part of https://github.com/flutter/flutter/issues/121989. 
See comment: https://github.com/flutter/flutter/issues/121989#issuecomment-1457104096

### DIFF
--- a/cron.yaml
+++ b/cron.yaml
@@ -7,9 +7,12 @@ cron:
   url: /api/vacuum-github-commits
   schedule: every 10 minutes
 
-- description: vacuum stale tasks
-  url: /api/scheduler/vacuum-stale-tasks
-  schedule: every 10 minutes
+# Disabled for https://github.com/flutter/flutter/issues/121989
+# TODO(keyonghan): will delete if `In Progress` hanging issue is resolved:
+# https://github.com/flutter/flutter/issues/120395#issuecomment-1444810718
+#- description: vacuum stale tasks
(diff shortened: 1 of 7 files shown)


cron.yaml
```diff
@@ -7,9 +7,12 @@ cron:
   url: /api/vacuum-github-commits
   schedule: every 10 minutes
 
-- description: vacuum stale tasks
-  url: /api/scheduler/vacuum-stale-tasks
-  schedule: every 10 minutes
+# Disabled for https://github.com/flutter/flutter/issues/121989
```
keyonghan, line 10: These are the related code changes. All others are format changes.

godofredoc (approved):

CaseyHillers:
FYI, this seems to have gotten us back to the state we were in originally. No benchmarks are being run anymore as some tasks will get stuck, blocking the backfiller.

image

keyonghan:
> FYI, this seems to have gotten us back to the bad state we were in originally. No benchmarks are being run anymore as some tasks will get stuck, blocking the backfiller.
> 
> image

That bad row happened days ago. Several potentially related issues have been fixed and merged these two days, and more logs are in as well to help triage. 
I would like to use https://github.com/flutter/flutter/issues/122117 to track the root cause. If this still severely affects our tree and we haven't found a root fix sooner, we can bring this task back, but maybe with a lower running frequency.

PR #2519, opened 2023-03-08
Update CI_YAML.MD regarding taget updates

keyonghan:
Following https://github.com/flutter/flutter/pull/120912#discussion_r1128793254 to make doc clear.

### DIFF
--- a/CI_YAML.md
+++ b/CI_YAML.md
@@ -171,7 +171,10 @@ tags: >
     - Send PR, wait for the checks to go green (**the change takes effect on both presubmit and postsubmit as cocoon scheduling**
     **fetches latest change and applies it to new builds immediately**)
 3. If the check is red, add patches to get it green
-4. Once the PR has landed, infrastructure may take 1 or 2 commits to apply the latest properties for incoming PRs to see them.
+4. Once the PR has landed, infrastructure may take 1 or 2 commits to apply the latest properties
+   1. PRs/commits that have rebased on the changing PR do not need to wait
+   2. PRs/commits that have not rebased on the changing PR need to wait
+   3. Local LED runs need to wait
 
 **Note:** updates on other entries except `properties` will not take effect immediately. Ths PR needs
 to be landed first to wait for changes propagated in infrastructure.
--- a/CI_YAML.md
+++ b/CI_YAML.md
@@ -168,9 +168,10 @@ tags: >
           ]
         timeout: 60
       ```
-    - Send PR, wait for the checks to go green (**the change takes effect on both presubmit and postsubmit**)
+    - Send PR, wait for the checks to go green (**the change takes effect on both presubmit and postsubmit as cocoon scheduling**
+    **fetches latest change and applies it to new builds immediately**)
 3. If the check is red, add patches to get it green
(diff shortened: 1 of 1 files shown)


CI_YAML.md
```diff
@@ -165,11 +167,14 @@ tags: >
               {"dependency": "open_jdk", "version": "11"}
           ]
         timeout: 60
-    ```
-    - Send PR, wait for the checks to go green (the change takes effect on presubmit)
+      ```
+    - Send PR, wait for the checks to go green (**the change takes effect on presubmit**)
 3. If the check is red, add patches to get it green
 4. Once the PR has landed, infrastructure may take 1 or 2 commits to apply the latest properties
```
stuartmorgan, line 173: I'm confused; it takes effect immediately for presubmit, but not for postsubmit? So property changes still can't be landed in the same PR as changes that require them (like removing bring up)?

Or is this not the case for `properties`?

stuartmorgan: So item 4 here should be removed, right?

keyonghan: We still need it. The `properties` will be seen by the current PR/commit immediately, but there are cases where we need to wait, such incoming PRs/commits, which haven't rebased on top of that changing PR, and local LED runs.

Expanded a bit. Please let me know if this makes sense to you.

stuartmorgan (approved):
LGTM

I don't understand why something that hasn't rebased would need to "wait 1 or 2 commits" rather than having to rebase, but I'll assume that's me not understanding the infrastructure.

keyonghan:
> LGTM
> 
> I don't understand why something that hasn't rebased would need to "wait 1 or 2 commits" rather than having to rebase, but I'll assume that's me not understanding the infrastructure.

Suppose we have a property change PR (pr1) and an incoming PR (pr2):
Note that cocoon schedules builds based on the current commit `sha`.
If a PR (pr2) hasn't rebased, cocoon will use pr2's target properties (without changes from pr1) on top of current infrastructure configs (pr1's changes either propagated or not) to schedule. If the pr1's changes haven't been propagated, they will not be reflected in pr2's scheduled builds. Yeah, this is some backend infra logics:-)

stuartmorgan:
> on top of current infrastructure configs (pr1's changes either propagated or not)

Wait, so does the infra behavior distinguish between "the state of .ci.yaml in the PR" and "the *changes* to .ci.yaml in the PR". I.e., looking at deltas in the PR specifically?

keyonghan:
> > on top of current infrastructure configs (pr1's changes either propagated or not)
> 
> Wait, so does the infra behavior distinguish between "the state of .ci.yaml in the PR" and "the _changes_ to .ci.yaml in the PR". I.e., looking at deltas in the PR specifically?

That's correct. If no changes in `.ci.yaml`, cocoon will trigger builds with whatever current infra configs are. If any changes, cocoon will fetch them and override corresponding values in existing infra configs. That's why the changes (properties) in the PR could be reflected in the current PR right away.

CI_YAML.md
CaseyHillers: Another case is caches. While dependencies can be upgraded, there may be some slow downs on the initial roll out due to cache invalidation

keyonghan: Added one more item to cover.

PR #2523, opened 2023-03-09
Limit postsubmit checkrun update logic to `postsubmitSupportedRepos` only

keyonghan:
This is a follow up of https://github.com/flutter/cocoon/pull/2516, which disabled postsubmit `checkrun` creation on other repos. This PR disables postsumbmit `checkrun` update on other repos.

Fixes: https://github.com/flutter/flutter/issues/122314

### DIFF
--- a/app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
+++ b/app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
@@ -99,7 +99,7 @@ class PostsubmitLuciSubscription extends SubscriptionHandler {
     }
 
     // Only update GitHub checks if target is not bringup
-    if (target.value.bringup == false) {
+    if (target.value.bringup == false && config.postsubmitSupportedRepos.contains(target.slug)) {
       log.info('Updating check status for ${target.getTestName}');
       await githubChecksService.updateCheckStatus(
         buildPushMessage,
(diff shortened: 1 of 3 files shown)


app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
```diff
@@ -99,7 +99,7 @@ class PostsubmitLuciSubscription extends SubscriptionHandler {
     }
 
     // Only update GitHub checks if target is not bringup
-    if (target.value.bringup == false) {
+    if (target.value.bringup == false && config.postsubmitSupportedRepos.contains(target.slug)) {
```
yusuf-goog, line 102: where is it restricting to `packages` only?

ricardoamador: https://github.com/flutter/cocoon/blob/main/app_dart/lib/src/service/config.dart#L61

ricardoamador: There are only two supported repos in that postsubmitSupportedRepos list.

ricardoamador (approved):
LGTM

app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
yusuf-goog: Then is this wrong? description says to limit to packages only.

keyonghan: Now with `plugins` repo merged to `packages` repo, the `postsubmitSupportedRepos` is actually only `packages`. We need a clean up. Created https://github.com/flutter/flutter/issues/122334 to track.

Also updated title to avoid confusion.

yusuf-goog (approved):

PR #2524, opened 2023-03-09
Wait all futures complete before return when there are errors in `batch backfiller`

keyonghan:
Batch backfiller API now bundles all suitable back filling targets in a Future list, and exits with error whenever an error happens in any Future (with `eagerError: true`). This has caused hanging yellow boxes without builds scheduled (https://github.com/flutter/flutter/issues/122117).

This PR disables `eagerError`, this way all other Futures will continue even there is any error occurring.

### DIFF
--- a/app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
+++ b/app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
@@ -92,7 +92,7 @@ class BatchBackfiller extends RequestHandler {
       });
       // Schedule all builds asynchronously.
       // Schedule after db updates to avoid duplicate scheduling when db update fails.
-      await Future.wait<void>(backfillRequestList(backfill), eagerError: true);
+      await Future.wait<void>(backfillRequestList(backfill));
     } catch (error) {
       log.severe('Failed to update tasks when backfilling: $error');
     }


app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
```diff
@@ -92,7 +92,7 @@ class BatchBackfiller extends RequestHandler {
       });
       // Schedule all builds asynchronously.
       // Schedule after db updates to avoid duplicate scheduling when db update fails.
-      await Future.wait<void>(backfillRequestList(backfill), eagerError: true);
+      await Future.wait<void>(backfillRequestList(backfill));
```
CaseyHillers, line 95: @christopherfujino is much more familiar with better handling of future errors, but I think you want something like:

```dart
final errors = [];
await Future.wait<void>(backfillRequestList(backfill).onError((e) => errors.add(e));
if (errors.isNotEmpty) {
  log.severe(errors);
}
```

Your original approach will consume all errorrs, and the catch will never be run.

keyonghan:
@CaseyHillers I see you added this flag in https://github.com/flutter/cocoon/pull/2455. Please correct me if this is used for something else.

app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
keyonghan: Based on doc, only the first error will be returned and it should trigger the catch?
```
If any future completes with an error, then the returned future completes with that error. If further futures also complete with errors, those errors are discarded.
```
But it would be good to catch all errors.

christopherfujino: I think the issue here is, if we're not `await`-ing `backfillRequestList`, will we even hit the `catch` clause? Let me try to test this out to confirm...

christopherfujino: Tested this with:

```dart
Future<void> main() async {
  try {
    await Future.wait(_getFutures());
  } catch (err) {
    print('caught $err');
  }
  print('done');
}

List<Future<void>> _getFutures() => <Future<void>>[
  Future<void>.delayed(const Duration(milliseconds: 40), () => throw Exception('error 1')),
  Future<void>.delayed(const Duration(milliseconds: 80), () => throw Exception('error 2')),
  Future<void>.delayed(const Duration(milliseconds: 300), () => print('finished third future')),
];
```

And it logged the expected:

```
finished third future
caught Exception: error 1
done
```

So this LGTM

christopherfujino (approved):

app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
christopherfujino: And good catch! This was a subtle bug.

keyonghan: Thanks for the validation @christopherfujino !

PR #2529, opened 2023-03-13
Enforce `backfillerTargetLimit` when backfilling

keyonghan:
This is an attempt to fix issue https://github.com/flutter/flutter/issues/122117 by limiting the number of back filler targets, which limits the number of pub/sub requests in each API.


ricardoamador (approved):
LGTM

PR #2532, opened 2023-03-14
Add licenses to Dart projects

CaseyHillers:
* Remove code comment blocks from LICENSE
* Make all LICENSEs in Cocoon consistent
  * Add check verifying all match the golden LICENSE at root


keyonghan:
Is a LICENSE file expected for sub dirs of a project considering it's already under root? /cc @godofredoc @Hixie

CaseyHillers:
> Is a LICENSE file expected for sub dirs of a project considering it's already under root? /cc @godofredoc @Hixie

That's a good question, and IANAL. What I observed is flutter/packages includes it individually, and some of our newer projects seem to have their own copy in the directory (codesign and autosubmit). I sent this as it's easier than writing a custom copybara workflow

CaseyHillers:
@keyonghan I updated the LICENSE files. If there's interest, I can do a follow up PR to make all the LICENSE headers in the code files have the same year.

keyonghan (approved):
LGTM

PR #2543, opened 2023-03-17
[app_dart] consider all files under flutter_tools/test to be tests

christopherfujino:
Fixes https://github.com/flutter/flutter/issues/122924

For an example, see the PR https://github.com/flutter/flutter/pull/122856/files, which updated packages/flutter_tools/test/integration.shard/test_utils.dart


christopherfujino:
@jmagman do you remember if there was a reason we didn't do this before?

jmagman (approved):
> @jmagman do you remember if there was a reason we didn't do this before?

No, seems like a good idea though. I guess files in this directory are usually caught by the `_test.dart` part.

jmagman (commented):
Formatter isn't happy.

christopherfujino:
argh, didn't run dartfmt

PR #2547, opened 2023-03-20
Add autosubmit configuration for configurable autosubmit.

ricardoamador:
Add the configuration file for configurable autosubmit. This will not do anything at the moment unless the configurable autosubmit is deployed to flutter-dashboard.

I plan on deploying a test instance after hours just in case to validate since I am hitting issues  with the list-teams api.

## Pre-launch Checklist

- [x] I signed the [CLA].


.github/autosubmit.yml
```diff
@@ -0,0 +1,16 @@
+default_branch: 'main'
```
CaseyHillers, line 1: In yaml, main is fine. None of these need to be string encapsulated as it's implicit

.github/autosubmit.yml
```diff
@@ -0,0 +1,16 @@
...
+  - 'engine-flutter-autoroll'
+  - 'dependabot[bot]'
+  - 'dependabot'
+  - 'DartDevtoolWorkflowBot'
+approving_reviews: 2
+approval_group: 'flutter-hackers'
+run_ci: true
+support_no_review_revert: true
+required_checkruns:
+  - 'ci.yaml validation'
```
CaseyHillers, line 16: Flutter style guide is to include new line before EOF

.github/autosubmit.yml
```diff
@@ -0,0 +1,16 @@
+default_branch: 'main'
+issues_repository:
+  owner: 'flutter'
+  repo: 'flutter'
+auto_approval_accounts:
+  - 'skia-flutter-autoroll'
```
CaseyHillers, line 6: Are all of these bots needed for Cocoon?

.github/autosubmit.yml
```diff
@@ -0,0 +1,16 @@
+default_branch: 'main'
+issues_repository:
+  owner: 'flutter'
+  repo: 'flutter'
+auto_approval_accounts:
```
godofredoc, line 5: Do we need to split this config into two different places? depending on how this file will be read there are several dangerous values:

Assuming the file will be read from the PR commit:

* I can change the auto_approval_accounts to include my username and bypass approval
* I can change the # of approvals to 0 to bypass codereview
* I can change run_ci or required_checkruns to land immediately

Assuming the file will be read from main ToT:

* This will require adding owners approval to the autosubmit.yml to ensure someone who understands the implications of the values reviews changes to it, although it can still be error prone.

Maybe we can read the file from .github repository instead. This repo is usually well protected because it contains org level configurations and only allow to override non dangerous values per repository.

.github/autosubmit.yml
ricardoamador: That I am not sure about. I will check.

.github/autosubmit.yml
ricardoamador: Yeah I thought of this as well. Right now I just want to get it running as there are quite a few changes that needed to come in order to make this work. But as far as your options I am definitely not reading from the PR commit. There are several reasons why this would not be beneficial for us. I chose to always read from ToT to reduce network calls and to utilize caching. I am fine reading from .github repo as that is the same amount of network calls thought it does prevent other teams from adopting more quickly and would require them to have access to that repo correct?

godofredoc: Not necessarily, usually .github is readonly for org members and write access to a few individuals. Everybody can send PRs but only the people with write access can +2 and submit.

Let's give it a try, can you try to create a PR in https://github.com/flutter/.github?

ricardoamador: Yeah let me try that.

.github/autosubmit.yml
keyonghan: We need only `dependabot` for cocoon.

PR #2548, opened 2023-03-21
Skip tests check for `dependabot[bot]`

keyonghan:
Some PRs from roller `dependabot` are with `user.login` as `dependabot[bot]` in webhook payload. This PR covers it to skip tests check for repos (packages, engine, framework).

Fixes: https://github.com/flutter/flutter/issues/108129

### DIFF
--- a/app_dart/lib/src/service/config.dart
+++ b/app_dart/lib/src/service/config.dart
@@ -317,6 +317,7 @@ class Config {
         'skia-flutter-autoroll',
         'engine-flutter-autoroll',
         'dependabot',
+        'dependabot[bot]',
       };
 
   Future<String> generateJsonWebToken() async {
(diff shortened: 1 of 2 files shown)


app_dart/lib/src/service/config.dart
```diff
@@ -317,6 +317,7 @@ class Config {
         'skia-flutter-autoroll',
         'engine-flutter-autoroll',
         'dependabot',
+        'dependabot[bot]',
```
godofredoc, line 320: Do we need to remove the one above?

stuartmorgan (approved):
LGTM

app_dart/lib/src/service/config.dart
keyonghan: We do have earlier PRs based on `dependabot` which didn't complain about tests: https://github.com/flutter/engine/pull/40447

PR #2550, opened 2023-03-22
add new error message pattern to ios_debug_symbol_doctor

christopherfujino:
Related to https://github.com/flutter/flutter/issues/123257

Treat messages that look like:

`! Error: iPhone 11 is busy: Preparing iPhone 11 for development. Xcode will continue when iPhone 11 is finished. (code -10)`

As debug symbol errors that should be recovered by opening Xcode.


yusuf-goog (approved):

yusuf-goog:
Should we bother trying to make that regex even more general to try and future proof?

jmagman:
> > Should we bother trying to make that regex even more general to try and future proof?
> 
> As I recall, there are some types of annoying errors that we don't care about, is that right @jmagman ?

Yeah very likely.  Could check for the `-10` code instead.

PR #2568, opened 2023-03-24
Bump batch backfiller commit row limit

keyonghan:
Fixes: https://github.com/flutter/flutter/issues/123405

### DIFF
--- a/app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
+++ b/app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
@@ -49,7 +49,8 @@ class BatchBackfiller extends RequestHandler {
 
   Future<void> backfillRepository(RepositorySlug slug) async {
     final DatastoreService datastore = datastoreProvider(config.db);
-    final List<FullTask> tasks = await (datastore.queryRecentTasks(slug: slug)).toList();
+    final List<FullTask> tasks =
+        await (datastore.queryRecentTasks(slug: slug, commitLimit: config.backfillerCommitLimit)).toList();
 
     // Construct Task columns to scan for backfilling
     final Map<String, List<FullTask>> taskMap = <String, List<FullTask>>{};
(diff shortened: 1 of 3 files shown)


app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
```diff
@@ -49,7 +49,8 @@ class BatchBackfiller extends RequestHandler {
 
   Future<void> backfillRepository(RepositorySlug slug) async {
     final DatastoreService datastore = datastoreProvider(config.db);
-    final List<FullTask> tasks = await (datastore.queryRecentTasks(slug: slug)).toList();
+    final List<FullTask> tasks =
+        await (datastore.queryRecentTasks(slug: slug, commitLimit: config.backfillerCommitLimit)).toList();
```
ricardoamador, line 53: I see that you are passing the limit in so was the limit hardcoded before this change?

app_dart/lib/src/service/config.dart
```diff
@@ -154,6 +154,9 @@ class Config {
   /// Upper limit of targets to be backfilled in API call.
   int get backfillerTargetLimit => 50;
```
ricardoamador, line 155: What is the difference between this constant and the new one that was added?

keyonghan: `backfillerTargetLimit` is the number limit of targets to be call in each API call. Most of the cases, we have hundreds of targets to schedule, but enforce this limit to avoid potential backend errors.

The new `backfillerCommitLimit` is the number of commit rows to scan (each row contains the to-be-scheduled targets). From task execution point of view we hope to scan as many as possible when bots are idle. But it would be a waste of resource to check say 1000 rows each time. 50 should be a big enough number to start with.

Updated the doc to explain.

ricardoamador: Ah okay. Thanks for adding the doc.

PR #2589, opened 2023-04-05
Use cache + retry when batch backfilling to mitigate server errors

keyonghan:
This PR is to mitigate https://github.com/flutter/flutter/issues/124018

Changes include:
1) Cache builder list when calling the buildbucket API. This should help with 
  - Failed to update tasks when backfilling: BuildBucketException: [502]
  - Failed to update tasks when backfilling: BuildBucketException: [504]: Request deadline exceeded
  - Failed to update tasks when backfilling: Connection closed before full header was received
2) Retry when pub/sub error happens. This should help with the `DetailedApiRequestError`.

### DIFF
--- a/app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
+++ b/app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
@@ -108,6 +108,7 @@ class BatchBackfiller extends RequestHandler {
     }
   }
 
+  /// Schedules tasks with retry when hitting pub/sub server errors.
   Future<void> _scheduleWithRetries(List<Tuple<Target, FullTask, int>> backfill) async {
     const RetryOptions retryOptions = Config.schedulerRetry;
     try {
@@ -127,6 +128,11 @@ class BatchBackfiller extends RequestHandler {
     }
   }
 
+  /// Updates the [backfill] list with those that fail to get scheduled.
+  ///
+  /// [tupleLists] maintains the same tuple order as those in [backfill].
+  /// Each element from [backfill] is encapsulated as a list in [tupleLists] to prepare for
+  /// [scheduler.luciBuildService.schedulePostsubmitBuilds].
   _updateBackfill(List<Tuple<Target, FullTask, int>> backfill, List<List<Tuple<Target, Task, int>>> tupleLists) {
     final List<Tuple<Target, FullTask, int>> updatedBackfill = <Tuple<Target, FullTask, int>>[];
     for (int i = 0; i < tupleLists.length; i++) {
--- a/app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
+++ b/app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
@@ -133,7 +133,8 @@ class BatchBackfiller extends RequestHandler {
   /// [tupleLists] maintains the same tuple order as those in [backfill].
   /// Each element from [backfill] is encapsulated as a list in [tupleLists] to prepare for
   /// [scheduler.luciBuildService.schedulePostsubmitBuilds].
-  _updateBackfill(List<Tuple<Target, FullTask, int>> backfill, List<List<Tuple<Target, Task, int>>> tupleLists) {
+  List<Tuple<Target, FullTask, int>> _updateBackfill(
+      List<Tuple<Target, FullTask, int>> backfill, List<List<Tuple<Target, Task, int>>> tupleLists) {
     final List<Tuple<Target, FullTask, int>> updatedBackfill = <Tuple<Target, FullTask, int>>[];
     for (int i = 0; i < tupleLists.length; i++) {
       if (tupleLists[i].isNotEmpty) {
--- a/app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
+++ b/app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
@@ -10,10 +10,13 @@ import 'package:cocoon_service/src/service/scheduler/policy.dart';
 import 'package:gcloud/db.dart';
 import 'package:github/github.dart';
 import 'package:meta/meta.dart';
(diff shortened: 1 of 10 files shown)


drewroengoogle (approved):
LGTM!

app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
```diff
@@ -99,15 +102,50 @@ class BatchBackfiller extends RequestHandler {
...
+      log.severe('Failed to backfill ${backfill.length} targets.');
+    }
+  }
+
+  /// Updates the [backfill] list with those that fail to get scheduled.
+  ///
+  /// [tupleLists] maintains the same tuple order as those in [backfill].
+  /// Each element from [backfill] is encapsulated as a list in [tupleLists] to prepare for
+  /// [scheduler.luciBuildService.schedulePostsubmitBuilds].
+  _updateBackfill(List<Tuple<Target, FullTask, int>> backfill, List<List<Tuple<Target, Task, int>>> tupleLists) {
```
ricardoamador, line 136: I did not know you could omit the return type in dart like this.

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -53,6 +58,9 @@ class LuciBuildService {
   static const int kDefaultPriority = 30;
   static const int kRerunPriority = 29;
 
+  /// Name of the subcache to store luci build related values in redis.
+  static const String subCacheName = 'luci';
```
ricardoamador, line 62: It looks like the cache size for in-memory is 256. I am not sure if that is sub-caches or if this is total elements. How much will doing this affect the cache size?

ricardoamador (approved):
LGTM with some questions.

app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
```diff
@@ -99,15 +102,50 @@ class BatchBackfiller extends RequestHandler {
...
       // Schedule after db updates to avoid duplicate scheduling when db update fails.
-      await Future.wait<void>(backfillRequestList(backfill));
+      await _scheduleWithRetries(backfill);
     } catch (error) {
       log.severe('Failed to update tasks when backfilling: $error');
     }
   }
 
+  /// Schedules tasks with retry when hitting pub/sub server errors.
+  Future<void> _scheduleWithRetries(List<Tuple<Target, FullTask, int>> backfill) async {
```
ricardoamador, line 112: Do you think this will have an impact on the secondary rate limit with github? I worry about us just retrying everything as we hit these limits regularly.

app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
```diff
@@ -99,15 +102,50 @@ class BatchBackfiller extends RequestHandler {
...
+    const RetryOptions retryOptions = Config.schedulerRetry;
+    try {
+      await retryOptions.retry(
```
ricardoamador, line 115: If we begin a backfill and get 90 percent through it where some were successful but then we hit a pub/sub error will it reattempt the successful tasks that were scheduled already?

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -398,10 +406,31 @@ class LuciBuildService {
...
+  ///
+  /// Returns cache if existing. Otherwise make the RPC call to fetch list.
   Future<Set<String>> getAvailableBuilderSet({
     String project = 'flutter',
     String bucket = 'prod',
+  }) async {
+    final Uint8List? cacheValue = await cache.getOrCreate(
+      subCacheName,
+      'builderlist',
+      createFn: () => _getAvailableBuilderSet(project: project, bucket: bucket),
```
ricardoamador, line 419: Please correct me if i am wrong but the builder set can be a fairly large list, are there size limits to entries in the cache?

app_dart/lib/src/service/luci_build_service.dart
keyonghan: Are you referring to `inMemoryMaxNumberEntries`? This is for `inMemory = true`. For our production server, they are using redis cache. Also this number is for number of entries, and we are consuming one entry for this purpose. Even with inMemory, we should have enough entries left.

app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
keyonghan: Good catch. This is not a good style. Added the return type.

app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
keyonghan: From `backfillRequestList`, you can see we are handling each task as a separate request. This means the failed request affects only the included target itself. So we do not have issue scheduling duplicated successful tasks.

app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
keyonghan: This retry doesn't communicate with GitHub at all.
Though there is some check run logic in the scheduler, they are meant only for packages repo, which will not be called in the batch_backfiller api. This API handles only repo that is using batch scheduling logic, whereas packages is using `guaranteed` policy.
Additionally, this retries only tasks that fail to get scheduled. all happen in framework repo which will not touch the check run API.

app_dart/lib/src/service/luci_build_service.dart
keyonghan: I see only a limit on the cache entries rather than a specific cache size.. For this question, I am not 100% sure. Though the list is big, the size is several kb I guess? I would expect redis handles this with no problem.. But needs to validate. I will monitor it closely tomorrow morning if this lands today.

ricardoamador:
Thanks for answering my questions. Send it!

PR #2594, opened 2023-04-06
[codesign] white list for debug tools

XilaiZhang:
skip codesign for engine binaries that are debug symbols and not executable, which don't need to be codesigned or notarized.

Context: https://github.com/flutter/flutter/issues/124322

### DIFF
--- a/codesign/lib/src/file_codesign_visitor.dart
+++ b/codesign/lib/src/file_codesign_visitor.dart
@@ -64,6 +64,7 @@ class FileCodesignVisitor {
 
   Set<String> fileWithEntitlements = <String>{};
   Set<String> fileWithoutEntitlements = <String>{};
+  Set<String> fileWhitelist = <String>{};
   Set<String> fileConsumed = <String>{};
   Set<String> directoriesVisited = <String>{};
   Map<String, String> availablePasswords = {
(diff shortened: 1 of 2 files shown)


codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -71,6 +71,7 @@ class FileCodesignVisitor {
     'CODESIGN_TEAM_ID': '',
     'APP_SPECIFIC_PASSWORD': ''
   };
+  Set<String> whitelist = {'FlutterEmbedder.framework.zip', 'FlutterMacOS.dSYM.zip'};
```
godofredoc, line 74: I would prefer to have something closer to the artifact creation. Doing this way will mean that we need to update this tool every time a new artifact is added.

godofredoc: Maybe just logging that a file was not code signed because it didn't have code sign metadata?

godofredoc: @christopherfujino what do you think?

XilaiZhang: Ah i see. Good point! yeah need to change BUILD.gn to add another file

christopherfujino: Yeah, I think the build rules is the right place to add this.

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -64,6 +64,7 @@ class FileCodesignVisitor {
 
   Set<String> fileWithEntitlements = <String>{};
   Set<String> fileWithoutEntitlements = <String>{};
+  Set<String> fileWhitelist = <String>{};
```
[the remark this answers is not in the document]
XilaiZhang: umm when I googled the term whitelist it says its a computing word for "a list of people or things considered to be acceptable or trustworthy"

Anyways we will be abandoning this PR in favor of the namelist approach of zip files in recipe

XilaiZhang:
We will use namelist approach of zip files in recipe implemented by sir Godofredo

PR #2603, opened 2023-04-11
Extend cronjob to vacuum github commits to every 6 hours

keyonghan:
Part of https://github.com/flutter/flutter/issues/124603.


CaseyHillers (commented):
I left a comment on the linked issue to better understand why we're removing this first. You'll need to extend the log filter to the past year to see this is working as it's only used during GitHub outages.

keyonghan:
Thanks for the context. Updated the frequency to every 6 hours.

CaseyHillers (approved):

PR #2606, opened 2023-04-12
Adding an await to function definition.

ricardoamador:
I am investigating an issue related to this error: https://github.com/dart-lang/pub-dev/issues/3878
I have done some digging around the internets and found that we are possibly not waiting for a Future to complete which can lead to this issue. This is a stretch but I found some code that looks like it could be responsible in the cache service where we submit a createFn that runs if the value is not found.

The create function is not explicitly defined async when it is passed by value to the getOrCreate call. Though it seems that we do wait (await) when attempting to actually call it if the value does not exist in the cache_service. I am wondering if the change I made would have any effect on the call. 

Fixes https://github.com/flutter/flutter/issues/124613

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/app_dart/lib/src/request_handling/cache_request_handler.dart
+++ b/app_dart/lib/src/request_handling/cache_request_handler.dart
@@ -56,7 +56,7 @@ class CacheRequestHandler<T extends Body> extends RequestHandler<T> {
     final Uint8List? cachedResponse = await cache.getOrCreate(
       responseSubcacheName,
       responseKey,
-      createFn: () => getBodyBytesFromDelegate(delegate),
+      createFn: () async { return await getBodyBytesFromDelegate(delegate); },
       ttl: ttl,
     );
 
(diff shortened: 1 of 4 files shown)


app_dart/lib/src/request_handling/cache_request_handler.dart
```diff
@@ -56,7 +56,7 @@ class CacheRequestHandler<T extends Body> extends RequestHandler<T> {
     final Uint8List? cachedResponse = await cache.getOrCreate(
       responseSubcacheName,
       responseKey,
-      createFn: () => getBodyBytesFromDelegate(delegate),
+      createFn: () async { return await getBodyBytesFromDelegate(delegate); },
```
christopherfujino, line 59: In this example, the semantic meaning is the same, the anonymous function we're creating will immediately invoke `getBodyBytesFromDelegate()`, which will immediately return a future, and then the anonymous function will immediately return that Future, which will be completed at some time in the future.

ricardoamador: Okay so it does not need the explicit async and await keywords and the code that calls that function will allow the future to complete if it awaits the call?

Basically this is doing nothing correct?

CaseyHillers: Correct. The issue appears to be related directly to the package:neat_cache, and you may want to send more debugging PRs there.

I suspect you could write an integration test that spawns a redis instance, and spawn a 1000 threads to make operations against it to reproduce this.

ricardoamador: Cool! Thanks for clarifying. It seems that this might not be totally thread-safe.

PR #2612, opened 2023-04-12
[autosubmit] add PR body to commit message when auto-submitting

christopherfujino:
Issues:

* Fixes: https://github.com/flutter/flutter/issues/124742
* Related: https://github.com/flutter/flutter/issues/124554

https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#merge-a-pull-request

### DIFF
--- a/auto_submit/lib/service/validation_service.dart
+++ b/auto_submit/lib/service/validation_service.dart
@@ -334,7 +334,17 @@ Exception: ${exception.message}
     final github.RepositorySlug slug = messagePullRequest.base!.repo!.slug();
     final int number = messagePullRequest.number!;
     // Pass an explicit commit message from the PR title otherwise the GitHub API will use the first commit message.
-    final String commitMessage = _sanitizePrBody(messagePullRequest.body ?? '');
+    const String revertPattern = 'Revert "Revert';
+    String messagePrefix = '';
+    if (messagePullRequest.title!.contains(revertPattern)) {
+      // Cleanup auto-generated revert messages.
+      messagePrefix = '''
+${messagePullRequest.title!.replaceFirst('Revert "Revert', 'Reland')}
+
+''';
+    }
+    final String prBody = _sanitizePrBody(messagePullRequest.body ?? '');
+    final String commitMessage = '$messagePrefix$prBody';
 
     try {
       github.PullRequestMerge? result;
--- a/auto_submit/lib/service/validation_service.dart
+++ b/auto_submit/lib/service/validation_service.dart
@@ -7,7 +7,6 @@ import 'package:auto_submit/model/big_query_pull_request_record.dart';
 import 'package:auto_submit/model/big_query_revert_request_record.dart';
 import 'package:auto_submit/model/pull_request_change_type.dart';
 import 'dart:async';
-import 'dart:convert';
 
 import 'package:auto_submit/service/bigquery.dart';
 import 'package:auto_submit/service/config.dart';
@@ -335,13 +334,7 @@ Exception: ${exception.message}
     final github.RepositorySlug slug = messagePullRequest.base!.repo!.slug();
     final int number = messagePullRequest.number!;
     // Pass an explicit commit message from the PR title otherwise the GitHub API will use the first commit message.
-    final String prTitle =
-        messagePullRequest.title!.replaceFirst('Revert "Revert', 'Reland'); // Cleanup auto-generated revert messages.
-    final String commitMessage = _sanitizePrBody('''
-$prTitle
-
-${messagePullRequest.body ?? ''}
-''');
+    final String commitMessage = _sanitizePrBody(messagePullRequest.body ?? '');
 
     try {
       github.PullRequestMerge? result;
--- a/auto_submit/lib/service/validation_service.dart
+++ b/auto_submit/lib/service/validation_service.dart
@@ -7,6 +7,7 @@ import 'package:auto_submit/model/big_query_pull_request_record.dart';
 import 'package:auto_submit/model/big_query_revert_request_record.dart';
 import 'package:auto_submit/model/pull_request_change_type.dart';
 import 'dart:async';
+import 'dart:convert';
 
 import 'package:auto_submit/service/bigquery.dart';
 import 'package:auto_submit/service/config.dart';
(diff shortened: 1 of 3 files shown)


christopherfujino:
Should we cap this or otherwise filter it?

ricardoamador (approved):
LGTM

auto_submit/lib/service/validation_service.dart
```diff
@@ -334,8 +335,13 @@ Exception: ${exception.message}
     final github.RepositorySlug slug = messagePullRequest.base!.repo!.slug();
     final int number = messagePullRequest.number!;
     // Pass an explicit commit message from the PR title otherwise the GitHub API will use the first commit message.
-    String commitMessage = messagePullRequest.title!;
-    commitMessage = commitMessage.replaceFirst('Revert "Revert', 'Reland'); // Cleanup auto-generated revert messages.
+    final String prTitle =
+        messagePullRequest.title!.replaceFirst('Revert "Revert', 'Reland'); // Cleanup auto-generated revert messages.
```
Hixie, line 339: i understand this is not really part of your PR so feel free to ignore, but:

fwiw i would avoid "Reland" as well (also if i'm not misreading this code it'll cause a trailing quote mark).
would be ideal if the commit message was just the original one and the pr/commit description mentioned that it was a reland (ideally with a link to the revert PR).

auto_submit/test/service/validation_service_test.dart
```diff
@@ -691,7 +691,33 @@ void main() {
...
-      expect(result.message, 'Reland "My first PR!"');
+      expect(result.message, contains('Reland "My first PR!"'));
+    });
+
+    test('Includes PR description', () async {
+      final PullRequest pullRequest = generatePullRequest(
+        prNumber: 0,
+        repoName: slug.name,
+        title: 'PR title',
+        // This will be interpolated into JSON, so escape newlines
```
Hixie, line 702: can you elaborate on this comment?

Hixie (approved):
LGTM

auto_submit/test/service/validation_service_test.dart
christopherfujino: this is specifically for the test helper function `generatePullRequest()`.

christopherfujino: I can flesh out the comment

christopherfujino:
I'd like final approval from @CaseyHillers on if you think anything bad might happen if we send too long a commit message

auto_submit/test/service/validation_service_test.dart
christopherfujino: done

ricardoamador:
@christopherfujino this might be helpful: https://github.com/refined-github/refined-github/issues/2178 if you are specifically talking about the merge commit message.

christopherfujino:
> @christopherfujino this might be helpful: [refined-github/refined-github#2178](https://github.com/refined-github/refined-github/issues/2178) if you are specifically talking about the merge commit message.

nice

ditman:
@christopherfujino IMO this should also include the text of the commits that were squashed during the merge, like so:

```
Title of the PR (#ID_OF_THE_PR)

Full description of the PR

Commits:
* Commit message 1
* Commit message 2
* ...
* Commit message n
```

I'd like to have the commit texts so we don't lose the `Co-Authored-By` lines that github adds when a user sends a "suggestion" to a PR, or somebody else commits to a PR in progress, and there's multiple authors to a single PR.

I've left some examples of commits in my issue (which this PR would also fix! 🥳) here:

* https://github.com/flutter/flutter/issues/124554

auto_submit/lib/service/validation_service.dart
CaseyHillers: For context, I rely on our commit log daily to find good branch points without having to wait hours for tests.

Relands indicate to me that a commit is risky to rely on, and it saves me time to  avoid including it in the branch. Unfortunately, I need to read ~80 commits a day, so I'd prefer not to read the whole message.

auto_submit/lib/service/validation_service.dart
```diff
@@ -334,8 +335,13 @@ Exception: ${exception.message}
...
+    final String commitMessage = ('''
```
CaseyHillers, line 340: Should this start with a newline? I thought the first line is the commit title

```suggestion
    final String commitMessage = ('''$prTitle
```

auto_submit/lib/service/validation_service.dart
```diff
@@ -334,8 +335,13 @@ Exception: ${exception.message}
...
+$prTitle
+
+${messagePullRequest.body ?? ''}
```
CaseyHillers, line 343: Should we filter out the checkboxes in the default pr template?

auto_submit/lib/service/validation_service.dart
christopherfujino: If the first line of a multiline string consists solely of whitespace, it's ignored by the parser, including the newline: https://github.com/dart-lang/language/blob/master/specification/dartLangSpec.tex#L9112

auto_submit/lib/service/validation_service.dart
christopherfujino: considered this. rather than matching the full text (which would be brittle), what about skipping any line that starts with `- [ ]` or `- [x]`?

CaseyHillers: That SGTM, that should remove most of the noise

auto_submit/lib/service/validation_service.dart
christopherfujino: So, I think I found a compromise:

The initial change that introduced this functionality is actually what regressed https://github.com/flutter/flutter/issues/124554. It also introduced the weird behavior where the commit now has the title twice (GitHub was still reading the PR title, appending the PR number, and setting that as the first line of the commit message, since github.dart is passing a null commit_title to the API).

Also, it seems the string `Revert "Revert` appears relatively infrequently in our PR titles (this was the only example I can find across framework and engine that appeared to actually use the substitution https://github.com/flutter/engine/commit/56bebec98b81).

So now, we only include the title in the message IFF the title matches the `Revert "Revert` pattern. (If we ever want to really set the title for every merge commit, I think we'll need to upstream a change to github.dart to set the `commit_title` field in the JSON we send to the github API).

PR #2614, opened 2023-04-13
[codesign] skip symlink files

XilaiZhang:
context: fixes https://github.com/flutter/flutter/issues/124656 , where symlink is followed and an exception was thrown.

Instead, if a file is a symlink that is pointing to another file, we will skip code signing this file.

### DIFF
--- a/codesign/lib/src/file_codesign_visitor.dart
+++ b/codesign/lib/src/file_codesign_visitor.dart
@@ -205,8 +205,13 @@ update these file paths accordingly.
       );
     }
     directoriesVisited.add(directory.absolute.path);
-    final List<FileSystemEntity> entities = await directory.list().toList();
+    final List<FileSystemEntity> entities = await directory.list(followLinks: false).toList();
     for (FileSystemEntity entity in entities) {
+      if (entity is io.Link) {
+        log.info("current file or direcotry ${entity.path} is a symlink to ${await (entity as io.Link).target()}, "
+            "codesign is therefore skipped for the current file or directory.");
+        continue;
+      }
       if (entity is io.Directory) {
         await visitDirectory(
           directory: directory.childDirectory(entity.basename),
@@ -270,13 +275,6 @@ update these file paths accordingly.
   Future<void> visitBinaryFile({required File binaryFile, required String parentVirtualPath}) async {
     final String currentFileName = binaryFile.basename;
     final String entitlementCurrentPath = joinEntitlementPaths(parentVirtualPath, currentFileName);
-
-    final String symlinkOrFilename = await symlink(binaryFile.path, processManager);
-    if (symlinkOrFilename != binaryFile.path) {
-      log.info("current file ${binaryFile.path} is a symlink to $symlinkOrFilename, "
-          "codesign is therefore skipped for the current file");
-      return;
-    }
 
     if (!fileWithEntitlements.contains(entitlementCurrentPath) &&
         !fileWithoutEntitlements.contains(entitlementCurrentPath)) {
--- a/codesign/lib/src/file_codesign_visitor.dart
+++ b/codesign/lib/src/file_codesign_visitor.dart
@@ -105,7 +105,7 @@ Codesign test failed.
 
 We compared binary files in engine artifacts with those listed in
 entitlement.txt and withoutEntitlements.txt, and the binary files do not match.
-*entitlements.txt is the configuartion file encoded in engine artifact zip,
+*entitlements.txt is the configuration file encoded in engine artifact zip,
 built by BUILD.gn and Ninja, to detail the list of entitlement files.
 Either an expected file was not found in *entitlements.txt, or an unexpected
 file was found in entitlements.txt.
@@ -280,7 +280,7 @@ update these file paths accordingly.
     if (!fileWithEntitlements.contains(entitlementCurrentPath) &&
         !fileWithoutEntitlements.contains(entitlementCurrentPath)) {
       log.severe('The system has detected a binary file at $entitlementCurrentPath.'
-          'but it is not in the entitlements configuartion files you provided.'
+          'but it is not in the entitlements configuration files you provided.'
           'if this is a new engine artifact, please add it to one of the entitlements.txt files');
       throw CodesignException(fixItInstructions);
     }
--- a/codesign/lib/src/file_codesign_visitor.dart
+++ b/codesign/lib/src/file_codesign_visitor.dart
@@ -199,12 +199,6 @@ update these file paths accordingly.
     required String parentVirtualPath,
   }) async {
     log.info('Visiting directory ${directory.absolute.path}');
-    final String symlinkOrFilename = await symlink(directory.absolute.path, processManager);
-    if (symlinkOrFilename != directory.absolute.path) {
-      log.info("current direcotry is a symlink to $symlinkOrFilename, "
-          "codesign is therefore skipped for the current directory");
-      return;
-    }
     if (directoriesVisited.contains(directory.absolute.path)) {
       log.warning(
         'Warning! You are visiting a directory that has been visited before, the directory is ${directory.absolute.path}',
@@ -276,6 +270,13 @@ update these file paths accordingly.
   Future<void> visitBinaryFile({required File binaryFile, required String parentVirtualPath}) async {
     final String currentFileName = binaryFile.basename;
     final String entitlementCurrentPath = joinEntitlementPaths(parentVirtualPath, currentFileName);
+
+    final String symlinkOrFilename = await symlink(binaryFile.path, processManager);
+    if (symlinkOrFilename != binaryFile.path) {
+      log.info("current file ${binaryFile.path} is a symlink to $symlinkOrFilename, "
+          "codesign is therefore skipped for the current file");
+      return;
+    }
 
     if (!fileWithEntitlements.contains(entitlementCurrentPath) &&
         !fileWithoutEntitlements.contains(entitlementCurrentPath)) {
--- a/codesign/lib/src/file_codesign_visitor.dart
+++ b/codesign/lib/src/file_codesign_visitor.dart
@@ -280,9 +280,10 @@ update these file paths accordingly.
 
     if (!fileWithEntitlements.contains(entitlementCurrentPath) &&
         !fileWithoutEntitlements.contains(entitlementCurrentPath)) {
-      log.severe('The system has detected a binary file at $entitlementCurrentPath.'
-          'but it is not in the entitlements configuration files you provided.'
-          'if this is a new engine artifact, please add it to one of the entitlements.txt files');
+      log.severe('the binary file $currentFileName is causing an issue');
+      log.severe('The system has detected a binary file at $entitlementCurrentPath. '
+          'But it is not in the entitlements configuration files you provided. '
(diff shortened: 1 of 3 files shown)


XilaiZhang:
The logic has been tested locally by running the code signing script, and the output has been verified with 'codesign -vvv'.

codesign/lib/src/utils.dart
```diff
@@ -30,6 +30,18 @@ enum FileType {
   }
 }
 
+Future<String> symlink(String folderPath, ProcessManager processManager) async {
```
godofredoc, line 33: Docs?

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -199,6 +199,12 @@ update these file paths accordingly.
     required String parentVirtualPath,
   }) async {
     log.info('Visiting directory ${directory.absolute.path}');
+    final String symlinkOrFilename = await symlink(directory.absolute.path, processManager);
+    if (symlinkOrFilename != directory.absolute.path) {
+      log.info("current direcotry is a symlink to $symlinkOrFilename, "
+          "codesign is therefore skipped for the current directory");
```
godofredoc, line 205: File or directory?

godofredoc (commented):
Fix typo in line 108 and 243 - configuartion 

Remove the dot at the end of line 282 to make it easier to read.

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: umm I implemented it as a check on directory path
for example:
```
xilaizhang-macbookpro:~ xilaizhang$ readlink -f Downloads/FlutterMacOS.framework/Versions/Current
/Users/xilaizhang/Downloads/FlutterMacOS.framework/Versions/A
xilaizhang-macbookpro:~ xilaizhang$ readlink -f Downloads/FlutterMacOS.framework/Versions
/Users/xilaizhang/Downloads/FlutterMacOS.framework/Versions
```
the directory which starts to become a symlink will return a different path when `readlink` is invoked

codesign/lib/src/utils.dart
XilaiZhang: Sure, added doc and corrected typo in the newest commit. Thanks for the catch!

codesign/lib/src/file_codesign_visitor.dart
godofredoc: I believe the same logic works for files. In the future we may have both directory and file symlinks

XilaiZhang: Ah true, yeah sir Godofredo is right. it works for both directory and files. Umm would we have a scenario where /directory is not a symlink but /directory/file is a symlink?

If so, I need to update the logic to process the files as well. I will work on symlink files logic once I arrive at office tomorrow. thanks for the catch!

godofredoc: No sure if we need it now but it is probably easier to fix that now rather than spending cycles later to triage the problem when we introduce file symlinks

christopherfujino: we should not be calling readlink, but instead using `dart:io`'s APIs. Let me write up some sample code...

christopherfujino: ahh but reading more, I see the other problem is that we need to be more robust about what entities we check

christopherfujino: I'll follow up with a PR later to migrate this off of readlink, but that is unrelated to this discussion.

XilaiZhang: thanks for the insights! just to verify, umm does this mean it is okay to use readlink for now? I am halfway through getting it done with readlink

XilaiZhang: Updated the logic to check symlink in visitBinary. verified locally and get the expected output
```
[INFO] codesign: current file /var/folders/2p/d9w0246x2kd8_49q4xs4055c00xkgf/T/conductor_codesignZIT6l1/single_artifact/FlutterMacOS.framework/Versions/Current/FlutterMacOS is a symlink to /private/var/folders/2p/d9w0246x2kd8_49q4xs4055c00xkgf/T/conductor_codesignZIT6l1/single_artifact/FlutterMacOS.framework/Versions/A/FlutterMacOS
...
[INFO] codesign: current file /var/folders/2p/d9w0246x2kd8_49q4xs4055c00xkgf/T/conductor_codesignZIT6l1/single_artifact/FlutterMacOS.framework/FlutterMacOS is a symlink to /private/var/folders/2p/d9w0246x2kd8_49q4xs4055c00xkgf/T/conductor_codesignZIT6l1/single_artifact/FlutterMacOS.framework/Versions/A/FlutterMacOS
, codesign is therefore skipped for the current file
```

XilaiZhang: Actually there is something wrong with the current logic. it skips the non-symlink binary as well, inspecting. 
```
[INFO] codesign: current file /var/folders/2p/d9w0246x2kd8_49q4xs4055c00xkgf/T/conductor_codesignZIT6l1/single_artifact/FlutterMacOS.framework/Versions/A/FlutterMacOS is a symlink to /private/var/folders/2p/d9w0246x2kd8_49q4xs4055c00xkgf/T/conductor_codesignZIT6l1/single_artifact/FlutterMacOS.framework/Versions/A/FlutterMacOS
```

XilaiZhang: Looks like the system temp directory is a symlink itself, (mapping from /var/folders -> /private/var/folders), thinking how to resolve this

christopherfujino: I'm seeing a newline at the end of the filename, I think you should trim the stdout in the `symlink()` function (although amazingly this still works?)

XilaiZhang: Replaced `readlink` with [list](https://api.dart.dev/stable/2.19.6/dart-io/Directory-class.html) (`followLinks: false`) from the `dart:io` family. The [Link](https://api.dart.dev/stable/2.19.6/dart-io/Link-class.html) is an implementation of [filesystem entity](https://api.dart.dev/stable/2.19.6/dart-io/FileSystemEntity-class.html) so we could check it with [isLink](https://api.dart.dev/stable/2.19.6/dart-io/FileSystemEntity/isLink.html).

Verified locally that the output are correct and expected
```
[INFO] codesign: signing file at path /var/folders/2p/d9w0246x2kd8_49q4xs4055c00xkgf/T/conductor_codesignEPQEqT/single_artifact/FlutterMacOS.framework/Versions/A/FlutterMacOS
...
[INFO] codesign: current file /var/folders/2p/d9w0246x2kd8_49q4xs4055c00xkgf/T/conductor_codesignEPQEqT/single_artifact/FlutterMacOS.framework/Resources is a symlink to Versions/Current/Resources, codesign is therefore skipped for the current file
[INFO] codesign: current file /var/folders/2p/d9w0246x2kd8_49q4xs4055c00xkgf/T/conductor_codesignEPQEqT/single_artifact/FlutterMacOS.framework/Versions/Current is a symlink to A, codesign is therefore skipped for the current file
[INFO] codesign: current file /var/folders/2p/d9w0246x2kd8_49q4xs4055c00xkgf/T/conductor_codesignEPQEqT/single_artifact/FlutterMacOS.framework/Headers is a symlink to Versions/Current/Headers, codesign is therefore skipped for the current file
[INFO] codesign: current file /var/folders/2p/d9w0246x2kd8_49q4xs4055c00xkgf/T/conductor_codesignEPQEqT/single_artifact/FlutterMacOS.framework/Modules is a symlink to Versions/Current/Modules, codesign is therefore skipped for the current file
[INFO] codesign: current file /var/folders/2p/d9w0246x2kd8_49q4xs4055c00xkgf/T/conductor_codesignEPQEqT/single_artifact/FlutterMacOS.framework/FlutterMacOS is a symlink to Versions/Current/FlutterMacOS, codesign is therefore skipped for the current file
```

This approach should resolve all the previously mentioned problems.

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -205,8 +205,13 @@ update these file paths accordingly.
       );
     }
     directoriesVisited.add(directory.absolute.path);
-    final List<FileSystemEntity> entities = await directory.list().toList();
+    final List<FileSystemEntity> entities = await directory.list(followLinks: false).toList();
     for (FileSystemEntity entity in entities) {
+      if (entity is io.Link) {
+        log.info("current file or direcotry ${entity.path} is a symlink to ${await (entity as io.Link).target()}, "
```
christopherfujino, line 211: A call to `link.targetSync()` should be fast, and it will avoid needing to have an asynchronous suspension:

```suggestion
        log.info("current file or directory ${entity.path} is a symlink to ${(entity as io.Link).targetSync()}, "
```

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -273,9 +278,10 @@ update these file paths accordingly.
 
     if (!fileWithEntitlements.contains(entitlementCurrentPath) &&
         !fileWithoutEntitlements.contains(entitlementCurrentPath)) {
-      log.severe('The system has detected a binary file at $entitlementCurrentPath.'
-          'but it is not in the entitlements configuartion files you provided.'
-          'if this is a new engine artifact, please add it to one of the entitlements.txt files');
+      log.severe('the binary file $currentFileName is causing an issue');
```
christopherfujino, line 281: Do we need to log out both `currentFileName` AND `entitlementCurrentPath`? Would this be confusing to the person reading these logs?

codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -205,8 +205,13 @@ update these file paths accordingly.
       );
     }
     directoriesVisited.add(directory.absolute.path);
-    final List<FileSystemEntity> entities = await directory.list().toList();
+    final List<FileSystemEntity> entities = await directory.list(followLinks: false).toList();
```
christopherfujino, line 208: Awesome!

christopherfujino (approved):
LGTM

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: Umm they are both logged currently. It is just that `entitlementCurrentPath` is logged on line 282 and `currentFileName` is logged on line 281. umm maybe we should log them on the same line to be clearer?

XilaiZhang: Changed to logging `entitlementCurrentPath` twice. Open to suggestions on how to make it more readable

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: TIL. Didn't know targetSync is fast. Thanks for pointing this out Christopher!

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: thank you!

godofredoc:
thanks @XilaiZhang and @christopherfujino for the change. I'll give it a second try and report back any findings.

PR #2618, opened 2023-04-13
[device_doctor] ios_debug_symbol_doctor recover exit early if subprocess fails

christopherfujino:
Fixes https://github.com/flutter/flutter/issues/124749

Before this change, if the subprocess `xcrun xcodebuild -runFirstLaunch` exits non-zero, we'd log about it and then continue going, wasting CI time.


jmagman (approved):
LGTM

PR #2637, opened 2023-04-20
Follow up Pull Request Revert issues are no longer created

ricardoamador:
This pull request removes the feature of creating a review issue when a revert pull request is generated. Previously a tracking issue was created so that a flutter-hacker member would follow up on a review of the reverted request. This has not been a popular feature and has caused more of a headache than anything else.

Feedback the devs.

## Pre-launch Checklist

- [x] I signed the [CLA].


cron.yaml
```diff
@@ -54,8 +54,4 @@ cron:
   url: /check-pull-request
   target: auto-submit
   schedule: every 1 minutes
-
-- description: check revert review issues for closed issues to update in our database
-  url: '/api/update-revert-issues'
-  target: auto-submit
-  schedule: every 24 hours
+
```
yusuf-goog, line 57: newline at EOF

yusuf-goog (approved):

godofredoc:
@Hixie FYI you're going to like this PR

cron.yaml
ricardoamador: It is odd, I have the newline in my editor and when it gets sent here it does not show. Maybe github is not showing the newline on EOF?

Hixie:
ah cool. Does it just require a normal review before landing now, like other patches?

ricardoamador:
No you would still not need a review up front to merge it. Otherwise it would just be a regular commit from the point of view of autosubmit. 
 I am in the process of reworking this entirely where you would only need to add the label to a closed and merged pull request; then the bot would take care of everything.

Hixie:
do we verify that the commit is precisely a revert of a (very recent) PR?
can we make this ping a channel on discord?
we should never have any code (revert or not) land without two party approval.

ricardoamador:
We verify the change counts in the two pull requests.
Yes we can make this ping discord. I have run a quick test from my local to
ping the tree status channel and it works fine. Just not sure about when it
is deployed to gcp but I can test that.
If that is the case then there is no special path needed for revert.
Everything would just go through autosubmit.

ricardoamador:
To add to this, there has been pushback from the developers that say they
need to be able to revert changes faster which is why the tracking review
issue on reverts was created. So there would be visibility on the revert
and to perform a review. We can still allow that but enforce a two party
review.

Hixie:
the change counts alone isn't going to verify it's a perfect revert. it would be easy to sneak in a vulnerability if all that has to match is the change counts.

ricardoamador:
Yeah, I agree. That is why I am getting to an update to prevent the user
from even needing to do anything aside from adding a label.

Hixie:
to be clearer:

- if we can 100% guarantee that a revert is a precise revert of a PR that landed in the last ~24 hours then I think it's ok to let that go through with just one party approval (the flutter-hackers member adding the label). In theory people could carefully construct a pair of PRs where reverting the first after landing the second opens up a vulnerability but frankly code review isn't going to catch that anyway.
- anything else imho should just use the normal two-party review system we have for any PR. If it's older than a day then it's clearly not that urgent any more and can wait the extra 15 minutes to find a reviewer. If it's more complicated than a straight review then similarly it can get a more thorough review.

If we can make the revert system just be "add a label to a recent PR, and that PR will get autoreverted by a bot whose integrity we can verify" then that sounds fantastic. It would guarantee the revert is correct. (I would recommend still posting an FYI to the #tree-status channel, since people might want to know what is happening to fix the tree.) (Also we should still limit that to "recent" PRs, for some definition of "recent" that's something like "past 24 hours".)

PR #2642, opened 2023-04-21
Missing async keyword not allowing cocoon to wait for build collection

ricardoamador:
A change was made to collect the try tasks and cancel them for a pull request when it is updated, closed, or merged. The function that gets the tasks was missing the 'async' keyword which caused us not to wait for the results of the task collection. It would only work rarely but it should be collecting all of the targets for the pull request each time unless nothing was truly scheduled or run yet.

Fixes [125278](https://github.com/flutter/flutter/issues/125278)

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/app_dart/lib/src/service/luci_build_service.dart
+++ b/app_dart/lib/src/service/luci_build_service.dart
@@ -135,6 +135,12 @@ class LuciBuildService {
         ],
       ),
     );
+
+    log.info('Reponses from get builds batch request = ${batch.responses!.length}');
+    for (Response response in batch.responses!) {
+      log.info('Found a response: ${response.toString()}');
+    }
+
     final Iterable<Build> builds = batch.responses!
         .map((Response response) => response.searchBuilds)
         .expand((SearchBuildsResponse? response) => response!.builds ?? <Build>[]);
--- a/app_dart/lib/src/service/luci_build_service.dart
+++ b/app_dart/lib/src/service/luci_build_service.dart
@@ -86,7 +86,7 @@ class LuciBuildService {
   /// Returns an Iterable of try Buildbucket [Build]s for a given [PullRequest].
   Future<Iterable<Build>> getTryBuildsByPullRequest(
     github.PullRequest pullRequest,
-  ) {
+  ) async {
     final github.RepositorySlug slug = pullRequest.base!.repo!.slug();
     final Map<String, List<String>> tags = <String, List<String>>{
       'buildset': <String>['pr/git/${pullRequest.number}'],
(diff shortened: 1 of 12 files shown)


app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -86,7 +86,7 @@ class LuciBuildService {
   /// Returns an Iterable of try Buildbucket [Build]s for a given [PullRequest].
   Future<Iterable<Build>> getTryBuildsByPullRequest(
     github.PullRequest pullRequest,
-  ) {
```
CaseyHillers, line 89: This isn't necessary.

https://dart-lang.github.io/linter/lints/unnecessary_await_in_return.html

ricardoamador: But I am not awaiting a return.

ricardoamador:
Closing this as there is some other issue with our code. Adding the async did not prevent the issue.

app_dart/lib/src/service/luci_build_service.dart
ricardoamador: @CaseyHillers you are right I just proved it through a deployment. Something is up, because I see this working when I have tested it.

CaseyHillers: See the example:

```
GOOD:
Future<int> f2() {
  return future;
}
```

These lints aren't intended to explain the concept, but to enforce best practices. We should add the lint so this PR fails on additions like this.

See https://dart.dev/guides/libraries/library-tour#future for more info on why, but whenever possible, the callers of methods should do await, not the method itself.

ricardoamador: Man I think we do this a lot throughout the code too. But this is good. I will add this and do some cleanup. Thank you for the extra info here.

ricardoamador:
I may have spoke too soon.

app_dart/lib/src/service/luci_build_service.dart
ricardoamador: Okay I am going to disagree with you on this change. Adding the linter rule does not flag the async keyword. And it appears to be working every time now except for when a PR is first opened. This is the fix.

CaseyHillers: FYI I created https://github.com/flutter/cocoon/pull/2644. Seems like several analysis options are failing at HEAD due to this check not running in CI.

\cc @christopherfujino have you seen this issue before? I don't see anything functionally wrong that would require async to be in the method, as this is only returning a future.

PR #2650, opened 2023-04-25
Disable device doctor in recipes CQ

keyonghan:
The device doctor tasks are not necessary to run in recipes CQ. Same recipes is being validated by other target `Mac codesign`.

At the same time `Mac device doctor` is failing recipes CQ (https://flutter-review.googlesource.com/c/recipes/+/42520). To unblock, removing cq.


keyonghan:
Update mac dart path following dart sdk change.

godofredoc (approved):

PR #2651, opened 2023-04-25
Switch dart-internal results from pubsub endpoint to http endpoint

drewroengoogle:
* Change `api/dart-internal-subscription` to `api/dart-internal-task-results`
* Switch the dart-internal endpoint to be accessible by http
* This allows us to directly call cocoon from recipes instead of going through pubsub.
  * The reason we are blocked from publishing results through pubsub is because of a couple of things preventing us from using pubsub in the engine_v2 recipes (Linux has a maximum byte amount which is exceeded by the build results, and Windows gcloud from Fuchsia does not seem to be working at all)
  * Using curl to directly call the endpoint will allow us to specify the build results in a file
  * CC: @godofredoc 

## Pre-launch Checklist

- [x] I signed the [CLA].


keyonghan (commented):
>> Linux has a maximum byte amount which is exceeded by the build results

>Has this been determined as the way to proceed? I may miss some context. Comparing to direct `curl`, pubsub has some advantages like auto retry when api fails due to server side issue.

This is a terminal limitation in all the platforms. From the recipes point of view there is no difference on wether we are using the gcloud client or curl as the gcloud client is sending http requests to the pubsub APIs.   


>> Windows gcloud from Fuchsia does not seem to be working at all

>What is this limit? Could we just forward the necessary build info instead of the whole build details? From dashboard side, we need only several entries, like name, build status, commit, etc.

The gcloud pubsub client does not support standard input and sending the entire message body as a parameter in the terminal is hitting max char sizes. 


> What is the error? Fuchsia gcloud module is pretty simple. Any chance we mock a bit and fix the issue?

The module is not the problem but the gcloud client itself. To fix this issue gcloud pubsub needs to support standard input. Another alternative is to just send the buildbucket id of the message with the status and let the backend query buildbucket to get all the additional details in the backend.

godofredoc:
Can we add to the PR description that this is only the skeleton of the handler and the additional functionality will come in subsequent PRs?



>

drewroengoogle:
> This is a terminal limitation in all the platforms. From the recipes point of view there is no difference on wether we are using the gcloud client or curl as the gcloud client is sending http requests to the pubsub APIs.

I'm not sure I follow the response here - this PR skips pubsub completely, so to Keyong's point, we do lose the functionality of automatically retrying a failed call.

> Windows gcloud from Fuchsia does not seem to be working at all
@keyonghan to give a bit more insight here, it seems that any call to the Windows `gcloud` cipd package fails. I know after bringing it up with @godofredoc we attempted to look at fuchsia/gcloud to try and see if we could modify the module and fix it, but we were unsuccessful.

> Another alternative is to just send the buildbucket id of the message with the status and let the backend query buildbucket to get all the additional details in the backend.
This still leaves us with the issue of Windows gcloud failing to send any pubsub message. What are your thoughts on me spending more time trying to fix `gcloud` for Windows? I sent a message in the internal LUCI chat with no response.

godofredoc:
I'll send a fix for the gcloud independently of whether we use it or not.

keyonghan:
> > This is a terminal limitation in all the platforms. From the recipes point of view there is no difference on wether we are using the gcloud client or curl as the gcloud client is sending http requests to the pubsub APIs.
> 
> I'm not sure I follow the response here - this PR skips pubsub completely, so to Keyong's point, we do lose the functionality of automatically retrying a failed call.
> 
> > Windows gcloud from Fuchsia does not seem to be working at all
> 
> @keyonghan to give a bit more insight here, it seems that any call to the Windows `gcloud` cipd package fails. I know after bringing it up with @godofredoc we attempted to look at fuchsia/gcloud to try and see if we could modify the module and fix it, but we were unsuccessful.
> 
> > Another alternative is to just send the buildbucket id of the message with the status and let the backend query buildbucket to get all the additional details in the backend.
> 
> This still leaves us with the issue of Windows gcloud failing to send any pubsub message. What are your thoughts on me spending more time trying to fix `gcloud` for Windows? I sent a message in the internal LUCI chat with no response.
> 
> The overall context here is that when attempting to update the recipe for reporting the build results, we ran into two main issues: Windows gcloud not working at all, and Linux having a maximum byte limit, preventing the full Buildbucket result from being published. This solution came from syncing with @godofredoc to try and solve these issues.

In that case, can we reuse the existing API: `/api/update-task-status`?

drewroengoogle:
> I'll send a fix for the gcloud independently of whether we use it or not.

Talked 

> /api/update-task-status

I talked with Godofredo and it sounds like he has a potential CL to fix gcloud in Windows, which I didn't realize. Using that, we could use the pub/sub strategy and just send the buildbucket id, and then get the build data within the endpoint. If that's okay with you, I can close this PR. The only difference from the initial design doc would be that we acquire the build data from the build id instead of getting it directly from a pub/sub message.

keyonghan:
> > > > This is a terminal limitation in all the platforms. From the recipes point of view there is no difference on wether we are using the gcloud client or curl as the gcloud client is sending http requests to the pubsub APIs.
> > > 
> > > 
> > > I'm not sure I follow the response here - this PR skips pubsub completely, so to Keyong's point, we do lose the functionality of automatically retrying a failed call.
> > > > Windows gcloud from Fuchsia does not seem to be working at all
> > > 
> > > 
> > > @keyonghan to give a bit more insight here, it seems that any call to the Windows `gcloud` cipd package fails. I know after bringing it up with @godofredoc we attempted to look at fuchsia/gcloud to try and see if we could modify the module and fix it, but we were unsuccessful.
> > > > Another alternative is to just send the buildbucket id of the message with the status and let the backend query buildbucket to get all the additional details in the backend.
> > > 
> > > 
> > > This still leaves us with the issue of Windows gcloud failing to send any pubsub message. What are your thoughts on me spending more time trying to fix `gcloud` for Windows? I sent a message in the internal LUCI chat with no response.
> > > The overall context here is that when attempting to update the recipe for reporting the build results, we ran into two main issues: Windows gcloud not working at all, and Linux having a maximum byte limit, preventing the full Buildbucket result from being published. This solution came from syncing with @godofredoc to try and solve these issues.
> > 
> > 
> > In that case, can we reuse the existing API: `/api/update-task-status`?
> 
> I talked with Godofredo and it sounds like he has a potential CL to fix gcloud in Windows, which I didn't realize. Using that, we could use the pub/sub strategy and just send the buildbucket id, and then get the build data within the endpoint. If that's okay with you, I can close this PR. The only difference from the initial design doc would be that we acquire the build data from the build id instead of getting it directly from a pub/sub message.

SG. Thanks for the update!

PR #2652, opened 2023-04-25
Add doc for updating target platform in .ci.yaml

keyonghan:
Related https://github.com/flutter/flutter/issues/125492

### DIFF
--- a/CI_YAML.md
+++ b/CI_YAML.md
@@ -182,9 +182,15 @@ to be landed first to wait for changes propagated in infrastructure.
 
 #### Update target platform
 
+Target depends on the prefix platform in its `name` to decide which platform to run on. This should match
+to an existing platform under `platform_properties`.
+
 If one target needs to switch running platforms, e.g. from a devicelab bot to a host only bot:
 1. Keep the old target entry
-2. Add a new entry under the new platform as `bringup: true` with necessary dependencies
+2. Add a new entry under the new platform with
+  1. `bringup: true`
+  2. necessary dependencies
+  3. corresponding tags (tags will only be used for infra metrics analysis)
 3. Land the change with the new entry
 4. If the new target under the new platform passes in postsubmit
   1. Remove the old target entry and mark the new target as `bringup: false`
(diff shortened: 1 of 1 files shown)


CI_YAML.md
```diff
@@ -180,6 +180,38 @@ tags: >
...
+2. Add a new entry under the new platform as `bringup: true` with necessary dependencies
+3. Land the change with the new entry
+4. If the new target under the new platform passes in postsubmit
+  1. Remove the old target entry and mark the new target as `bringup: false`
+
+Example: say one wants to switch `Linux_android web_size__compile_test` to a vm.
+
+Existing config:
+```yaml
+- name: Linux_android web_size__compile_test
```
zanderso, line 196: Could you explain the part of this change that specifically causes the new version to become host-only? Is it removing the `_android` from the name, or is it some other part of the config?

(As an aside, encoding the hardware requirements in the config name seems confusing to me, if that's the case. Should it be explicitly specified in the `properties` instead?)

keyonghan: Yes, by updating the prefix platform in the target name. See the target `name` described above: https://github.com/flutter/cocoon/blob/main/CI_YAML.md#ciyaml

This dependency of platform in the target name is some legacy thing. https://github.com/flutter/flutter/issues/114905 tracks deprecating the platform dependency on target names.

zanderso: I'd suggest including that explanation in the docs in this PR. Under `Linux_android web_size__compile_test` I'm currently also seeing:
```
properties:
      tags: >
        ["devicelab", "android", "linux"]
```
Do any of those need to stay or be removed in order to make the test host-only?

keyonghan: Updated the doc accordingly, including the `tags`.

CI_YAML.md
```diff
@@ -180,6 +180,49 @@ tags: >
...
+
+Target depends on the prefix platform in its `name` to decide which platform to run on. This should match
+to an existing platform under `platform_properties`.
+
+If one target needs to switch running platforms, e.g. from a devicelab bot to a host only bot:
+1. Keep the old target entry
+2. Add a new entry under the new platform with
+  1. `bringup: true`
+  2. necessary dependencies
+  3. corresponding tags (tags will only be used for infra metrics analysis)
```
ricardoamador, line 193: Aren't tags also used for performance tests or is that what you mean here?

zanderso (approved):
Thanks!

PR #2657, opened 2023-04-26
Refactor Autosubmit so that it can be configured by Org and Repository

ricardoamador:
## Summary

The purpose of these changes are to allow teams outside of Flutter to use autosubmit in their repositories. This change makes it so that autosubmit will now read it's configuration from two places, the Org level repository and the Repository where Autosubmit is installed.

This change will enable auto submit to be configured at the organization level and per repository. The Organization level configuration can be overridden if the allow_overrides flag is set. This means that any values supplied in the repository level can override the values set in the Org level. The following rules apply:
  * The default branch cannot be overridden as it is collected.
  * If the default branch is supplied then it will not be automatically collected.
  * The number of approving reviews cannot be less than what is set in the organization repo.
  * Any values that are arrays are additive. This means you cannot override auto approval accounts or required check runs on revert. The Org level config should have the lowest common denominator when it comes to list values.

Approving authors team membership is now used vs OWNER/MEMBER in the previous version. 

Fixes https://github.com/flutter/flutter/issues/120742

Also addresses
Fixes https://github.com/flutter/flutter/issues/123665

Validation has been done with both Flutter and Dart-lang repositories and is ongoing after hours.

## Open Questions
Do you think the endpoints for reading a config and forcing a config reload are valuable?

## Pre-launch Checklist

- [x] I signed the [CLA].


yusuf-goog (approved):

yusuf-goog (commented):
LGTM, but should get another pair of eyes on it, due to the large size.

auto_submit/lib/configuration/repository_configuration_manager.dart
```diff
@@ -0,0 +1,174 @@
...
+/// The [RepositoryConfigurationManager] is responsible for fetching and merging
+/// the autosubmit configuration from the Org level repository and if needed
+/// fetching the override configuration from the pull request repository.
+///
+/// It will attempt to access the cache first before repulling the configuraiton
+/// from the repositories. This is currently set at a 10 minute TTL.
+class RepositoryConfigurationManager {
+  final Mutex _mutex = Mutex();
+
+  static const String fileSeparator = '/';
```
godofredoc, line 24: Is there a constant we can use rather than defining our own?

auto_submit/lib/configuration/repository_configuration_manager.dart
```diff
@@ -0,0 +1,174 @@
...
+  /// Entries will be stored in the cache as config/slug/autosubmit.yaml
+  Future<RepositoryConfiguration> readRepositoryConfiguration(
+    RepositorySlug slug,
+  ) async {
+    await _mutex.acquire();
+    try {
+      // Get the contents from the cache or go to github.
+      final cacheValue = await cache['${slug.fullName}$fileSeparator$fileName'].get(
+        () async => _getConfiguration(slug),
+        const Duration(minutes: 10),
```
godofredoc, line 50: Can we make add this as a constant and document it somewhere? should we add a README file to explain the different configurations?

auto_submit/lib/configuration/repository_configuration_manager.dart
```diff
@@ -0,0 +1,174 @@
...
+      );
+      // final String yamlContents = String.fromCharCodes(cacheValue!);
+      final String cacheYaml = String.fromCharCodes(cacheValue);
+      log.info('Converting yaml to RepositoryConfiguration: $cacheYaml');
+      return RepositoryConfiguration.fromYaml(cacheYaml);
```
godofredoc, line 55: Is this always reading from master?

auto_submit/lib/configuration/repository_configuration_manager.dart
```diff
@@ -0,0 +1,174 @@
...
+    RepositorySlug slug,
+  ) async {
+    await _mutex.acquire();
+    try {
+      // Get the contents from the cache or go to github.
+      final cacheValue = await cache['${slug.fullName}$fileSeparator$fileName'].get(
+        () async => _getConfiguration(slug),
+        const Duration(minutes: 10),
+      );
+      // final String yamlContents = String.fromCharCodes(cacheValue!);
```
godofredoc, line 52: rm?

auto_submit/lib/configuration/repository_configuration_manager.dart
```diff
@@ -0,0 +1,174 @@
...
+    // Read the org level configuraiton file first.
+    log.info('Getting org level configuration.');
+    // Get the Org level configuration.
+    final RepositorySlug orgSlug = RepositorySlug(slug.owner, orgRepository);
+    final GithubService githubService = await config.createGithubService(orgSlug);
+    final String orgLevelConfig = await githubService.getFileContents(orgSlug, '$dirName$fileSeparator$fileName');
+    final RepositoryConfiguration globalRepositoryConfiguration = RepositoryConfiguration.fromYaml(orgLevelConfig);
+
+    // Collect the default branch if it was not supplied.
+    if (globalRepositoryConfiguration.defaultBranch!.isEmpty) {
```
godofredoc, line 75: Do we allow the configuration to contradict github in terms of default branch? or what is the use case for globalRepositoryConfiguration.defaultBranch?

auto_submit/lib/configuration/repository_configuration_manager.dart
```diff
@@ -0,0 +1,174 @@
...
+        localRepositoryConfigurationYaml =
+            await localGithubService.getFileContents(slug, '$dirName$fileSeparator$fileName');
+        final RepositoryConfiguration localRepositoryConfiguration =
+            RepositoryConfiguration.fromYaml(localRepositoryConfigurationYaml);
+        final RepositoryConfiguration mergedRepositoryConfiguration = mergeConfigurations(
+          globalRepositoryConfiguration,
+          localRepositoryConfiguration,
+        );
+        return mergedRepositoryConfiguration.toString().codeUnits;
+      } on Exception {
```
godofredoc, line 97: Is there a more specific exception to track?

godofredoc: Also should this log the exception and fail rather than hiding the error?

auto_submit/lib/requests/github_webhook.dart
```diff
@@ -59,7 +60,7 @@ class GithubWebhook extends RequestHandler {
 
     if (hasAutosubmit || hasRevertLabel) {
       log.info('Found pull request with auto submit and/or revert label.');
-      await pubsub.publish('auto-submit-queue', pullRequest);
+      await pubsub.publish(Config.pubsubPullRequestTopic, pullRequest);
```
godofredoc, line 63: is this a static property? or should we use an instance of config rather than the class?

auto_submit/lib/requests/github_webhook.dart
```diff
@@ -69,6 +70,7 @@ class GithubWebhook extends RequestHandler {
     String? signature,
     List<int> requestBody,
   ) async {
+    log.info('Validating request...');
```
godofredoc, line 73: Some info about the request being validated can be useful when debugging.

auto_submit/lib/service/approver_service.dart
```diff
@@ -3,9 +3,12 @@
 // found in the LICENSE file.
 
 import 'package:auto_submit/model/auto_submit_query_result.dart';
+import 'package:auto_submit/service/github_service.dart';
 import 'package:auto_submit/service/log.dart';
-import 'package:github/github.dart' as gh;
+import 'package:github/github.dart' as github;
+import 'package:github/github.dart';
```
godofredoc, line 9: rm? double import

auto_submit/lib/validations/validation_filter.dart
```diff
@@ -0,0 +1,100 @@
...
+  final RepositoryConfiguration repositoryConfiguration;
+
+  @override
+  Set<Validation> getValidations() {
+    final Set<Validation> validationsToRun = {};
+
+    validationsToRun.add(Revert(config: config));
+    // TODO(ricardoamador) add this back when validations for revert have been refactored.
+    // Validate required check runs if they are requested.
+    // if (repositoryConfiguration.requiredCheckRunsOnRevert!.isNotEmpty) {
```
godofredoc, line 94: rm?

godofredoc (commented):
Small PRs are preferred for faster reviews and quick turn around. It is very easy to get lost on PRs containing tens of files.

auto_submit/lib/configuration/repository_configuration_manager.dart
ricardoamador: Looks like there is a Platform.pathSeparator getter/constant.

auto_submit/lib/configuration/repository_configuration_manager.dart
ricardoamador: Yeah I can do that.

auto_submit/lib/configuration/repository_configuration_manager.dart
ricardoamador: Yes. Removed.

auto_submit/lib/configuration/repository_configuration_manager.dart
ricardoamador: Yes this assumes master. I am not providing a ref to the underlying Github lib which assumes that it is reading from the default branch. I would think that is what we want here right?

auto_submit/lib/configuration/repository_configuration_manager.dart
ricardoamador: Hmm, this may be from a convenience for myself from early testing when I was assuming that default branch was being supplied to us. Let me refactor this.

auto_submit/lib/configuration/repository_configuration_manager.dart
ricardoamador: Ah yeah for this we can fail. This was mostly due to testing as we had it configured and dart did not so I just left it as a warning knowing that we do not have overrides happening. If it is allowed and we find nothing would it be harmful to just use the global configuration?

auto_submit/lib/requests/github_webhook.dart
ricardoamador: It is currently static. We have a mix but I can make it an instance field if that is preferable?

auto_submit/lib/service/approver_service.dart
ricardoamador: Good catch. Fixed.

auto_submit/lib/service/validation_service.dart
```diff
@@ -39,37 +35,25 @@ class ValidationService {
     /// Validates a PR marked with the reverts label.
     revertValidation = Revert(config: config);
     approverService = ApproverService(config);
-
-    validations.addAll({
```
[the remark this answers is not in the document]
ricardoamador: Yes. I refactored this to the ValidationFilter class. It allows us to manage the validations done on a particular type of pull request rather than putting a bunch of if/else clauses. These are all still there.

auto_submit/lib/service/validation_service.dart
```diff
@@ -227,7 +225,25 @@ class ValidationService {
     required String ackId,
     required PubSub pubsub,
   }) async {
-    final ValidationResult revertValidationResult = await revertValidation!.validate(result, messagePullRequest);
+    // TODO(ricardoamador) rework this part as the validations should just be iterated through
+    // like when a pull request is processed.
+
+    // final RepositoryConfiguration repositoryConfiguration = await config.getRepositoryConfiguration(slug);
```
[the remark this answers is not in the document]
ricardoamador: Yeah, this will all go away when revert is refactored.

auto_submit/lib/validations/validation_filter.dart
ricardoamador: Yes. Thanks!

PR #2699, opened 2023-05-08
Mark canceled builds as infra failure to close the tree

keyonghan:
Fix: https://github.com/flutter/flutter/issues/125012

Related: https://github.com/flutter/flutter/issues/126294

### DIFF
--- a/app_dart/lib/src/model/appengine/task.dart
+++ b/app_dart/lib/src/model/appengine/task.dart
@@ -353,7 +353,7 @@ class Task extends Model<int> {
       case Result.success:
         return status = statusSucceeded;
       case Result.canceled:
-        return status = statusCancelled;
+        return status = statusInfraFailure;
       case Result.infraFailure:
         return status = statusInfraFailure;
       case Result.failure:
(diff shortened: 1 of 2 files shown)


ricardoamador (approved):
LGTM

godofredoc (commented):
Please do not land, this is not the correct place to add the functionality.

app_dart/lib/src/model/appengine/task.dart
```diff
@@ -353,7 +353,7 @@ class Task extends Model<int> {
       case Result.success:
         return status = statusSucceeded;
       case Result.canceled:
-        return status = statusCancelled;
+        return status = statusInfraFailure;
```
godofredoc, line 356: Please add the Task.Cancelled state here https://cs.opensource.google/flutter/cocoon/+/main:app_dart/lib/src/service/build_status_provider.dart;l=140. That way we can keep the cancelled state in the task model.

keyonghan: Good catch. Updated.

godofredoc (approved):

PR #2702, opened 2023-05-09
[codesign] publish codesign arm cipd

XilaiZhang:
Add codesign cipd built on arm machine. marked as "bringup: true" as it is a new target

Context: codesign runs on the same builder that compiled it, and we would need to run it on arm machines.

This is aiming to address [codesign failure on beta branch](https://ci.chromium.org/ui/p/flutter/builders/prod/Mac%20mac_host_engine/3525/overview)


.ci.yaml
```diff
@@ -64,6 +64,18 @@ targets:
       - codesign/**
       - .ci.yaml
 
+  - name: Mac arm_codesign
```
christopherfujino, line 67: in what way will this build be any different from the intel one?

christopherfujino: oh wait, nvm, i see cpu under properties, lol

XilaiZhang: umm I guess a `cpu:arm64` will have this cipd be built on an arm64 machine?

godofredoc (approved):

PR #2704, opened 2023-05-09
[codesign] remove bringup

XilaiZhang:
Removes the bringup property on codesign arm target so it is added to the prod pool instead of the staging pool. 

Context: targets with bringup will be added to the staging pool on post submit.


christopherfujino (approved):
LGTM

PR #2725, opened 2023-05-12
Limit flake issue/pr creation number in each API call

keyonghan:
Fixes: https://github.com/flutter/flutter/issues/126604

### DIFF
--- a/app_dart/lib/src/request_handlers/file_flaky_issue_and_pr.dart
+++ b/app_dart/lib/src/request_handlers/file_flaky_issue_and_pr.dart
@@ -49,6 +49,7 @@ class FileFlakyIssueAndPR extends ApiRequestHandler<Body> {
     final String testOwnerContent = await gitHub.getFileContent(slug, kTestOwnerPath);
     final Map<String?, Issue> nameToExistingIssue = await getExistingIssues(gitHub, slug);
     final Map<String?, PullRequest> nameToExistingPR = await getExistingPRs(gitHub, slug);
+    int filedIssueAndPRNumber = 0;
     for (final BuilderStatistic statistic in builderStatisticList) {
       // Skip if ignore_flakiness is specified.
       if (getIgnoreFlakiness(statistic.name, ciYaml)) {
@@ -58,7 +59,7 @@ class FileFlakyIssueAndPR extends ApiRequestHandler<Body> {
         continue;
       }
       final BuilderType type = getTypeForBuilder(statistic.name, ciYaml);
-      await _fileIssueAndPR(
+      final bool issueAndPRFiled = await _fileIssueAndPR(
         gitHub,
         slug,
         builderDetail: BuilderDetail(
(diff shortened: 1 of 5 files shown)


app_dart/lib/src/request_handlers/file_flaky_issue_and_pr.dart
```diff
@@ -49,6 +49,7 @@ class FileFlakyIssueAndPR extends ApiRequestHandler<Body> {
     final String testOwnerContent = await gitHub.getFileContent(slug, kTestOwnerPath);
     final Map<String?, Issue> nameToExistingIssue = await getExistingIssues(gitHub, slug);
     final Map<String?, PullRequest> nameToExistingPR = await getExistingPRs(gitHub, slug);
+    int filedIssueAndPRNumber = 0;
```
ricardoamador, line 52: nit: consider renaming this to filedIssueAndPRCount. Number makes me think this is tracking the PR number.

keyonghan: Updated.

ricardoamador:
Just something to consider with this, is that it may not directly be the call to create the issue but all of the calls to GitHub that we make during processing this request. We could limit the number of concurrent calls or space out the calls with waits between them for now.

cron.yaml
```diff
@@ -36,7 +36,7 @@ cron:
 
 - description: detect and flag tests with high flaky rates
   url: /api/file_flaky_issue_and_pr?threshold=0.02
-  schedule: every wednesday 9:00
+  schedule: every wednesday 9:00,9:30,10:00,10:30,11:00
```
ricardoamador, line 39: I saw your comment in the issue this is fixing; why not make this run in off peak hours? Does it need to be run during hours? What about every wednesday at 4:00 am?

ricardoamador (approved):
LGTM with questions.

keyonghan:
> Just something to consider with this, is that it may not directly be the call to create the issue but all of the calls to GitHub that we make during processing this request. We could limit the number of concurrent calls or space out the calls with waits between them for now.

This PR is limiting the number of "concurrent" calls.

This reminds me of https://github.com/flutter/cocoon/pull/2201, where I added a delay between PR creations for `check_flaky_builders` API. (which doesn't seem to hit this issue, but we may not have that many number of PRs to deflake).

Instead of this change, let me add a delay for this api as well to see how it goes. It it helps, we can skip this change.

cron.yaml
keyonghan: The time is UTC. Converting to PST these would be 1:00am - 3:00am.

app_dart/lib/src/service/config.dart
```diff
@@ -164,6 +164,12 @@ class Config {
   /// are idle, we hope to scan as many commit rows as possible.
   int get backfillerCommitLimit => 50;
 
+  /// Upper limit of issue/PRs allowed each API call.
+  ///
+  /// GitHub enforces a secondary rate limit on frequency API calls. This causes
+  /// our API failure when many issues/PRs are created in a short time.
+  int get issueAndPRLimit => 2;
```
chunhtai, line 171: can also consider turning this into query parameter

chunhtai (approved):
LGTM

keyonghan:
There was one PR (totally 5 this week) hitting this issue. I am landing this PR to help.

PR #2758, opened 2023-05-16
[codesign] remove metadata files for now

XilaiZhang:
Remove codesign metadata after parsing them.

Context: fix https://github.com/flutter/flutter/issues/126705 by removing codesign metadata after parsing them. When https://github.com/flutter/flutter/pull/126875 becomes ready, we could add back codesign metadata inside engine zips?

Sorry for the delay, was busy landing different cps during the day

### DIFF
--- a/codesign/lib/src/file_codesign_visitor.dart
+++ b/codesign/lib/src/file_codesign_visitor.dart
@@ -335,13 +335,10 @@ update these file paths accordingly.
 
     final Set<String> fileWithEntitlements = <String>{};
 
-    try {
-      fileWithEntitlements.addAll(await fileSystem.file(entitlementFilePath).readAsLines());
-    } finally {
-      // TODO(xilaizhang) : add back metadata information after https://github.com/flutter/flutter/issues/126705
-      // is resolved.
-      await fileSystem.file(entitlementFilePath).delete();
-    }
+    fileWithEntitlements.addAll(await fileSystem.file(entitlementFilePath).readAsLines());
+    // TODO(xilaizhang) : add back metadata information after https://github.com/flutter/flutter/issues/126705
+    // is resolved.
+    await fileSystem.file(entitlementFilePath).delete();
 
     return fileWithEntitlements;
   }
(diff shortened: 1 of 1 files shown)


codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -334,7 +334,15 @@ update these file paths accordingly.
...
     final Set<String> fileWithEntitlements = <String>{};
-    fileWithEntitlements.addAll(await fileSystem.file(entitlementFilePath).readAsLines());
+
+    try {
+      fileWithEntitlements.addAll(await fileSystem.file(entitlementFilePath).readAsLines());
+    } finally {
+      // TODO(xilaizhang) : add back metadata information after https://github.com/flutter/flutter/issues/126705
+      // is resolved.
+      await fileSystem.file(entitlementFilePath).delete();
+    }
```
christopherfujino, line 344:
```suggestion
    fileWithEntitlements.addAll(await fileSystem.file(entitlementFilePath).readAsLines());
    // TODO(xilaizhang) : add back metadata information after https://github.com/flutter/flutter/issues/126705
    // is resolved.
    await fileSystem.file(entitlementFilePath).delete();
```

You don't need the `try` - `finally` if you're not doing any error handling.

christopherfujino (approved):
nit about removing try - finally, otherwise LGTM

christopherfujino (approved):
nit about removing try - finally, otherwise LGTM

codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: thanks for the review! umm initially I don't have the try finally block, but I was worried that the second await statement could get executed before the first await statement: i.e., the file could have got deleted before it is read. My understanding is that there might not be an enforced order among the await statements? umm I added the try finally block to enforce a sequential order

XilaiZhang: Never mind, awaits are sequential inside the same function. TIL, thank you Christopher!

PR #2644, opened 2023-04-21
Run analyze on Cocoon

CaseyHillers:
Several analysis options have been failing as it's not running in CI.

PR #2761, opened 2023-05-16
Switch flutter_test_runner to --no-fatal-infos

goderbauer:
As per policy (https://github.com/flutter/tests#adding-more-tests):

> The tests must pass even if there are analysis 'info' level issues in the code. Generally, this means that if the test performs static analysis, it does so ignoring info level items (i.e., flutter analyze --no-fatal-infos).


goderbauer:
@CaseyHillers You added the --fatal-into in https://github.com/flutter/cocoon/pull/2644. Was that deliberate? Or did you mean to say --no-fatal-into?

test_utilities/bin/flutter_test_runner.bat
goderbauer, line 1: BTW, looks like the windows version of this script is currently unused in `flutter/tests`:

https://github.com/flutter/tests/blob/3b9e47a41d38e88c94c327d93e59450f45b48e05/registry/flutter_cocoon.test

Should we add it there?

goderbauer:
Check failure will be resolved by pkg:collection bump in https://github.com/flutter/cocoon/pull/2760.

test_utilities/bin/flutter_test_runner.bat
CaseyHillers: Can we remove the windows scripts completely? For context, we had at some point a windows bot in this repo, but it wasn't generating value. Thanks!

CaseyHillers (approved):

CaseyHillers:
Thanks for the clean up!

test_utilities/bin/flutter_test_runner.bat
goderbauer: Deleted.

PR #2780, opened 2023-05-19
do not require tests for whitespace only changes

christopherfujino:
See for example this change that only touches comments and whitespace: https://github.com/flutter/flutter/pull/127227

### DIFF
--- a/app_dart/lib/src/request_handlers/github/webhook_subscription.dart
+++ b/app_dart/lib/src/request_handlers/github/webhook_subscription.dart
@@ -686,10 +686,17 @@ class GithubWebhookSubscription extends SubscriptionHandler {
     // would require the full file and non-trivial parsing. Also doesn't handle
     // end-of-line comments (e.g., "int x = 0; // Info about x").
     final RegExp commentRegex = RegExp(r'[+-]\s*//');
+    final RegExp onlyWhitespaceRegex = RegExp(r'[+-]\s*$');
     for (String line in patch.split('\n')) {
       if (!line.startsWith('+') && !line.startsWith('-')) {
         continue;
       }
+
+      if (onlyWhitespaceRegex.hasMatch(line)) {
+        // whitespace only changes don't require tests
+        continue;
+      }
+
       if (!commentRegex.hasMatch(line)) {
         return false;
       }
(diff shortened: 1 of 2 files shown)


hellohuanlin (approved):
is this only for the engine? what about the framework?

app_dart/lib/src/request_handlers/github/webhook_subscription.dart
```diff
@@ -686,10 +686,17 @@ class GithubWebhookSubscription extends SubscriptionHandler {
     // would require the full file and non-trivial parsing. Also doesn't handle
     // end-of-line comments (e.g., "int x = 0; // Info about x").
     final RegExp commentRegex = RegExp(r'[+-]\s*//');
+    final RegExp onlyWhitespaceRegex = RegExp(r'[+-]\s*$');
```
stuartmorgan, line 689: I forgot to start the regex with a `^`, can you fix that here and in the line above while your are here?

Right now a line like
```
+      anotherElementForALongMathExpression +  // Some explanation
```

would be considered a comment.

christopherfujino:
> is this only for the engine? what about the framework?

I could add another test, but the same code runs regardless of the repo

app_dart/lib/src/request_handlers/github/webhook_subscription.dart
christopherfujino: Good catch

christopherfujino: Now I have to be careful though whether these lines start with the +/-, or if there might be a whitespace prefix...

stuartmorgan: No, there's already a skip of any line that doesn't start with + or - (a few lines up), so that this only looks at change lines, not context, headers, etc

christopherfujino: Ahh, good point

christopherfujino: done

stuartmorgan (approved):
Thanks for cleaning up my mistake 🙂

PR #2801, opened 2023-05-24
Add check for DevToolsSecurity

vashworth:
Adding to help debug https://github.com/flutter/flutter/issues/120808.

Currently it does not do anything other than print out the status. In the future, we may want to try a) changing the status or b) opening Xcode based off the status. First, we just want to observe what the status is when tests run.

It's currently piggybacking on the `ios_debug_symbol_doctor.dart diagnose` command. If this is something we want to keep and use, we should rename `ios_debug_symbol_doctor` or make this its own command.

I filed https://github.com/flutter/flutter/issues/127517 to make sure we follow up.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/device_doctor/lib/src/ios_debug_symbol_doctor.dart
+++ b/device_doctor/lib/src/ios_debug_symbol_doctor.dart
@@ -55,7 +55,8 @@ class DiagnoseCommand extends Command<bool> {
     return true;
   }
 
-  Future<bool> checkDevToolsSecurity() async {
+  /// Log the status of DevToolsSecurity.
+  Future<void> checkDevToolsSecurity() async {
     final List<String> command = <String>['xcrun', 'DevToolsSecurity', '--status'];
     final io.ProcessResult result = await processManager.run(
       command,
@@ -64,12 +65,9 @@ class DiagnoseCommand extends Command<bool> {
       logger.severe(
         '$command failed with exit code ${result.exitCode}\n${result.stderr}',
       );
-      return false;
     }
     final String stdout = result.stdout as String;
     logger.info(stdout);
-
-    return true;
   }
 }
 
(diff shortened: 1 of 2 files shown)


device_doctor/lib/src/ios_debug_symbol_doctor.dart
```diff
@@ -50,8 +50,25 @@ class DiagnoseCommand extends Command<bool> {
...
   }
+
+  /// Log the status of DevToolsSecurity.
+  Future<void> checkDevToolsSecurity() async {
+    final List<String> command = <String>['xcrun', 'DevToolsSecurity', '--status'];
+    final io.ProcessResult result = await processManager.run(
+      command,
+    );
+    if (result.exitCode != 0) {
+      logger.severe(
```
christopherfujino, line 65: is your intention for this just to be diagnostic, or should this block the build?

device_doctor/lib/src/ios_debug_symbol_doctor.dart
```diff
@@ -50,8 +50,25 @@ class DiagnoseCommand extends Command<bool> {
       return false;
```
christopherfujino, line 50: Maybe rather than returning `false`, we should collect the results of both this AND `DevToolsSecurity` and report both?

device_doctor/lib/src/ios_debug_symbol_doctor.dart
vashworth: Diagnostic, does `severe` block the build?

device_doctor/lib/src/ios_debug_symbol_doctor.dart
vashworth: Sure, I'll just do the `DevToolsSecurity` check first

device_doctor/lib/src/ios_debug_symbol_doctor.dart
christopherfujino: nope it doesn't. if this is diagnostic only then this approach looks fine

christopherfujino (approved):
LGTM

PR #2802, opened 2023-05-24
Split `file_flaky_issue_and_pr` to multiple cron jobs

keyonghan:
This is a follow up of https://github.com/flutter/cocoon/pull/2725, which scheduled the cronjob as 
```
schedule: every wednesday 9:00,9:30,10:00,10:30,11:00
```

However, we can either schedule using `Sub-daily intervals`:
```
schedule: every 30 minutes from 9:00 to 11:00
```
or schedule using `Custom interval` at a specific time:
```
schedule: every wednesday 9:00
```

Unfortunately we can not schedule mixing these two to run on multiple times within an interval on a specific day. This PR splits the schedule to multiple entries based on the custom internal.

Context: https://cloud.google.com/appengine/docs/legacy/standard/python/config/cronref#defining_the_cron_job_schedule

### DIFF
--- a/cron.yaml
+++ b/cron.yaml
@@ -34,9 +34,25 @@ cron:
   url: /api/public/github-rate-limit-status
   schedule: every 1 minutes
 
-- description: detect and flag tests with high flaky rates
+- description: detect and flag tests with high flaky rates 1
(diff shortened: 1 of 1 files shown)


ricardoamador (approved):
LGTM

PR #2812, opened 2023-05-26
[dashboard] Remove View Logs button

CaseyHillers:
This button is no longer used in LUCI world

A few new users to the build dashboard were confused by the view logs button. They expected it would show a link of all the builds for that commit, but it redirects to the builder page. This can already be done via the column header icon.


ricardoamador:
Can you please add more context to the description here? The VIEW LOGS button takes you currently to the overview of that task with all the builds, for example:

image

So I am not sure what you mean by:

> is no longer used in LUCI world

CaseyHillers:
I added some comment in the description. This is legacy from pre-LUCI days when Cirrus and the legacy DeviceLab allowed you to download all logs for a given commit.

Clicking the column header takes to the full builds.

image

Let me know what you think, but I can change it to recent builds as well.

ricardoamador:
Ah thanks for the context. I am fine if you remove it since clicking the icon takes you to the same page. No sense in keeping the button.

ricardoamador (approved):
LGTM - Let's remove the button to avoid confusion.

PR #2826, opened 2023-06-07
WIP: Process buildbucket results

drewroengoogle:
Still WIP (Need to refactor)
https://github.com/flutter/flutter/issues/124629

## Pre-launch Checklist

- [ ] I signed the [CLA].


app_dart/lib/src/request_handlers/dart_internal_subscription.dart
```diff
@@ -2,31 +2,208 @@
...
+  final RetryOptions retryOptions;
+
   @override
   Future<Body> post() async {
-    log.fine('Consumed message: ${message.data}');
+    final DatastoreService datastore = datastoreProvider(config.db);
+
+    log.info("Converting message data to non nullable int");
+
+    final int buildbucketId = int.parse(message.data.toString().replaceAll("'", ""));
```
godofredoc, line 51: Should we pass the messages as json? that way we don't need to do this adhoc parsing.

app_dart/lib/src/request_handlers/dart_internal_subscription.dart
```diff
@@ -2,31 +2,208 @@
...
+    );
+    late Build build;
+    try {
+      build = await retryOptions.retry(
+        () async {
+          return _getBuildFromBuildbucket(request);
+        },
+        retryIf: (Exception e) => e is UnfinishedBuildException,
+      );
+    } catch (e) {
```
godofredoc, line 67: Do we need the try catch if we are throwing below?

app_dart/lib/src/request_handlers/dart_internal_subscription.dart
```diff
@@ -2,31 +2,208 @@
...
+      if (e is! UnfinishedBuildException) {
```
godofredoc, line 68: Why do we need this if?

app_dart/lib/src/request_handlers/dart_internal_subscription.dart
```diff
@@ -2,31 +2,208 @@
...
+          "Failed to get build data for build $buildbucketId with the exception: ${e.toString()}",
+        );
+        throw InternalServerError(
+          'Failed to get build number $buildbucketId from Buildbucket. Error: $e',
+        );
+      }
+    }
+
+    log.info("Checking for existing task in datastore");
+    late Task? existingTask;
```
godofredoc, line 79: rm?

app_dart/lib/src/request_handlers/dart_internal_subscription.dart
```diff
@@ -2,31 +2,208 @@
...
+    existingTask = await _getExistingTaskFromDatastore(build, datastore);
```
godofredoc, line 80: final Task? existingTask = ...

app_dart/lib/src/request_handlers/dart_internal_subscription.dart
```diff
@@ -2,31 +2,208 @@
...
+
+    late Task taskToInsert;
+    if (existingTask != null) {
+      log.info("Creating Task from existing Task");
```
godofredoc, line 84: Updating task?

app_dart/lib/src/request_handlers/dart_internal_subscription.dart
```diff
@@ -2,31 +2,208 @@
...
+    log.info("Inserting Task into the datastore: ${taskToInsert.toString()}");
+    await datastore.insert(<Task>[taskToInsert]);
+
+    return Body.forJson(taskToInsert.toString());
+  }
+
+  Future<Build> _getBuildFromBuildbucket(
+    GetBuildRequest request,
+  ) async {
+    late Build build;
```
godofredoc, line 102: Here and everywhere, do not use late when you are initializing in the next line.

app_dart/lib/src/request_handlers/dart_internal_subscription.dart
```diff
@@ -2,31 +2,208 @@
...
+    build = await buildBucketClient.getBuild(request);
+
+    if (build.endTime == null) {
```
godofredoc, line 105: Would it be easier to use state rather than end time?

app_dart/lib/src/request_handlers/dart_internal_subscription.dart
```diff
@@ -2,31 +2,208 @@
...
+        "Build is not finished",
+      );
+      throw UnfinishedBuildException(
+        "Build is not finished",
+      );
+    }
+    return build;
+  }
+
+  Future<Task> _createTaskFromBuildbucketResult(
```
godofredoc, line 116: would task be a better place for this function?

app_dart/lib/src/request_handlers/dart_internal_subscription.dart
```diff
@@ -2,31 +2,208 @@
...
+    final String hash = build.input!.gitilesCommit!.hash!;
+    log.fine("Hash: $hash");
+
+    final RepositorySlug slug = RepositorySlug("flutter", repository);
+    log.fine("Slug: ${slug.toString()}");
+
+    final String id = 'flutter/${slug.name}/$branch/$hash';
+    final Key<String> commitKey = datastore.db.emptyKey.append<String>(Commit, id: id);
+
+    try {
```
godofredoc, line 188: Why we need a try catch?

app_dart/lib/src/request_handlers/dart_internal_subscription.dart
```diff
@@ -2,31 +2,208 @@
...
+      return await Task.fromDatastore(datastore: datastore, commitKey: commitKey, name: build.builderId.builder);
+    } catch (e) {
+      return null;
+    }
+  }
 
-    return Body.empty;
+  String _convertStatusToString(Status status) {
```
godofredoc, line 195: Does this belong to the model classes?

app_dart/lib/src/request_handlers/dart_internal_subscription.dart
drewroengoogle: My intention with this is to be able to try getting the build a couple times in case it hasn't finished. However, in the case that the build is still not complete, but the build doesn't represent that, I didn't want to fail the build.

Retry does not support retrying based on the returned object, would it make sense to add a custom retry instead?

app_dart/lib/src/request_handlers/dart_internal_subscription.dart
drewroengoogle: Currently, if we call fromDatastore and it cannot find a task, it will throw an error. If we don't find an existing task, we don't want it to fail, we just want it to return null since there isn't an existing task.

app_dart/lib/src/request_handlers/dart_internal_subscription.dart
drewroengoogle: I've updated it to fail if we are unable to get a complete build after three attempts.

app_dart/lib/src/request_handlers/dart_internal_subscription.dart
drewroengoogle: Try/catch removed

app_dart/lib/src/model/appengine/task.dart
```diff
@@ -129,6 +131,79 @@ class Task extends Model<int> {
...
+      attempts: 1,
+      buildNumber: build.number,
+      buildNumberList: build.number.toString(),
+      builderName: build.builderId.builder,
+      commitKey: key,
+      createTimestamp: startTime,
+      endTimestamp: endTime,
+      luciBucket: build.builderId.bucket,
+      name: build.builderId.builder,
+      stageName: "dart-internal",
```
godofredoc, line 177: Is this something we need to pass to the method instead of hardcoding?

app_dart/lib/src/model/appengine/task.dart
```diff
@@ -129,6 +131,79 @@ class Task extends Model<int> {
...
   }
 
+  /// Creates a task based on a buildbucket build.
+  static Future<Task> fromBuildbucketResult(
+    bb.Build build,
+    DatastoreService datastore,
+  ) async {
+    log.fine("Creating task from buildbucket result: ${build.toString()}");
+
+    final String repository = build.input!.gitilesCommit!.project!.split('/')[1];
```
ricardoamador, line 141: I would assume that you are trying to split the full repo name here but can you it might make sense to add a comment for others to state specifically what you are doing.

app_dart/lib/src/model/appengine/task.dart
```diff
@@ -129,6 +131,79 @@ class Task extends Model<int> {
...
+    log.fine("Repository: $repository");
+
+    final String branch = build.input!.gitilesCommit!.ref!.split('/')[2];
```
ricardoamador, line 144: Same comment with this guy too. But this one is less intuitive. Also you might want to add in a comment the format of the thing you are trying to split. I am guessing it is a string separated by '/'.

app_dart/lib/src/model/appengine/task.dart
```diff
@@ -129,6 +131,79 @@ class Task extends Model<int> {
...
+    log.fine("Start/end time (ms): $startTime, $endTime");
+
+    final Key<String> key = Commit.createKey(
+      db: datastore.db,
+      slug: slug,
+      gitBranch: branch,
+      sha: hash,
+    );
+
+    final String id = 'flutter/${slug.name}/$branch/$hash';
```
ricardoamador, line 164: This could be shortened to '${slug.fullname}/$branch/$hash'.

app_dart/lib/src/request_handlers/dart_internal_subscription.dart
ricardoamador: Can we add the specific Exception/Error here?

ricardoamador: It would also be good to log the exception/error.

app_dart/lib/src/request_handlers/dart_internal_subscription.dart
```diff
@@ -2,31 +2,126 @@
...
+  final RetryOptions retryOptions;
+
   @override
   Future<Body> post() async {
-    log.fine('Consumed message: ${message.data}');
+    final DatastoreService datastore = datastoreProvider(config.db);
+
+    log.info("Converting message data to non nullable int");
+
+    final int buildbucketId = int.parse(json.decode(message.data.toString())['buildbucket_id']);
```
ricardoamador, line 52: I might have missed something but where is message defined previous to this access?

app_dart/lib/src/request_handlers/dart_internal_subscription.dart
```diff
@@ -2,31 +2,126 @@
...
+
+    return Body.forJson(taskToInsert.toString());
+  }
+
+  Future<Build> _getBuildFromBuildbucket(
+    GetBuildRequest request,
+  ) async {
+    final Build build = await buildBucketClient.getBuild(request);
+
+    if (build.status == Status.started) {
```
ricardoamador, line 92: What are the possible states here? I would assume more than one terminal state? Might be better to look for those instead if you are testing for finished.

app_dart/lib/src/request_handlers/dart_internal_subscription.dart
```diff
@@ -2,31 +2,126 @@
...
+
+    final int buildbucketId = int.parse(json.decode(message.data.toString())['buildbucket_id']);
+
+    log.info("Creating build request object");
+    final GetBuildRequest request = GetBuildRequest(id: buildbucketId.toString());
+
+    log.info(
+      "Calling buildbucket api to get build data for build $buildbucketId",
+    );
+    final Build build = await retryOptions.retry(
```
ricardoamador, line 60: Should be able to push the retry logic down in the _getBuildFromBuildBucket function to make this a little cleaner.

app_dart/lib/src/request_handlers/dart_internal_subscription.dart
```diff
@@ -2,31 +2,126 @@
...
+
+    log.info("Checking for existing task in datastore");
+    final Task? existingTask = await _getExistingTaskFromDatastore(build, datastore);
+
+    late Task taskToInsert;
+    if (existingTask != null) {
+      log.info("Updating Task from existing Task");
+      existingTask.buildNumber = build.number;
+      existingTask.buildNumberList = '${build.number},${existingTask.buildNumberList}';
+      taskToInsert = existingTask;
```
keyonghan, line 75: can we reuse `task.updateFromBuild`?

app_dart/lib/src/request_handlers/dart_internal_subscription.dart
```diff
@@ -2,31 +2,126 @@
...
+        "Build is not finished",
+      );
+      throw UnfinishedBuildException(
+        "Build is not finished",
+      );
+    }
+    return build;
+  }
+
+  Future<Task?> _getExistingTaskFromDatastore(Build build, DatastoreService datastore) async {
```
keyonghan, line 103: I would put this logic to `lib/src/service/datastore.dart`

app_dart/lib/src/request_handlers/dart_internal_subscription.dart
keyonghan: Is this the message published to pubsub from recipes side? It would be good to add some doc/link.

app_dart/lib/src/request_handlers/dart_internal_subscription.dart
drewroengoogle: Good point. I updated it to check for success/failure/canceled instead.

app_dart/lib/src/model/appengine/task.dart
```diff
@@ -129,6 +131,79 @@ class Task extends Model<int> {
...
+    log.fine("Hash: $hash");
+
+    final RepositorySlug slug = RepositorySlug("flutter", repository);
+    log.fine("Slug: ${slug.toString()}");
+
+    final int startTime = build.startTime != null ? build.startTime!.millisecondsSinceEpoch : 0;
+    final int endTime = build.endTime != null ? build.endTime!.millisecondsSinceEpoch : 0;
+    log.fine("Start/end time (ms): $startTime, $endTime");
+
+    final Key<String> key = Commit.createKey(
```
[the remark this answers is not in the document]
drewroengoogle: They are basically the same, but there should only be one. I've removed `key` and now use `commitKey`.

app_dart/lib/src/request_handlers/dart_internal_subscription.dart
drewroengoogle: I've added a comment above this line linking to the `engine_v2` recipe that publishes the build message!

app_dart/lib/src/service/exceptions.dart
```diff
@@ -11,3 +11,13 @@ class NoBuildFoundException implements Exception {
   @override
   String toString() => cause;
 }
+
+class UnfinishedBuildException implements Exception {
```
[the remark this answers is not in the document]
drewroengoogle: Ideally the retry package would be able to handle that, but it currently only supports retrying on a specific `Exception`, so creating `UnfinishedBuildException` lets us retry when the build isn't yet complete.

app_dart/lib/src/request_handlers/dart_internal_subscription.dart
drewroengoogle: I've updated it to catch only `InternalServerError`, and log that error if it occurs!

app_dart/lib/src/request_handlers/dart_internal_subscription.dart
keyonghan: Yeah, you are right. 
Do you need to update other task info, like status, createTime, etc?

drewroengoogle: To solve this, I ended up moving the update logic to `task.dart` under a new method, `updateFromBuildBucket`, so now the existing task from the datastore is directly updated based on the build, and inserted back into the datastore. Let me know what you think

app_dart/lib/src/model/appengine/task.dart
```diff
@@ -339,6 +408,27 @@ class Task extends Model<int> {
...
+      final Set<String> buildNumberSet = buildNumberList!.split(',').toSet();
+      buildNumberSet.add(buildNumber.toString());
+      buildNumberList = buildNumberSet.join(',');
+    }
+
+    createTimestamp = build.startTime?.millisecondsSinceEpoch ?? 0;
+    startTimestamp = build.startTime?.millisecondsSinceEpoch ?? 0;
+    endTimestamp = build.endTime?.millisecondsSinceEpoch ?? 0;
+
+    attempts = (attempts ?? 0) + 1;
```
keyonghan, line 427: Is each update here a rerun? We do not want to bump the `attempts` if this is just a status change, say `scheduled` -> `started`.

Usually we increase the `attempts` only when schedule/rerun a target.

app_dart/lib/src/request_handlers/dart_internal_subscription.dart
keyonghan: From the linked example, the injected message is simply the id `api.buildbucket.build.id`? Here from the subscription, it seems expecting a json.

app_dart/lib/src/service/exceptions.dart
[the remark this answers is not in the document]
ricardoamador: I agree with this as you are throwing a distinct exception. I do this in Autosubmit as it keeps the code clean and tells other devs what you are specifically doing.

app_dart/lib/src/request_handlers/dart_internal_subscription.dart
drewroengoogle: Ah good catch. https://flutter-review.googlesource.com/c/recipes/+/45980 should be merged soon, I'll update the comment to point to this instead!

PR #2829, opened 2023-06-08
Add retries to builds that are cancelled

drewroengoogle:
https://github.com/flutter/flutter/issues/128469

* Retries builds that are cancelled
* Adds a unit test to check for retrying a task on an infra failure

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
+++ b/app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
@@ -87,7 +87,7 @@ class PostsubmitLuciSubscription extends SubscriptionHandler {
     final Commit commit = await datastore.lookupByValue<Commit>(commitKey);
     final CiYaml ciYaml = await scheduler.getCiYaml(commit);
     final Target target = ciYaml.postsubmitTargets.singleWhere((Target target) => target.value.name == task!.name);
-    if (task.status == Task.statusFailed || task.status == Task.statusInfraFailure) {
+    if (task.status == Task.statusFailed || task.status == Task.statusInfraFailure || _taskTimedOut(task)) {
       log.fine('Trying to auto-retry...');
       final bool retried = await scheduler.luciBuildService.checkRerunBuilder(
         commit: commit,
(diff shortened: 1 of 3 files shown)


ricardoamador:
It might be wise to limit the number of reruns.

app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
```diff
@@ -87,7 +87,7 @@ class PostsubmitLuciSubscription extends SubscriptionHandler {
     final Commit commit = await datastore.lookupByValue<Commit>(commitKey);
     final CiYaml ciYaml = await scheduler.getCiYaml(commit);
     final Target target = ciYaml.postsubmitTargets.singleWhere((Target target) => target.value.name == task!.name);
-    if (task.status == Task.statusFailed || task.status == Task.statusInfraFailure) {
+    if (task.status == Task.statusFailed || task.status == Task.statusInfraFailure || _taskTimedOut(task)) {
```
keyonghan, line 90: We should use `Task.statusCancelled` instead.

keyonghan:
> It might be wise to limit the number of reruns. Or is this limited elsewhere?

It is controlled by `maxLuciTaskRetries`:
https://github.com/flutter/cocoon/blob/main/app_dart/lib/src/service/config.dart#L297

app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
drewroengoogle: Do we want to always retry if a build is canceled? Eg: If I went in and manually canceled a build, should we expect it to be retried?

drewroengoogle: That probably isn't a reasonable use case, since that would block the tree. I've updated it to check for `Task.statusCancelled` instead!

ricardoamador (approved):
LGTM

app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
keyonghan: Thanks for the update. Yeah, post-submit CI is not expecting much human behavior, except the `manual retry`.
Some other `cancelled` cases are like `bot_died`, luci incident`, etc, and we want to treat them same as `failures` and and to auto retry to green the tree.

PR #701, opened 2020-03-20
Remove all stray prints in tests.

Hixie:

PR #2835, opened 2023-06-12
Initial Flutter Releases App

CaseyHillers:
https://github.com/flutter/flutter/issues/132082

## Pre-launch Checklist

- [X] I signed the [CLA].


CaseyHillers:
\cc @ricardoamador

releases/lib/common/router.dart
```diff
@@ -0,0 +1,111 @@
...
+import 'dart:async';
+import 'dart:io';
+
+import 'package:meta/meta.dart';
+import 'package:shelf/shelf_io.dart';
+
+import '../service/log.dart';
+import 'handler.dart';
+
+class Router {
```
CaseyHillers, line 14: @ricardoamador I think most of the common source code would be applicable to a microservice package. I tried to make it light weight to set up the server based off how some of the golang server frameworks I've used.

ricardoamador:
Taking a look.

releases/bin/server.dart
```diff
@@ -0,0 +1,23 @@
...
+import 'dart:async';
+
+import 'package:appengine/appengine.dart';
+import 'package:releases/common/handler.dart';
+import 'package:releases/common/router.dart';
+import 'package:releases/requests/github_webhook.dart';
+import 'package:releases/requests/readiness_check.dart';
+
+Future<void> main() async {
+  await withAppEngineServices(() async {
```
ricardoamador, line 14: Since the motivation for this library is to make a package that is externally accessible then we should add the ability to register endpoints with this router.

releases/lib/common/context.dart
```diff
@@ -0,0 +1,12 @@
+// Copyright 2019 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/// Context for the running server, providing injections for various services.
+class Context {
```
ricardoamador, line 6: Are you envisioning this as having the same roll as the Config.dart class has currently?

releases/lib/common/handler.dart
```diff
@@ -0,0 +1,38 @@
...
+  Future<Response> get(Context context, Request request) async {
+    return Response(501, body: 'GET ${request.url.path} not implemented');
+  }
+
+  /// Services a POST request.
+  ///
+  /// Subclasses should override this method if they support POST requests.
+  ///
+  /// An unimplemented response will be returned if there is no concrete implementation.
+  Future<Response> post(Context context, Request request) async {
```
ricardoamador, line 35: Might want to add the other verbs (PUT, PATCH, DELETE) to make this complete.

releases/lib/common/context.dart
CaseyHillers: Yes, but making it more as the way dependencies are injected. I still see the need for a `Config` service for having dynamic feature rollout.

Next PR we can add `context.getDatastore()` and `context.getGithub()`, which can be reusable in tests.

releases/lib/common/handler.dart
CaseyHillers: I'm not sure we'd use the HTTP methods. We only use GET and POST in Cocoon now. I think if we need it, we can add it in future PRs.

https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it

releases/bin/server.dart
CaseyHillers: This releases project won't be the common library.

My expectation is that `bin/server.dart` will still list all the server specific handlers to be added, and `Router` + `Handler` can come from `package:cocoon`

ricardoamador: Okay that makes sense and does create a delineation between what you are doing and what I am seeking to deliver.

cloud_build/deploy_releases.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+# Copyright 2020 The Flutter Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+# Deploy a new releases version to google cloud.
+
+pushd releases > /dev/null
+# auto_submit
```
keyonghan, line 9: Update.

cloud_build/deploy_releases.sh
```diff
@@ -0,0 +1,15 @@
...
+gcloud app deploy --project "$1" --version "version-$2" -q --no-promote --no-stop-previous-version --image-url "us-docker.pkg.dev/$1/appengine/auto-submit.version-$2"
```
keyonghan, line 10:
```suggestion
gcloud app deploy --project "$1" --version "version-$2" -q --no-promote --no-stop-previous-version --image-url "us-docker.pkg.dev/$1/appengine/releases.version-$2"
```

keyonghan (commented):
The `app.yaml` is missing.

cloud_build/deploy_releases.sh
```diff
@@ -0,0 +1,15 @@
...
+gcloud app services set-traffic releases --splits version-$2=1 --quiet
+# Google Cloud quota only allows 30 versions.
+# For daily builds, this will keep the 10 most recent versions, but wipe the rest.
+gcloud app versions list --format="value(version.id)" --sort-by="~version.createTime"  --service=auto-submit | tail -n +20 | xargs -r gcloud app versions delete --quiet
```
ricardoamador, line 14: Just do a find and replace on these files for auto-submit.

releases/lib/common/handler.dart
```diff
@@ -0,0 +1,38 @@
+// Copyright 2019 The Flutter Authors. All rights reserved.
```
ricardoamador, line 1: Update the dates in the headers.

releases/lib/service/log.dart
```diff
@@ -0,0 +1,7 @@
+// Copyright 2022 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:logging/logging.dart';
+
+final Logger log = Logger('auto_submit');
```
ricardoamador, line 7: Change this to 'releases'

releases/lib/common/handler.dart
CaseyHillers: Since this is derivative of previous work in the repo, it should be based on the original dates.

See https://github.com/flutter/cocoon/pull/701#discussion_r395447338

yusuf-goog:
ping on this

ricardoamador:
Looks like there are a few feedbacks to address but otherwise it should be good to go.

CaseyHillers:
Converting to draft. I'm playing around with Google Testing using gRPC to see how that is before I commit to it here.

PR #2846, opened 2023-06-14
Updated dashboard/test/widgets tests to handle useMaterial3:true by default

HansMuller:
Updated all of the golden image file tests in dashboard/test/widgets to specify `useMaterial3: false` to ensure that the images continue to match when the default for useMaterial3 is changed to true.


christopherfujino (approved):
LGTM

PR #2847, opened 2023-06-15
[app_dart] Skip needs tests on branches

CaseyHillers:
Fixes https://github.com/flutter/flutter/issues/127808

### DIFF
--- a/app_dart/lib/src/request_handlers/github/webhook_subscription.dart
+++ b/app_dart/lib/src/request_handlers/github/webhook_subscription.dart
@@ -27,10 +27,10 @@ const List<String> kNotActuallyATest = <String>[
 ];
 
 /// List of repos that require check for tests.
-const Set<String> kNeedsTests = <String>{
-  'flutter/engine',
-  'flutter/flutter',
-  'flutter/packages',
+Set<RepositorySlug> kNeedsTests = <RepositorySlug>{
+  Config.engineSlug,
+  Config.flutterSlug,
+  Config.packagesSlug,
 };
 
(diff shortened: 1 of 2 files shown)


app_dart/lib/src/request_handlers/github/webhook_subscription.dart
```diff
@@ -253,17 +253,19 @@ class GithubWebhookSubscription extends SubscriptionHandler {
...
-        } else if (repo == 'flutter/engine') {
-          await _applyEngineRepoLabels(gitHubClient, eventAction, pr);
-        } else if (repo == 'flutter/plugins' || repo == 'flutter/packages') {
-          await _applyPackageTestChecks(gitHubClient, eventAction, pr);
+        switch (slug.name) {
+          case 'flutter':
+            return _applyFrameworkRepoLabels(gitHubClient, eventAction, pr);
+          case 'engine':
+            return _applyEngineRepoLabels(gitHubClient, eventAction, pr);
+          case 'packages':
```
XilaiZhang, line 267: Umm why did we remove the plugins repo

XilaiZhang (approved):

app_dart/lib/src/request_handlers/github/webhook_subscription.dart
CaseyHillers: See https://flutter.dev/go/flutter-plugins-repo-migration

The plugins repo was migrated to packages. There shouldn't be any migrations for the plugins other than removing outdated tests.

XilaiZhang: thanks for explaining Casey!

sorry for bothering Casey when Casey is on vacation

PR #2872, opened 2023-06-28
Adjust to new priority schemes

Hixie:
We recently changed the priority labels. This updates code in cocoon accordingly.


keyonghan (approved):
LGTM

PR #2875, opened 2023-06-28
Replacing luci-flutter with tree-status as context name in GitHub presubmit

keyonghan:
Part of https://github.com/flutter/flutter/issues/92931

Follow up is to clean up old `luci-flutter`/`luci-engine` names from auto-submit.


ricardoamador (approved):
LGTM

PR #2906, opened 2023-06-30
Add team label to flaky pull requests

keyonghan:
Fixes: https://github.com/flutter/flutter/issues/112891

### DIFF
--- a/app_dart/lib/src/service/github_service.dart
+++ b/app_dart/lib/src/service/github_service.dart
@@ -153,6 +153,20 @@ class GithubService {
     );
   }
 
+  /// Adds labels to an issue.
+  ///
+  /// A pull request is an issue. This works for pull requests as well.
+  Future<List<IssueLabel>> addIssueLabels(
+    RepositorySlug slug,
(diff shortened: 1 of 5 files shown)


app_dart/lib/src/service/github_service.dart
```diff
@@ -153,6 +153,20 @@ class GithubService {
...
 
+  /// Adds labels to an issue.
+  ///
+  /// A pull request is an issue. This works for pull requests as well.
+  Future<List<IssueLabel>> addIssueLabels(
+    RepositorySlug slug,
+    int issueNumber,
+    List<String> labels,
+  ) async {
+    ArgumentError.checkNotNull(slug);
```
ricardoamador, line 164: Does this throw? I am assuming this is similar to using assert?

ricardoamador (approved):
LGTM

app_dart/lib/src/service/github_service.dart
keyonghan: Yes. Exactly.

PR #2916, opened 2023-07-06
Fix the Merge check from using stale data before attempting to merge in Autosubmit

ricardoamador:
Almost all places in autosubmit are using the newest data via a query to the graphql api, however one spot was causing an issue with mergeability as it was still using stale data from the original PR message. This fix addresses that so that we are not blocking on the very last mergeability check.

Fixes https://github.com/flutter/flutter/issues/130089

## Pre-launch Checklist

- [x] I signed the [CLA].


auto_submit/lib/service/validation_service.dart
```diff
@@ -353,35 +369,60 @@ ${messagePullRequest.title!.replaceFirst('Revert "Revert', 'Reland')}
...
-    bool result = true;
-    String message = 'Pull request ${slug.fullName}/$pullRequestNumber is mergeable';
-    if (pullRequest.mergeable == null) {
-      message =
-          'Mergeability of pull request ${slug.fullName}/$pullRequestNumber could not be determined at time of merge.';
-      result = false;
-    } else if (pullRequest.mergeable == false) {
-      result = false;
-      message = 'Pull request ${slug.fullName}/$pullRequestNumber is not in a mergeable state.';
+  // Future<ProcessMergeResult> isMergeable(github.RepositorySlug slug, int pullRequestNumber) async {
```
yusuf-goog, line 385: update doc: Future<ProcessMergeResult>

ricardoamador: Ah good catch! Thank you.

auto_submit/lib/service/validation_service.dart
```diff
@@ -194,12 +194,19 @@ class ValidationService {
     }
 
     // If we got to this point it means we are ready to submit the PR.
-    final ProcessMergeResult processed = await processMerge(
+    final ValidationResult processed = await processMerge(
```
keyonghan, line 197: Shall we create a new validation for it as other validations to simplify the logic? considering here it is being treated as a validation.

ricardoamador: You mean a new wrapper per se? Yeah I can do that so it is not confused with the ValidationResult.

ricardoamador: Done. Added a Record!

keyonghan: I meant moving the mergeability logic as a `validation` check. Seems we already have a `conflicting` validation, shall we merge the new logic with it?

keyonghan: https://github.com/flutter/cocoon/blob/main/auto_submit/lib/validations/conflicting.dart

ricardoamador: Yeah it technically already is. I thought about this and if you are fine with it I will remove this pre-merge altogether. This was added because there was a request to make a check right before the merge.

https://github.com/flutter/flutter/issues/112018

Otherwise with the previous changes we are essentially doing this if we move it to the validations. We check unknown mergeable and conflicting mergeable already which leaves only the positive state Mergeable.

ricardoamador: I'd prefer removing code if possible.

keyonghan: Yeah, I would move the logic to `conflicting` validation and rename it to say `mergeability`.

With the proposed logic here:
```
switch (mergeableState) {
      case MergeableState.MERGEABLE:
        return (
          result: true,
          action: Action.REMOVE_LABEL,
          message: 'Pull request ${slug.fullName}/$pullRequestNumber is mergeable',
        );
      case MergeableState.UNKNOWN:
        return (
          result: false,
          action: Action.IGNORE_TEMPORARILY,
          message:
              'Mergeability of pull request ${slug.fullName}/$pullRequestNumber could not be determined at time of merge.',
        );
      case MergeableState.CONFLICTING:
        return (
          result: false,
          action: Action.REMOVE_LABEL,
          message: 'Pull request ${slug.fullName}/$pullRequestNumber is not in a mergeable state.',
        );
    }
```
it is enough to cover the the request from https://github.com/flutter/flutter/issues/112018. If it's unknown, then we ignore_temporarily.

Does this make sense to you?

ricardoamador: Yeah this makes sense. This will be a huge cleanup and will simplify a lot of things.

auto_submit/lib/validations/mergeable.dart
```diff
@@ -0,0 +1,41 @@
...
+        );
+      case MergeableState.CONFLICTING:
+        return ValidationResult(
+          false,
+          Action.REMOVE_LABEL,
+          'Pull request ${slug.fullName}/$pullRequestNumber is not in a mergeable state.',
+        );
+    }
+  }
+}
```
keyonghan: Nit: an extra line.

keyonghan (approved):
LGTM. Just a tiny nit.

PR #2934, opened 2023-07-14
Skip scheduling recent failed tasks if an earlier task is already InProgress

keyonghan:
Fixes: https://github.com/flutter/flutter/issues/110716

### DIFF
--- a/app_dart/lib/src/service/scheduler/policy.dart
+++ b/app_dart/lib/src/service/scheduler/policy.dart
@@ -2,6 +2,8 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'dart:math' as math;
+
 import 'package:cocoon_service/src/service/datastore.dart';
 
 import '../../model/appengine/task.dart';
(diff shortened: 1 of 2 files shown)


ricardoamador (approved):
LGTM

app_dart/lib/src/service/scheduler/policy.dart
```diff
@@ -56,6 +56,11 @@ class BatchPolicy implements SchedulerPolicy {
     required DatastoreService datastore,
   }) async {
     final List<Task> recentTasks = await datastore.queryRecentTasksByName(name: task.name!).toList();
+    // Skip scheduling if there is already a running task.
+    if (recentTasks.any((Task task) => task.status == Task.statusInProgress)) {
```
ricardoamador, line 60: Not a code review comment but is it possible that a running task can still be active outside the "last 6 records" fetched? Meaning a task is running but the most recent 6 show a status other than Task.statusInProgress?

keyonghan: That's a very good point, and the answer is yes.
We will skip scheduling that case as well. At any specific time, we want only one task is running (In Progress).
If we have an old task running (beyond the 6), we will skip it in the webhook scheduling. When it finishes, we rely on the back filler to pick up the most recent New task to run.

ricardoamador: Ah okay cool! Thanks for the answer!

PR #2950, opened 2023-07-25
Jobs not progressing 131192

ricardoamador:
Removing the explicit commit seems to have fixed the issue. There must be change to the underlying datastore implementation that must be committing the transaction. Removing the explicit transaction ceases the transaction aborted errors. We must have been attempting to issue the commit on its own resulting in the transaction aborted error.

Running a test instance for 4+ hours and have not seen the 500 issue.

Last 6 hours on test instance:
image

Last 6 hours on non test instance:
image

Will check again tomorrow in the morning on error count. Splitting traffic 50/50 for overnight.

Part of https://github.com/flutter/flutter/issues/131192

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/app_dart/lib/src/service/datastore.dart
+++ b/app_dart/lib/src/service/datastore.dart
@@ -247,15 +247,10 @@ class DatastoreService {
   Future<void> insert(List<Model<dynamic>> rows) async {
     final List<List<Model<dynamic>>> shards = await shard(rows);
     for (List<Model<dynamic>> shard in shards) {
-      await runTransactionWithRetries(
-        () async {
-          await db.withTransaction<void>((Transaction transaction) async {
+      // Simplify the transaction and commit and remote the retries for now.
+      await db.withTransaction<void>((Transaction transaction) async {
             transaction.queueMutations(inserts: shard);
-            await transaction.commit();
           });
-        },
-        retryOptions: retryOptions,
-      );
     }
   }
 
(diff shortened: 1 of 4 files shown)


app_dart/lib/src/service/datastore.dart
```diff
@@ -247,15 +247,10 @@ class DatastoreService {
   Future<void> insert(List<Model<dynamic>> rows) async {
     final List<List<Model<dynamic>>> shards = await shard(rows);
     for (List<Model<dynamic>> shard in shards) {
-      await runTransactionWithRetries(
```
keyonghan, line 250: Can we keep the retry and remove the `transaction.commit()` only?

ricardoamador: I suspect that we might be getting a commit from somewhere else or that Datastore is committing these transactions. I have not found for sure why this specific commit is causing the issue but that intuitively makes sense. 

Now with the retries I am not sure this makes any difference since if we look at the code we retryIf on AbortedTransactionError which we see and which means since are not wrapping the full call we are just issuing the commit every time. I would rather track it down first than just add the retry back in at this point.

keyonghan: Huh, I am revisiting why we added the transaction.commit() logic, which was intended to fix the similar/same issue: https://github.com/flutter/flutter/issues/121413 `The transaction was aborted.`

Let's land it here as it mitigates our issues now. But something sounds we need to follow up with Datastore team to find out the culprit.

keyonghan (approved):
LGTM

ricardoamador:
Closing as this is not a fix to the issue.

PR #2964, opened 2023-07-31
Fix datastore task update for post-submit builds

keyonghan:
This PR fixes a couple of things:

1) skips task updates for build `scheduled` case
    a) All tasks will be marked as `in progress` whenever scheduled. This skips the 400 errors: `Failed to process Instance of 'PushMessage'. (400) null is unknown`
    b) This skips datastore updates when a `scheduled` message comes after a `completed` message. Our pub/sub subscriber was created with `Message ordering: Disabled`, and we need to make sure [`idempotency`](https://en.wikipedia.org/wiki/Idempotence#Computer_science_meaning) from our consumer side.
2) skips task updates for task which has already finished
3) Removes the buggy logic to reset task status as `New`. 
4) Adds explicit task status in log before and after the update.

Helps: https://github.com/flutter/flutter/issues/131192

### DIFF
--- a/app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
+++ b/app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
@@ -80,9 +80,19 @@ class PostsubmitLuciSubscription extends SubscriptionHandler {
     }
     log.fine('Found $task');
 
+    // No need to process if the build is `scheduled`. Task is marked as `In Progress`
+    // whenever scheduled, either from scheduler/backfiller/rerun. We need to update
+    // task in datastore only for
+    //   1) `started`: update info like builder number.
+    //   2) `completed`: update info like status.
+    if (build.status == Status.scheduled) {
+      log.fine('skip processing for status: scheduled.');
+      return Body.empty;
+    }
+    final String oldTaskStatus = task.status;
     task.updateFromBuild(build);
     await datastore.insert(<Task>[task]);
-    log.fine('Updated datastore');
+    log.fine('Updated datastore from $oldTaskStatus to ${task.status}');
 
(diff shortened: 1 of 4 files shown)


app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
```diff
@@ -80,9 +80,21 @@ class PostsubmitLuciSubscription extends SubscriptionHandler {
     }
     log.fine('Found $task');
 
+    // No need to process if
+    // 1) the build is `scheduled`. Task is marked as `In Progress`
```
ricardoamador, line 84: What is marking it as in-progress?

ricardoamador: Ah I see, the next time around this check will be made.

ricardoamador (approved):
LGTM but lets wait to merge this to see what happens with the deployed test instance so we don't have too many variables to deal with right now.

app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
keyonghan: 1. For retry: https://github.com/flutter/cocoon/blob/main/app_dart/lib/src/service/luci_build_service.dart#L672
2. For scheduler: https://github.com/flutter/cocoon/blob/main/app_dart/lib/src/service/scheduler.dart#L147
3. For backfiller: https://github.com/flutter/cocoon/blob/main/app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart#L236

keyonghan:
> LGTM but lets wait to merge this to see what happens with the deployed test instance so we don't have too many variables to deal with right now.

SGTM.

PR #2970, opened 2023-08-01
Restore auto rerun on failures

keyonghan:
Follow up of https://github.com/flutter/cocoon/pull/2964, which exit early also for the failed case. It causes auto-retry skipped.

This PR reorg the logics and restores the rerun logic back for the failed case.


ricardoamador (approved):
LGTM

PR #2972, opened 2023-08-02
Add Doxygen build script

gspencergoog:
## Description

This adds a build script for creating a doxygen executable that can be placed into a CIPD package.

## Related Issues
 - https://github.com/flutter/flutter/issues/124833


keyonghan (approved):
LGTM

gspencergoog:
@keyonghan Can you please take another look: I changed some things to make it pass the presubmits, and clean up some things in the repo checks.

licenses/check_licenses.dart
```diff
@@ -124,7 +124,7 @@ Future<void> verifyNoMissingLicense(String workingDirectory, {bool checkMinimums
     workingDirectory,
     'sh',
     overrideMinimumMatches ?? 1,
-    '#!/bin/bash\n' + _generateLicense('# '),
+    '#!.*sh\n' + _generateLicense('# '),
```
keyonghan, line 127: Can we use the full path for both to make sure an exact match?  Or instead, have the `build.sh` to follow the old format, and then create a followup PR to update all .sh to use the new path? Which seems the way from the framework repo.

gspencergoog: Okay, although it seems like it could even just be `.*\n`, since it seems odd to me that a script for checking licenses is enforcing which shells are allowed in scripts.

gspencergoog: How about if I just update all of the existing scripts to use `#!/usr/bin/env bash`, but leave this as-is? I don't think we should be limiting the allowed shells, or at least we shouldn't be doing it here.

gspencergoog: (And I'll move it into another PR.)

keyonghan: Updating to use `#!/usr/bin/env bash` sounds good to me. But for the license check, can we follow what framework is doing: https://github.com/flutter/flutter/blob/master/dev/bots/analyze.dart#L642

I am not sure about the guidance here. maybe @Hixie ?

gspencergoog: Done, but I do think that if we want to restrict the shells used, we should be doing it in the analyze script, not in the license check.

gspencergoog: I reverted all the shebang-line changes, and moved them to another PR: https://github.com/flutter/cocoon/pull/2974

PR #2973, opened 2023-08-02
Counts go_router and go_router_builder custom test folders as tests

chunhtai:
related https://github.com/flutter/packages/pull/4627

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/app_dart/lib/src/request_handlers/github/webhook_subscription.dart
+++ b/app_dart/lib/src/request_handlers/github/webhook_subscription.dart
@@ -430,9 +430,6 @@ class GithubWebhookSubscription extends SubscriptionHandler {
       // See https://github.com/flutter/flutter/wiki/Plugin-Tests for discussion
       // of various plugin test types and locations.
       if (filename.endsWith('_test.dart') ||
-          // Test files in package-specific test folders.
-          filename.contains('packages/packages/go_router/test_fixes/') ||
-          filename.contains('packages/packages/go_router_builder/test_inputs/') ||
           // Native iOS/macOS tests.
           filename.contains('RunnerTests/') ||
           filename.contains('RunnerUITests/') ||
@@ -444,7 +441,10 @@ class GithubWebhookSubscription extends SubscriptionHandler {
           // Native Windows tests.
           filename.endsWith('_test.cpp') ||
           // Pigeon native tests.
-          filename.contains('/platform_tests/')) {
+          filename.contains('/platform_tests/') ||
+          // Test files in package-specific test folders.
+          filename.contains('go_router/test_fixes/') ||
+          filename.contains('go_router_builder/test_inputs/')) {
         hasTests = true;
       }
     }


app_dart/lib/src/request_handlers/github/webhook_subscription.dart
```diff
@@ -423,13 +423,14 @@ class GithubWebhookSubscription extends SubscriptionHandler {
...
           // coverage.
-          !filename.endsWith('tool/run_tests.dart') &&
           !filename.endsWith('run_tests.sh')) {
         needsTests = !_allChangesAreCodeComments(file);
       }
       // See https://github.com/flutter/flutter/wiki/Plugin-Tests for discussion
       // of various plugin test types and locations.
       if (filename.endsWith('_test.dart') ||
+          // Test files in custom package-specific test folders.
+          filename.contains(RegExp('packages/packages/[^/]+/tool/')) ||
```
stuartmorgan, line 433: It's not the case that everything in the tool directory is a test, so this is too broad. `tool` is just the convention for non-public scripts: https://dart.dev/tools/pub/package-layout#public-tools

E.g., a code generation script might live in `tool`, but not be a test.

stuartmorgan (changes requested):

app_dart/lib/src/request_handlers/github/webhook_subscription.dart
chunhtai: What about we match any file that has word `test` anywhere in this folder? something like this
'packages/packages/[^/]+/tool/.*test.*'

Or do you prefer more package specific exception here? like
`packages/packages/go_router/tool/test_fixes`

stuartmorgan: > What about we match any file that has word `test` anywhere in this folder? something like this
> 'packages/packages/[^/]+/tool/.+test.+'

There are currently only three examples of that pattern, at at least one is a false positive (`pigeon/tool/test.dart`, which is just a script for running manual tests locally, so provides no automated test coverage).

The missing-test bot should err on the side of false positives rather than false negatives.

chunhtai: sounds good I will make this more specific to package instead. If we have more and more use cases, we can come up with a more systematic solution later.

app_dart/lib/src/request_handlers/github/webhook_subscription.dart
```diff
@@ -430,6 +430,9 @@ class GithubWebhookSubscription extends SubscriptionHandler {
       // See https://github.com/flutter/flutter/wiki/Plugin-Tests for discussion
       // of various plugin test types and locations.
       if (filename.endsWith('_test.dart') ||
+          // Test files in package-specific test folders.
+          filename.contains('packages/packages/go_router/test_fixes/') ||
+          filename.contains('packages/packages/go_router_builder/test_inputs/') ||
```
stuartmorgan, line 435: Nit: Could you put these at the end, so the generic rules continue to be first?

app_dart/lib/src/request_handlers/github/webhook_subscription.dart
```diff
@@ -430,6 +430,9 @@ class GithubWebhookSubscription extends SubscriptionHandler {
       // See https://github.com/flutter/flutter/wiki/Plugin-Tests for discussion
       // of various plugin test types and locations.
       if (filename.endsWith('_test.dart') ||
+          // Test files in package-specific test folders.
+          filename.contains('packages/packages/go_router/test_fixes/') ||
```
stuartmorgan, line 434: The `packages/packages/` part of these parts unnecessarily ties these patterns to the specific repo structure; we should just start with the package name.

stuartmorgan (changes requested):

PR #2976, opened 2023-08-02
Fix typo in Linux Doxygen builder and enable

gspencergoog:
## Description

Fixes a typo in the script path for the Linux Doxygen builder


keyonghan:
You may want to retry after this build finishes: https://ci.chromium.org/ui/p/flutter/builders/prod/Linux%20ci_yaml%20roller/1551/overview

gspencergoog:
There was a typo in the path to the build script in the `.ci.yaml`, which I fixed, but do I now have to wait for another PR to land before it'll run the job that syncs the `.ci.yaml`?

gspencergoog:
And do I have to land it first with `bringup:true` before it will sync the typo fix?

keyonghan:
Let's land the typo fix with `bringup: true` first.

keyonghan:
The recipes needs a change to reflect the new path: https://flutter-review.googlesource.com/c/recipes/+/48520.

An led based on the above recipes change and this PR: https://luci-milo.appspot.com/raw/build/logs.chromium.org/flutter/led/keyonghan_google.com/e0417daf31545dae8c24de62a411f39c5edd2f2b8a0a491b949f83a9bf32cac6/+/build.proto

keyonghan:
It passes. And just double checked that it only pushes the package to CIPD when it's from `prod` (post-submit).

keyonghan:
This should be unblocked after https://flutter-review.googlesource.com/c/recipes/+/48520 lands.

gspencergoog:
Okay, so with the recipe change, do you still want me to revert to `bringup: true` and change that in another PR?

keyonghan:
> Okay, so with the recipe change, do you still want me to revert to `bringup: true` and change that in another PR?

With `bringup: true`, it will be same as the led run I just validated. So no need. Let's land as it is now.
But one thing to note: packages write is allowed from the builder for `flutter` namespace only, as specified in the target config: cipd_name: **flutter**/doxygen/linux-amd64

You need a recipes change to update to point to the new flutter location: https://flutter.googlesource.com/recipes/+/refs/heads/main/recipe_modules/flutter_deps/api.py#313

keyonghan (approved):
LGTM

gspencergoog:
Ahh, so it can't write to flutter_internal/tools/doxygen?  So the recipe should be something like this?:

```python
doxygen.add_package('flutter/doxygen/${platform}', version)
```

keyonghan:
> Ahh, so it can't write to `flutter_internal/tools/doxygen`? So the recipe should be something like this?:
> 
> ```python
> doxygen.add_package('flutter/doxygen/${platform}', version)
> ```

That's correct.

PR #2990, opened 2023-08-11
Add autosubmit code owners

CaseyHillers:
Follow up from [Formalize Cocoon Contributions](https://docs.google.com/document/d/1SxYP41p3CZY_BrUNb_zSfufvhUiSN3qxGl7QEjjTqHk/edit#heading=h.lgp3erb6u85n)

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -4,6 +4,10 @@
 # These names are just suggestions. It is fine to have your changes
 # reviewed by someone else.
 
+app_dart                                           @keyonghan
+app_dart                                           @ricardoamador
+auto_submit                                        @keyonghan
+auto_submit                                        @ricardoamador
 cipd_packages/codesign/**                          @XilaiZhang
 cipd_packages/device_doctor/**                     @yusuf-goog
 cipd_packages/doxygen/**                           @gspencergoog
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -4,9 +4,6 @@
 # These names are just suggestions. It is fine to have your changes
 # reviewed by someone else.
 
-app_dart                                           @keyonghan
-app_dart                                           @ricardoamador
-auto_submit                                        @keyonghan
 auto_submit                                        @ricardoamador
 cipd_packages/codesign/**                          @XilaiZhang
 cipd_packages/device_doctor/**                     @yusuf-goog
--- a/app_dart/CODEOWNERS
+++ b/app_dart/CODEOWNERS
@@ -1,2 +1,0 @@
-@keyonghan
-@ricardoamador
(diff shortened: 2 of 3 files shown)


ricardoamador:
There is already a top level codeowners file. Do we want to just use that one.

CaseyHillers:
> There is already a top level codeowners file. Do we want to just use that one.

Good point. I'm used to how we do it in Google which is next to the closest subdirectory.

@keyonghan do you have opinions?

keyonghan:
I would use the top level OWNERSHIP file to keep it consistent with our other GitHub repos.
As for app_dart, shall we define the ownership on API level?

CaseyHillers:
Thanks for the context, done! Updated to use top level owners

CODEOWNERS
```diff
@@ -4,6 +4,10 @@
 # These names are just suggestions. It is fine to have your changes
 # reviewed by someone else.
 
+app_dart                                           @keyonghan
+app_dart                                           @ricardoamador
+auto_submit                                        @keyonghan
+auto_submit                                        @ricardoamador
```
ricardoamador, line 10: Can we just use the infra team for this instead of people directly?

CaseyHillers: Happy to! What group is it? I don't see it in https://github.com/orgs/flutter/teams

keyonghan: i am a bit reluctant to add the team here. The ownership here will trigger review request for any changes happened from the Dir.
@ricardoamador Are you okay to own the whole auto_submit dir? If not, let's divide it into smaller pieces. Same for the app_dart.
The reason I put the ownership on dir level for cipd packages is they are relatively a small self-contained process. As for app_dart, it would be too large to give a single ownership entry on the dir level. But as for `auto_submit`, up to you @ricardoamador

ricardoamador: I will defer to Keyong.

keyonghan: As for auto_submit, I prefer to keep only @ricardoamador (assuming okay:-) as the owner considering he has been the main contact/contributor of this app and the app is not that big. 
As for app_dart, let's skip it for now. We want to break it into smaller pieces on API/file level.

CaseyHillers: I would go the opposite way. Can we add the entire engprod team as owners? Having a single person be the owner creates a bottleneck, and the best practice is to ensure owners have redundancy.

keyonghan: CODEOWNERS file here is behaving differently from Google. It just auto adds the owner as a code reviewer, and will not be a bottleneck. People can add anybody else as a code reviewer.

CaseyHillers: Done. Applied feedback for now to only add the autosubmit CODEOWNERS

ricardoamador (approved):
LGTM

PR #2991, opened 2023-08-14
Add logic to handle a task for a commit that is missing

drewroengoogle:
https://github.com/flutter/flutter/issues/132066
* Adds logic to dart-internal-subscription endpoint, where if there is no commit for the associated task from buildbucket, then create it and add it to the datastore. This is done by adding an optional parameter to the method that creates a Task from a buildbucket build, where you can specify a specific Commit you want use.

Small additional changes:
* Add a commit deep copy method `Commit.from()`
* Move away from the `hash` variable to `sha` since `hash` is reserved

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/app_dart/lib/src/model/appengine/commit.dart
+++ b/app_dart/lib/src/model/appengine/commit.dart
@@ -108,6 +108,19 @@ class Commit extends Model<String> {
       ..write(')');
     return buf.toString();
   }
+
+  static Commit from(Commit commit) {
+    return Commit(
+      key: commit.key,
+      sha: commit.sha,
+      timestamp: commit.timestamp,
+      author: commit.author,
+      authorAvatarUrl: commit.authorAvatarUrl,
+      message: commit.message,
+      repository: commit.repository,
+      branch: commit.branch,
+    );
+  }
 }
 
 /// The serialized representation of a [Commit].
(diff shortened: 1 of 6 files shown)


keyonghan (commented):
Do we expect the commit to exist for these dart internal targets? If yes, do we know why we are missing the commit? I would try fixing there instead if possible.

drewroengoogle:
> Do we expect the commit to exist for these dart internal targets? If yes, do we know why we are missing the commit? I would try fixing there instead if possible.

The commit exists referencing main, but not the candidate branch. For example:

Commit hash `abcdef` on `main` has a commit in the datastore `flutter/flutter/main/abcdef`. However, when a new branch is cut from `abcdef` on `flutter-candidate-1`, there is no commit in the datastore for `flutter/flutter/flutter-candidate-1/abcdef`.

I'm not super familiar with github webhooks, is there one for when a new branch is created? We could check for candidate branches and create a new commit there instead of going this route.

keyonghan:
> EDIT: Is this what we're looking for? https://docs.github.com/en/webhooks-and-events/webhooks/webhook-events-and-payloads#create

Seems yes. 

When we create a new branch, no corresponding PR will be created but tasks will be triggered. Is this correct? If yes, then a webhook event would solve our problem here. For that event, we just need to call the existing `_addCommit` to create the commit and tasks entries in datastore.

drewroengoogle:
Closing due to https://github.com/flutter/cocoon/pull/3005

PR #3004, opened 2023-08-22
Add support for run_if_not.

godofredoc:
This is required to filter fuchsia changes in presubmit.

Bug: https://github.com/flutter/flutter/issues/93261

## Pre-launch Checklist

- [X] I signed the [CLA].


godofredoc:
This is now ready for review.

app_dart/lib/src/foundation/utils.dart
```diff
@@ -124,17 +131,33 @@ Future<List<Target>> getTargetsToRun(Iterable<Target> targets, List<String?> fil
...
-    }
-
-    for (String glob in globs) {
-      glob = glob.replaceAll('**', '[A-Za-z0-9_/.]+');
-      glob = glob.replaceAll('*', '[A-Za-z0-9_.]+');
-      // If a file is found within a pre-set dir, the builder needs to run. No need to check further.
-      final RegExp regExp = RegExp('^$glob\$');
-      if (glob.isEmpty || files.any((String? file) => regExp.hasMatch(file!))) {
+      // Evaluate run_if_not.
+      final List<String> negativeGlobs = target.value.runIfNot;
```
ricardoamador, line 135: Rather than getting the value could you not just create a runIfNot getter?

app_dart/lib/src/foundation/utils.dart
```diff
@@ -124,17 +131,33 @@ Future<List<Target>> getTargetsToRun(Iterable<Target> targets, List<String?> fil
...
+      if (negativeGlobs.isEmpty) {
         targetsToRun.add(target);
-        break;
+        continue;
+      }
+      bool shouldAdd = true;
+      for (String glob in negativeGlobs) {
+        // If a file is found within a pre-set dir, the builder needs to run. No need to check further.
```
ricardoamador, line 142: Not sure this comment applies here any longer.

CI_YAML.md
```diff
@@ -42,7 +42,10 @@ targets:
 # presubmit: Whether to run this target on presubmit (defaults to true).
 # postsubmit: Whether to run this target on postsubmit (defaults to true).
 # run_if: List of path regexes that can trigger this target on presubmit.
-#         If none are passed, will always run in presubmit.
+#         If none are passed, it will evaluare run_if_not. If both are empty the target
+#         will always run in presubmit.
+# run_if_not: List of path regexes. The target will be run only if the files changed
+#         do not match any paths in this list.
```
ricardoamador, line 48: What happens if someone supplies both of these properties?

It looks as if both are supplied it will ignore the run_if_not in that case. Someone will try this for sure; is this desired behavior?

CI_YAML.md
```diff
@@ -42,7 +42,10 @@ targets:
 # presubmit: Whether to run this target on presubmit (defaults to true).
 # postsubmit: Whether to run this target on postsubmit (defaults to true).
 # run_if: List of path regexes that can trigger this target on presubmit.
-#         If none are passed, will always run in presubmit.
+#         If none are passed, it will evaluare run_if_not. If both are empty the target
+#         will always run in presubmit.
+# run_if_not: List of path regexes. The target will be run only if the files changed
```
ricardoamador, line 47: You should also mention that this applies only to presubmit targets. Same as run_if does.

app_dart/lib/src/foundation/utils.dart
```diff
@@ -102,6 +102,13 @@ FutureOr<String> getUrl(
...
 }
 
+/// Expands globs string to a regex for evaluation.
+Future<RegExp> parseGlob(String glob) async {
+  glob = glob.replaceAll('**', '[A-Za-z0-9_/.]+');
+  glob = glob.replaceAll('*', '[A-Za-z0-9_.]+');
+  return RegExp('^$glob\$');
+}
+
 /// Returns a LUCI [builder] list that covers changed [files].
```
ricardoamador, line 112: Update this doc to include the run_if_not option.

ricardoamador:
This seems like it was only meant for fuchsia, can we restrict this use towards those targets only? I could see where this and run_if could be abused.

app_dart/lib/src/foundation/utils.dart
godofredoc: Updated!

CI_YAML.md
godofredoc: Documented that if both are provided and non empty run_if_not will be ignored.

app_dart/lib/src/foundation/utils.dart
godofredoc: Removed!

app_dart/lib/src/foundation/utils.dart
godofredoc: From the existing code Target was designed as a wrapper to the proto class potentially to avoid having to keep the wrapper and the proto in sync.

godofredoc:
> This seems like it was only meant for fuchsia, can we restrict this use towards those targets only? I could see where this and run_if could be abused.

From the initial bug it seems like the same pattern can be used for the other embedders to filter out targets for android, ios etc. In the future we may need to provide support for expression evaluations on top the path based ones.

ricardoamador (approved):
LGTM!

PR #3005, opened 2023-08-23
Insert a new commit on branch creation

drewroengoogle:
Issue: https://github.com/flutter/flutter/issues/132066

This is based on comments from https://github.com/flutter/cocoon/pull/2991, but the code was significantly different so it made more sense to create a new branch.

When a branch is created, get the associated commit and push it to the datastore with the new branch. This will allow dart-internal builders, which run on a new candidate branch creation, to have a commit to save their task results to.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/app_dart/lib/src/request_handlers/github/branch_subscription.dart
+++ b/app_dart/lib/src/request_handlers/github/branch_subscription.dart
@@ -47,7 +47,11 @@ class GithubBranchWebhookSubscription extends SubscriptionHandler {
     log.fine('Processing ${webhook.event}');
     final CreateEvent createEvent = CreateEvent.fromJson(json.decode(webhook.payload) as Map<String, dynamic>);
     await branchService.handleCreateRequest(createEvent);
-    await commitService.handleCreateGithubRequest(createEvent);
+
+    final RegExp candidateBranchRegex = RegExp(r'flutter-\d+\.\d+-candidate\.\d+');
+    if (candidateBranchRegex.hasMatch(createEvent.ref!)) {
+      await commitService.handleCreateGithubRequest(createEvent);
+    }
 
     return Body.empty;
   }
--- a/app_dart/lib/src/request_handlers/github/branch_subscription.dart
+++ b/app_dart/lib/src/request_handlers/github/branch_subscription.dart
@@ -4,6 +4,7 @@
 
 import 'dart:convert';
 
+import 'package:cocoon_service/src/service/commit_service.dart';
 import 'package:github/hooks.dart';
 import 'package:meta/meta.dart';
 
@@ -25,9 +26,11 @@ class GithubBranchWebhookSubscription extends SubscriptionHandler {
     required super.cache,
     required super.config,
     required this.branchService,
+    required this.commitService,
   }) : super(subscriptionName: 'github-webhook-branches');
 
   final BranchService branchService;
+  final CommitService commitService;
 
   @override
   Future<Body> post() async {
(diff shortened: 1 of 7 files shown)


app_dart/lib/src/service/commit_service.dart
```diff
@@ -0,0 +1,72 @@
...
+      log.info('commit does not exist in datastore, inserting into datastore');
+      await datastore.insert(<Commit>[commit]);
+    }
+  }
+
+  Future<Commit> _createCommitFromBranchEvent(DatastoreService datastore, RepositorySlug slug, String branch) async {
+    final GithubService githubService = await config.createDefaultGitHubService();
+    final GitReference gitRef = await githubService.getReference(slug, 'heads/$branch');
+    final String sha = gitRef.object!.sha!;
+    final Response response = await githubService.github.request(
```
ricardoamador, line 50: You might want to push this into the githubService rather than accessing the services fields.

ricardoamador: If you want to get a commit the github lib we use has a method for this already. Just use that instead of running the http request.

app_dart/lib/src/service/commit_service.dart
```diff
@@ -0,0 +1,72 @@
...
+      '/repos/${slug.fullName}/commits/$sha',
+    );
+
+    final Map<String, dynamic> commit = jsonDecode(response.body) as Map<String, dynamic>;
+
+    final String id = '${slug.fullName}/$branch/$sha';
+    final Key<String> key = datastore.db.emptyKey.append<String>(Commit, id: id);
+    return Commit(
+      key: key,
+      timestamp: DateTime.parse(commit['commit']['author']['date'] as String).millisecondsSinceEpoch,
```
ricardoamador, line 61: Can you instantiate the Commit from the json rather than looking through the map? Something like Commit.fromJson()

ricardoamador:
Can you provide more context on why this is needed and how it seems to affect only dart internal builders?

drewroengoogle:
> Can you provide more context on why this is needed and how it seems to affect only dart internal builders?

Yep! When a new branch is created, cocoon targets are not run due to those targets already being run on whatever branch the branch was cut from (main or master generally). Dart-internal targets aren't run on main/master, only on candidate branches - therefore, when a new candidate branch is created, the LUCI scheduler triggers the dart-internal tasks, but there is no commit in the datastore to link them to, so the task results from these dart-internal builders will constantly fail to save to the datastore.

The intention of this PR is to create a commit right at branch creation time that the dart-internal tasks can be saved to.

app_dart/lib/src/service/commit_service.dart
drewroengoogle: Updated this to use `githubService.github.repositories.getCommit`

app_dart/lib/src/service/commit_service.dart
drewroengoogle: Updated this to use the `RepositoryCommit` object, so we are no longer looking through a map directly!

app_dart/lib/src/request_handlers/github/branch_subscription.dart
```diff
@@ -45,6 +48,11 @@ class GithubBranchWebhookSubscription extends SubscriptionHandler {
     final CreateEvent createEvent = CreateEvent.fromJson(json.decode(webhook.payload) as Map<String, dynamic>);
     await branchService.handleCreateRequest(createEvent);
 
+    final RegExp candidateBranchRegex = RegExp(r'flutter-\d+\.\d+-candidate\.\d+');
+    if (candidateBranchRegex.hasMatch(createEvent.ref!)) {
```
drewroengoogle, line 52: @ricardoamador FYI, here is the logic for only creating a commit on candidate branches.

ricardoamador (approved):
LGTM

PR #3018, opened 2023-08-29
Auto retry presubmit failures from rollers

keyonghan:
Fixes: https://github.com/flutter/flutter/issues/126556

This PR does:
1) retry failures only once
2) retry failures only when that happen under roller accounts
3) inject `user_login` to `user_data` to help identify the user
4) inject `retry: true` to a build's tags when this is a retry to avoid unlimited rerunning

/cc @zanderso

### DIFF
--- a/app_dart/lib/src/service/github_checks_service.dart
+++ b/app_dart/lib/src/service/github_checks_service.dart
@@ -68,8 +68,9 @@ class GithubChecksService {
   Future<bool> updateCheckStatus(
     push_message.BuildPushMessage buildPushMessage,
     LuciBuildService luciBuildService,
-    github.RepositorySlug slug,
-  ) async {
+    github.RepositorySlug slug, {
+    bool rescheduled = false,
+  }) async {
     final push_message.Build? build = buildPushMessage.build;
     if (buildPushMessage.userData.isEmpty) {
       return false;
@@ -96,15 +97,8 @@ class GithubChecksService {
     final String? url = status == github.CheckRunStatus.completed ? checkRun.detailsUrl : buildPushMessage.build!.url;
     github.CheckRunOutput? output;
     // If status has completed with failure then provide more details.
-    if (status == github.CheckRunStatus.completed && failedStatesSet.contains(conclusion)) {
-      if (shouldRerun(buildPushMessage)) {
-        final String builderName = buildPushMessage.build!.tagsByName('builder').single;
-        log.fine('Rerun a failed task: $builderName');
-        await luciBuildService.reschedulePresubmitBuild(
-          builderName: builderName,
-          buildPushMessage: buildPushMessage,
-          retry: true,
-        );
+    if (taskFailed(buildPushMessage)) {
+      if (rescheduled) {
         status = github.CheckRunStatus.queued;
         conclusion = null;
       } else {
@@ -127,18 +121,25 @@ class GithubChecksService {
     return true;
   }
 
-  /// Check if we should rerun a failed build.
+  /// Check if task has completed with failure.
+  bool taskFailed(push_message.BuildPushMessage buildPushMessage) {
+    final push_message.Build? build = buildPushMessage.build;
+    final github.CheckRunStatus status = statusForResult(build!.status);
+    final github.CheckRunConclusion? conclusion =
+        (buildPushMessage.build!.result != null) ? conclusionForResult(buildPushMessage.build!.result) : null;
+    return status == github.CheckRunStatus.completed && failedStatesSet.contains(conclusion);
+  }
+
+  /// Returns current reschedule attempt.
   ///
-  /// A presubmit retry happens if and only if
-  ///   1) the user is an auto roller account
-  ///   2) the same check run has never been retried before
-  bool shouldRerun(push_message.BuildPushMessage buildPushMessage) {
+  /// It returns 1 if this is the first run, and +1 with each reschedule.
+  int currentAttempt(push_message.BuildPushMessage buildPushMessage) {
     final push_message.Build build = buildPushMessage.build!;
-    if (!buildPushMessage.userData.containsKey('user_login') ||
-        !config.rollerAccounts.contains(buildPushMessage.userData['user_login'])) {
-      return false;
+    if (build.tagsByName('current_attempt').isEmpty) {
+      return 1;
+    } else {
+      return int.parse(build.tagsByName('current_attempt').single);
     }
-    return build.tagsByName('retry').isEmpty || build.tagsByName('retry').contains('false');
   }
 
   /// Appends triage wiki page to `summaryMarkdown` from LUCI build so that people can easily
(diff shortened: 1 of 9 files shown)


app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -281,27 +282,33 @@ class LuciBuildService {
   /// The buildset, user_agent, and github_link tags are applied to match the
   /// original build. The build properties and user data from the original build
   /// are also preserved.
-  Future<Build> rescheduleBuild({
-    required String commitSha,
+  Future<Build> reschedulePresubmitBuild({
```
ricardoamador, line 285: nit: might want to leave this and keep it generic (not specify presubmit) as there has been some work to do that. There is also a cleanup issue for this as well.

app_dart/lib/src/service/github_checks_service.dart
```diff
@@ -114,6 +127,20 @@ class GithubChecksService {
...
+  /// A presubmit retry happens if and only if
+  ///   1) the user is an auto roller account
+  ///   2) the same check run has never been retried before
+  bool shouldRerun(push_message.BuildPushMessage buildPushMessage) {
+    final push_message.Build build = buildPushMessage.build!;
+    if (!buildPushMessage.userData.containsKey('user_login') ||
+        !config.rollerAccounts.contains(buildPushMessage.userData['user_login'])) {
+      return false;
+    }
+    return build.tagsByName('retry').isEmpty || build.tagsByName('retry').contains('false');
```
ricardoamador, line 141: I am not sure I follow here. Are we tracking the number of retries through tags? or are we looking at a specific tag?

keyonghan: I am using the `retry` tag to track if the current build is a rerun. Put it this way considering we only want to retry once.

ricardoamador (approved):
LGTM

app_dart/lib/src/service/luci_build_service.dart
keyonghan: Huh, I intentionally update it specifically for presubmit as this logic will only by consumed by presubmit builds.
For post submit builds, they are handled in a different way as no check run id is needed.

ricardoamador: Yeah no problem. Not a big deal as I labeled it a nit. I am fine if you leave it as is.

godofredoc:
Please do not merge yet. Are we missing the proto changes? the retries should be configured in a per build basis rather than retrying everything.

keyonghan:
> Please do not merge yet. Are we missing the proto changes? the retries should be configured in a per build basis rather than retrying everything.

1) Do we want to retry only a limited number of targets from the auto roller? If yes, then we are missing the proto support.
2) Do we want to retry multiple times? The original request asks for only one-time retry. In any ways, we can make it general to support >1 reruns.

godofredoc:
> > Please do not merge yet. Are we missing the proto changes? the retries should be configured in a per build basis rather than retrying everything.
> 
> 1. Do we want to retry only a limited number of targets from the auto roller? If yes, then we are missing the proto support.
Yes, please take a look at my comments in the bug.
> 2. Do we want to retry multiple times? The original request asks for only one-time retry. In any ways, we can make it general to support >1 reruns.
This should be configurable. As long as we have a way to associate a bug to the configuration change.

app_dart/lib/src/service/github_checks_service.dart
```diff
@@ -89,19 +89,32 @@ class GithubChecksService {
...
         (buildPushMessage.build!.result != null) ? conclusionForResult(buildPushMessage.build!.result) : null;
     // Do not override url for completed status.
     final String? url = status == github.CheckRunStatus.completed ? checkRun.detailsUrl : buildPushMessage.build!.url;
     github.CheckRunOutput? output;
     // If status has completed with failure then provide more details.
     if (status == github.CheckRunStatus.completed && failedStatesSet.contains(conclusion)) {
-      final Build build =
-          await luciBuildService.getBuildById(buildPushMessage.build!.id, fields: 'id,builder,summaryMarkdown');
-      output = github.CheckRunOutput(title: checkRun.name!, summary: getGithubSummary(build.summaryMarkdown));
+      if (shouldRerun(buildPushMessage)) {
```
godofredoc, line 100: Please move this logic to a separate method as this method is supposed to just update the check status.

app_dart/lib/src/service/github_checks_service.dart
godofredoc: Retries should be configurable based on a target property.

app_dart/lib/src/service/luci_build_service.dart
godofredoc: Why do we need different logic for presubmit and postsubmit?

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -281,27 +282,33 @@ class LuciBuildService {
...
     required String builderName,
     required push_message.BuildPushMessage buildPushMessage,
+    retry = false,
```
godofredoc, line 288: The method is called reschedule which I'd assume is a retry. What is the difference between reschedule and retry in this context?

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -281,27 +282,33 @@ class LuciBuildService {
...
     final String bucketName = buildPushMessage.build!.bucket!.split('.').last;
+    final Map<String, List<String>> tags = <String, List<String>>{
+      'buildset': buildPushMessage.build!.tagsByName('buildset'),
+      'user_agent': buildPushMessage.build!.tagsByName('user_agent'),
+      'github_link': buildPushMessage.build!.tagsByName('github_link'),
+      'cipd_version': buildPushMessage.build!.tagsByName('cipd_version'),
+      'github_checkrun': buildPushMessage.build!.tagsByName('github_checkrun'),
+    };
+    if (retry) {
+      tags['retry'] = <String>['true'];
```
godofredoc, line 302: Update to retry count?

app_dart/lib/src/service/github_checks_service.dart
```diff
@@ -114,6 +127,20 @@ class GithubChecksService {
...
   }
 
+  /// Check if we should rerun a failed build.
+  ///
+  /// A presubmit retry happens if and only if
+  ///   1) the user is an auto roller account
+  ///   2) the same check run has never been retried before
+  bool shouldRerun(push_message.BuildPushMessage buildPushMessage) {
+    final push_message.Build build = buildPushMessage.build!;
+    if (!buildPushMessage.userData.containsKey('user_login') ||
```
godofredoc, line 137: Please avoid as much as possible this type of logic as it introduces just another path of execution for a subset of users. Let's make the implementation generic and consistent for everybody.

godofredoc (approved):

godofredoc:
We are missing the proto changes for this change to work.

ricardoamador:
Ah okay. We can wait for Keyong then. I was unaware this required further changes.

PR #3021, opened 2023-08-30
Show release channels in build dashboard

CaseyHillers:
* The previous way can show the past 60 days worth of branches
* Instead, only show HEAD, stable, beta, dev

### DIFF
--- a/app_dart/lib/src/service/branch_service.dart
+++ b/app_dart/lib/src/service/branch_service.dart
@@ -151,6 +151,10 @@ class BranchService {
     );
     return <Map<String, String>>[
       {
+        "branch": Config.defaultBranch(slug),
+        "name": "HEAD",
+      },
+      {
         "branch": stableName,
         "name": "stable",
       },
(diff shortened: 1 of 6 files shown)


XilaiZhang (approved):
Thanks for the change Casey!
images (1)

app_dart/lib/src/service/branch_service.dart
```diff
@@ -160,8 +164,8 @@ class BranchService {
       },
       {
         "branch": latestCandidateBranch,
-        "name": "latestCandidateBranch",
-      }
+        "name": "dev",
+      },
```
godofredoc, line 168: Aren't we over engineering this solution? why don't we just run a query on the Checklist entity?

CaseyHillers: What type of query do you imagine? We could simplify this with enums to dedupe the code, but it's still going to require lookups to GitHub to find what RC maps to beta and stable. Files from GitHub should be aggressively cached by Cocoon, so this should be a cheap lookup.

yusuf-goog:
@godofredoc @CaseyHillers are the questions resolved?

PR #3042, opened 2023-09-06
Add a retry for checking for the GoB commit after PR merged

ricardoamador:
We consistently get 500 errors when checking for commits on GoB. This is most likely due to some latency so adding a retry for up to a second should help in reducing the number of 500 errors we see from this.

image

Fixes https://github.com/flutter/flutter/issues/134146

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/app_dart/lib/src/request_handlers/github/webhook_subscription.dart
+++ b/app_dart/lib/src/request_handlers/github/webhook_subscription.dart
@@ -194,20 +194,15 @@ class GithubWebhookSubscription extends SubscriptionHandler {
     final RepositorySlug slug = pr.base!.repo!.slug();
     final String sha = pr.mergeCommitSha!;
     GerritCommit? gobCommit;
-    try {
-      await Config.gobRetryOptions.retry(
-        () async {
-          gobCommit = await gerritService.findMirroredCommit(slug, sha);
-          if (gobCommit == null) {
-            throw const NotFoundException('Gob commit not found');
-          }
-        },
-        retryIf: (e) => e is NotFoundException,
-      );
-    } on NotFoundException {
-      // do nothing here as we are simply using the exception to retry but do
-      // not want to throw from here.
-    }
+    await Config.gobRetryOptions.retry(
+      () async {
+        gobCommit = await gerritService.findMirroredCommit(slug, sha);
+        if (gobCommit == null) {
+          throw const InternalServerError('Gob commit not found');
+        }
+      },
+      retryIf: (e) => e is InternalServerError,
+    );
     return gobCommit != null;
   }
 
--- a/app_dart/lib/src/request_handlers/github/webhook_subscription.dart
+++ b/app_dart/lib/src/request_handlers/github/webhook_subscription.dart
@@ -194,11 +194,11 @@ class GithubWebhookSubscription extends SubscriptionHandler {
     final RepositorySlug slug = pr.base!.repo!.slug();
     final String sha = pr.mergeCommitSha!;
     GerritCommit? gobCommit;
-    await Config.gobRetryOptions.retry(
+    await config.getGobRetryOptions.retry(
       () async {
         gobCommit = await gerritService.findMirroredCommit(slug, sha);
         if (gobCommit == null) {
-          throw const InternalServerError('Gob commit not found');
+          throw InternalServerError('$sha was not found on GoB. Failing so this event can be retried...');
         }
       },
       retryIf: (e) => e is InternalServerError,
--- a/app_dart/lib/src/request_handlers/github/webhook_subscription.dart
+++ b/app_dart/lib/src/request_handlers/github/webhook_subscription.dart
@@ -193,7 +193,21 @@ class GithubWebhookSubscription extends SubscriptionHandler {
   Future<bool> _commitExistsInGob(PullRequest pr) async {
     final RepositorySlug slug = pr.base!.repo!.slug();
     final String sha = pr.mergeCommitSha!;
-    final GerritCommit? gobCommit = await gerritService.findMirroredCommit(slug, sha);
+    GerritCommit? gobCommit;
+    try {
+      await Config.gobRetryOptions.retry(
+        () async {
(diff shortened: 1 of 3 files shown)


app_dart/lib/src/request_handlers/github/webhook_subscription.dart
```diff
@@ -193,7 +193,21 @@ class GithubWebhookSubscription extends SubscriptionHandler {
   Future<bool> _commitExistsInGob(PullRequest pr) async {
     final RepositorySlug slug = pr.base!.repo!.slug();
     final String sha = pr.mergeCommitSha!;
-    final GerritCommit? gobCommit = await gerritService.findMirroredCommit(slug, sha);
+    GerritCommit? gobCommit;
+    try {
```
godofredoc, line 197: Should we just allow the exception to propagate? as it is it will fail silently.

app_dart/lib/src/service/config.dart
```diff
@@ -83,6 +83,13 @@ class Config {
     return defaultBranches[slug] ?? kDefaultBranchName;
   }
 
+  // GoB retry options
+  static const RetryOptions gobRetryOptions = RetryOptions(
+    maxAttempts: 4,
+    maxDelay: Duration(seconds: 1),
```
yusuf-goog, line 89: Should we actually make this longer? Is this issue due to the commit not sync'ing (which may take longer) or is this just some transient connection type issue?

app_dart/lib/src/service/config.dart
```diff
@@ -83,6 +83,13 @@ class Config {
...
+    delayFactor: Duration(milliseconds: 300),
```
yusuf-goog, line 90: Should we also add the randomizationFactor in here to keep things spicy?

app_dart/lib/src/request_handlers/github/webhook_subscription.dart
ricardoamador: Not exactly. This exception here is just for the retry. We throw from a call up in the call stack. If the commit is not found then it will return null and we do throw an internal server error here: https://github.com/flutter/cocoon/blob/main/app_dart/lib/src/request_handlers/github/webhook_subscription.dart#L155

app_dart/lib/src/service/config.dart
ricardoamador: It appears that the commit is not present. The call simply looks for it there. I would try starting with this but if we need more time we can revisit.

app_dart/lib/src/service/config.dart
ricardoamador: I believe there is a default randomization factor that is set to 25% as the default. Do you want to go higher than that? My take is that the higher the factor the longer the wait in between the calls.

app_dart/lib/src/request_handlers/github/webhook_subscription.dart
godofredoc: If the caller is throwing the exception anyway there is no need for the try/catch here. Removing the logic will simplify the code and it will make easier to find potential errors with the real stacktrace in the logs.

app_dart/lib/src/service/config.dart
yusuf-goog: Thats fine, we can stick with default

app_dart/lib/src/service/config.dart
yusuf-goog: If its a gob sync issue, that could take up to a minute. Not really sure where in the stack this comes in to know what the delay is. Your call on whats appropriate

ricardoamador: Ah okay. that is some helpful information. Let me adjust the timing then.

app_dart/lib/src/request_handlers/github/webhook_subscription.dart
ricardoamador: I see what you mean. I can just change the exception and throw from here.

ricardoamador: Done. Changed the exception to an internal server error which we throw for simplification.

app_dart/lib/src/service/config.dart
```diff
@@ -86,8 +86,7 @@ class Config {
   // GoB retry options
   static const RetryOptions gobRetryOptions = RetryOptions(
     maxAttempts: 4,
-    maxDelay: Duration(seconds: 1),
-    delayFactor: Duration(milliseconds: 300),
+    delayFactor: Duration(milliseconds: 500),
```
yusuf-goog, line 89: what is your expected total wait for this?

app_dart/lib/src/service/config.dart
ricardoamador: Done. removed the max delay in order to wait for a realistic amount of time for the commit to propagate.

app_dart/lib/src/service/config.dart
ricardoamador: The wait is 500, 1000, 1500 milliseconds. Based on the info you gave for 1 minute this should be on either side of that. The maxDelay doesn't make sense with this as we would just be limiting the delay factor

yusuf-goog: You specified milliseconds not minutes?

ricardoamador: Ah sorry. I misread what you said about the wait time.

app_dart/test/src/datastore/fake_config.dart
```diff
@@ -106,6 +107,12 @@ class FakeConfig implements Config {
   Set<gh.RepositorySlug>? supportedReposValue;
   Set<gh.RepositorySlug>? postsubmitSupportedReposValue;
   Duration? githubRequestDelayValue;
+  
+  // GoB retry options
+  RetryOptions _gobRetryOptions = const RetryOptions(
+    maxAttempts: 1,
+    delayFactor: Duration(seconds: 0),
```
yusuf-goog, line 114: What does this do?

yusuf-goog (approved):

PR #3046, opened 2023-09-07
Revert "Auto retry presubmit failures from rollers"

ricardoamador:
Reverts flutter/cocoon#3018

We still need some protobuf plumbing for this. We can reland when Keyong is back.

cc/ @yusuf-goog since you are on call.


fluttergithubbot (approved):

ricardoamador:
cc/ @keyonghan We'll have to re-land this.

PR #3096, opened 2023-09-22
[app_dart] Add FRoB Webhook endpoint

CaseyHillers:
b/238326337 - Google Testing isn't receiving related "rerequested" check run events.

Adding this here, and we'll create dedicated FROB pubsubs for that application instead of reusing the flutter-dashboard events.

### DIFF
--- a/app_dart/bin/server.dart
+++ b/app_dart/bin/server.dart
@@ -90,6 +90,14 @@ Future<void> main() async {
       '/api/github-webhook-pullrequest': GithubWebhook(
         config: config,
         pubsub: const PubSub(),
+        secret: config.webhookKey,
+        topic: 'github-webhooks',
+      ),
+      '/api/github/frob-webhook': GithubWebhook(
+        config: config,
+        pubsub: const PubSub(),
+        secret: config.frobWebhookKey,
+        topic: 'frob-webhooks',
       ),
       '/api/github/webhook-subscription': GithubWebhookSubscription(
         config: config,
--- a/app_dart/bin/server.dart
+++ b/app_dart/bin/server.dart
@@ -93,6 +93,7 @@ Future<void> main() async {
         secret: config.webhookKey,
         topic: 'github-webhooks',
       ),
+      // TODO(chillers): Move to release service. https://github.com/flutter/flutter/issues/132082
       '/api/github/frob-webhook': GithubWebhook(
         config: config,
         pubsub: const PubSub(),
--- a/app_dart/lib/src/request_handlers/github_webhook.dart
+++ b/app_dart/lib/src/request_handlers/github_webhook.dart
@@ -26,7 +26,7 @@ class GithubWebhook extends RequestHandler<Body> {
   });
 
   final PubSub pubsub;
-  
+
   /// PubSub topic to publish authenticated requests to.
   final String topic;
 
--- a/app_dart/lib/src/request_handlers/github_webhook.dart
+++ b/app_dart/lib/src/request_handlers/github_webhook.dart
@@ -21,9 +21,17 @@ class GithubWebhook extends RequestHandler<Body> {
   GithubWebhook({
     required super.config,
     required this.pubsub,
+    required this.secret,
+    required this.topic,
   });
 
   final PubSub pubsub;
+  
+  /// PubSub topic to publish authenticated requests to.
+  final String topic;
+
+  /// Future that resolves to the GitHub apps webhook secret.
+  final Future<String> secret;
 
   @override
   Future<Body> post() async {
@@ -42,7 +50,7 @@ class GithubWebhook extends RequestHandler<Body> {
       ..event = event
       ..payload = requestString;
     log.fine(message);
-    await pubsub.publish('github-webhooks', message.writeToJsonMap());
+    await pubsub.publish(topic, message.writeToJsonMap());
 
     return Body.empty;
   }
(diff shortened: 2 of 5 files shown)


app_dart/bin/server.dart
```diff
@@ -90,6 +90,14 @@ Future<void> main() async {
...
         config: config,
         pubsub: const PubSub(),
+        secret: config.webhookKey,
+        topic: 'github-webhooks',
+      ),
+      '/api/github/frob-webhook': GithubWebhook(
+        config: config,
+        pubsub: const PubSub(),
+        secret: config.frobWebhookKey,
+        topic: 'frob-webhooks',
```
Jasguerrero, line 100: Do you see this new topic consuming more than just the re-request event in the future? Just curious on why we are making a new one instead of making the github-webhooks process the rerequest

Jasguerrero (approved):
LGTM

app_dart/bin/server.dart
CaseyHillers: The new topic should have all events for the Google Testing app, which may differ from the flutter-dashboard app based on their configs. Both should have the pull request events.

Can you clarify "making the github-webhooks process the rerequest"? I went with this approach as having the GitHub webhooks handler process webhooks from multiple apps would require more work.

Jasguerrero: > I went with this approach as having the GitHub webhooks handler process webhooks from multiple apps would require more work.

This pretty much answer my question. Thank you

godofredoc:
@ricardoamador can you please take a look.

ricardoamador:
What is calling this?

CaseyHillers:
> What is calling this? The Github application publishes to only a single endpoint the '/api/github-webhook-pullrequest.' So what is pushing to your new webhook handler? It seems like you would want the webhook to publish to a different topic as you stated in the buganizer:

The Flutter Roll on Borg GitHub application will be calling this. We'll need to create a new topic to separate the FRoB webhooks from the flutter-dashboard webhooks. If this looks good to you, I'll go ahead and request AOD to set up the new pubsub topic, and I'll file a GitHub admin ticket to connect the FRoB application.

> We can update flutter/cocoon to publish it's GitHub webhooks to pubsub. This makes Cocoon itself more resilient, and enables FROB to have access to the webhooks (as GitHub can't send requests to an internal borg service).

This part is already implemented.

ricardoamador:
Do you have a design doc somewhere that offers more context with what this is going to be used for?

CaseyHillers:
> Do you have a design doc somewhere that offers more context with what this is going to be used for?

See https://flutter.dev/go/google-testing-v2. This functionality has been in production for 2 quarters.

Due to a recent roll out of the checks feature, we realized we're not getting the Google testing rerun events from GitHub, which prevents us from supporting the feature. This PR enables us to get the correct events from "Flutter Roll on Borg" instead of "flutter-dashboard"

ricardoamador:
I'm still not clear as to what is pushing to this endpoint? And if you are simply using github events how is that information different from what we already get here? Why don't you just read the action from the header and push to both topics if you really want a separate topic?

CaseyHillers:
> I'm still not clear as to what is pushing to this endpoint?

The GitHub "Flutter Roll on Borg" will send events to this endpoint. Similar to how the flutter-dashboard app gets events to /api/github-pull-requests. Once this is created, I can update our app to send events here as GitHub needs to receive a HTTP 200 when adding the webhook URL.

> And if you are simply using github events how is that information different from what we already get here?

Check Run rerun events are sent based on the GitHub app. flutter-dashboard receives all the LUCI rerun events, but not the FRoB scheduled Google testing reruns.

In reverse, when we connect the Flutter roll on borg app, FRoB should not receive the LUCI check run events, and only the Google testing events.

> Why don't you just setup a subscription to get what you want?

We already have a [subscription set up](https://pantheon.corp.google.com/cloudpubsub/subscription/detail/github-webhooks-frob?project=flutter-dashboard) that works mostly. We use it today to process when pull requests are merged.

This subscription contains the flutter-dashboard events, which do not contain rerun events for Google testing. Because of that, I need to setup a new pubsub topic for the Flutter Roll on Borg webhooks, and migrate our subscription to use the new topic.

I'll set up time for us to discuss and add any additional context you need \cc @keyonghan

ricardoamador:
So the "Flutter Roll on Borg" will be setup to send it's subscribed to Github events to cocoon?

> Check Run rerun events are sent based on the GitHub app. 

which are events generated in the repo not specific to the app as you install the application in the target repositories. 

> flutter-dashboard receives all the LUCI rerun events, but not the FRoB scheduled Google testing reruns.

I'm not sure what this means but there is currently no ability to rerun Google Testing from Github. Don't you need to implement that?

It sounds like you are going to modify the checkrun outside of cocoon by having your service get events published to it through cocoon as there is no other logic, unless I am missing something, that acts on this event in Cocoon.

I'll wait for the meeting on Monday. I have more questions about how this will affect the current CI workflow with existing services.

ricardoamador (approved):
LGTM - Casey has confirmed that this will eventually be moved to a release app in gcp.

ricardoamador:
Oh can you create a comment and follow up issue for the eventual move to the release app please?

CaseyHillers:
> Oh can you create a comment and follow up issue for the eventual move to the release app please?

Done! Linked to https://github.com/flutter/flutter/issues/132082

PR #3105, opened 2023-09-26
Waitadb

yusuf-goog:
Bug:https://github.com/flutter/flutter/issues/135035

## Pre-launch Checklist

- [x] I signed the [CLA].


cipd_packages/device_doctor/lib/src/android_device.dart
```diff
@@ -380,7 +381,54 @@ class AndroidDevice implements Device {
...
+      'true',
+    ];
+    final int maxTries = 4;
+    final int sleepSecs = 5;
+    for (var i = 0; i < maxTries; i++) {
+      final Process proc = await startProcess('adb', waitForDeviceArgs, processManager: processManager);
+      if (0 == await proc.exitCode) {
+        break;
+      }
+      await Future.delayed(Duration(seconds: sleepSecs));
```
christopherfujino, line 401: I think you'll need to use `Future.wait`, otherwise your `Future.delayed` will never time out your `adb` process if it's hanging:

```suggestion
    bool success = false;
    for (var i = 0; i < maxTries; i++) {
      final Process proc = await startProcess('adb', waitForDeviceArgs, processManager: processManager);
      if (0 == await proc.exitCode) {
        break;
      }
      await Future.delayed(Duration(seconds: sleepSecs));
      await Future.wait(<Future<void>>[
        proc.exitCode.then((int code) {
          if (code == 0) success = true;
        }),
        Future.delayed(Duration(seconds: sleepSecs)),
      ]);
      if (sucess) {
        break;
      }
```

cipd_packages/device_doctor/lib/src/android_device.dart
```diff
@@ -380,7 +381,54 @@ class AndroidDevice implements Device {
...
+    }
+  }
+
+  @visibleForTesting
+  Future<bool> cleanDevice({ProcessManager? processManager}) async {
+    processManager ??= LocalProcessManager();
+    // Ensure device is rebooted after deleting cache and packages.
+    final List<String> deletePackageCacheArgs = [
+      'shell pm list packages -f',
+      '|',
```
christopherfujino, line 411: I'm pretty sure you can't create pipes like this with dart:io. What I'd do is something like:

```dart
  final io.ProcessResult result = io.Process.runSync('adb', <String>['shell', 'pm', 'list', 'packages', '-f']);
  if (result.exitCode != 0) {
    handleError(result.stderr as String);
  }
  // Or whatever the dart equivalent of your awk script would be
  final String pid = (result.stdout as String).split('=')[2];
  // or whatever the equivalent of your xargs would be...
  io.Process.runSync('adb', <String>['shell', 'pm', 'clear', pid]);
```

cipd_packages/device_doctor/lib/src/android_device.dart
christopherfujino: oh wait, I was wrong, you ACTUALLY want Future.any: https://api.dart.dev/stable/3.1.2/dart-async/Future/any.html

Future.wait will wait for ALL of your futures, Future.any any one of them.

cipd_packages/device_doctor/lib/src/android_device.dart
```diff
@@ -380,7 +381,68 @@ class AndroidDevice implements Device {
...
+  Future<void> deletePackageCache() async {
+    final ProcessResult result = Process.runSync('adb', <String>['shell', 'pm', 'list', 'packages']);
+
+    if (result.exitCode != 0) {
+      throw Exception(result.stderr as String);
+    }
+    final packages = <String>[];
+
+    // Listen to the stdout and stderr streams
+    await result.stdout.transform(utf8.decoder).forEach((data) {
```
christopherfujino, line 397: you shouldn't need to do this. unless you override the encoding, the default type of `result.stdout` should be a string. from there you can do `(result.stdout as String).split('\n').forEach((data) {`

cipd_packages/device_doctor/lib/src/android_device.dart
```diff
@@ -380,7 +381,68 @@ class AndroidDevice implements Device {
...
+      final packageMatch = RegExp(r'package:(.+)$').firstMatch(data);
+      if (packageMatch != null) {
+        final packageName = packageMatch.group(1);
+        if (packageName != null) {
+          packages.add(packageName);
+        }
+      }
+    });
+    for (final p in packages) {
+      Process.runSync('adb', <String>['shell', 'pm', 'clear', p]);
```
christopherfujino, line 407: should we check the output? even if we don't fail on non-zero, we could at least leave a debug print.

cipd_packages/device_doctor/lib/src/android_device.dart
```diff
@@ -380,7 +381,68 @@ class AndroidDevice implements Device {
...
+      final packageMatch = RegExp(r'package:(.+)$').firstMatch(data);
+      if (packageMatch != null) {
+        final packageName = packageMatch.group(1);
+        if (packageName != null) {
+          packages.add(packageName);
+        }
+      }
+    });
+    for (final p in packages) {
+      Process.runSync('adb', <String>['shell', 'pm', 'uninstall', p]);
```
christopherfujino, line 430: same comment here, I'd suggest at least leaving a debug print if exitcode != 0

christopherfujino (approved):
two nits about checking the exit codes, otherwise LGTM, ship it.

cipd_packages/device_doctor/lib/src/android_device.dart
yusuf-goog: done

cipd_packages/device_doctor/lib/src/android_device.dart
yusuf-goog: done

PR #3106, opened 2023-09-26
Revert "Add a retry for checking for the GoB commit after PR merged"

ricardoamador:
Reverts flutter/cocoon#3042

Fixes https://github.com/flutter/flutter/issues/135518

This was preventing messages from being processed as if copybara is latent we can get in a loop of retrying for the commit while blocking reprocessing from pubsub. This will result in unacked messages and will cause them to be retried on exponential backoff.

before_retry_removal

Removing the retry we can see that the unacked messages are being retried now as we are not passing the ack deadline.

image


yusuf-goog (approved):

keyonghan (approved):
LGTM

ricardoamador:
Associated issue for tracking work to update the 500 error: https://github.com/flutter/flutter/issues/135533

PR #3108, opened 2023-09-27
Skip scheduling `bringup: true` from webhook

keyonghan:
Fixes: https://github.com/flutter/flutter/issues/131501, https://github.com/flutter/flutter/issues/127138

When a new target is created, it should not be scheduled before its configs rolled to LUCI (via ciyaml_roller). Currently these new targets are treated same way as any other existing targets, and will just hang forever.

This PR skips scheduling all `bringup: true` targets from `webhook` side. This way:
1) newly created targets will not be scheduled before configs ready (avoid hanging forever)
2) existing `bringup: true` targets will be covered by backfiller scheduling.

### DIFF
--- a/app_dart/lib/src/service/scheduler.dart
+++ b/app_dart/lib/src/service/scheduler.dart
@@ -142,7 +142,11 @@ class Scheduler {
         policy = GuaranteedPolicy();
       }
       final int? priority = await policy.triggerPriority(task: task, datastore: datastore);
-      if (priority != null) {
+      // Skip scheduling `bringup: true` targets. They may be newly created and
+      // their corresponding LUCI builder configs may not be ready yet considering we
+      // disabled the `ci_yaml roller` backfill. Existing `bringup: true` targets
+      // can rely on backfiller to get tasks scheduled and executed.
+      if (priority != null && !target.value.bringup) {
         // Mark task as in progress to ensure it isn't scheduled over
         task.status = Task.statusInProgress;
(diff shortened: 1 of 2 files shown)


yusuf-goog (approved):

app_dart/lib/src/service/scheduler.dart
```diff
@@ -142,7 +142,11 @@ class Scheduler {
         policy = GuaranteedPolicy();
       }
       final int? priority = await policy.triggerPriority(task: task, datastore: datastore);
-      if (priority != null) {
+      // Skip scheduling `bringup: true` targets. They may be newly created and
+      // their corresponding LUCI builder configs may not be ready yet considering we
+      // disabled the `ci_yaml roller` backfill. Existing `bringup: true` targets
+      // can rely on backfiller to get tasks scheduled and executed.
+      if (priority != null && !target.value.bringup) {
```
ricardoamador, line 149: Good catch!

ricardoamador: What is the need for the priority not being null check? What if it is null?

ricardoamador (approved):
LGTM with some clarifying questions.

app_dart/lib/src/service/scheduler.dart
keyonghan: For batch scheduler policy, if there is a running (in progress) task, it will return null and the tot ones will not be scheduled right away.

PR #3110, opened 2023-09-28
Revert not correctly collecting default branch

ricardoamador:
This is a fix to the way the default branch is collected. The code was wrongly checking for if the default branch was empty after the config was collected. It was never empty because a default value of main was assigned so that was used each time.

This changes the default branch to have a default value of 'default' and then we check against that value to get true default branch from the repo.

Example pull request of the fix: https://github.com/flutter/flutter/pull/135692

Fixes https://github.com/flutter/flutter/issues/135670

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/auto_submit/lib/configuration/repository_configuration.dart
+++ b/auto_submit/lib/configuration/repository_configuration.dart
@@ -21,14 +21,14 @@ class RepositoryConfiguration {
   static const String requiredCheckRunsOnRevertKey = 'required_checkruns_on_revert';
 
   // Default values
-  static const bool allowConfigOverrideDefault = false;
-  static const String defaultBranchDefault = 'main';
-  static const Set<String> autoApprovalAccountsDefault = <String>{};
-  static const int approvingReviewsDefault = 2;
-  static const String approvalGroupDefault = 'flutter-hackers';
-  static const bool runCiDefault = true;
-  static const bool supportNoReviewRevertsDefault = true;
-  static const Set<String> requiredCheckRunsOnRevertDefault = <String>{};
+  // static const bool allowConfigOverrideDefault = false;
+  // static const String defaultBranchDefault = 'main';
+  // static const Set<String> autoApprovalAccountsDefault = <String>{};
+  // static const int approvingReviewsDefault = 2;
+  // static const String approvalGroupDefault = 'flutter-hackers';
+  // static const bool runCiDefault = true;
+  // static const bool supportNoReviewRevertsDefault = true;
+  // static const Set<String> requiredCheckRunsOnRevertDefault = <String>{};
 
   RepositoryConfiguration({
     allowConfigOverride,
@@ -40,7 +40,7 @@ class RepositoryConfiguration {
     supportNoReviewReverts,
     requiredCheckRunsOnRevert,
   })  : allowConfigOverride = allowConfigOverride ?? false,
-        defaultBranch = defaultBranch ?? 'main',
+        defaultBranch = defaultBranch ?? 'default',
         autoApprovalAccounts = autoApprovalAccounts ?? <String>{},
         approvingReviews = approvingReviews ?? 2,
         approvalGroup = approvalGroup ?? 'flutter-hackers',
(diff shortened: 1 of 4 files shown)


sealesj (approved):
LGTM with question - why are the other default values commented out here?

auto_submit/lib/configuration/repository_configuration.dart
```diff
@@ -21,14 +21,14 @@ class RepositoryConfiguration {
...
   // Default values
-  static const bool allowConfigOverrideDefault = false;
-  static const String defaultBranchDefault = 'main';
-  static const Set<String> autoApprovalAccountsDefault = <String>{};
-  static const int approvingReviewsDefault = 2;
-  static const String approvalGroupDefault = 'flutter-hackers';
-  static const bool runCiDefault = true;
-  static const bool supportNoReviewRevertsDefault = true;
-  static const Set<String> requiredCheckRunsOnRevertDefault = <String>{};
+  // static const bool allowConfigOverrideDefault = false;
```
godofredoc, line 24: rm if not needed anymore?

auto_submit/lib/configuration/repository_configuration.dart
```diff
@@ -40,7 +40,7 @@ class RepositoryConfiguration {
     supportNoReviewReverts,
     requiredCheckRunsOnRevert,
   })  : allowConfigOverride = allowConfigOverride ?? false,
-        defaultBranch = defaultBranch ?? 'main',
+        defaultBranch = defaultBranch ?? 'default',
```
godofredoc, line 43: config.repoDefaultBranch()?

auto_submit/lib/configuration/repository_configuration_manager.dart
```diff
@@ -69,7 +69,7 @@ class RepositoryConfigurationManager {
     final RepositoryConfiguration globalRepositoryConfiguration = RepositoryConfiguration.fromYaml(orgLevelConfig);
 
     // Collect the default branch if it was not supplied.
-    if (globalRepositoryConfiguration.defaultBranch.isEmpty) {
+    if (globalRepositoryConfiguration.defaultBranch == 'default') {
       globalRepositoryConfiguration.defaultBranch = await githubService.getDefaultBranch(slug);
```
godofredoc, line 73: Why not calling githubService.getDefaultBranch(slug) on repository_configuration?

godofredoc: Also would we need to cache the results of githubService.getDefaultBranch(slug)?

auto_submit/lib/configuration/repository_configuration.dart
ricardoamador: Done!

auto_submit/lib/configuration/repository_configuration.dart
ricardoamador: I am trying to avoid the config dependency.

auto_submit/lib/configuration/repository_configuration_manager.dart
ricardoamador: The reason is that you can actually set the default branch here and to avoid too many responsibilities the manager makes the call as it has the config and access to the Github service.

The results are cached so we are not making this call every time.

ricardoamador:
@sealesj did not see your earlier comment. They were from something else that I had done but I guess they were never removed. They were not being used anywhere in the code.

auto_submit/lib/configuration/repository_configuration_manager.dart
godofredoc: In that case let's define the 'default' string as a shared constant to avoid issues when updating in one file but not the other.

ricardoamador: Ah good point!

ricardoamador: Done!

godofredoc:
Are we still missing the constant update? I still see "default".

ricardoamador:
Sorry, forgot to push the change.

godofredoc (approved):

PR #3111, opened 2023-09-28
Adding a log for builders found vs build name in presubmit subscription

ricardoamador:
Occasionally we hit an issue where we cannot get an element from the ciYaml and then cannot update an existing task in presubmit. This implies that we found none or more than one task. This log message will let us know which it is if hit again.

Related to https://github.com/flutter/flutter/issues/135640

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/app_dart/lib/src/request_handlers/presubmit_luci_subscription.dart
+++ b/app_dart/lib/src/request_handlers/presubmit_luci_subscription.dart
@@ -119,8 +119,19 @@ class PresubmitLuciSubscription extends SubscriptionHandler {
     } else {
       ciYaml = await scheduler.getCiYaml(commit);
     }
-    final Target target = ciYaml.presubmitTargets.where((element) => element.value.name == builderName).single;
+
+    Target target;
+    try {
+      target = ciYaml.presubmitTargets.where((element) => element.value.name == builderName).single;
+    } on Exception {
+      log.warning('Did not find builder with name: $builderName in ciYaml for ${commit.sha}');
+      final List<String> availableBuilderList = ciYaml.presubmitTargets.map((Target e) => e.value.name).toList();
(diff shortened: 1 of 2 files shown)


ricardoamador:
On a side note, @keyonghan would it be safe to just assume no reschedule and return 1 on this error so we can update the failed task?

app_dart/lib/src/request_handlers/presubmit_luci_subscription.dart
```diff
@@ -119,8 +119,29 @@ class PresubmitLuciSubscription extends SubscriptionHandler {
...
+      final List<String> availableBuilderList = ciYaml.presubmitTargets.map((Target e) => e.value.name).toList();
+      log.warning('ciYaml presubmit targets found: $availableBuilderList');
+      rethrow;
+    }
+
+    // alternative solution, just not reschedule.
+    // This might be allowable. We are checking tip of tree but I don't think the
+    // target has been deleted.
+
+    // if (!ciYaml.presubmitTargets.any((element) => element.value.name == builderName)) {
```
keyonghan, line 137: Let's go with this solution, and add logs here. This way pub/sub messages will block builds update.

keyonghan:
> On a side note, @keyonghan would it be safe to just assume no reschedule and return 1 on this error so we can update the failed task?

This should be safe. I tried taking a look at the failed case, but didn't find why it complains. With this log, we should get some good idea.

app_dart/lib/src/request_handlers/presubmit_luci_subscription.dart
ricardoamador: Okay no problem. I will add it.

keyonghan (commented):
Could you add a unit test to cover this corner case?

keyonghan (approved):
LGTM

PR #3124, opened 2023-10-04
Fix exception handling in /dart-internal-subscription to throw internal server error

ricardoamador:
The subscription handler runs its request zoned and as a result if the error returned is not an http exception/error the request will not be handled correctly. We will also be leaving a write lock in the cache.

Fixes https://github.com/flutter/flutter/issues/135991

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/app_dart/lib/src/request_handlers/dart_internal_subscription.dart
+++ b/app_dart/lib/src/request_handlers/dart_internal_subscription.dart
@@ -5,6 +5,7 @@
 import 'dart:convert';
 
 import 'package:cocoon_service/src/model/luci/buildbucket.dart';
+import 'package:cocoon_service/src/request_handling/exceptions.dart';
 import 'package:meta/meta.dart';
 
 import '../../cocoon_service.dart';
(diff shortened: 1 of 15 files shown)


app_dart/lib/src/request_handlers/dart_internal_subscription.dart
```diff
@@ -83,24 +84,31 @@ class DartInternalSubscription extends SubscriptionHandler {
...
-    } else {
-      log.info('Creating Task from Buildbucket result');
-      taskToInsert = await Task.fromBuildbucketBuild(build, datastore);
-    }
-
-    log.info('Inserting Task into the datastore: ${taskToInsert.toString()}');
-    await datastore.insert(<Task>[taskToInsert]);
 
-    return Body.forJson(taskToInsert.toString());
+    try {
```
keyonghan, line 88: Based on the backend log:
```
Uncaught error in request handler: NoSuchMethodError: Class 'KeyNotFoundException' has no instance getter 'statusCode'.
Receiver: Instance of 'KeyNotFoundException'
Tried calling: statusCode
#0      Object.noSuchMethod (dart:core-patch/object_patch.dart:38:5)
#1      SubscriptionHandler.service.<anonymous closure>.<anonymous closure> (package:cocoon_service/src/request_handling/subscription_handler.dart:142:65)
#2      RequestHandler.service.<anonymous closure> (package:cocoon_service/src/request_handling/request_handler.dart:69:26)
<asynchronous suspension>
#3      SubscriptionHandler.service (package:cocoon_service/src/request_handling/subscription_handler.dart:138:5)
<asynchronous suspension>
#4      main.<anonymous closure>.<anonymous closure> (file:///app/bin/server.dart:290:11)
<asynchronous suspension>
```

https://github.com/flutter/cocoon/blob/main/app_dart/lib/src/request_handling/request_handler.dart#L69 runs on `HttpStatusException`, which should have `statusCode` though?

Not sure if the try logic helps. With the try logic, I am wondering if the specified `InternalServerError` has any difference from the potential exception (from getbuild).

ricardoamador: I believe that is the assumption. Thought what happens when the exception thrown is not an HttpStatusException and not an error (which appears to be a different class of errors than exception) as this https://github.com/flutter/cocoon/blob/main/app_dart/lib/src/request_handling/request_handler.dart#L61 is the first pass at exception catching while the one you linked should be the catch all for whatever comes from those? 

If the exception is uncaught that means it is neither an HttpStatusException or an Error as is the case of the KeyNotFoundException which comes from the datastore.

ricardoamador: Though it is running the onError clause...

ricardoamador: I can duplicate this situation thought I am finding it hard to get the debugger to cooperate.
image

keyonghan: Taking a deeper look, it is due to commit not found. 
https://github.com/flutter/cocoon/blob/main/app_dart/lib/src/request_handlers/dart_internal_subscription.dart#L98
https://github.com/flutter/cocoon/blob/main/app_dart/lib/src/model/appengine/task.dart#L161

With this patch, the API will throw internal error any way. I am wondering if we should consider fixing from datastore side. Like why the commit doesn't exist; Is it expected to exist?

ricardoamador: Yeah I am seeking to catch any non HttpRequestException here. Not just the keyNotFound exception. I don't think the datastore is wrong here in the implementation. It just means that whatever we are sending it is not in the datastore and from fromBuildBucketBuild we do not catch any exceptions, which is probably where we should actually fix this.

ricardoamador: I could catch datastore exceptions there and return a nullable Task which we could process to an internal server error.

keyonghan: Do do still recall https://github.com/flutter/cocoon/pull/3005? Which seems expected to insert the branch commit.

ricardoamador: Yes, but it does a check to see if the commit already exists. Which is where it is failing. But now I see what you are talking about. When this is called it is assumed that the commit should exist. Okay let me investigate more.

ricardoamador:
Reassigning to release team.

PR #3139, opened 2023-10-11
Schedule bringup true targets for guaranteed policy repos

keyonghan:
Fixes: https://github.com/flutter/flutter/issues/136361


app_dart/lib/src/service/scheduler.dart
```diff
@@ -169,6 +165,22 @@ class Scheduler {
     await _uploadToBigQuery(commit);
   }
 
+  /// Checks whether the has should be scheduled.
```
yusuf-goog, line 168: typo

yusuf-goog (approved):

PR #3159, opened 2023-10-13
Fix scheduling issue via priority control for new targets

keyonghan:
Fixes: https://github.com/flutter/flutter/issues/136487

Context:
https://github.com/flutter/cocoon/pull/3108 skipped scheduling bringup: true targets from webook for all repos
https://github.com/flutter/cocoon/pull/3139 restores scheduling bringup:true targets for guranteed policy repos (engine/pacakges)

With above two PR patches, we still hit https://github.com/flutter/flutter/issues/136487

This PR proposes a more concise way to solve via priority control.
1) restores the above two PRs
2) returns `null` priority when existing number of tasks < `kBatchSize`.

### DIFF
--- a/app_dart/lib/src/service/scheduler/policy.dart
+++ b/app_dart/lib/src/service/scheduler/policy.dart
@@ -64,8 +64,8 @@ class BatchPolicy implements SchedulerPolicy {
     // Ensure task isn't considered in recentTasks
     recentTasks.removeWhere((Task t) => t.commitKey == task.commitKey);
     if (recentTasks.length < kBatchSize) {
-      log.warning('${task.name} has less than $kBatchSize, triggerring all builds regardless of policy');
-      return LuciBuildService.kDefaultPriority;
+      log.warning('${task.name} has less than $kBatchSize, skip scheduling to wait for ci.yaml roll.');
+      return null;
     }
 
     // Prioritize tasks that recently failed.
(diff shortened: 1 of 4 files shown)


keyonghan:
A friendly ping 🙂

app_dart/lib/src/service/scheduler/policy.dart
```diff
@@ -64,8 +64,8 @@ class BatchPolicy implements SchedulerPolicy {
     // Ensure task isn't considered in recentTasks
     recentTasks.removeWhere((Task t) => t.commitKey == task.commitKey);
     if (recentTasks.length < kBatchSize) {
-      log.warning('${task.name} has less than $kBatchSize, triggerring all builds regardless of policy');
-      return LuciBuildService.kDefaultPriority;
+      log.warning('${task.name} has less than $kBatchSize, skip scheduling to wait for ci.yaml roll.');
+      return null;
```
keyonghan, line 68: This is the main change.

ricardoamador: qq: What does null represent in terms of scheduling?

keyonghan: `null` here means no schedule. See https://github.com/flutter/cocoon/blob/c93ec99b2c0d40f410368cfad2d62e2ca19608c5/app_dart/lib/src/service/scheduler.dart#L145

ricardoamador (approved):
LGTM

PR #3176, opened 2023-10-18
[dashboard] Redirect index to build dashboard

CaseyHillers:
* There's no other screens, redirect to the build page
* This will make it easy for the TV apps to load the build dashboard


keyonghan (approved):
LGTM

PR #3189, opened 2023-10-23
Update buildbucket to 1.6.0, failed to expose the request objects in 1.5.0

ricardoamador:

## Pre-launch Checklist

- [ ] I signed the [CLA].


keyonghan (approved):
LGTM

packages/buildbucket-dart/tool/regenerate.sh
```diff
@@ -5,14 +5,24 @@
 
 set -ex
 
-mkdir -p lib/src/generated
+NO_CLONE=1
```
keyonghan, line 8: Nit: seems a bool variable is enough.

packages/buildbucket-dart/tool/regenerate.sh
```diff
@@ -5,14 +5,24 @@
...
+if [[ $1 == '-no_clone' ]]; then
```
keyonghan, line 9: What is the use case for `no clone`? Do we need to regenerate the whole package each time?

packages/buildbucket-dart/tool/regenerate.sh
ricardoamador: There are no booleans in bash unless something has changed. 0 is true and 1 is false.

packages/buildbucket-dart/tool/regenerate.sh
ricardoamador: Ah I added this to save myself some time when trying to build the protos. Instead of blowing away the directory every time where nothing has changed I just added the flag so it would not reclone all of the code. If you run it as intended then it won't do anything. I also did not want to build out a full on args parser just for this but I can if that is preferred.

`sh tool/regenerate.sh` will do everything
`sh tool/regenerate.sh -no_clone` will not blow away the dir and clone everything. This is much faster when you are trying to add variables. But I get this might not be needed so I am fine with removing if we don't.

ricardoamador: Actually. I will change it back.

keyonghan (approved):
LGTM

PR #3190, opened 2023-10-24
Skip backfiller for recently created new targets

keyonghan:
Fixes https://github.com/flutter/flutter/issues/136487
Context: https://github.com/flutter/flutter/issues/136487#issuecomment-1778101873

### DIFF
--- a/app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
+++ b/app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
@@ -81,8 +81,8 @@ class BatchBackfiller extends RequestHandler {
         continue;
       }
       final FullTask? backfillTask = _backfillTask(target, taskColumn);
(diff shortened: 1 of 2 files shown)


app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
```diff
@@ -200,10 +200,16 @@ class BatchBackfiller extends RequestHandler {
 
   /// Returns priority for back filled targets.
   ///
+  /// Skips scheduling newly created targets whose available entries are
+  /// less than
```
ricardoamador, line 204: Is this comment complete?

ricardoamador (approved):
LGTM with nit.

app_dart/lib/src/request_handlers/scheduler/batch_backfiller.dart
keyonghan: Ah, good catch! I missed the `BatchPolicy.kBatchSize`. Updated.

PR #3196, opened 2023-10-27
Fix build dashboard task status for infra failures

keyonghan:
Fixes: https://github.com/flutter/flutter/issues/136908


app_dart/lib/src/model/appengine/task.dart
```diff
@@ -449,10 +449,12 @@ class Task extends Model<int> {
         return status = statusSucceeded;
       case Result.canceled:
         return status = statusCancelled;
-      case Result.infraFailure:
```
ricardoamador, line 452: Is infraFailure not needed? I would think since it is a separate ordinal it would still be viable to check for?

keyonghan: It turns out that `Result` does not support `INFRA_FAILURE`, so we need to determine the infra failure case with the combination of `Result.failure` and `FailureReason.infraFailure`
See more details: https://github.com/flutter/flutter/issues/136908#issuecomment-1783273728

ricardoamador: Gotcha! Might be good to add a note here. Though this is likely to change with the migration to buildbucket v2 so up to you.

ricardoamador (approved):
LGTM

app_dart/lib/src/model/appengine/task.dart
keyonghan: Added a doc to make is clear. Also removed the existent `Result.infraFailure`.

PR #3212, opened 2023-11-06
Add log warning when GitHub API usage exceeds half of the limit

keyonghan:
Part of https://github.com/flutter/flutter/issues/132616.

This is to help triage and prepare for alert setup in GCP.

### DIFF
--- a/app_dart/lib/src/request_handlers/github_rate_limit_status.dart
+++ b/app_dart/lib/src/request_handlers/github_rate_limit_status.dart
@@ -38,9 +38,11 @@ class GithubRateLimitStatus extends RequestHandler<Body> {
 
     final int remainingQuota = quotaUsage['remaining'] as int;
     const int quotaLimitSLO = 7500; // Half of the quota limit 15000.
-    if(remainingQuota < quotaLimitSLO) {
-      log.warning('Remaining GitHub quota is $remainingQuota, which is less than $quotaLimitSLO (half of the quotaLimit 15000).');
+    if (remainingQuota < quotaLimitSLO) {
+      log.warning(
+          'Remaining GitHub quota is $remainingQuota, which is less than $quotaLimitSLO (half of the quotaLimit 15000).');
     }
+
     /// Insert quota usage to BigQuery
     const String githubQuotaTable = 'GithubQuotaUsage';
     await insertBigquery(githubQuotaTable, quotaUsage, await config.createTabledataResourceApi());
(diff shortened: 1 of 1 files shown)


app_dart/lib/src/request_handlers/github_rate_limit_status.dart
```diff
@@ -38,9 +38,11 @@ class GithubRateLimitStatus extends RequestHandler<Body> {
 
     final int remainingQuota = quotaUsage['remaining'] as int;
     const int quotaLimitSLO = 7500; // Half of the quota limit 15000.
-    if(remainingQuota < quotaLimitSLO) {
-      log.warning('Remaining GitHub quota is $remainingQuota, which is less than $quotaLimitSLO (half of the quotaLimit 15000).');
+    if (remainingQuota < quotaLimitSLO) {
+      log.warning(
+          'Remaining GitHub quota is $remainingQuota, which is less than $quotaLimitSLO (half of the quotaLimit 15000).');
```
yusuf-goog, line 43: Remove the 15000, it will make this more future proof.

app_dart/lib/src/request_handlers/github_rate_limit_status.dart
```diff
@@ -35,10 +36,16 @@ class GithubRateLimitStatus extends RequestHandler<Body> {
     final Map<String, dynamic> quotaUsage = (await githubService.getRateLimit()).toJson();
     quotaUsage['timestamp'] = DateTime.now().toIso8601String();
 
+    final int remainingQuota = quotaUsage['remaining'] as int;
+    const int quotaLimitSLO = 7500; // Half of the quota limit 15000.
```
yusuf-goog, line 40: Can we get this number programatically?

app_dart/lib/src/request_handlers/github_rate_limit_status.dart
ricardoamador: I thought the quota was 16000?

app_dart/lib/src/request_handlers/github_rate_limit_status.dart
ricardoamador: It's in the headers and the github library does expose these.

app_dart/lib/src/request_handlers/github_rate_limit_status.dart
keyonghan: use variable instead.

app_dart/lib/src/request_handlers/github_rate_limit_status.dart
keyonghan: Updated to use the value from API data.

PR #3233, opened 2023-11-07
[dashboard] Add Android signing config

CaseyHillers:
Needed for uploading builds to the playstore.

I've uploaded the key file to valentine.

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/dashboard/android/key.properties
+++ b/dashboard/android/key.properties
@@ -1,0 +1,4 @@
+storePassword=dartdart
+keyPassword=dartdart
+keyAlias=upload
+storeFile=$HOME/upload-keystore.jks
--- a/dashboard/.gitignore
+++ b/dashboard/.gitignore
@@ -47,7 +47,6 @@ unlinked_spec.ds
 **/android/gradlew.bat
 **/android/local.properties
 **/android/**/GeneratedPluginRegistrant.java
-**/android/key.properties
 *.jks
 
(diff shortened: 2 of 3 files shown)


dashboard/android/key.properties
```diff
@@ -0,0 +1,4 @@
+storePassword=dartdart
```
CaseyHillers, line 1: AFAIK this isn't an actual password. It just verifies you're using the expected key file.

tvolkert: I think you're not supposed to check in key.properties? (and this may be the actual password?)

tvolkert: Ah right, this is the password to the key file, which contains the actual password...

dashboard/android/key.properties
```diff
@@ -0,0 +1,4 @@
...
+keyPassword=dartdart
+keyAlias=upload
+storeFile=$HOME/upload-keystore.jks
```
tvolkert, line 4: Make sure you store this file somewhere secure that's able to be ACL'd by other Googlers (like a password storage system)

tvolkert (approved):
LGTM

dashboard/android/key.properties
CaseyHillers: Done! I updated the README with instructions for the one time set up.

PR #3248, opened 2023-11-09
Default engine to requiring tests

stuartmorgan:
The current engine rules are the reverse of the framework and packages repos; those assume that any file that has not been explicitly exempted needs tests, whereas the engine only requires tests for a specific set of files. This is prone to false negatives (e.g., if/when we start adding Swift files, someone would have had to remember to come update this rule, which probably wouldn't have happened).

This flips the engine to be consistent with the other repositories. This will almost certainly cause a lot of false positives in the short term while the set of files that should be exempted is worked out.

Also fixes a bug where the removal-only case was only being applied to Dart files, not other languages.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/app_dart/lib/src/request_handlers/github/webhook_subscription.dart
+++ b/app_dart/lib/src/request_handlers/github/webhook_subscription.dart
@@ -371,6 +371,7 @@ class GithubWebhookSubscription extends SubscriptionHandler {
         filename.contains('analysis_options.yaml') ||
         filename.contains('AUTHORS') ||
         filename.contains('CODEOWNERS') ||
+        filename == 'DEPS' ||
         filename.contains('TESTOWNERS') ||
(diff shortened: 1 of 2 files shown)


stuartmorgan:
This logic came up in #hackers yesterday (since it was buggy for deletions); it's always bugged me that it's so different from framework and packages, and the recent discussion about more strictly applying the testing rule seemed like a good opportunity to fix it.

I'll solicit feedback in the engine channel about this in case there are any strong arguments for engine being opt-in rather than opt-out, and if there are obvious things we should exempt preemptively in this PR.

zanderso (approved):
LGTM

stuartmorgan:
Added gn/gni exemptions per Discord discussion.

PR #3257, opened 2023-11-15
[dashboard] Fix overlay box size

CaseyHillers:
* On web, it's inheriting the mobile version, which present the wrong size
* Use static sizes instead of inheriting
* Generalize the task overlay tests to make them easier to change when the TaskBox size changes

Fixes https://github.com/flutter/flutter/issues/137705

## Pre-launch Checklist

- [X] I signed the [CLA].


tvolkert:
This doesn't seem the right fix.  What was the failure mode?  Was it that it wasn't finding a `TaskBox` in the hierarchy and thus falling back to the default cell size?

CaseyHillers:
> This doesn't seem the right fix. What was the failure mode? Was it that it wasn't finding a TaskBox in the hierarchy and thus falling back to the default cell size?

Fixed! Moved the TaskBox to the top of the tree to ensure it's in the context when the OverlayEntry is created

tvolkert (approved):
LGTM!

PR #3260, opened 2023-11-16
Change `test: all` labels to also skip `runIf[Not]s`, since the goal is to run all 😎 

matanlurey:
See https://github.com/flutter/flutter/issues/138559.

The `test: all` label should not skip presubmits, since that sort of defeats the point :)

/cc @gaaclarke @dnfield


ricardoamador (approved):

matanlurey:
@CaseyHillers I am going to mark as auto-submit, but feel free to leave comments post-merge and I'll do them!

PR #3283, opened 2023-11-22
[dashboard] show error in dropdown and suppress snackbar error when branch file 404

XilaiZhang:
temporary fix before the next beta is pushed.
when branch version file such as https://github.com/flutter/flutter/blob/beta/bin/internal/release-candidate-branch.version isn't available, show error in the branch selection drop down of flutter dashboard, instead of popping up snackbar errors.

context: https://github.com/flutter/flutter/issues/138755, https://github.com/flutter/flutter/issues/138738, https://github.com/flutter/flutter/issues/138747

### DIFF
--- a/app_dart/lib/src/service/branch_service.dart
+++ b/app_dart/lib/src/service/branch_service.dart
@@ -141,8 +141,10 @@ class BranchService {
           ref: branchName,
         )
         .then((String value) => value.trim())
-        // return empty branch name if branch version file doesn't exist
-        .onError((e, _) => '');
+        .onError((e, _) {
+      log.severe('Could not fetch release version file: $e');
+      return '';
+    });
   }
 
   /// Retrieve the latest canidate branch from all candidate branches.
--- a/app_dart/lib/src/service/branch_service.dart
+++ b/app_dart/lib/src/service/branch_service.dart
@@ -134,17 +134,14 @@ class BranchService {
     required gh.RepositorySlug slug,
     required String branchName,
   }) async {
-    String branchNameFromFile;
-    try {
-      branchNameFromFile = await githubService.getFileContent(
-        slug,
-        'bin/internal/release-candidate-branch.version',
-        ref: branchName,
-      );
-    } catch (e) {
-      return '$branchName: 404 file not found';
-    }
-    return branchNameFromFile.trim();
+    return githubService
+        .getFileContent(
+          slug,
+          'bin/internal/release-candidate-branch.version',
+          ref: branchName,
+        )
+        .then((String value) => value.trim())
+        .onError((e, _) => '');
   }
 
   /// Retrieve the latest canidate branch from all candidate branches.
--- a/app_dart/lib/src/service/branch_service.dart
+++ b/app_dart/lib/src/service/branch_service.dart
@@ -141,6 +141,7 @@ class BranchService {
           ref: branchName,
         )
         .then((String value) => value.trim())
+        // return empty branch name if branch version file doesn't exist
         .onError((e, _) => '');
   }
 
--- a/app_dart/lib/src/service/branch_service.dart
+++ b/app_dart/lib/src/service/branch_service.dart
@@ -134,12 +134,17 @@ class BranchService {
     required gh.RepositorySlug slug,
     required String branchName,
   }) async {
-    return (await githubService.getFileContent(
-      slug,
-      'bin/internal/release-candidate-branch.version',
-      ref: branchName,
-    ))
-        .trim();
+    String branchNameFromFile;
+    try {
+      branchNameFromFile = await githubService.getFileContent(
+        slug,
+        'bin/internal/release-candidate-branch.version',
+        ref: branchName,
+      );
+    } catch (e) {
+      return '$branchName: 404 file not found';
+    }
+    return branchNameFromFile.trim();
   }
 
   /// Retrieve the latest canidate branch from all candidate branches.
--- a/app_dart/test/service/branch_service_test.dart
+++ b/app_dart/test/service/branch_service_test.dart
@@ -62,10 +62,6 @@ void main() {
       when(mockRepositoriesService.getContents(any, any)).thenAnswer((Invocation invocation) {
         throw Exception('404 file not found');
       });
-      // when(githubService.getBranchNameFromFile(any)).thenAnswer((Invocation invocation) {
-      //   return Stream.fromIterable([candidateBranch]);
-      // });
-      // Cocoon flug throws 404 file not found exceptions.
       final List<Map<String, String>> result =
           await branchService.getReleaseBranches(githubService: githubService, slug: Config.cocoonSlug);
       final betaBranch = result.singleWhere((Map<String, String> branch) => branch['name'] == 'beta');
--- a/app_dart/test/service/branch_service_test.dart
+++ b/app_dart/test/service/branch_service_test.dart
@@ -12,6 +12,7 @@ import 'package:mockito/mockito.dart';
 import 'package:retry/retry.dart';
 import 'package:test/test.dart';
 
+import '../request_handlers/check_flaky_builders_test_data.dart';
 import '../src/datastore/fake_datastore.dart';
 import '../src/service/fake_gerrit_service.dart';
 import '../src/service/fake_github_service.dart';
@@ -58,13 +59,19 @@ void main() {
       when(mockRepositoriesService.listBranches(any)).thenAnswer((Invocation invocation) {
         return Stream.fromIterable([candidateBranch]);
       });
+      when(mockRepositoriesService.getContents(any, any)).thenAnswer((Invocation invocation) {
+        throw Exception('404 file not found');
+      });
+      // when(githubService.getBranchNameFromFile(any)).thenAnswer((Invocation invocation) {
+      //   return Stream.fromIterable([candidateBranch]);
+      // });
       // Cocoon flug throws 404 file not found exceptions.
       final List<Map<String, String>> result =
           await branchService.getReleaseBranches(githubService: githubService, slug: Config.cocoonSlug);
       final betaBranch = result.singleWhere((Map<String, String> branch) => branch['name'] == 'beta');
-      expect(betaBranch['branch'], 'beta: 404 file not found');
+      expect(betaBranch['branch'], '');
       final stableBranch = result.singleWhere((Map<String, String> branch) => branch['name'] == 'stable');
-      expect(stableBranch['branch'], 'stable: 404 file not found');
+      expect(stableBranch['branch'], '');
     });
 
     test('return beta, stable, and latest candidate branches', () async {
@@ -77,6 +84,30 @@ void main() {
       final gh.Branch candidateBranchThree = generateBranch(6, name: 'flutter-0.5-candidate.0', sha: 'someZeroValues');
       final gh.Branch candidateCherrypickBranch =
           generateBranch(6, name: 'cherry-picks-flutter-3.11-candidate.3', sha: 'bad');
+
+      when(
+        mockRepositoriesService.getContents(
+          Config.flutterSlug,
+          'bin/internal/release-candidate-branch.version',
+          ref: 'beta',
+        ),
+      ).thenAnswer((Invocation invocation) {
+        return Future<gh.RepositoryContents>.value(
+          gh.RepositoryContents(file: gh.GitHubFile(content: gitHubEncode('flutter-3.2-candidate.5\n'))),
+        );
+      });
+
+      when(
+        mockRepositoriesService.getContents(
+          Config.flutterSlug,
+          'bin/internal/release-candidate-branch.version',
+          ref: 'stable',
+        ),
+      ).thenAnswer((Invocation invocation) {
+        return Future<gh.RepositoryContents>.value(
+          gh.RepositoryContents(file: gh.GitHubFile(content: gitHubEncode('flutter-2.13-candidate.0\n'))),
+        );
+      });
 
       when(mockRepositoriesService.listBranches(any)).thenAnswer((Invocation invocation) {
         return Stream.fromIterable([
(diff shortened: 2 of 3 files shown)


app_dart/lib/src/service/branch_service.dart
```diff
@@ -134,12 +134,17 @@ class BranchService {
...
-        .trim();
+    String branchNameFromFile;
+    try {
+      branchNameFromFile = await githubService.getFileContent(
+        slug,
+        'bin/internal/release-candidate-branch.version',
+        ref: branchName,
+      );
+    } catch (e) {
+      return '$branchName: 404 file not found';
```
CaseyHillers, line 145: It would be easier downstream to catch this with an empty string. Then the only check needed is `branch.isEmpty`

app_dart/test/src/service/fake_github_service.dart
```diff
@@ -77,6 +79,11 @@ class FakeGithubService implements GithubService {
   @override
   Future<String> getFileContent(RepositorySlug slug, String path, {String? ref}) async {
     if (path == 'bin/internal/release-candidate-branch.version') {
+      // Test error handling when branch version file isn't available.
+      // https://github.com/flutter/flutter/issues/138755
+      if (slug.name != 'flutter') {
```
CaseyHillers, line 84: Instead of hard coding this to fail on non-flutter files, can we inject the error?

app_dart/test/service/branch_service_test.dart
```diff
@@ -53,6 +53,20 @@ void main() {
       when(githubService.github.repositories).thenReturn(mockRepositoriesService);
     });
 
+    test('return 404 when branch version file does not exist', () async {
+      final gh.Branch candidateBranch = generateBranch(3, name: 'flutter-3.4-candidate.5', sha: '789dev');
+      when(mockRepositoriesService.listBranches(any)).thenAnswer((Invocation invocation) {
+        return Stream.fromIterable([candidateBranch]);
+      });
+      // Cocoon flug throws 404 file not found exceptions.
```
CaseyHillers, line 61:
```suggestion
      // Cocoon slug throws 404 file not found exceptions.
```

app_dart/lib/src/service/branch_service.dart
```diff
@@ -134,12 +134,17 @@ class BranchService {
     required gh.RepositorySlug slug,
     required String branchName,
   }) async {
-    return (await githubService.getFileContent(
-      slug,
-      'bin/internal/release-candidate-branch.version',
-      ref: branchName,
-    ))
-        .trim();
+    String branchNameFromFile;
```
CaseyHillers, line 137: Prefer using finals. We could rewrite this like

```dart
return githubService.getFileContent(
  slug,
  'bin/internal/release-candidate-branch.version',
   ref: branchName,
)
.then((String value) => value.trim())
.onError((_) => '');
```

https://dart.dev/tools/linter-rules/prefer_final_locals#:~:text=DO%20prefer%20declaring%20variables%20as,the%20compiler%20to%20do%20optimizations.

XilaiZhang: Synced with Casey and will return empty string, because other people might be able to create a branch with the nonempty name, and gain release branch privilege.

app_dart/lib/src/service/branch_service.dart
XilaiZhang: Synced with Casey and will return empty string, because other people might be able to create a branch with the nonempty name, and gain release branch privilege.

app_dart/test/src/service/fake_github_service.dart
XilaiZhang: refactored fake class and mock instances.

would the current implementation be considered error injection?

app_dart/test/service/branch_service_test.dart
XilaiZhang: Yeah good point, removed this section

app_dart/test/service/branch_service_test.dart
```diff
@@ -53,6 +54,22 @@ void main() {
       when(githubService.github.repositories).thenReturn(mockRepositoriesService);
     });
 
+    test('return 404 when branch version file does not exist', () async {
```
CaseyHillers, line 57:
```suggestion
    test('return empty when branch version file does not exist', () async {
```

app_dart/test/service/branch_service_test.dart
```diff
@@ -53,6 +54,22 @@ void main() {
...
+      final gh.Branch candidateBranch = generateBranch(3, name: 'flutter-3.4-candidate.5', sha: '789dev');
+      when(mockRepositoriesService.listBranches(any)).thenAnswer((Invocation invocation) {
+        return Stream.fromIterable([candidateBranch]);
+      });
+      when(mockRepositoriesService.getContents(any, any)).thenAnswer((Invocation invocation) {
+        throw Exception('404 file not found');
```
CaseyHillers, line 63: Instead of using generic exceptions, we should throw what is likely to be thrown

```suggestion
        throw GitHubError(github, '404 file not found');
```

app_dart/test/service/branch_service_test.dart
```diff
@@ -53,6 +54,22 @@ void main() {
...
+      });
+      final List<Map<String, String>> result =
+          await branchService.getReleaseBranches(githubService: githubService, slug: Config.cocoonSlug);
+      final betaBranch = result.singleWhere((Map<String, String> branch) => branch['name'] == 'beta');
+      expect(betaBranch['branch'], '');
```
CaseyHillers, line 68: Writing tests using behavior driven development makes them easier to follow

```suggestion
      expect(betaBranch['branch'].isEmpty, isTrue);
```

app_dart/test/service/branch_service_test.dart
```diff
@@ -53,6 +54,22 @@ void main() {
...
+      final stableBranch = result.singleWhere((Map<String, String> branch) => branch['name'] == 'stable');
+      expect(stableBranch['branch'], '');
```
CaseyHillers, line 70:
```suggestion
      expect(stableBranch['branch'].isEmpty, isTrue);
```

app_dart/test/service/branch_service_test.dart
```diff
@@ -53,6 +54,22 @@ void main() {
       when(githubService.github.repositories).thenReturn(mockRepositoriesService);
     });
 
+    test('return 404 when branch version file does not exist', () async {
+      final gh.Branch candidateBranch = generateBranch(3, name: 'flutter-3.4-candidate.5', sha: '789dev');
+      when(mockRepositoriesService.listBranches(any)).thenAnswer((Invocation invocation) {
+        return Stream.fromIterable([candidateBranch]);
+      });
+      when(mockRepositoriesService.getContents(any, any)).thenAnswer((Invocation invocation) {
```
CaseyHillers, line 62: `thenAnswer` is for futures, in this case you can use `thenReturn`

```suggestion
      when(mockRepositoriesService.getContents(any, any)).thenReturn((_) =>
```

app_dart/test/service/branch_service_test.dart
```diff
@@ -53,6 +54,22 @@ void main() {
       when(githubService.github.repositories).thenReturn(mockRepositoriesService);
     });
 
+    test('return 404 when branch version file does not exist', () async {
+      final gh.Branch candidateBranch = generateBranch(3, name: 'flutter-3.4-candidate.5', sha: '789dev');
+      when(mockRepositoriesService.listBranches(any)).thenAnswer((Invocation invocation) {
+        return Stream.fromIterable([candidateBranch]);
```
CaseyHillers, line 60:
```suggestion
        return Stream.value(candidateBranch);
```

app_dart/test/service/branch_service_test.dart
```diff
@@ -53,6 +54,22 @@ void main() {
       when(githubService.github.repositories).thenReturn(mockRepositoriesService);
     });
 
+    test('return 404 when branch version file does not exist', () async {
+      final gh.Branch candidateBranch = generateBranch(3, name: 'flutter-3.4-candidate.5', sha: '789dev');
+      when(mockRepositoriesService.listBranches(any)).thenAnswer((Invocation invocation) {
```
CaseyHillers, line 59:
```suggestion
      when(mockRepositoriesService.listBranches(any)).thenAnswer((_) =>
```

app_dart/test/service/branch_service_test.dart
```diff
@@ -64,6 +81,30 @@ void main() {
       final gh.Branch candidateCherrypickBranch =
           generateBranch(6, name: 'cherry-picks-flutter-3.11-candidate.3', sha: 'bad');
 
+      when(
+        mockRepositoriesService.getContents(
+          Config.flutterSlug,
+          'bin/internal/release-candidate-branch.version',
+          ref: 'beta',
+        ),
+      ).thenAnswer((Invocation invocation) {
```
CaseyHillers, line 90:
```suggestion
      ).thenAnswer((_) async => gh.RepositoryContents(file: gh.GitHubFile(content: gitHubEncode('flutter-3.2-candidate.5\n'))));
```

app_dart/test/service/branch_service_test.dart
```diff
@@ -64,6 +81,30 @@ void main() {
...
+      });
+
+      when(
+        mockRepositoriesService.getContents(
+          Config.flutterSlug,
+          'bin/internal/release-candidate-branch.version',
+          ref: 'stable',
+        ),
+      ).thenAnswer((Invocation invocation) {
+        return Future<gh.RepositoryContents>.value(
```
CaseyHillers, line 103: ditto about using arrow functions, async, and hiding invocation

CaseyHillers (commented):
LGTM! I defer final review to the infra team

app_dart/test/service/branch_service_test.dart
XilaiZhang: Sure, updated

app_dart/test/service/branch_service_test.dart
XilaiZhang: Sure, updated

app_dart/test/service/branch_service_test.dart
XilaiZhang: Oh good point, thank you Casey!

app_dart/test/service/branch_service_test.dart
XilaiZhang: umm Couldn't get `thenReturn` to work, ended up using `thenThrow` for the Error type

app_dart/test/service/branch_service_test.dart
XilaiZhang: Sure, updated

app_dart/test/service/branch_service_test.dart
XilaiZhang: Sure, updated. 

umm but I thought returning an empty string might be the behavior we are expecting?

app_dart/test/service/branch_service_test.dart
XilaiZhang: Oh TIL about `GithubError`

app_dart/test/service/branch_service_test.dart
XilaiZhang: Sure, updated

app_dart/test/service/branch_service_test.dart
CaseyHillers: `''` is harder to see than `isEmpty`. In BDD, making tests read like a sentence is part of the goal.

app_dart/test/src/service/fake_github_service.dart
CaseyHillers: Yep, nice!

ricardoamador:
Hey @XilaiZhang you wrote:

> temporary fix before the next beta is pushed.

What is the plan for the actual fix? Thanks.

XilaiZhang:
> Hey @XilaiZhang you wrote:
> 
> > temporary fix before the next beta is pushed.
> 
> What is the plan for the actual fix? Ideally we would like to avoid partial fixes. Thanks.
> 
> I don't see any issue with this change but would like a little more info.

yeah the actual fix would be when we restore the branch file, at the time of pushing the next beta release

the goal of this PR is to prevent a cocoon dashboard error, if branch files are deleted again by error in the future

and sure, more info were discussed and reported in https://github.com/flutter/flutter/issues/138755, https://github.com/flutter/flutter/issues/138738, https://github.com/flutter/flutter/issues/138747

ricardoamador:
So I would prefer the point you made and have a log of the error at least rather than swallowing the exception and saying nothing.

XilaiZhang:
> So I would prefer you log the error rather than rather than swallowing the exception and saying nothing if the UI snackbar error is being removed. At least we know the problem will still exist and there is a follow up to it.
> 
> Or at least add a TODO note as to why it is being done with a linked issue.

yeah i think Casey's vision was that the error would be reported and shown in the drop down button instead. when we go to cocoon dashboard and click on the 'branch' button to select branch, it should show an empty block for release branch if the version file is 404.

app_dart/lib/src/service/branch_service.dart
```diff
@@ -134,12 +134,14 @@ class BranchService {
...
-    ))
-        .trim();
+    return githubService
+        .getFileContent(
+          slug,
+          'bin/internal/release-candidate-branch.version',
+          ref: branchName,
+        )
+        .then((String value) => value.trim())
+        .onError((e, _) => '');
```
ricardoamador, line 144: Can we add a TODO here with the follow up issue linked. That may be easier than logging.

ricardoamador:
Okay but will changes need to be made here again in the future?

app_dart/lib/src/service/branch_service.dart
XilaiZhang: Sure, added a comment on why we return empty string. Yeah the expectation is that this would be the PR so that no follow up fix is needed.

XilaiZhang:
> Okay but will changes need to be made here again in the future?

right changes won't be needed here. I think the idea is that regardless of other things (restoring beta version file), the logic in this PR is needed so that it prevents spitting errors in the future.

ricardoamador (approved):
LGTM

PR #3311, opened 2023-12-05
[scheduler] Add default values to runIf

CaseyHillers:
I'll follow up in the separate repos ci.yamls once this has landed to clean them up.

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/app_dart/lib/src/model/ci_yaml/target.dart
+++ b/app_dart/lib/src/model/ci_yaml/target.dart
@@ -240,6 +240,16 @@ class Target {
     return value.bringup ? 'staging' : 'prod';
   }
 
+  /// Get the runIf globs for when this target should run.
+  ///
+  /// If not empty, will include default globs that should trigger all tests.
+  List<String> get runIf {
+    if (value.runIf.isEmpty) {
+      return <String>[];
+    }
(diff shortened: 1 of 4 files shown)


app_dart/lib/src/model/ci_yaml/target.dart
```diff
@@ -240,6 +240,16 @@ class Target {
...
   }
 
+  /// Get the runIf globs for when this target should run.
+  ///
+  /// If not empty, will include default globs that should trigger all tests.
+  List<String> get runIf {
+    if (value.runIf.isEmpty) {
+      return <String>[];
+    }
+    return <String>['.ci.yaml', 'DEPS', ...value.runIf];
```
keyonghan, line 250: Is this only for simplicity in .ci.yaml? It seems more clear to me to have `.ci.yaml/DEPS` listed in the target, rather than a `magic` logic handled in the backend logic.

keyonghan:
Does it make sense to put the default values on the platform level? That way it's visible to people and we also get the control.

CaseyHillers:
Closing in favor of https://github.com/flutter/cocoon/pull/3333

PR #3317, opened 2023-12-06
Unpin dependencies that are pinned by the Flutter SDK

Hixie:
These pins otherwise prevent us from rolling the pins in the Flutter SDK because this is a a presubmit test for that repo.

- also, cleanup boilerplate comments
- also, add breadcrumbs for the next person who wonders how these dependencies roll
- also, alphabetize the dependencies

### DIFF
--- a/dashboard/pubspec.yaml
+++ b/dashboard/pubspec.yaml
@@ -5,83 +5,37 @@
 name: flutter_dashboard
 description: The Flutter dashboard.
 publish_to: none
-
-# The following defines the version and build number for your application.
-# A version number is three numbers separated by dots, like 1.2.43
-# followed by an optional build number separated by a +.
-# Both the version and the builder number may be overridden in flutter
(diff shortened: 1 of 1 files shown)


parlough (approved):
This all looks good, and the cleanup is nice!

There's a [`sort_pub_dependencies`](https://dart.dev/tools/linter-rules/sort_pub_dependencies) lint that might be worth enabling :) 

Will dependabot keep the comments when updating versions?

Hixie:
> There's a [`sort_pub_dependencies`](https://dart.dev/tools/linter-rules/sort_pub_dependencies) lint that might be worth enabling :)

I only sort of alphabetised them (e.g. the SDK one is first). Not sure what the lint does. :-)


> Will dependabot keep the comments when updating versions?

No idea, I guess we'll find out...

Hixie:
thanks for the review

PR #3322, opened 2023-12-07
[codesign] hide codesign credentials from logs

XilaiZhang:
context: b/311427184

replace codesign credentials in logs with placeholders. 
The log will now look like `this command failed: blah blah ... --apple-id <appleID> --password <appSpecificPassword> --team-id <teamID> ... blah blah` instead of printing the literal passwords.

refactored test to be more stable and reliable.

### DIFF
--- a/cipd_packages/codesign/lib/src/file_codesign_visitor.dart
+++ b/cipd_packages/codesign/lib/src/file_codesign_visitor.dart
@@ -71,6 +71,7 @@ class FileCodesignVisitor {
     'CODESIGN_TEAM_ID': '',
     'APP_SPECIFIC_PASSWORD': '',
   };
+  Map<String, String> redactedCredentials = {};
 
   late final File entitlementsFile;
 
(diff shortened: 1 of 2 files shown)


cipd_packages/codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -416,23 +416,28 @@ update these file paths accordingly.
...
       '--apple-id',
       codesignAppstoreId,
+      '--password',
+      appSpecificPassword,
       '--team-id',
       codesignTeamId,
     ];
 
-    log.info('checking notary status with ${args.join(' ')}');
+    final String argsWithoutCredentials = args
```
CaseyHillers, line 427: Since there's a few of these special keys, should we make it an explicit list?

See https://github.com/flutter/cocoon/pull/1008 as an example.

CaseyHillers (approved):

cipd_packages/codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -416,23 +416,28 @@ update these file paths accordingly.
...
+        .join(' ')
+        .replaceFirst(codesignAppstoreId, '<appleID>')
```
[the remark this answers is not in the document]
XilaiZhang: Sure, changed all the variables to have a `-redacted` suffix

cipd_packages/codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: Sure, stored passwords in a map, and iterate through the map to replace the passwords. 

personally i think replacing one by one is easier but I am good with either approach. updated tests.

PR #3325, opened 2023-12-07
Replace hixie@ with @test-exemption-reviewers.

matanlurey:
More scaleable than listing people's names 1-by-1. Yay!


Hixie:
LGTM

PR #3332, opened 2023-12-08
Flag stale PRs when any github check hangs >2 hours

keyonghan:
This PR creates a warning to GCP log when PR hangs. 
A follow up to to setup an alert based on the flagged logs.

First part of https://github.com/flutter/flutter/issues/125451

### DIFF
--- a/auto_submit/lib/service/config.dart
+++ b/auto_submit/lib/service/config.dart
@@ -48,6 +48,9 @@ class Config {
   /// Labels autosubmit looks for on pull requests
   static const String kAutosubmitLabel = 'autosubmit';
 
+  /// GitHub check stale threshold.
+  static const int kGitHubCheckStaleThreshold = 2; // hours
+
   // Labels the bot looks for on revert requests.
   // TODO (ricardoamador) https://github.com/flutter/flutter/issues/134845:
   // add a link to a one page doc outlining the workflow that happens here.
--- a/auto_submit/lib/validations/ci_successful.dart
+++ b/auto_submit/lib/validations/ci_successful.dart
@@ -79,7 +79,7 @@ class CiSuccessful extends Validation {
     }
 
     /// Validate if all checkRuns have succeeded.
-    allSuccess = validateCheckRuns(slug, prNumber, checkRuns, failures, allSuccess);
+    allSuccess = validateCheckRuns(slug, prNumber, checkRuns, failures, allSuccess, author);
 
     if (!allSuccess && failures.isEmpty) {
       return ValidationResult(allSuccess, Action.IGNORE_TEMPORARILY, '');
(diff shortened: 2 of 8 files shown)


ricardoamador:
Hmm, I wonder if it might be good to add a label or a comment as well as flagging in the logs when a pr is hung? Up to you.

auto_submit/lib/service/config.dart
```diff
@@ -48,6 +48,9 @@ class Config {
   /// Labels autosubmit looks for on pull requests
   static const String kAutosubmitLabel = 'autosubmit';
 
+  /// GitHub check stale threshold.
+  static const int kGitHubCheckStaleThreshold = 2; // hours
```
ricardoamador, line 52: I wonder if this would be something to put into the autosubmit configuration as then a repository could specify if they want to change the stale check value. For example in cocoon it could 1 hour or in flutter it could be 3. Something to think about long term as things may change. Also keep in mind we support dart-lang as well.

ricardoamador (approved):
LGTM

auto_submit/lib/service/config.dart
keyonghan: Yeah, this is a good point. 
But the main motivation of this PR is to help ease engine gardener manually monitoring these roller PRs.
Non-engine roller PRs are expected to be monitored by their authors. I would prefer to focusing on these roller PRs.
If people find it useful to be alerted as well, then we can easily extend the logic by posting a comment.

To begin with, I updated the logic to be limited to engine roller specific.

ricardoamador: Yeah no problem. I am just thinking out a couple of quarters. I have some changes in mind to make to cleanup autosubmit further so not a big deal here.

auto_submit/lib/validations/ci_successful.dart
```diff
@@ -193,7 +208,21 @@ class CiSuccessful extends Validation {
...
   }
+
+  // Treat any GitHub check run as stale if created over [Config.kGitHubCheckStaleThreshold] hours ago.
+  bool isStale(DateTime dateTime) {
+    return dateTime.compareTo(DateTime.now().subtract(const Duration(hours: Config.kGitHubCheckStaleThreshold))) < 0;
+  }
+
+  bool isEngineRoller(Author author, github.RepositorySlug slug) {
+    return config.rollerAccounts.contains(author.login!) && slug == Config.engineSlug;
+  }
```
keyonghan, line 227: @zanderso My earlier change is for rolled PRs landing in Engine repo. I guess we also want to include the engine to framework ones?

zanderso: Yes.

PR #3354, opened 2023-12-18
Update and reenable periodic task vacuum

keyonghan:
We should enable the task vaccum process before https://github.com/flutter/flutter/issues/122117 is fixed, especially when there is limited gardener support.

This PR:
1) updates the vacuum logic to be based on `task.status` is `in progress` and `builder` is null, instead of being based on `created time` which caused https://github.com/flutter/flutter/issues/121989.
2) enables as a daily cronjob running 12:00 PST

### DIFF
--- a/app_dart/lib/src/request_handlers/scheduler/vacuum_stale_tasks.dart
+++ b/app_dart/lib/src/request_handlers/scheduler/vacuum_stale_tasks.dart
@@ -10,7 +10,6 @@ import 'package:meta/meta.dart';
 
 import '../../model/appengine/task.dart';
 import '../../service/datastore.dart';
-import '../../service/logging.dart';
 
 /// Vacuum stale tasks.
 ///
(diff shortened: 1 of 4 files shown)


app_dart/lib/src/request_handlers/scheduler/vacuum_stale_tasks.dart
```diff
@@ -53,32 +53,16 @@ class VacuumStaleTasks extends RequestHandler<Body> {
     final DatastoreService datastore = datastoreProvider(config.db);
 
     final List<FullTask> tasks = await datastore.queryRecentTasks(slug: slug).toList();
-    final Set<Task> tasksToBeReset = <Task>{};
+    final List<Task> tasksToBeReset = <Task>[];
```
godofredoc, line 56: Is this to allow duplicated entries?

app_dart/lib/src/request_handlers/scheduler/vacuum_stale_tasks.dart
```diff
@@ -53,32 +53,16 @@ class VacuumStaleTasks extends RequestHandler<Body> {
...
     for (FullTask fullTask in tasks) {
       final Task task = fullTask.task;
-      if (task.status != Task.statusInProgress) {
-        continue;
-      }
-
-      if (task.createTimestamp == null) {
-        log.fine('Vacuuming $task due to createTimestamp being null');
+      if (task.status == Task.statusInProgress && task.buildNumber == null) {
```
godofredoc, line 59: Do we need to check how long the task has been in progress before resetting?

cron.yaml
```diff
@@ -7,12 +7,11 @@ cron:
...
 
-# Disabled for https://github.com/flutter/flutter/issues/121989
 # TODO(keyonghan): will delete if `In Progress` hanging issue is resolved:
 # https://github.com/flutter/flutter/issues/120395#issuecomment-1444810718
-#- description: vacuum stale tasks
-#  url: /api/scheduler/vacuum-stale-tasks
-#  schedule: every 10 minutes
+- description: vacuum stale tasks
+ url: /api/scheduler/vacuum-stale-tasks
+ schedule: every day 08:00 # PST 12:00am
```
godofredoc, line 14: Do we need to vacuum more frequently?

app_dart/lib/src/request_handlers/scheduler/vacuum_stale_tasks.dart
keyonghan: The old logic to be based on `createTimestamp` caused duplicated scheduling, and that's why I switched to use `buildNumber==null`. Considering the running frequency is super low, I would not worry too much about the timestamp check.

app_dart/lib/src/request_handlers/scheduler/vacuum_stale_tasks.dart
keyonghan: We should not have duplicated entries. The query above returns recent tasks and all are different entries.

cron.yaml
keyonghan: Updated to run twice a day.

godofredoc (approved):

PR #3398, opened 2024-01-04
[codesign] Add additional retries on notarytool commands with exponential backoff

CaseyHillers:
Fixes https://github.com/flutter/flutter/issues/138636

* Migrate to package:retry instead of hand writing the retry logic
  * Enables easily adding exponential backoff
* Minor cleanup

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/cipd_packages/codesign/lib/src/file_codesign_visitor.dart
+++ b/cipd_packages/codesign/lib/src/file_codesign_visitor.dart
@@ -6,7 +6,9 @@ import 'dart:async';
 import 'dart:io' as io;
 
 import 'package:file/file.dart';
+import 'package:meta/meta.dart';
 import 'package:process/process.dart';
+import 'package:retry/retry.dart';
 
 import 'log.dart';
 import 'utils.dart';
@@ -34,6 +36,10 @@ class FileCodesignVisitor {
     required this.codesignAppstoreIDFilePath,
     required this.codesignTeamIDFilePath,
     this.dryrun = true,
+    @visibleForTesting this.retryOptions = const RetryOptions(
+      maxAttempts: 5,
+      delayFactor: Duration(seconds: 2),
+    ),
     this.notarizationTimerDuration = const Duration(seconds: 5),
   }) {
     entitlementsFile = rootDirectory.childFile('Entitlements.plist')..writeAsStringSync(_entitlementsFileContents);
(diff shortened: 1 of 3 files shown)


cipd_packages/codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -34,6 +36,10 @@ class FileCodesignVisitor {
     required this.codesignAppstoreIDFilePath,
     required this.codesignTeamIDFilePath,
     this.dryrun = true,
+    @visibleForTesting this.retryOptions = const RetryOptions(
+      maxAttempts: 5,
```
XilaiZhang, line 40: nit: umm do we need 5 attempts, maybe 3 is enough but its really nit

CaseyHillers: 3 was the previous number, so increasing to 5 with exponential back off should prevent another breakage.

cipd_packages/codesign/test/file_codesign_visitor_test.dart
```diff
@@ -944,46 +950,38 @@ status: Invalid''',
...
-            '--apple-id',
-            fakeAppleID,
-            '--password',
-            fakePassword,
-            '--team-id',
-            fakeTeamID,
-            '--verbose',
-          ],
-          stdout: '''Error uploading file.
+      for (int i = 0; i < 3; i++) {
```
XilaiZhang, line 959: umm should this be 5 attempts if we update max attempts from 3 to 5. 

nit: I guess the test always pass whether the input is 3 or 5. maybe checking the number of repetitions in records would catch this but its nit.

cipd_packages/codesign/lib/src/file_codesign_visitor.dart
XilaiZhang: umm I am not sure if retry is the culprit but I am good with adding more retries.

test below might need to be updated from 3 to 5 too. otherwise LGTM

XilaiZhang (approved):

cipd_packages/codesign/test/file_codesign_visitor_test.dart
CaseyHillers: Generally, refactors should avoid changing the tests too much. In this case, I kept it as 3 instead of 5 as that's probably a good low number to verify the retries pass through exceptions.

PR #3403, opened 2024-01-09
Handle override label in retry for packages repository

keyonghan:
Fixes: https://github.com/flutter/flutter/issues/140232

### DIFF
--- a/app_dart/lib/src/service/github_service.dart
+++ b/app_dart/lib/src/service/github_service.dart
@@ -153,6 +153,11 @@ class GithubService {
     );
   }
 
+  /// Gets label list for an issue.
+  Future<List<IssueLabel>>? getIssueLabels(RepositorySlug slug, int issueNumber) async {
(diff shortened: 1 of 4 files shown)


app_dart/lib/src/service/github_service.dart
```diff
@@ -153,6 +153,11 @@ class GithubService {
     );
   }
 
+  /// Gets label list for an issue.
+  Future<List<IssueLabel>>? getIssueLabels(RepositorySlug slug, int issueNumber) async {
+    return github.issues.listLabelsByIssue(slug, issueNumber).toList();
```
ricardoamador, line 158: Does github.issues.listLabelsByIssue return a nullable list?

ricardoamador (approved):
LGTM

app_dart/lib/src/service/github_service.dart
keyonghan: Good catch! If there are no labels, it returns an empty list. Updated.

PR #3404, opened 2024-01-11
Add null check when bumping github version

keyonghan:
Same as https://github.com/flutter/cocoon/pull/3386 with a fix.


ricardoamador:
Is this caused by a github change?

ricardoamador (approved):
LGTM

keyonghan:
> Is this caused by a github change?

I would expect yes - caused only by a version bump.

PR #3405, opened 2024-01-12
Make engine build configs test exempt

zanderso:
For example for PRs like https://github.com/flutter/engine/pull/49776

### DIFF
--- a/app_dart/lib/src/request_handlers/github/webhook_subscription.dart
+++ b/app_dart/lib/src/request_handlers/github/webhook_subscription.dart
@@ -404,6 +404,8 @@ class GithubWebhookSubscription extends SubscriptionHandler {
           !_isTestExempt(filename) &&
           // License goldens are auto-generated.
           !filename.startsWith('ci/licenses_golden/') &&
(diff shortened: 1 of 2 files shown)


app_dart/lib/src/request_handlers/github/webhook_subscription.dart
```diff
@@ -404,6 +404,8 @@ class GithubWebhookSubscription extends SubscriptionHandler {
           !_isTestExempt(filename) &&
           // License goldens are auto-generated.
           !filename.startsWith('ci/licenses_golden/') &&
```
keyonghan, line 406: Not related to this PR, but should this line cover https://github.com/flutter/engine/pull/49775?

keyonghan (approved):
LGTM

app_dart/lib/src/request_handlers/github/webhook_subscription.dart
zanderso: It would cover `ci/licenses_golden/excluded_files`, but this change is needed to cover the other file in that PR.

PR #3429, opened 2024-01-19
[codesign] codesign framework bundle

XilaiZhang:
**context:** 
https://github.com/flutter/flutter/issues/140934

**what I learnt:**
1. out of our current [release bundles downloaded from gcs](https://pantheon.corp.google.com/storage/browser/flutter_infra_release/flutter/4f18bb4dcb99d7e0d94aeff2ffa74e6ddef301cd;tab=objects?authuser=0&project=flutter-infra&prefix=&forceOnObjectsSortingFiltering=false), only Flutter.xcframework is not codesigned. Both FlutterMacOS.framework and Flutter.xcframework/*/Flutter.framework are already codesigned. I will explain in more details on why this happens in bullet point 5 below. <br />
The existing bundles are verified using `codesign --verify --verbose --display <path to codesign bundle>`. 

2. According to Eskimo's reply in [this post](https://developer.apple.com/forums/thread/661852) and [this other post](https://developer.apple.com/forums/thread/129678). For nested bundle / framework we should codesign them from the inside-out. And thus in this PR, we codesign the binaries and sub-bundles inside our bundle first, and then codesign the entire bundle.

3. Both .xcframework and .framework direcotry/folders qualify as bundles. <br />
According to the [in-depth developer doc on code signing](https://developer.apple.com/library/archive/documentation/Security/Conceptual/CodeSigningGuide/Procedures/Procedures.html):
`Bundles must have their Info.plist in the proper location. For app bundles, this is in Contents. For frameworks, this is in Versions/Current/Resources.`
So from the look of our folders, FluttMacOs.framework and Flutter.xcframework/*/Flutter.framework seem to qualify as bundles, but Flutter.xcframework doesn't. 
However, when running the codesign tool it recognizes Flutter.xcframework as bundle: `Flutter.xcframework/: signed bundle`.

4. We are currently signing version `Current` in the bundle, which I think would be good enough. <br />
According to the [in-depth developer doc on code signing](https://developer.apple.com/library/archive/documentation/Security/Conceptual/CodeSigningGuide/Procedures/Procedures.html):
`When you manually code sign a framework, only the current version, the one pointed at by the Versions/Current symbolic link, is signed by default.` For our bundle, we only have Version A and Version Current, and since Version Current is a symlink to Version A, we codesigned all of our versions. 

5. [my theory] some bundles are already codesigned because the main executable of the bundle is codesigned <br />
Eskimo hinted in [this post](https://developer.apple.com/forums/thread/129678) that `The only exception to this is the main executable of a bundle, where signing the bundle takes care of signing its main executable`. 
And [Apple Technical Note 2206](https://developer.apple.com/library/archive/technotes/tn2206/_index.html) pointed out that `Most frameworks contain a single version and can in fact be signed directly.`<br />
This could provide a theory on how our `FlutterMacOS.framework` bundle is already code signed because the major executable` Versions/A/FlutterMacOS` is codesigned. And our `Flutter.xcFramework/*/Flutter.framework` bundle is codesigned because the major executable `Flutter.xcFramework/*/Flutter.framework/Flutter` is code signed. 
However, .xcFramework itself is not already codesigned and will be included through the change in this PR.

**Tested**:
The output artifact of this PR is [built and tested](https://github.com/flutter/cocoon/tree/main/cipd_packages/codesign)
The output artifact contains a code signed xcFramework, verified using the command mentioned in bullet point 1.

**TODO**: 
deploy this current version of codesign and set cipd label to 'live' when this PR is merged.

### DIFF
--- a/cipd_packages/codesign/lib/src/file_codesign_visitor.dart
+++ b/cipd_packages/codesign/lib/src/file_codesign_visitor.dart
@@ -256,8 +256,8 @@ update these file paths accordingly.
       log.info('Child file of directory ${directory.basename} is ${entity.basename}');
     }
     final String directoryExtension = directory.basename.split('.').last;
-    if( directoryExtension == 'framework' || directoryExtension == 'xcframework'){
-        final List<String> args = <String>[
+    if (directoryExtension == 'framework' || directoryExtension == 'xcframework') {
+      final List<String> args = <String>[
         '/usr/bin/codesign',
         '--keychain',
         'build.keychain', // specify the keychain to look for cert
@@ -275,12 +275,12 @@ update these file paths accordingly.
           return;
         }
 
-      throw CodesignException(
-        'Failed to codesign bundle ${directory.absolute.path} with args: ${args.join(' ')}\n'
-        'stdout:\n${(result.stdout as String).trim()}'
-        'stderr:\n${(result.stderr as String).trim()}',
-      );
-    });
+        throw CodesignException(
+          'Failed to codesign bundle ${directory.absolute.path} with args: ${args.join(' ')}\n'
+          'stdout:\n${(result.stdout as String).trim()}'
+          'stderr:\n${(result.stderr as String).trim()}',
+        );
+      });
     }
   }
 
--- a/cipd_packages/codesign/lib/src/file_codesign_visitor.dart
+++ b/cipd_packages/codesign/lib/src/file_codesign_visitor.dart
@@ -257,30 +257,7 @@ update these file paths accordingly.
     }
     final String directoryExtension = directory.basename.split('.').last;
     if (directoryExtension == 'framework' || directoryExtension == 'xcframework') {
-      final List<String> args = <String>[
-        '/usr/bin/codesign',
-        '--keychain',
-        'build.keychain', // specify the keychain to look for cert
-        '-f', // force. Possible re-signing of FlutterMacOS.framework, Flutter.xcframework/ios-arm64/Flutter.framework etc.
-        '-s', // use the cert provided by next argument
-        codesignCertName,
-        directory.absolute.path,
-        '--timestamp', // add a secure timestamp
-      ];
-
-      await retryOptions.retry(() async {
-        log.info('Code signing framework bundle: ${args.join(' ')}\n');
-        final io.ProcessResult result = await processManager.run(args);
-        if (result.exitCode == 0) {
-          return;
-        }
-
-        throw CodesignException(
-          'Failed to codesign bundle ${directory.absolute.path} with args: ${args.join(' ')}\n'
-          'stdout:\n${(result.stdout as String).trim()}'
-          'stderr:\n${(result.stderr as String).trim()}',
-        );
-      });
+      await codesignAtPath(binaryOrBundlePath: directory.absolute.path);
     }
   }
 
@@ -341,17 +318,24 @@ update these file paths accordingly.
     if (dryrun) {
       return;
     }
+    await codesignAtPath(binaryOrBundlePath: binaryFile.absolute.path, entitlementCurrentPath: entitlementCurrentPath);
+  }
+
+  Future<void> codesignAtPath({
+    required String binaryOrBundlePath,
+    String? entitlementCurrentPath,
+  }) async {
     final List<String> args = <String>[
       '/usr/bin/codesign',
       '--keychain',
       'build.keychain', // specify the keychain to look for cert
-      '-f', // force
+      '-f', // force. Needed to overwrite signature if major executable of bundle is already signed before bundle is signed.
       '-s', // use the cert provided by next argument
       codesignCertName,
-      binaryFile.absolute.path,
+      binaryOrBundlePath,
       '--timestamp', // add a secure timestamp
       '--options=runtime', // hardened runtime
-      if (fileWithEntitlements.contains(entitlementCurrentPath)) ...<String>[
+      if (entitlementCurrentPath != '' && fileWithEntitlements.contains(entitlementCurrentPath)) ...<String>[
         '--entitlements',
         entitlementsFile.absolute.path,
       ],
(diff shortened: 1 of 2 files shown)


cipd_packages/codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -255,6 +255,33 @@ update these file paths accordingly.
       }
       log.info('Child file of directory ${directory.basename} is ${entity.basename}');
     }
+    final String directoryExtension = directory.basename.split('.').last;
+    if (directoryExtension == 'framework' || directoryExtension == 'xcframework') {
+      final List<String> args = <String>[
+        '/usr/bin/codesign',
+        '--keychain',
+        'build.keychain', // specify the keychain to look for cert
+        '-f', // force. Possible re-signing of FlutterMacOS.framework, Flutter.xcframework/ios-arm64/Flutter.framework etc.
```
jmagman, line 264: Does it re-sign them? Should it be using `--options=runtime` then (just needed for macOS)?  In which case maybe this the actual `codesign` args logic should be shared?
https://github.com/flutter/cocoon/blob/58930df6c1d291136bcb59a7a94532520eeb24f6/cipd_packages/codesign/lib/src/file_codesign_visitor.dart#L326C8-L326C25

XilaiZhang: Good point! looked up `--options=runtime` and it seems to protect runtime integrity. Will add this flag soon.

Sorry for the confusion, by potential re-signing I was referring to:
if the major executable of a bundle is already signed, then when we later code sign the bundle itself the signature of bundle might already exist.
e.g. if we code sign in an inside-out fashion, the main executable `FlutterMacOS.framework/Versions/A/FlutterMacOS` will be codesigned before we codesign `FlutterMacOS.framework`. code signing the latter might fail if a signature from the former already exists, but if we use the `--force` flag it will replace the existing signature if it exists.
I will try to make this comment clearer or delete it.

and sure I will look into refactoring the code

XilaiZhang: refactored and updated

cipd_packages/codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -314,17 +318,24 @@ update these file paths accordingly.
...
-      '-f', // force
+      '-f', // force. Needed to overwrite signature if major executable of bundle is already signed before bundle is signed.
       '-s', // use the cert provided by next argument
       codesignCertName,
-      binaryFile.absolute.path,
+      binaryOrBundlePath,
       '--timestamp', // add a secure timestamp
       '--options=runtime', // hardened runtime
-      if (fileWithEntitlements.contains(entitlementCurrentPath)) ...<String>[
+      if (entitlementCurrentPath != '' && fileWithEntitlements.contains(entitlementCurrentPath)) ...<String>[
```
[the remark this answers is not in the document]
XilaiZhang: yeah `entitlementCurrentPath` is null if we are code signing a bundle (L260), and is not null if we are signing a binary (L321)

1. if we are signing a binary, the args are the same as before, and `entitlementCurrentPath` is not null.
2. if we are signing a bundle, `entitlementCurrentPath` is empty, so we don't add the '--entitlement' flag.

jmagman (approved):
LGTM!

Can you let me know when this codesigns a xcframework and I can take a look at it and see if it looks right and Xcode is happy.

XilaiZhang:
Sure, I will label the cipd package after this PR is merged, and it should sign the framework in the next release (either stable or beta)

jmagman:
> Sure, I will label the cipd package after this PR is merged, and it should sign the framework in the next release (either stable or beta)

Is there a way I can check it before it's packaged for beta/stable?

XilaiZhang:
Sure:

1. download the cipd package from https://chrome-infra-packages.appspot.com/p/flutter/codesign (x86 or arm64)
2. find [codesign team ID](https://valentine.corp.google.com/#/show/1666993908871912?tab=metadata), [apple store ID](https://valentine.corp.google.com/#/show/1460636848884456?tab=metadata), and [app specific password](https://valentine.corp.google.com/#/show/1705967636315891?tab=metadata) from valentine
3. run cipd binary extracted from step one using [arguments listed on cocoon README](https://github.com/flutter/cocoon/tree/main/cipd_packages/codesign): something like
```
./codesign --dryrun
--codesign-cert-name="FLUTTER.IO LLC"
--codesign-team-id-file-path=/a/b/c.txt
--codesign-appstore-id-file-path=/a/b/b.txt
--app-specific-password-file-path=/a/b/a.txt
--input-zip-file-path=/a/input.zip
--output-zip-file-path=/b/output.zip
```
where each password file path is a local file that contains only the password extracted from step 2.
input zip file path is any zip file with unsigned framework. for example, we can input something from the [gcs bucket](https://pantheon.corp.google.com/storage/browser/flutter_infra_release/flutter?authuser=0&project=flutter-infra&pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&prefix=&forceOnObjectsSortingFiltering=false) 
and output file path can be anything, which will contain the signed zip.

Meanwhile, would be great if i could learn how to test framework using Xcode. Last time i was trying to build an app and import framework following [the verify origin guide](https://developer.apple.com/documentation/xcode/verifying-the-origin-of-your-xcframeworks) and [the medium post](https://medium.com/@fdchiu/how-to-add-3rd-party-framework-or-library-to-your-xcode-project-466ffa75b2e5), but my app didn't seem to build

cipd_packages/codesign/lib/src/file_codesign_visitor.dart
```diff
@@ -255,6 +255,10 @@ update these file paths accordingly.
       }
       log.info('Child file of directory ${directory.basename} is ${entity.basename}');
     }
+    final String directoryExtension = directory.basename.split('.').last;
+    if (directoryExtension == 'framework' || directoryExtension == 'xcframework') {
```
godofredoc, line 259: Can this be implemented in a generic way? e.g. only check if the path is a directory rather than using the extension and hardcoding to framework or xcfrmamework

XilaiZhang: Umm my understanding is that a generic directory might not qualify as a bundle.

Sorry for the verbosity, to summarize (3) in PR description, my understanding is that both .xcframework and .framework direcotry/folders qualify as bundles (based on info.plist and folder structure etc.), but a regular directory might not qualify (without info.plist and the required structure)

jmagman: He's right `.framework`s and `.xcframework`s are special with specific structures. Probably easier to check this faux "extension" than the bundle structure.

PR #3457, opened 2024-02-01
[scheduler] Parallelize ci.yaml scheduling requests

CaseyHillers:
The framework presubmit check takes about ~1 minute to schedule everything. The culprit appears to be running everything synchronously.

* Use `Future.wait` to process the various requests in parallel
* Small refactor to move the core logic into a re-usable function (_scheduleTryBuild)

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/app_dart/lib/src/service/luci_build_service.dart
+++ b/app_dart/lib/src/service/luci_build_service.dart
@@ -168,7 +168,6 @@ class LuciBuildService {
       return targets;
     }
 
-    final List<Request> requests = <Request>[];
     final List<String> branches = await gerritService.branches(
       'flutter-review.googlesource.com',
       'recipes',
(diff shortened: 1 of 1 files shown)


app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -180,53 +179,18 @@ class LuciBuildService {
...
+      futures.add(
+        _scheduleTryBuild(
+          target: target,
+          sha: sha,
+          pullRequest: pullRequest,
+          cipdVersion: cipdVersion,
         ),
       );
     }
+    final List<Request> requests = await Future.wait(futures);
```
keyonghan, line 193: I doubt this will help since there is no async operation in `_scheduleTryBuild` at all

CaseyHillers: githubChecksUtil.createCheckRun is async, and seems to take about 1 second

keyonghan: Yeah, you are right.

One concern is the `second rate limit`. With this change, I expect the API will be called more frequently within a short time and will increase the chance to hit the second rate limit.

IIUC, this PR aims to reduce the show up latency of these check runs, but doesn't seem to help with the LUCI builds scheduling/execution time.

CaseyHillers: For the secondary rate limit, my understanding is that is only an issue on authentication. This is for repeat calls to an existing client. Or is there some other logic happening under the hood now?

For the LUCI scheduling, that is published to the scheduler-reqeusts topic which should take ~10ms. We could parallelize it if there's interest, but I estimate it's <1% of the framework time.

CaseyHillers: Thanks for the link, I'll close this. However, Cocoon isn't implementing this best practice today, so I filed https://github.com/flutter/flutter/issues/142816 for tracking.

PR #3472, opened 2024-02-07
Start writing new commits/tasks to Firestore

keyonghan:
This PR
1) starts writing new commit/tasks to the new Firestore database
2) is a no-op to existing workflow. Will catch any exception without breaking existing logic.
3) puts generic logics (like documents to writes, comits/tasks to documents, etc) in the firestore service
4) adds extensive tests for each logic added. 


yusuf-goog (approved):

app_dart/lib/src/service/firestore.dart
```diff
@@ -27,4 +32,67 @@ class FirestoreService {
     );
     return FirestoreApi(client).projects.databases.documents;
   }
+
+  /// Writes [writes] to Firestore within a transaction.
+  ///
+  /// This is an atomic operation: either all writes succeed or all writes fail.
+  Future<CommitResponse> writeViaTransaction(List<Write> writes) async {
+    final ProjectsDatabasesDocumentsResource databasesDocumentsResource = await documentResource();
+    final BeginTransactionRequest beginTransactionRequest =
```
ricardoamador, line 41: Some databases have multiple commit modes. Is there anything in Firestore where you don't have to manage the commits manually like with with transactions?

app_dart/lib/src/service/firestore.dart
```diff
@@ -27,4 +32,67 @@ class FirestoreService {
...
+        BeginTransactionRequest(options: TransactionOptions(readWrite: ReadWrite()));
+    final BeginTransactionResponse beginTransactionResponse =
+        await databasesDocumentsResource.beginTransaction(beginTransactionRequest, kDatabase);
+    final CommitRequest commitRequest =
```
ricardoamador, line 45: If commit fails do we need a rollback?

app_dart/lib/src/service/firestore.dart
```diff
@@ -27,4 +32,67 @@ class FirestoreService {
...
+        CommitRequest(transaction: beginTransactionResponse.transaction, writes: writes);
+    return databasesDocumentsResource.commit(commitRequest, kDatabase);
+  }
+}
+
+/// Generates task documents based on targets.
+List<Document> targetsToTaskDocuments(Commit commit, List<Target> targets) {
+  final Iterable<Document> iterableDocuments = targets.map(
+    (Target target) => Document(
+      name: '$kDatabase/documents/tasks/${commit.sha}_${target.value.name}_1',
```
ricardoamador, line 55: What is the significance of the '_1'?

app_dart/lib/src/service/firestore.dart
```diff
@@ -27,4 +32,67 @@ class FirestoreService {
...
+      fields: <String, Value>{
+        'builderNumber': Value(integerValue: null),
```
ricardoamador, line 57: Are we sure this can be nullable? Is this the main identifier?

ricardoamador: Ah this might be updated at a later date.

app_dart/lib/src/service/firestore.dart
keyonghan: For a transaction operation, a commit is explicitly needed. There are other operations like `batchWrites`, which doesn't need a commit.

app_dart/lib/src/service/firestore.dart
keyonghan: Transaction acts as an atomic operation. If it fails, everything will auto roll back.

app_dart/lib/src/service/firestore.dart
keyonghan: I am using `_1` to track the old `attempt` logic. We want to keep all the records of retries, instead of overriding using a single record.

app_dart/lib/src/service/firestore.dart
keyonghan: Exactly. We will update the entry later when status changes.

ricardoamador (approved):
LGTM!

PR #2585, opened 2023-04-03
Do not enforce flutter-gold checks on auto roller authored pull requests

ricardoamador:
When an autoroller account generates a pull request, we do not need for it to wait for the flutter-gold status check.

*List which issues are fixed by this PR. You must list at least one issue.*
Fixes https://github.com/flutter/flutter/issues/123979

*If you had to change anything in the [flutter/tests] repo, include a link to the migration guide as per the [breaking change policy].*

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read the [Flutter Style Guide] _recently_, and have followed its advice.
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[test-exempt]: https://github.com/flutter/flutter/wiki/Tree-hygiene#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat

PR #3487, opened 2024-02-13
Treat `flutter-gold` failures in `flutter/engine` as failures.

matanlurey:
Closes https://github.com/flutter/flutter/issues/143352.

In practice this reverts the functionality in https://github.com/flutter/flutter/issues/123979 (https://github.com/flutter/cocoon/pull/2585). Sorry for the back-and-forth, we tried this model but ended up with (currently) [~170 untriaged failures](https://flutter-engine-gold.skia.org/), so we want to go back to treating these as failures and improve our tools/process if needed instead.

/cc @jonahwilliams and @Piinks

### DIFF
--- a/auto_submit/lib/validations/ci_successful.dart
+++ b/auto_submit/lib/validations/ci_successful.dart
@@ -153,7 +153,7 @@ class CiSuccessful extends Validation {
         if (status.state == STATUS_FAILURE && !notInAuthorsControl.contains(name)) {
           failures.add(FailureDetail(name!, status.targetUrl!));
         }
-        if (status.state == STATUS_PENDING && isStale(status.createdAt!) && supportStale(author, slug)) {
+        if (status.state == STATUS_PENDING && status.createdAt != null && isStale(status.createdAt!) && supportStale(author, slug)) {
           staleStatuses.add(status);
         }
       }
--- a/auto_submit/lib/validations/ci_successful.dart
+++ b/auto_submit/lib/validations/ci_successful.dart
@@ -79,7 +79,7 @@ class CiSuccessful extends Validation {
     }
 
     /// Validate if all checkRuns have succeeded.
-    allSuccess = validateCheckRuns(slug, prNumber, checkRuns, failures, allSuccess);
+    allSuccess = validateCheckRuns(slug, prNumber, checkRuns, failures, allSuccess, author);
 
     if (!allSuccess && failures.isEmpty) {
       return ValidationResult(allSuccess, Action.IGNORE_TEMPORARILY, '');
@@ -140,6 +140,7 @@ class CiSuccessful extends Validation {
     final String overrideTreeStatusLabel = config.overrideTreeStatusLabel;
     log.info('Validating name: ${slug.name}/$prNumber, statuses: $statuses');
 
+    final List<ContextNode> staleStatuses = <ContextNode>[];
     for (ContextNode status in statuses) {
       // How can name be null but presumed to not be null below when added to failure?
       final String? name = status.context;
@@ -152,7 +153,15 @@ class CiSuccessful extends Validation {
         if (status.state == STATUS_FAILURE && !notInAuthorsControl.contains(name)) {
           failures.add(FailureDetail(name!, status.targetUrl!));
         }
+        if (status.state == STATUS_PENDING && isStale(status.createdAt!) && supportStale(author, slug)) {
+          staleStatuses.add(status);
+        }
       }
+    }
+    if (staleStatuses.isNotEmpty) {
+      log.warning(
+        'Pull request https://github.com/${slug.fullName}/pull/$prNumber from ${slug.name} repo auto roller has been running over ${Config.kGitHubCheckStaleThreshold} hours due to: ${staleStatuses.map((e) => e.context).toList()}',
+      );
     }
 
     return allSuccess;
@@ -168,9 +177,11 @@ class CiSuccessful extends Validation {
     List<github.CheckRun> checkRuns,
     Set<FailureDetail> failures,
     bool allSuccess,
+    Author author,
   ) {
     log.info('Validating name: ${slug.name}/$prNumber, checkRuns: $checkRuns');
 
+    final List<github.CheckRun> staleCheckRuns = <github.CheckRun>[];
     for (github.CheckRun checkRun in checkRuns) {
       final String? name = checkRun.name;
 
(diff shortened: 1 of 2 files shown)


Piinks (commented):
Definitely check with @mdebbar, I am less familiar here.

I think if the expectation that the engine sheriff will triage images is not working, then this is a good option. It is how we treat the same situation in the framework. 

Alternatively, I think Gold has an endpoint to see "are there unapproved images at head?" If we wanted to try making ownership by the sheriff work, we could have a separate check or alert that notifies them. IIRC that endpoint had rate limitations though. @kjlubick may be able to expand more on this. :)

kjlubick:
https://flutter-engine-gold.skia.org/search is the best way to see all unapproved images at head. It should be reasonably performant (as well as the underlying RPC call, but please try to avoid calling those directly, since we may break you).

I think the rate limitation thing @Piinks was thinking of was related to the changes for a particular PR - we use some extra caching for HEAD that isn't there for PRs.

mdebbar:
I'm not familiar with the recent developments in the engine golden story. But one thing that came to mind a couple days ago as I was looking into some golden problems was this:
https://github.com/flutter/cocoon/blob/f163e347d2a966ed6a524d9ca6c4b54f3cf80db0/app_dart/lib/src/request_handlers/push_gold_status_to_github.dart#L122-L127

I think that `if` condition is stale and we might be skipping the golden checks on some PRs that DO introduce golden changes. Thoughts on removing the condition and doing golden checks on ALL PRs? (cc @Piinks and @kjlubick in case this breaks some Gold API quota?)

kjlubick:
I would like a quick write up on which RPCs would be affected on the Gold side before making a call about quotas.

Piinks:
> I would like a quick write up on which RPCs would be affected on the Gold side before making a call about quotas.

@kjlubick  [This doc](https://docs.google.com/document/d/1H3CDqT7zBUt4Je2HPQpleYA-drwj2oy0mdPlSdf2d4A/edit?usp=sharing) lists all the RPCs we use, where and how. It is linked here in the wiki: https://github.com/flutter/flutter/wiki/Writing-a-golden-file-test-for-package:flutter and cane be used to write this up if we want to make changes.

> I think that if condition is stale and we might be skipping the golden checks on some PRs that DO introduce golden changes. Thoughts on removing the condition and doing golden checks on ALL PRs? (cc @Piinks and @kjlubick in case this breaks some Gold API quota?)

@mdebbar  That's super intentional (https://github.com/flutter/cocoon/pull/672 for context).

1. Contributors complained that they did not want to be bothered with a flutter-gold check on their PR if the change had nothing to do with goldens. And I mean, they felt really **really** strongly about this.
2. By not running the flutter-gold check on PRs that do not execute golden file tests, we were able to reduce the amount of calls to Gold significantly. Changing that back would greatly increase traffic.

If the engine has deviated from the expected pattern for naming test shards that execute golden file tests, then that's something we can easily fix.

matanlurey:
Just so I understand correctly @Piinks, [the configuration](https://github.com/flutter/cocoon/blob/f163e347d2a966ed6a524d9ca6c4b54f3cf80db0/app_dart/lib/src/request_handlers/push_gold_status_to_github.dart#L122-L127) pointed out by @mdebbar is `flutter/flutter` specific, right?

matanlurey:
Also while I appreciate the chat here, I'd love to still get this PR approved/landed unless I'm missing something?

auto_submit/lib/validations/ci_successful.dart
```diff
@@ -201,40 +184,10 @@ class CiSuccessful extends Validation {
...
-      log.warning(
-        'Pull request https://github.com/${slug.fullName}/pull/$prNumber from ${slug.name} repo auto roller has been running over ${Config.kGitHubCheckStaleThreshold} hours due to: ${staleCheckRuns.map((e) => e.name).toList()}',
-      );
-    }
 
     return allSuccess;
   }
-
-  // Treat any GitHub check run as stale if created over [Config.kGitHubCheckStaleThreshold] hours ago.
-  bool isStale(DateTime dateTime) {
```
ricardoamador, line 221: I am not sure about the removal of this here. @keyonghan what do you think?

ricardoamador: If I recall correctly we are trying to flag PRs that are over 2 hours old so I think this code needs to stay if you are only removing stuff related to flutter-gold.

auto_submit/lib/validations/ci_successful.dart
```diff
@@ -140,17 +140,10 @@ class CiSuccessful extends Validation {
     final String overrideTreeStatusLabel = config.overrideTreeStatusLabel;
     log.info('Validating name: ${slug.name}/$prNumber, statuses: $statuses');
 
-    final List<ContextNode> staleStatuses = <ContextNode>[];
     for (ContextNode status in statuses) {
       // How can name be null but presumed to not be null below when added to failure?
       final String? name = status.context;
 
-      // If the account author is a roller account do not block merge on flutter-gold check.
-      if (isToEngineRoller(author, slug) && name == 'flutter-gold') {
```
ricardoamador, line 149: I think this is the only code that needs to be removed here to avoid skipping and enforcing the failure.

ricardoamador:
Overall I think this is good but I would like to get @keyonghan to look over the stale checkRun flagging as I am not sure if he is tracking metrics there or we have alerting around that.

auto_submit/lib/validations/ci_successful.dart
matanlurey: Ah thinks Ricardo, this is my bad - it looks like I had an older branch. Let me redo this fix and tag you.

matanlurey:
> Overall I think this is good but I would like to get @keyonghan to look over the stale checkRun flagging as I am not sure if he is tracking metrics there or we have alerting around that.

Thanks @ricardoamador. I actually made a mistake and branched from an older `main` locally.

The change to `ci_successful.dart` should be ~1-line-ish, re-upped.

Piinks:
> Just so I understand correctly @Piinks, [the configuration](https://github.com/flutter/cocoon/blob/f163e347d2a966ed6a524d9ca6c4b54f3cf80db0/app_dart/lib/src/request_handlers/push_gold_status_to_github.dart#L122-L127) pointed out by @mdebbar is `flutter/flutter` specific, right?

No. That whole file is the flutter-gold implementation intended to cover flutter/flutter and flutter/engine.

auto_submit/lib/validations/ci_successful.dart
matanlurey: This was my bad, I had branched from a stale PR. Reverted.

ricardoamador (approved):
LGTM, thanks @matanlurey !

yjbanov:
A note about the 170 untriaged web goldens. They all come from Safari + HTML renderer. All of them exhibit bimodality. There are exactly 2 golden variants per test. One hypothesis from @eyebrowsoffire is that our macOS fleet is made of Arm and x64 machines, and our tests get one of them randomly. If Safari is sensitive to Arm vs x64, that would lead to this behavior. Essentially, you send a PR, triage goldens, then submit. However, what you just did is triage a random combination of Arm and x64 goldens. After the PR is submitted, the post-submit instance of LUCI will again randomize the bots that run web tests, and report different golden outcomes in post-submit.

matanlurey:
Thanks for explaining! I'm happy to hold this PR until we have consensus on how to handle those diffs.

jonahwilliams:
Goldens are allowed to have multiple images per configuration, so if it was truly bimodal (only), then you'd eventually triage all of them.

matanlurey:
I don't have consensus on how to land this yet, so holding off until I chat more with the teams that will be impacted (I'll loop in infra as well).

mdebbar (approved):

PR #3488, opened 2024-02-13
Start updating post-submit tasks in Firestore

keyonghan:
This PR
1) creates a Firestore task model
2) adds getDocument and batchWriteDocuments to firestore service
3) starts updating documents in Firestore
4) no op to existing production workflow.

### DIFF
--- a/app_dart/lib/src/model/firestore/task.dart
+++ b/app_dart/lib/src/model/firestore/task.dart
@@ -1,0 +1,214 @@
+// Copyright 2019 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:googleapis/firestore/v1.dart' hide Status;
+
+import '../../request_handling/exceptions.dart';
+import '../../service/firestore.dart';
+import '../../service/logging.dart';
+import '../luci/push_message.dart';
+
+class Task extends Document {
+  /// Lookup [Task] from Firestore.
+  ///
+  /// `documentName` follows `/projects/{project}/databases/{<}database}/documents/{document_path}`
+  static Future<Task> fromFirestore({
+    required FirestoreService firestoreService,
+    required String documentName,
+  }) async {
+    final Document document = await firestoreService.getDocument(documentName);
+    return Task.fromDocument(taskDocument: document);
+  }
+
+  /// Create [Task] from a task Document.
+  static Task fromDocument({
+    required Document taskDocument,
+  }) {
+    return Task()
+      ..fields = taskDocument.fields!
+      ..name = taskDocument.name!;
+  }
+
+  /// The task was cancelled.
+  static const String statusCancelled = 'Cancelled';
+
+  /// The task is yet to be run.
+  static const String statusNew = 'New';
+
+  /// The task failed to run due to an unexpected issue.
+  static const String statusInfraFailure = 'Infra Failure';
+
+  /// The task is currently running.
+  static const String statusInProgress = 'In Progress';
+
+  /// The task was run successfully.
+  static const String statusSucceeded = 'Succeeded';
+
+  /// The task failed to run successfully.
+  static const String statusFailed = 'Failed';
+
+  /// The task was skipped or canceled while running.
+  ///
+  /// This status is only used by LUCI tasks.
+  static const String statusSkipped = 'Skipped';
+
+  /// The list of legal values for the [status] property.
+  static const List<String> legalStatusValues = <String>[
+    statusCancelled,
+    statusFailed,
+    statusInfraFailure,
+    statusInProgress,
+    statusNew,
+    statusSkipped,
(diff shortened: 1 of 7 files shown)


app_dart/lib/src/model/firestore/task.dart
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+  /// The task was cancelled.
+  static const String statusCancelled = 'Cancelled';
+
+  /// The task is yet to be run.
+  static const String statusNew = 'New';
+
+  /// The task failed to run due to an unexpected issue.
+  static const String statusInfraFailure = 'Infra Failure';
```
ricardoamador, line 40: I feel like I see these enums in a lot of places...

app_dart/lib/src/model/firestore/task.dart
```diff
@@ -0,0 +1,214 @@
...
+  /// The task failed to run successfully.
+  static const String statusFailed = 'Failed';
+
+  /// The task was skipped or canceled while running.
+  ///
+  /// This status is only used by LUCI tasks.
+  static const String statusSkipped = 'Skipped';
+
+  /// The list of legal values for the [status] property.
+  static const List<String> legalStatusValues = <String>[
```
ricardoamador, line 57: Would it make sense to define an enum with the string names and the legal status? Something like:

```
enum TaskStatus {

  statusFailed ('Failed', true, true),
  statusInProgress('In progress', true, false),
  ...
  statusInfraFailure('Infra Failure', true, true);

  const({
    required this.name,
    required this.isLegalStatus,
   required this.isTerminalStatus,
  });

  final String name;
  final bool isLegalStatus;
  final bool isTerminalStatus;
}
```

app_dart/lib/src/model/firestore/task.dart
```diff
@@ -0,0 +1,214 @@
...
+      throw ArgumentError('Invalid state: "$taskStatus"');
+    }
+    return taskStatus;
+  }
+
+  String setStatus(String value) {
+    if (!legalStatusValues.contains(value)) {
+      throw ArgumentError('Invalid state: "$value"');
+    }
+    fields!['status'] = Value(stringValue: value);
```
ricardoamador, line 140: This fields! everywhere, is this coming from the parent Document class?

app_dart/lib/src/model/firestore/task.dart
```diff
@@ -0,0 +1,214 @@
...
+      return setStatus(statusInProgress);
+    }
+    switch (build.result) {
+      case Result.success:
+        return setStatus(statusSucceeded);
+      case Result.canceled:
+        return setStatus(statusCancelled);
+      case Result.failure:
+        // Note that `Result` does not support `infraFailure`:
+        // https://github.com/luci/luci-go/blob/main/common/api/buildbucket/buildbucket/v1/buildbucket-gen.go#L247-L251
```
ricardoamador, line 178: With this require an update when the buildbucket v2 implementation is merged in?

ricardoamador (approved):
LGTM with some nonblocking questions.

app_dart/lib/src/model/firestore/task.dart
keyonghan: Yes. The `Task` is extending `Document`, where the `fields` is defined.

app_dart/lib/src/model/firestore/task.dart
keyonghan: This sounds a good idea. I expect it will need a bit refactoring of multiple places, and prefer to skip it from this PR. Created https://github.com/flutter/flutter/issues/143465 to track.

app_dart/lib/src/model/firestore/task.dart
keyonghan: We have different statuses in different contexts: the flutter build dashboard task statuses, the buildbucket build statuses, GitHub statuses, etc. 
But here is a direct map from datastore/task statuses, which would be deprecated when firestore logic is ready.

app_dart/lib/src/model/firestore/task.dart
ricardoamador: Gotcha! Thanks for clarifying.

app_dart/lib/src/model/firestore/task.dart
ricardoamador: Yeah no problem.

app_dart/lib/src/model/firestore/task.dart
ricardoamador: Got it!

app_dart/lib/src/model/firestore/task.dart
keyonghan: That's a good point. Could we help check how the status look like for v2? Maybe the LUCI team has a better idea how the build message looks like when switched to v2. If anything changes, we need to handle them accordingly here.

PR #3489, opened 2024-02-14
Add logs to help triage Firestore task update failure

keyonghan:
Help with https://github.com/flutter/flutter/issues/143497

### DIFF
--- a/app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
+++ b/app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
@@ -88,7 +88,7 @@ class PostsubmitLuciSubscription extends SubscriptionHandler {
       task.updateFromBuild(build);
       await datastore.insert(<Task>[task]);
       try {
-        await updateFirestore(build, commitKey.id!, task.name!);
+        await updateFirestore(build, rawCommitKey, task.name!);
       } catch (error) {
         log.warning('Failed to update task in Firestore: $error');
       }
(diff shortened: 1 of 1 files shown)


app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
```diff
@@ -144,12 +144,16 @@ class PostsubmitLuciSubscription extends SubscriptionHandler {
   }
 
   /// Queries the task document and updates based on the latest build data.
-  Future<void> updateFirestore(Build build, String sha, String taskName) async {
+  Future<void> updateFirestore(Build build, String commitKeyId, String taskName) async {
```
yusuf-goog, line 147: Should we name the arg consistently (rawCommitKey)?

yusuf-goog (approved):
LGTM with one question

app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
keyonghan: From the user_data, it uses `rawCommitKey`, but this is actually the Key.id of a Commit.

app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
```diff
@@ -144,12 +144,16 @@ class PostsubmitLuciSubscription extends SubscriptionHandler {
...
     final FirestoreService firestoreService = await config.createFirestoreService();
+    final String sha = commitKeyId.split('/').last;
     final String documentName = '$kDatabase/documents/tasks/${sha}_${taskName}_1';
+    log.info('getting firestore document: $documentName');
     final f.Task firestoreTask =
         await f.Task.fromFirestore(firestoreService: firestoreService, documentName: documentName);
+    log.info('updating firestoreTask based on build');
     firestoreTask.updateFromBuild(build);
+    log.info('finished updating firestoreTask based on builds');
```
ricardoamador, line 156: For these two log messages can we add some kind of identifier? Something like the commit for easier debugging?

keyonghan: The `commit` info is included in line 151. These two logs are just used as a way to tell which step in between fails if any. In other words, if we see line 154 or line 156, line 151 should already show up right above.

PR #3524, opened 2024-02-27
Fix null sha/task values when resetting prod task

keyonghan:
Fixes: https://github.com/flutter/flutter/issues/144236


ricardoamador (approved):
LGTM

PR #3544, opened 2024-03-05
Refactor to move logics closer to data models

keyonghan:
This PR doesn't add new functionalities, but refactor logics to closer to data models.

1) moves task `variables` and `functions` from `firestore.dart` to `firestore/task.dart`.
2) moves commit `variables` and `functions` from `firestore.dart` to `firestore/commit.dart`.
3) moves github gold status `variables` and `functions` from `firestore.dart` to `firestore/github_gold_status.dart`.
4) updates related reference points from other places.

### DIFF
--- a/app_dart/lib/src/model/firestore/task.dart
+++ b/app_dart/lib/src/model/firestore/task.dart
@@ -285,7 +285,7 @@ List<Task> targetsToTaskDocuments(Commit commit, List<Target> targets) {
 }
 
 /// Generates task document based on datastore task data model.
-Task taskToTaskDocument(datastore.Task task) {
+Task taskToDocument(datastore.Task task) {
   final String commitSha = task.commitKey!.id!.split('/').last;
   return Task.fromDocument(
     taskDocument: Document(
--- a/app_dart/lib/src/model/firestore/task.dart
+++ b/app_dart/lib/src/model/firestore/task.dart
@@ -8,7 +8,23 @@ import 'package:googleapis/firestore/v1.dart' hide Status;
 import '../../request_handling/exceptions.dart';
 import '../../service/firestore.dart';
 import '../../service/logging.dart';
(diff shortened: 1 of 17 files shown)


app_dart/lib/src/model/firestore/task.dart
```diff
@@ -245,3 +261,45 @@ class Task extends Document {
     return buf.toString();
   }
 }
+
+/// Generates task documents based on targets.
+List<Task> targetsToTaskDocuments(Commit commit, List<Target> targets) {
```
ricardoamador, line 266: nit: I think the name of the function is a little confusing and might be clearer as 'targetsToDocuments'.

ricardoamador: Are we converting targets or tasks?

app_dart/lib/src/model/firestore/task.dart
```diff
@@ -245,3 +261,45 @@ class Task extends Document {
...
+          kTaskCommitShaField: Value(stringValue: commit.sha),
+        },
+      ),
+    ),
+  );
+  return iterableDocuments.toList();
+}
+
+/// Generates task document based on datastore task data model.
+Task taskToTaskDocument(datastore.Task task) {
```
ricardoamador, line 288: nit: same here with the naming. maybe 'taskToDocument'?

ricardoamador (approved):
LGTM

app_dart/lib/src/model/firestore/task.dart
keyonghan: This is to convert targets to tasks

app_dart/lib/src/model/firestore/task.dart
keyonghan: Updated.

PR #3550, opened 2024-03-06
Use repository and prNumber as goldstatusstatus document name

keyonghan:
This is to make it consistent with old datastore logic: for each pr of a repo, we keep a single record.

Namely, this PR replaces the following entries into a single one.
Screenshot 2024-03-06 at 10 30 36 AM


app_dart/lib/src/model/firestore/github_gold_status.dart
```diff
@@ -83,9 +83,10 @@ class GithubGoldStatus extends Document {
 
 /// Generates GithubGoldStatus document based on datastore GithubGoldStatusUpdate data model.
 GithubGoldStatus githubGoldStatusToDocument(GithubGoldStatusUpdate githubGoldStatus) {
+  final String repo = githubGoldStatus.repository!.replaceAll('/', '_');
```
ricardoamador, line 86: nit: add a comment about why this replaceAll is being done.

ricardoamador (approved):
LGTM

PR #3553, opened 2024-03-06
Add a wrapper for Firestore query to simplify logics

keyonghan:

This PR:
1) adds a query wrapper to simplify logics
2)switch existing commitTask query to use the new wrapper

### DIFF
--- a/app_dart/lib/src/service/firestore.dart
+++ b/app_dart/lib/src/service/firestore.dart
@@ -8,13 +8,18 @@ import 'package:cocoon_service/cocoon_service.dart';
 import 'package:googleapis/firestore/v1.dart';
 import 'package:http/http.dart';
 
-import '../model/firestore/task.dart' as firestore;
+import '../model/firestore/task.dart';
 import 'access_client_provider.dart';
 import 'config.dart';
 
 const String kDatabase = 'projects/${Config.flutterGcpProjectId}/databases/${Config.flutterGcpFirestoreDatabase}';
 const String kDocumentParent = '$kDatabase/documents';
 const String kFieldFilterOpEqual = 'EQUAL';
+const String kCompositeFilterOpAnd = 'AND';
+
+const Map<String, String> kRelationMapping = <String, String>{
+  '=': 'EQUAL',
+};
 
 class FirestoreService {
   const FirestoreService(this.accessClientProvider);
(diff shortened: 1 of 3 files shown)


app_dart/lib/src/service/firestore.dart
```diff
@@ -67,24 +72,88 @@ class FirestoreService {
...
+          fieldFilter: FieldFilter(
+            field: FieldReference(fieldPath: name),
+            op: comparison,
+            value: value,
+          ),
+        ),
+      );
+    });
+    Filter? finalFilter;
+    if (filters.length == 1) {
```
ricardoamador, line 117: Can you use a composite filter with a single entry?

app_dart/lib/src/service/firestore.dart
```diff
@@ -67,24 +72,88 @@ class FirestoreService {
...
+    }
+    return Value(stringValue: comparisonOject as String);
+  }
+
+  /// Generates Firestore query filter based on "human" read conditions.
+  Filter generateFilter(Map<String, Object> filterMap, String compositeFilterOp) {
+    final List<Filter> filters = <Filter>[];
+    filterMap.forEach((filterString, comparisonOject) {
+      final List<String> parts = filterString.split(' ');
+      if (parts.length != 2 || !kRelationMapping.containsKey(parts[1])) {
```
ricardoamador, line 99: nit: prefer constants for these integers.

app_dart/lib/src/service/firestore.dart
```diff
@@ -67,24 +72,88 @@ class FirestoreService {
...
+      );
+    }
+    return finalFilter;
+  }
+
+  /// Wrapper to simplify Firestore query.
+  ///
+  /// The [filterMap] follows format:
+  ///   {
+  ///     'fieldInt =': 1,
```
ricardoamador, line 134: Is this convention from firestore?

keyonghan: No. It's a convention from datastore. This is actually a user side behavior. Using the same as before to make it same for user experience.
https://github.com/flutter/cocoon/blob/main/app_dart/lib/src/service/datastore.dart#L215
https://github.com/dart-lang/gcloud/blob/master/lib/src/db/db.dart#L185

app_dart/lib/src/service/firestore.dart
keyonghan: Good point! Just validated a composite with a single field filter works. Updated.

app_dart/lib/src/service/firestore.dart
keyonghan: Done.

app_dart/lib/src/service/firestore.dart
```diff
@@ -67,24 +72,88 @@ class FirestoreService {
...
+      '$kTaskCommitShaField = ': commitSha,
+    };
+    final List<Document> documents = await query(kTaskCollectionId, filterMap);
+    return documents.map((Document document) => Task.fromDocument(taskDocument: document)).toList();
+  }
+
+  /// Returns Firestore [Value] based on corresponding object type.
+  Value getValueFromFilter(Object comparisonOject) {
+    if (comparisonOject is int) {
+      return Value(integerValue: comparisonOject.toString());
```
ricardoamador, line 90: Is this meant to be toString()? Seems odd that you would need to supply the integer value as a string but boolean as it normally appears.

app_dart/lib/src/service/firestore.dart
ricardoamador: Thanks!

ricardoamador (approved):
LGTM

app_dart/lib/src/service/firestore.dart
ricardoamador: Ewww. thanks for clarifying.

PR #3558, opened 2024-03-08
Fix null check error when querying Firestore

keyonghan:
Fixes: https://github.com/flutter/flutter/issues/144856


Piinks (commented):
Can you add a test for this scenario?

keyonghan:
> Can you add a test for this scenario?

Added for the gold status API code. 

For the Firestore piece of codes, I am not sure how to test it via a unit test as it connects with the gcp server. But I have validated the logic works via a local service.

keyonghan:
Okay. Separated the logic as a function, and added corresponding unit tests to cover.

keyonghan:
A friendly ping:-)

Piinks (approved):
Flutter_LGTM

PR #3571, opened 2024-03-13
Add a caution callout to the GitHub Flutter Gold message.

matanlurey:
Mitigation towards https://github.com/flutter/flutter/issues/145043.

### DIFF
--- a/app_dart/lib/src/service/config.dart
+++ b/app_dart/lib/src/service/config.dart
@@ -269,7 +269,13 @@ class Config {
       'resolve the images on the Flutter Gold dashboard, consider marking this PR '
       'as a draft pull request above. You will still be able to view image results '
       'on the dashboard, commenting will be silenced, and the check will not try to resolve itself until '
-      'marked ready for review.\n\n';
+      'marked ready for review.\n\n'
+      '> [!CAUTION]\n'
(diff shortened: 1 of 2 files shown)


app_dart/lib/src/service/config.dart
```diff
@@ -269,7 +269,13 @@ class Config {
       'resolve the images on the Flutter Gold dashboard, consider marking this PR '
       'as a draft pull request above. You will still be able to view image results '
       'on the dashboard, commenting will be silenced, and the check will not try to resolve itself until '
-      'marked ready for review.\n\n';
+      'marked ready for review.\n\n'
```
Piinks, line 272: This will only notify PRs that have been detected as generating untriaged images, it would not have caught today's scenario. We need a new action in the push_gold_status* bot that will notify on any PR that is generating golden file images.

matanlurey: Gotcha, PTAL - I think this should be applied now anytime that a comment otherwise would be left.

matanlurey:
Thinking about this with fresh eyes, it seems much easier to just have try-bots fail on negatives than try to teach folks via bot messages not to use negatives. At least on `flutter/engine` we're going to use that approach, so I'll close this PR for now.

If the framework team would like a similar change, I'd be happy to send/review a PR there as well.

PR #3574, opened 2024-03-13
Support Firestore in build status service

keyonghan:
This is to prepare query logics switch to Firestore. 

This PR is a no op for existing workflow, but just inject the firestore service to prepare for future logic updates.

### DIFF
--- a/app_dart/lib/src/service/build_status_provider.dart
+++ b/app_dart/lib/src/service/build_status_provider.dart
@@ -4,6 +4,7 @@
 
 import 'dart:async';
 
+import 'package:cocoon_service/cocoon_service.dart';
 import 'package:github/github.dart';
 import 'package:meta/meta.dart';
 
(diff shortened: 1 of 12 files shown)


app_dart/lib/src/service/build_status_provider.dart
```diff
@@ -14,20 +15,27 @@ import '../model/appengine/task.dart';
 import 'datastore.dart';
 
 /// Function signature for a [BuildStatusService] provider.
-typedef BuildStatusServiceProvider = BuildStatusService Function(DatastoreService datastoreService);
+typedef BuildStatusServiceProvider = BuildStatusService Function(
+  DatastoreService datastoreService,
+  FirestoreService firestoreService,
```
keyonghan, line 20: This is the main change. All the other files are updated to be consistent here.

yusuf-goog (approved):

PR #3583, opened 2024-03-19
Support Firestore user data for postsubmit build

keyonghan:

To migrate postsubmit build API fully to firestore, we need to support user data servicing firestore data models. This PR adds corresponding document names.


ricardoamador (approved):
LGTM

PR #3584, opened 2024-03-19
Switch to use firestore task for postsubmit_luci_subscription API

keyonghan:

Following https://github.com/flutter/cocoon/pull/3583, this PR:
1) switches logics to use firestore task
2) keeps querying/writing task status for both Datastore and Firestore, in case fallback happens
3) updates all unit tests
4) removes the condition when `taskKey` doesn't exist (seems all builds are enforced with the `userData`, and confirmed no call for the past week)

### DIFF
--- a/app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
+++ b/app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
@@ -57,8 +57,9 @@ class PostsubmitLuciSubscription extends SubscriptionHandler {
 
     final String? rawTaskKey = buildPushMessage.userData['task_key'] as String?;
     final String? rawCommitKey = buildPushMessage.userData['commit_key'] as String?;
-    if (rawCommitKey == null) {
-      throw const BadRequestException('userData does not contain commit_key');
+    final String? taskDocumentName = buildPushMessage.userData['firestore_task_document_name'] as String?;
+    if (taskDocumentName == null) {
+      throw const BadRequestException('userData does not contain firestore_task_document_name');
     }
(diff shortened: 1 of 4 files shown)


app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
```diff
@@ -67,57 +68,48 @@ class PostsubmitLuciSubscription extends SubscriptionHandler {
...
     Task? task;
-    final String? taskName = build.buildParameters?['builder_name'] as String?;
-    if (rawTaskKey == null || rawTaskKey.isEmpty || rawTaskKey == 'null') {
-      log.fine('Pulling builder name from parameters_json...');
-      log.fine(build.buildParameters);
-      if (taskName == null || taskName.isEmpty) {
-        throw const BadRequestException('task_key is null and parameters_json does not contain the builder name');
-      }
-      final List<Task> tasks = await datastore.queryRecentTasksByName(name: taskName).toList();
-      task = tasks.singleWhere((Task task) => task.parentKey?.id == commitKey.id);
```
keyonghan, line 78: As said in the PR description: this logic has been obsolete.

yusuf-goog (approved):

app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
ricardoamador: Nice! I like this but will be extra work for me with the buildbucket merge.

ricardoamador (approved):
LGTM

PR #3589, opened 2024-03-22
Use Firestore data when reseting prod task and rerunning a build

keyonghan:

This PR:
1) supports `taskDocumentName` arg for reset-prod-task API. This is to prepare frontend logic switched to Firestore
2) uses Firestore data when checking rerun, instead of Datastore data
3) moves `taskFailStatusSet` to Task models
4) moves Firestore logics out of try/catch block to break API if failing

### DIFF
--- a/app_dart/lib/src/service/luci_build_service.dart
+++ b/app_dart/lib/src/service/luci_build_service.dart
@@ -16,6 +16,7 @@ import 'package:googleapis/pubsub/v1.dart';
 import '../foundation/github_checks_util.dart';
 import '../model/appengine/commit.dart';
 import '../model/appengine/task.dart';
+import '../model/firestore/commit.dart' as firestore_commit;
 import '../model/firestore/task.dart' as firestore;
 import '../model/ci_yaml/target.dart';
 import '../model/github/checks.dart' as cocoon_checks;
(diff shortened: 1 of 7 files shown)


ricardoamador:
Quick question, this does not change functionality does it?

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -26,12 +27,6 @@ import '../service/logging.dart';
 import 'exceptions.dart';
 import 'github_service.dart';
 
-const Set<String> taskFailStatusSet = <String>{
```
ricardoamador, line 29: Ha! I did something similar in my branch!

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -727,19 +708,28 @@ class LuciBuildService {
...
+
     return true;
   }
 
   /// Check if a builder should be rerun.
   ///
   /// A rerun happens when a build fails, the retry number hasn't reached the limit, and the build is on TOT.
-  Future<bool> _shouldRerunBuilder(Task task, Commit commit, DatastoreService? datastore) async {
-    if (!taskFailStatusSet.contains(task.status)) {
+  Future<bool> _shouldRerunBuilderFirestore(firestore.Task task, FirestoreService firestoreService) async {
```
keyonghan, line 731: This is the main change: use firestore task to determine whether rerun.

keyonghan:
> Quick question, this does not change functionality does it?

Added a comment for the main change.

ricardoamador (approved):
Can we wait until Monday to merge this? I have some testing I would like to get in today before needing to refactor on top of this.

keyonghan:
> Can we wait until Monday to merge this? I have some testing I would like to get in today before needing to refactor on top of this.

You got it. Will add the label Monday.

ricardoamador:
> > Can we wait until Monday to merge this? I have some testing I would like to get in today before needing to refactor on top of this.
> 
> You got it. Will add the label Monday.

Awesome! Thank you!

keyonghan:
Landing now.

PR #3597, opened 2024-03-25
Save commit/tasks to firestore first before scheduling LUCI builds

keyonghan:
This is to resolve issues:
1) data mismatch between datastore and firestore when LUCI schedule fails
2) postsubmit-luci-scheduler failures if LUCI schedules a task before the data is saved to firestore


ricardoamador (approved):
LGTM

PR #3634, opened 2024-04-10
Add logics to fetch commit status from Firestore

keyonghan:
This is part of https://github.com/flutter/flutter/issues/142951

This PR adds logics to fetch commit tasks status from Firestore based on newly added Firestore data model protos. After this lands, next step is to switch the build logic to reference this PR's changes from Datastore.

This PR is a no-op change by this point.

### DIFF
--- a/dashboard/lib/service/appengine_cocoon.dart
+++ b/dashboard/lib/service/appengine_cocoon.dart
@@ -12,10 +12,13 @@ import 'package:http/http.dart' as http;
 
 import '../logic/qualified_task.dart';
 import '../model/build_status_response.pb.dart';
+import '../model/commit_firestore.pb.dart';
 import '../model/commit.pb.dart';
 import '../model/commit_status.pb.dart';
+import '../model/commit_tasks_status.pb.dart';
 import '../model/key.pb.dart';
 import '../model/task.pb.dart';
+import '../model/task_firestore.pb.dart';
 import 'cocoon.dart';
 
 /// CocoonService for interacting with flutter/flutter production build data.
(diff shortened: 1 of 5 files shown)


dashboard/lib/service/appengine_cocoon.dart
```diff
@@ -67,6 +70,38 @@ class AppEngineCocoonService implements CocoonService {
...
+        '/api/public/get-status-firestore returned ${response.statusCode}',
+      );
+    }
+
+    try {
+      final Map<String, dynamic> jsonResponse = jsonDecode(response.body);
+      return CocoonResponse<List<CommitTasksStatus>>.data(
+        _commitStatusesFromJsonFirestore(jsonResponse['Statuses']),
+      );
+    } catch (error) {
```
ricardoamador, line 100: Is there a specific error we can catch?

dashboard/lib/service/appengine_cocoon.dart
```diff
@@ -224,11 +259,34 @@ class AppEngineCocoonService implements CocoonService {
...
+    return statuses;
+  }
+
   String? _branchFromJson(Map<String, dynamic> jsonChecklist) {
     final Map<String, dynamic> checklist = jsonChecklist['Checklist'];
     return checklist['Branch'] as String?;
   }
 
+  String? _branchFromJsonFirestore(Map<String, dynamic> jsonCommit) {
+    return jsonCommit['Branch'] as String?;
```
ricardoamador, line 287: Consider making constants for these map keys like 'Branch", "Commit" etc.

dashboard/lib/service/appengine_cocoon.dart
```diff
@@ -249,6 +307,21 @@ class AppEngineCocoonService implements CocoonService {
     return result;
   }
 
+  CommitDocument _commitFromJsonFirestore(Map<String, dynamic> jsonCommit) {
+    final CommitDocument result = CommitDocument()
+      ..documentName = jsonCommit['DocumentName']
+      ..createTimestamp = Int64() + jsonCommit['CreateTimestamp']!
```
ricardoamador, line 313: This line is odd, can you just instantiate the Int64 with the jsonCommit value here?

ricardoamador: Yeah! See your code below.

dashboard/lib/service/appengine_cocoon.dart
keyonghan: I am not sure about potential error types here. At this stage, I think it would be good to keep logics the same as before. We can identify different error types and do different actions when everything sets down. Does this sound good to you?

dashboard/lib/service/appengine_cocoon.dart
```diff
@@ -224,11 +259,34 @@ class AppEngineCocoonService implements CocoonService {
     return statuses;
   }
 
+  List<CommitTasksStatus> _commitStatusesFromJsonFirestore(List<dynamic>? jsonCommitStatuses) {
+    assert(jsonCommitStatuses != null);
+    // TODO(chillers): Remove adapter code to just use proto fromJson method. https://github.com/flutter/cocoon/issues/441
```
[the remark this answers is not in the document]
keyonghan: Removed.

dashboard/lib/service/appengine_cocoon.dart
keyonghan: Done for both commit and task.

dashboard/lib/service/appengine_cocoon.dart
keyonghan: Fixed!

dashboard/lib/service/appengine_cocoon.dart
ricardoamador: That works for me.

ricardoamador (approved):
LGTM

PR #3635, opened 2024-04-11
Add API `api/public/get-status-firestore`

keyonghan:

This PR adds api entry point for `api/public/get-status-firestore`. This is prepare frontend migration to Firestore.

This change is a no-op, as the frontend is still calling `api/public/get-status`.

### DIFF
--- a/app_dart/lib/src/request_handlers/get_status_firestore.dart
+++ b/app_dart/lib/src/request_handlers/get_status_firestore.dart
@@ -83,8 +83,6 @@ class GetStatusFirestore extends RequestHandler<Body> {
   }
 }
 
-/// The serialized representation of a [CommitStatus].
-// TODO(tvolkert): Directly serialize [CommitStatus] once frontends migrate to new format.
 class SerializableCommitTasksStatus {
   const SerializableCommitTasksStatus(this.status);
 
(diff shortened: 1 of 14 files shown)


app_dart/lib/src/request_handlers/get_status_firestore.dart
```diff
@@ -0,0 +1,97 @@
...
+        orElse: () => throw NotFoundException('Failed to find commit with key $ownerKey'),
+      );
+
+      lastCommitTimestamp = commit.timestamp!;
+    }
+    return lastCommitTimestamp;
+  }
+}
+
+class SerializableCommitTasksStatus {
```
ricardoamador, line 86: Should we consider moving this to its own class file? Will it potentially have uses outside of this class and do you think it makes sense to move it closer to the CommitTaskStatus class?

ricardoamador (approved):
LGTM with a non blocking question.

app_dart/lib/src/request_handlers/get_status_firestore.dart
keyonghan: Good point. Created a new model file `commit_tasks_status.dart`, and moved logics related to `CommitTasksStatus` there.

PR #3666, opened 2024-04-16
Frontend support for firestore migration

keyonghan:


dashboard/lib/logic/qualified_task.dart
```diff
@@ -4,59 +4,61 @@
...
+const Set<String> dartInternalTasks = <String>{
+  'Linux engine_release_builder',
+  'Linux flutter_release',
+  'Linux flutter_release_builder',
+  'Linux packaging_release_builder',
+  'Mac engine_release_builder',
+  'Mac packaging_release_builder',
+  'Windows engine_release_builder',
+  'Windows packaging_release_builder',
+};
```
keyonghan, line 25: @godofredoc Do we expect to have more dart internal tasks? I have been thinking about whether it deserves the effort to add a new `Task` field to identify these targets, similar to the old `stage`. If these targets are relatively steady, then a list like this may be enough. Or is there a pattern I can use here to determine whether a target is from dart-internal? like `*flutter_release*`/`*engine_release*`/`*packaging_release*`?

PR #3669, opened 2024-04-17
Add `BuildList` field to ease frontend logics

keyonghan:
Frontend needs `BuildList` to handle case of multiple reruns. This PR updates the returned results from.

```
statuses:
  commit,
  tasks
```

to 

```
statuses:
  Commit,
  Tasks
    Task
    BuildList
```

Part of https://github.com/flutter/flutter/issues/142951.

PR #3670, opened 2024-04-17
Add buildList field in Frontend task model

keyonghan:
This is a follow up of https://github.com/flutter/cocoon/pull/3669 to support `buildList` in frontend data model.


ricardoamador (approved):
LGTM

PR #3685, opened 2024-04-23
Clean up Frontend task stages

keyonghan:
It has been a technical debt to support Task `stages` in both dashboard and backend: we used to support `cirrus`, `travis`, `chrome`, `google`, `LUCI`, `cocoon` as stages. Over past years, we have been refactoring everything in LUCI, and there is no need to maintain stages.

Now the only `stage` tasks we support is `dart-internal` ones, which are still LUCI though. To distinguish these tests, we reuse the concept of `stage`, but this PR simplifies corresponding logics.

This PR prepares for Frontend switch over to Firestore, which doesn't support `stage` any more.

### DIFF
--- a/dashboard/lib/logic/qualified_task.dart
+++ b/dashboard/lib/logic/qualified_task.dart
@@ -6,57 +6,59 @@ import 'package:flutter/foundation.dart';
 
 import '../model/task.pb.dart';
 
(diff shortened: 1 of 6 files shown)


keyonghan:
This also fixes: https://github.com/flutter/flutter/issues/145549

PR #3697, opened 2024-04-30
Reland buildbucket v2 migration parts 3&4

stuartmorgan:
Reverts https://github.com/flutter/cocoon/pull/3686, re-landing the most recent parts of the buildbucket v2 migration.

The one difference relative to the original PRs (see commit 2 in the PR) is that in addition to duplicating all the services and handlers, the original PR duplicated the *cron jobs*. My initial impression of the scheduler code is that there's a non-trivial risk of race conditions in the current design, and so my speculation is that having two cron jobs scheduled at the same time, both trying to run backfill, tripped race conditions that put the databases into bad states.

This is definitely speculative, but given that my understanding is that v2 was successfully handling traffic for at least a day, a race condition rather than a fundamental problem with the implementation of the v2 communication protocol is certainly plausible. My proposal is that we deploy this to test that theory, and to get new data for failure analysis if it doesn't work, but be prepared to roll back and manually adjust the DB again if necessary.

## Pre-launch Checklist

- [x] I signed the [CLA].


eyebrowsoffire (approved):
I have not done a comprehensive review of this entire 15k line change, but it was previously reviewed by knowledgable infra team members, and I think your theory about double cron jobs seems reasonable to me.

stuartmorgan:
> I have not done a comprehensive review of this entire 15k line change

FWIW I did some spot checks of the diffs between the original code and the added v2 copies of the code, and in most of it, it's exactly the same code except using the `V2` scheduler class instead of the original scheduler class, and those two are also quite similar except for the actual LUCI interactions. I'm pretty sure an abstraction around the V1 and V2 scheduler class could remove a ton of code duplication here, but since the goal is to quickly get to an all-V2 state and remove the V1 copies I didn't want to start wading into the PR and reworking it given that it's already essentially done. (If the cron job theory is wrong and I need to dig further, I may well go the refactoring route at that point.)

christopherfujino (approved):
RSLGTM

christopherfujino:
> FWIW I did some spot checks of the diffs between the original code and the added v2 copies of the code, and in most of it, it's exactly the same code except using the `V2` scheduler class instead of the original scheduler class, and those two are also quite similar except for the actual LUCI interactions. I'm pretty sure an abstraction around the V1 and V2 scheduler class could remove a ton of code duplication here, but since the goal is to quickly get to an all-V2 state and remove the V1 copies I didn't want to start wading into the PR and reworking it given that it's already essentially done. (If the cron job theory is wrong and I need to dig further, I may well go the refactoring route at that point.)

This makes sense to me, including landing this as is now and monitoring the db.

PR #3699, opened 2024-05-02
Vacuum stale In Progress tasks more frequently

stuartmorgan:
https://github.com/flutter/flutter/issues/122117 is still happening frequently. Currently the cron job is only running every 12 hours, which means that when it happens we can have large numbers of stuck columns on the dashboard (it seems that when this happens it sometimes happens to an entire batch of backfill scheduling, which is currently 50) for up to 12 hours.

Looking at the history of this task:
- Originally it ran every 10 minutes, and was much more aggressive in what tasks it reset, causing https://github.com/flutter/flutter/issues/121989 and was subsequently disabled in https://github.com/flutter/cocoon/pull/2515
- It was re-enabled in https://github.com/flutter/cocoon/pull/3354, more narrowly targetted so it wouldn't reset jobs that had been assigned builds. However, it was re-landed with a much more conservative 12 hour cycle. That version *replaced* the time-based check, however, instead of adding to it, which means it's subject to a race when it runs. The justification for removing it was that since the job frequency was low the race wouldn't be an issue (which is only somewhat true).

This re-adds a time-based check to avoid races, and increases the frequency to one hour (which may still cause some tree delays; we can evaluate once it's been running in this mode for a while if we want to make it more often and make the timestamp check less conservative). It also removes the reset of the creation timestamp to 0 so that the timestamp check will work; that appears to be a legacy of the earliest version of this job which was trying to clean up a slightly different case involving null creation timestamps.

Mitigation for https://github.com/flutter/flutter/issues/122117

## Pre-launch Checklist

- [x] I signed the [CLA].


app_dart/lib/src/request_handlers/scheduler/vacuum_stale_tasks.dart
```diff
@@ -32,8 +32,9 @@ class VacuumStaleTasks extends RequestHandler<Body> {
   /// For testing, can be used to inject a deterministic time.
   final DateTime? nowValue;
 
-  /// Tasks that are in progress for this duration will be reset.
-  static const Duration kTimeoutLimit = Duration(hours: 3);
+  /// Tasks that are in progress without a build for this duration will be
+  /// reset.
+  static const Duration kTimeoutLimit = Duration(minutes: 30);
```
christopherfujino, line 37: This seems a little aggressive, does this need to be lowered?

stuartmorgan: The code only triggers if the build ID is `null`. Maybe I'm wrong about the protocol here, but I thought we would expect to get a build ID back from LUCI pretty much right away, and if that's the case than 30 minutes is a very long time.

I can put it back to 3 hours, but that means that when this case happens where an entire block of 50 backfill builds never gets scheduled, it'll be 3-4 hours before we even start to recover those columns, which seemed undesirable. The 3 hour timeout was from the original code that didn't check for `null` builds, which meant that it would restart any job older than 3 hours that was in progress *even if we knew it was running*, which I assume is why it was so high.

Are there expected cases where we wouldn't get a build number from LUCI within 30 minutes?

eyebrowsoffire (approved):
LGTM!

app_dart/lib/src/request_handlers/scheduler/vacuum_stale_tasks.dart
christopherfujino: is the build ID filled in if it's pending but not started? If so, then 30 minutes SGTM.

stuartmorgan: I asked LUCI folks about the potential max delay here, and it turns out that we should actually be getting the build ID back *immediately* when we do the RPC, we don't actually have to wait for the pubsub like we did with v1. I started to look and it didn't seem that we were actually using that information, at least for backfill.

So the build ID exists ~immediately, but I don't know when we get it via pubsub. I'll adjust this back to three hours for now to make the change more conservative, land it for incremental tree health improvement, and then start looking into using the RPC response data so it's more robust.

christopherfujino: Cool, SGTM

PR #3700, opened 2024-05-03
Remove most Buildbucket v1 code

stuartmorgan:
This eliminates most of the Buildbucket v1 code:
- Removes a number of v1 REST handlers with the v2 version (making the two versioned endpoints for these actually call identical code internally):
  - batch backfiller service—the code was identical other than using the v2 scheduler
  - github vacuum service—the code was identical other than using the v2 scheduler
  - try job reset service—the code was identical other than using the v2 scheduler
  - prod job reset service—the code was identical other than using the v2 scheduler, and the v2 version didn't get the logging from https://github.com/flutter/cocoon/pull/3636 so I copied that over.
- Removes the v1 scheduler abstraction, replacing all use with the v2 version.
- Removes the unused portions of the v1 GitHub checks service.
- Removes the v1 LCUI build service abstraction, replacing almost all use with the v2 version.
- Removes the no-longer-used v1 Buildbucker methods (in buildbucket.dart)

The v1 code that remains is code for which v2 services can't be used as a drop-in replacement, which is places where the legacy v1 pubsub subscription handlers take incoming v1 messages and use them to reschedule builds. To replace that with v2 would require writing a complex v1-data-to-v2-data translation layer, which shouldn't be necessary because nothing should be coming in on the v1 pubsub subscription in the first place. However, because that's harder to reason directly about (since it depends on more than the local code), I'm leaving removing that for a later follow-up step; once this PR is deployed, we can guarantee that we aren't accidentally generating any new v1 requests, so then we can wait, watch server traffic, and then do a follow-up to simply remove the v1 subscription logic and all the code it uses when we see that there is in fact none. (Since that will be a separate PR, it will also be easy to roll back via traffic redirection if we somehow miss something.)

Even where the v1 versions of classes are removed, the v2 class and file naming is left for now. Cleaning that will be done in a later follow-up since diffs will then correctly show it as a move, whereas if I moved them over the deleted code in the same PR, it would look like this PR was deleting the v2 version and doing major rewrites to v1, making the PR much harder to reason about.

Part of https://github.com/flutter/flutter/issues/135934

## Pre-launch Checklist

- [x] I signed the [CLA].


app_dart/test/service/github_checks_service_v2_test.dart
stuartmorgan, line 1: Apparently no v2 test file was created for the v2 service. This is a duplicate of the v1 test file, modified to fake/mock the correct bits for the v2 variant, so that we have test coverage of the code we are using.

app_dart/lib/src/request_handlers/presubmit_luci_subscription.dart
```diff
@@ -126,4 +129,60 @@ class PresubmitLuciSubscription extends SubscriptionHandler {
     }
     return properties['presubmit_max_attempts'] as int;
   }
+
+  /// Sends [ScheduleBuildRequest] using information from a given build's
```
stuartmorgan, line 133: All of the "new" code here is moved here inline from v1 service objects that I deleted, to make it easier to reason about the fact that the legacy code is only used in this one location. (The alternative would be to leave partial versions of the v1 service classes; I can unwind that if there's a strong preference to do it that way.)

app_dart/lib/src/request_handlers/scheduler/batch_backfiller_v2.dart
```diff
@@ -138,9 +137,9 @@ class BatchBackfillerV2 extends RequestHandler {
     }
     final List<Tuple<Target, FullTask, int>> filteredBackfill = <Tuple<Target, FullTask, int>>[];
     final List<Tuple<Target, FullTask, int>> highPriorityBackfill =
-        backfill.where((element) => element.third == LuciBuildService.kRerunPriority).toList();
+        backfill.where((element) => element.third == LuciBuildServiceV2.kRerunPriority).toList();
```
stuartmorgan, line 140: The constants were duplicated when the service class was duplicated, but none of the code had updated to use the new copy of the constants. I checked that the values were all the same.

app_dart/lib/src/service/github_checks_service.dart
stuartmorgan, line 1: Code left in this file is used in both `PresubmitLuciSubscription` and `PostsubmitLuciSubscription` (both of which are v1 pubsub handlers), so I didn't inline it.

christopherfujino:
Taking a look at this now...

app_dart/lib/src/request_handlers/scheduler/batch_backfiller_v2.dart
```diff
@@ -7,7 +7,6 @@ import 'package:cocoon_service/src/model/appengine/task.dart';
 import 'package:cocoon_service/src/model/firestore/task.dart' as firestore;
 import 'package:cocoon_service/src/service/datastore.dart';
 import 'package:cocoon_service/src/service/scheduler/policy.dart';
-import 'package:cocoon_service/src/service/scheduler_v2.dart';
```
christopherfujino, line 10: Can this library and `class BatchBackfillerV2` be renamed to exclude the v2?

stuartmorgan: This is addressed at the end of the PR description:

> Even where the v1 versions of classes are removed, the v2 class and file naming is left for now. Cleaning that will be done in a later follow-up since diffs will then correctly show it as a move, whereas if I moved them over the deleted code in the same PR, it would look like this PR was deleting the v2 version and doing major rewrites to v1, making the PR much harder to reason about.

christopherfujino: sorry, missed that

app_dart/lib/src/service/github_checks_service.dart
```diff
@@ -205,35 +172,4 @@ class GithubChecksService {
         throw StateError('unreachable');
     }
   }
-
-  /// Given a [headSha] and [checkSuiteId], finds the [PullRequest] that matches.
-  Future<github.PullRequest?> findMatchingPullRequest(
```
christopherfujino, line 210: Was this dead code?

stuartmorgan: The only client was v1 `Scheduler`, so removing that removed the need for this. v2 `Scheduler` uses the v2 version of `GithubChecksService`, which has its own copy of this.

(There's a lot of duplication right now; this PR is removing a lot of it, and then the rest will come out in a follow-up later.)

stuartmorgan:
I'll keep an eye on the tree tomorrow once this rolls out, and be prepared to roll traffic back. This will shift some REST handlers to the v2 codepath, so while I don't expect breakage, it's not actually a no-op.

app_dart/lib/src/service/github_checks_service.dart
christopherfujino: SGTM. I spot checked this for obvious copy paste mistakes, but I figure if any subtle bugs we'll catch in prod :)

christopherfujino:
> I'll keep an eye on the tree tomorrow once this rolls out, and be prepared to roll traffic back. This will shift some REST handlers to the v2 codepath, so while I don't expect breakage, it's not actually a no-op.

SGTM

PR #3706, opened 2024-05-06
Remove semantics from the cocoon dashboard.

matanlurey:
Closes https://github.com/flutter/flutter/issues/145849.

This was originally added by Casey so he could use some Chrome extension to navigate the dashboard (https://github.com/flutter/cocoon/pull/1729), but it appears to cause serious problems for mouse-users.

I've filed https://github.com/flutter/flutter/issues/147881 for the web team to take a look.

### DIFF
--- a/dashboard/lib/main.dart
+++ b/dashboard/lib/main.dart
@@ -67,8 +67,6 @@ void main([List<String> args = const <String>[]]) async {
       child: Now(child: const MyApp()),
     ),
   );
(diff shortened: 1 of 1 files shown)


stuartmorgan:
What are the accessibility implications of this? Will it break screenreaders, for instance?

stuartmorgan:
Okay, in that case I'm fine with trying this for now then, and gathering actual feedback about what (if anything) regresses that people are using.

PR #3707, opened 2024-05-08
Remove remaining v1 Buildbucket code

stuartmorgan:
This eliminates the remaining Buildbucket v1 code:
- Removes several v1 REST handlers that should no longer be used:
  - `api/presubmit-luci-subscription` - callback from LUCI for v1 requests, which we should no longer be making
  - `api/postsubmit-luci-subscription` - callback from LUCI for v1 requests, which we should no longer be making
  - `api/schedul/batch-request-subscription` - called by the v1 scheduler, which was removed
- Removes v1 helper code that was used only by those removed endpoints
- Removes the low-level v1 buildbucket abstractions (`buildbucket.dart` and `push_message.dart`)
- Removes the v1 service handler base class, moving the one remaining thing on it to the v2 base (requiring only minor changes to the code inspecting the message due to using the v2 low-level abstractions; it's not a LUCI handler, so the actual data didn't change).

Server logs show these three endpoints dormant since May 2nd, except for a blip this morning when I created activity by re-trying a job. I think that it was because I was using a window that hadn't refreshed since before this morning's deployment (since I can't repro now, and the timing lines up), but I'll watch the logs for several more days before landing this just in case.

Even where the v1 versions of classes are removed, the v2 class and file naming is left for now. Cleaning that will be done in a later follow-up since diffs will then correctly show it as a move, whereas if I moved them over the deleted code in the same PR, it would look like this PR was deleting the v2 version and doing major rewrites to v1, making the PR much harder to reason about.

Follow-up to https://github.com/flutter/cocoon/pull/3700
Part of https://github.com/flutter/flutter/issues/135934

## Pre-launch Checklist

- [x] I signed the [CLA].


christopherfujino (approved):
LGTM

stuartmorgan:
All three subscriptions have continued to be dormant, as expected, so landing. I'll leave the subscriptions themselves in GCP for now, for easier unwinding in case we need to revert for any reason.

PR #3777, opened 2024-06-11
Decode BuildsV2PubSub.build_large_fields and merge it with the rest of the build data

jason-simmons:
https://chromium.googlesource.com/infra/luci/luci-go/+/refs/heads/main/buildbucket/proto/notification.proto says that some parts of the Buildbucket build proto are stored in a separate compressed field.

This must be merged with the rest of the build proto so that consumers of the build will see fields such as build.input.properties

See https://github.com/flutter/flutter/issues/149851

### DIFF
--- a/app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
+++ b/app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
@@ -3,6 +3,7 @@
 // found in the LICENSE file.
 
 import 'dart:convert';
+import 'dart:io';
 
 import 'package:buildbucket/buildbucket_pb.dart' as bbv2;
 import 'package:cocoon_service/ci_yaml.dart';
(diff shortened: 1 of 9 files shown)


app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
```diff
@@ -82,9 +83,13 @@ class PostsubmitLuciSubscription extends SubscriptionHandler {
     }
 
     final bbv2.Build build = buildsPubSub.build;
+
     // Note that result is no longer present in the output.
     log.fine('Updating buildId=${build.id} for result=${build.status}');
 
+    // Add build fields that are stored in a separate compressed buffer.
+    build.mergeFromBuffer(ZLibCodec().decode(buildsPubSub.buildLargeFields));
```
christopherfujino, line 91: Have we been building the wrong commits post-submit too?!

christopherfujino: Although, I suppose the git_ref field is only used in presubmit. Still, I wonder if there are any other correctness issues we might have with prod binaries....

christopherfujino (approved):
LGTM

app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
jason-simmons: I don't know if there have been any specific problems affecting postsubmit.  But IIUC any consumer of `BuildsV2PubSub` should be merging the data from `build_large_fields`.

christopherfujino:
dart format failure, i'll try to fix

Hixie (approved):
man the formatter really did a number on that code

LGTM I guess

PR #3675, opened 2024-04-19
Fix postsubmit rerun based on checkrun

keyonghan:
Fixes: https://github.com/flutter/flutter/issues/147007

PR #3780, opened 2024-06-11
Fix postsubmit rerun based on checkrun - v2

stuartmorgan:
This is an update of https://github.com/flutter/cocoon/pull/3675, adjusted for v2.

Fixes: https://github.com/flutter/flutter/issues/147007


stuartmorgan:
I don't have a token that is authorized to push to other people's PRs for flutter/cocoon; rather than set up a new token just for this I opted for making a new PR instead.

This may actually be better though, since it allows for an easy side-by-side of the diffs in the original to see what I changed relative to that one to adapt to v2.

stuartmorgan:
(I don't currently know how to manually test this, but the original was apparently manually tested, and this version passes the unit tests, so I'm moderately confident. I'll make sure this lands separately from anything else so I can test it once it rolls out, and then roll it back right away if needed.)

stuartmorgan:
@christopherfujino Ping?

christopherfujino (approved):
Oops sorry I forgot about this, LGTM

PR #3803, opened 2024-07-03
Add target property `flakiness_threshold`

jtmcdole:
Targets should be able to define higher or lower threshold limits.

Fixes flutter/flutter#151195

### DIFF
--- a/app_dart/lib/src/model/ci_yaml/ci_yaml.dart
+++ b/app_dart/lib/src/model/ci_yaml/ci_yaml.dart
@@ -5,6 +5,7 @@
 import 'dart:convert';
 
 import 'package:cocoon_service/cocoon_service.dart';
+import 'package:collection/collection.dart';
 import 'package:github/github.dart';
 
 import '../proto/internal/scheduler.pb.dart' as pb;
@@ -81,6 +82,12 @@ class CiYaml {
     // filter if release_build true if current branch is a release candidate branch.
     enabledTargets = _filterReleaseBuildTargets(enabledTargets);
     return enabledTargets;
+  }
+
+  /// Gets the first [Target] matching [builderName] or null.
+  Target? getFirstTarget(String builderName) {
+    return _targets
+        .singleWhereOrNull((Target target) => target.value.name == builderName);
   }
 
   /// Filters post submit targets to remove targets we do not want backfilled.
--- a/app_dart/lib/src/model/ci_yaml/ci_yaml.dart
+++ b/app_dart/lib/src/model/ci_yaml/ci_yaml.dart
@@ -85,8 +85,8 @@ class CiYaml {
   }
 
   /// Gets the first [Target] matching [builderName] or null.
-  Target? getFirstTarget(String builderName) {
-    return _targets.singleWhereOrNull((Target target) => target.value.name == builderName);
+  Target? getFirstPostsubmitTarget(String builderName) {
+    return postsubmitTargets.singleWhereOrNull((Target target) => target.value.name == builderName);
   }
 
   /// Filters post submit targets to remove targets we do not want backfilled.
--- a/app_dart/lib/src/model/ci_yaml/ci_yaml.dart
+++ b/app_dart/lib/src/model/ci_yaml/ci_yaml.dart
@@ -86,8 +86,7 @@ class CiYaml {
 
   /// Gets the first [Target] matching [builderName] or null.
   Target? getFirstTarget(String builderName) {
-    return _targets
-        .singleWhereOrNull((Target target) => target.value.name == builderName);
+    return _targets.singleWhereOrNull((Target target) => target.value.name == builderName);
   }
 
   /// Filters post submit targets to remove targets we do not want backfilled.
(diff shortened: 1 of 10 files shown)


app_dart/lib/src/model/ci_yaml/target.dart
```diff
@@ -41,6 +41,8 @@ class Target {
 
   static String kIgnoreFlakiness = 'ignore_flakiness';
 
+  static const String kFalkinessThreshold = 'flakiness_threshold';
```
jmagman, line 44:
```suggestion
  static const String kFlakinessThreshold = 'flakiness_threshold';
```

app_dart/lib/src/model/ci_yaml/target.dart
```diff
@@ -208,6 +210,9 @@ class Target {
       return value.replaceAll('\\n', '\n');
     } else if (int.tryParse(value) != null) {
       return int.parse(value);
+    } else if (double.tryParse(value) != null) {
```
jmagman, line 213: Is the comment right above relevant, that the changes need to be also made in the infra repo?
https://github.com/flutter/cocoon/blob/4369cde1522bf88dfc2b6f99fb62c8a45108ae3c/app_dart/lib/src/model/ci_yaml/target.dart#L194-L197

app_dart/lib/src/request_handlers/file_flaky_issue_and_pr.dart
```diff
@@ -99,7 +99,8 @@ class FileFlakyIssueAndPR extends ApiRequestHandler<Body> {
       return true;
     }
     // Skips if the flaky percentage is below the threshold.
-    if (statistic.flakyRate < _threshold) {
+    final threshold = ciYaml.getFirstTarget(statistic.name)?.flakinessThreshold ?? _threshold;
```
jmagman, line 102: It looks like `getIgnoreFlakiness` already has a pattern like this that uses `postsubmitTargets` instead of your new `getFirstTarget`?
https://github.com/flutter/cocoon/blob/4369cde1522bf88dfc2b6f99fb62c8a45108ae3c/app_dart/lib/src/request_handlers/check_flaky_builders.dart#L168-L172

app_dart/lib/src/model/ci_yaml/target.dart
jtmcdole: I'll investigate tomorrow and see

app_dart/lib/src/model/ci_yaml/target.dart
```diff
@@ -251,6 +256,15 @@ class Target {
     }
     return false;
   }
+
+  /// Returns the value of flakiness_threshold property or the [global].
+  double? get flakinessThreshold {
+    final Map<String, Object> properties = getProperties();
+    if (properties.containsKey(kFalkinessThreshold)) {
+      return properties[kFalkinessThreshold] as double;
```
christopherfujino, line 264: spelling here and on 263

app_dart/lib/src/model/ci_yaml/target.dart
jtmcdole: I think I need to update `_parse_attribute` from there, but I need to grok python now :)

jtmcdole: starlark does not support `float`; so if I do nothing on Luci - this will get parsed as a string (since `isdigit()` will return false).

I do not expect to do anything with this new flag on Luci and only on Cocoon; so I don't think I need to do this at all.

app_dart/lib/src/model/ci_yaml/target.dart
jtmcdole: Done!

app_dart/lib/src/request_handlers/file_flaky_issue_and_pr.dart
jtmcdole: I think you're right about using post submits, will update. I do need this to work in multiple places; so making another getter here wouldn't look great.

I'll refactor.

jtmcdole:
Updated from feedback. I think luci star files are a no-op.

app_dart/lib/src/model/ci_yaml/target.dart
jmagman: Maybe add a comment then, like "double is not currently implemented in LUCI" or something similar.

app_dart/lib/src/model/ci_yaml/ci_yaml.dart
```diff
@@ -83,6 +84,11 @@ class CiYaml {
     return enabledTargets;
   }
 
+  /// Gets the first [Target] matching [builderName] or null.
+  Target? getFirstPostsubmitTarget(String builderName) {
```
jmagman, line 88: Can `getIgnoreFlakiness` also be updated to use this?

jmagman (approved):
LGTM with nits

app_dart/lib/src/model/ci_yaml/ci_yaml.dart
jtmcdole: done

app_dart/lib/src/model/ci_yaml/target.dart
jtmcdole: done

PR #3823, opened 2024-07-17
exclude sky/tools python scripts from needing tests

christopherfujino:
The sky/tools/*.py files are essentially extensions of our *.gn build files.


cbracken (approved):
LGTM stamp from a Japanese personal seal

PR #3833, opened 2024-07-26
Allow using the `test: all` label in flutter/flutter.

matanlurey:
The label will still have to be created, but should now apply to both repos.


loic-sharma (approved):
Thanks!

christopherfujino:
Addresses: https://github.com/flutter/flutter/issues/151114

christopherfujino (approved):
LGTM

PR #3834, opened 2024-07-26
increase the go docker image version

christopherfujino:
Fix https://github.com/flutter/flutter/issues/152394
Because the slsa-verifier transitively depends on at least go sdk version 1.21


stuartmorgan (approved):
LGTM

PR #3835, opened 2024-07-26
run go mod tidy on tooling/go.mod and downgrade docker image

christopherfujino:
Addresses https://github.com/flutter/flutter/issues/152394

Last build failed with an error message to run `go mod tidy`, which i did on //tooling/go.mod. Also downgraded docker image to golang:1.21 since that's what upstream slsa-verifier uses.


jason-simmons (approved):

tooling/go.mod
```diff
@@ -1,6 +1,8 @@
 module cocoon-tooling
 
-go 1.18
+go 1.21
+
+toolchain go1.21.4
```
stuartmorgan, line 5: This appears to be breaking the build:

```
go: errors parsing go.mod:
/workspace/tooling/go.mod:5: unknown directive: toolchain
```

PR #3843, opened 2024-07-29
Use a consistent version of Golang in all cloud build configurations

jason-simmons:
See https://github.com/flutter/cocoon/pull/3835


stuartmorgan (approved):
LGTM

PR #3889, opened 2024-09-05
dart_test_runner: output expected dart format args

cbracken:
Currently on a failed run due to unformatted dart code, the output looks like:

    ######### dart format #########
    Formatted lib/src/file_codesign_visitor.dart
    Formatted 10 files (1 changed) in 0.65 seconds.

In the face of this output, my natural inclination would be to run

    dart format path/to/file

but that would be incorrect. What would be correct would be to locate our CI test script, read the code, and discover that we format to a width of 120 characters, then format with those args. 

Instead, we now inform the user of the args directly in the test output to save them some digging.

Also checks to see the user actually supplied an argument to the script and if not, emits an error message and usage info then fails, rather than silently running  `pushd ""`, resulting in no change to PWD, then failing when it realises that the top level of the repo is not a pub package.

## Pre-launch Checklist

- [X] I signed the [CLA].
- [X] I do not like it Sam-I-Am, I do not like cocoon log spam.

### DIFF
--- a/test_utilities/bin/dart_test_runner.sh
+++ b/test_utilities/bin/dart_test_runner.sh
@@ -8,17 +8,25 @@
 
 set -e
 
-# Build and analyze
 echo "###### dart_test_runner #######"
 echo "Directory: $1"
+
+if [ -z "$1" ]; then
+  echo "ERROR: no package directory supplied."
+  echo "usage: $0 PACKAGE_DIR_TO_TEST"
+  exit 1
+fi
+
+# Setup.
 pushd "$1" > /dev/null
 flutter clean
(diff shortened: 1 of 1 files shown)


jmagman (approved):

test_utilities/bin/dart_test_runner.sh
```diff
@@ -8,17 +8,25 @@
 
 set -e
```
christopherfujino, line 9: While we're at it

```suggestion
set -euo pipefail
```

christopherfujino (approved):
nit about options, but LGTM

PR #3900, opened 2024-09-10
flaky issue bot should not override existing priorities

christopherfujino:
Fixes https://github.com/flutter/flutter/issues/154963


chunhtai (commented):
While I can see the reasoning behind this, but we will risking the serious flake gets ignored due to the owner giving it a low priority.

If the problem is that the bot is too annoying at non critical flake, we can consider raising the threshold or have multiple threshold level for different priority. e.g. >10% is P0, 2%~10% is P1, below 2% is P2. WDYT?

christopherfujino:
Closing as stale

PR #3901, opened 2024-09-12
Remove `ThemeData.dialogBackgroundColor` usage

TahaTesser:
Related https://github.com/flutter/flutter/pull/155072

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read the [Flutter Style Guide] _recently_, and have followed its advice.
- [x] I signed the [CLA].
- [ ] I listed at least one issue that this PR fixes in the description above.
- [ ] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#overview
[Tree Hygiene]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md
[test-exempt]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/blob/master/docs/contributing/Chat.md

PR #3902, opened 2024-09-12
Filtered Android tasks should include mokey and pixel 7 pro in dashboard

jmagman:
Update the dashboard so "mokey" and "pixel_7pro"  tasks are filtered/iconed as Android.  https://github.com/flutter/flutter/issues/148085

For example, `Linux_mokey old_gallery__transition_perf` should show up in the dashboard as an Android test, but its icon and filtering counts it as a Linux test:

Screenshot 2024-09-12 at 9 57 52 AM
Screenshot 2024-09-12 at 9 57 48 AM

Also fix two unrelated warnings.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/dashboard/lib/main.dart
+++ b/dashboard/lib/main.dart
@@ -8,7 +8,6 @@ import 'package:firebase_core/firebase_core.dart';
 import 'package:firebase_crashlytics/firebase_crashlytics.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
-import 'package:flutter/rendering.dart';
 import 'package:flutter/services.dart';
 
 import 'build_dashboard_page.dart';
--- a/dashboard/lib/widgets/task_icon.dart
+++ b/dashboard/lib/widgets/task_icon.dart
@@ -51,9 +51,8 @@ class TaskIcon extends StatelessWidget {
     final String matchedName = qualifiedTask.task!.toLowerCase();
     final bool isWebTest = matchedName.contains('_web') || matchedName.contains('web_');
     final bool isToolTest = matchedName.contains('_tool') || matchedName.contains('tool_');
-    final bool isAndroidTest = matchedName.contains('_android') ||
-        matchedName.contains('_mokey') ||
-        matchedName.contains('_pixel_7pro');
+    final bool isAndroidTest =
+        matchedName.contains('_android') || matchedName.contains('_mokey') || matchedName.contains('_pixel_7pro');
 
     if (matchedName.contains('_fuchsia')) {
       return Padding(
--- a/dashboard/lib/widgets/task_icon.dart
+++ b/dashboard/lib/widgets/task_icon.dart
@@ -51,7 +51,9 @@ class TaskIcon extends StatelessWidget {
     final String matchedName = qualifiedTask.task!.toLowerCase();
     final bool isWebTest = matchedName.contains('_web') || matchedName.contains('web_');
     final bool isToolTest = matchedName.contains('_tool') || matchedName.contains('tool_');
-    final bool isAndroidTest = matchedName.contains('_android') || matchedName.contains('_mokey');
+    final bool isAndroidTest = matchedName.contains('_android') ||
+        matchedName.contains('_mokey') ||
+        matchedName.contains('_pixel_7pro');
 
     if (matchedName.contains('_fuchsia')) {
       return Padding(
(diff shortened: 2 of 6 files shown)


dashboard/lib/widgets/task_icon.dart
```diff
@@ -51,6 +51,7 @@ class TaskIcon extends StatelessWidget {
     final String matchedName = qualifiedTask.task!.toLowerCase();
     final bool isWebTest = matchedName.contains('_web') || matchedName.contains('web_');
     final bool isToolTest = matchedName.contains('_tool') || matchedName.contains('tool_');
+    final bool isAndroidTest = matchedName.contains('_android') || matchedName.contains('_mokey');
```
zanderso, line 54: All the instances of `_android` should be gone, and I think all we have now are `_mokey` and `_pixel_7pro`.

jmagman: Maybe next time we do this we use "_highend" and "_lowend" or something similar so the builder names don't change.

dashboard/lib/main.dart
```diff
@@ -8,7 +8,6 @@ import 'package:firebase_core/firebase_core.dart';
 import 'package:firebase_crashlytics/firebase_crashlytics.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
-import 'package:flutter/rendering.dart';
```
jmagman, line 11: Unrelated to this change but fix the warning:
```
info: The import of 'package:flutter/rendering.dart' is unnecessary because all of the used elements are also provided by the import of 'package:flutter/material.dart'. (unnecessary_import at [flutter_dashboard] lib/main.dart:11)
```

dashboard/test/widgets/task_grid_test.dart
```diff
@@ -878,5 +878,10 @@ Future<void> expectTaskBoxColorWithMessage(WidgetTester tester, String message,
   expect(pixels!.lengthInBytes, (cellPixelArea * 4).round());
   const double padding = 4.0;
   final int rgba = pixels.getUint32((((cellPixelSize * (cellSize + padding)) + cellSize + padding).ceil()) * 4);
-  expect((rgba >> 8) | (rgba << 24) & 0xFFFFFFFF, expectedColor.value);
+  final Color actualColor = Color((rgba >> 8) | (rgba << 24) & 0xFFFFFFFF);
```
jmagman, line 881: Unrelated to this change but fix the warning:
```
info: 'value' is deprecated and shouldn't be used. Use component accessors like .r or .g. (deprecated_member_use at [flutter_dashboard] test/widgets/task_grid_test.dart:881)
```

TahaTesser: I've same issue in https://github.com/flutter/cocoon/pull/3901#issuecomment-2348290865. I provided some details on how it seems to be handled in the framework and tagged @gaaclarke.

zanderso (approved):

dashboard/lib/widgets/task_icon.dart
zanderso: Changing the builder name is our current mechanism for starting a new series in SkiaPerf, which we want when we change the kind of device a benchmark runs on. We could investigate a new mechanism for that, and then we could leave the builder names the same.

dashboard/test/widgets/task_grid_test.dart
jmagman: @TahaTesser do you have any concerns about how I handled it here, checking `shade500`?

jmagman: You make a good point though, it would be good to have `isEquivalentTo` available outside the framework for anyone else hitting this deprecation.

dashboard/lib/widgets/task_icon.dart
jmagman: Or they could be different, but have some kind of stable way to identify them as Android from the name agnostic of the device hardware, so this doesn't need to be updated every time.

dashboard/test/widgets/task_grid_test.dart
TahaTesser: Your fix is great! This is an ongoing change so I just wanted to loop in for this edge case :)

jmagman: Ah ha, https://github.com/flutter/website/pull/11132 says use `isSameColorAs` matcher, which works on this PR.

TahaTesser: Thanks!

PR #3980, opened 2024-10-22
add webhook processing for merge_group

yjbanov:
Adds handling of the `merge_group` Github request. The handling only logs the request. It does not kick off any CI jobs just yet.


christopherfujino (approved):
LGTM

app_dart/lib/src/request_handlers/github/webhook_subscription.dart
```diff
@@ -211,6 +214,49 @@ class GithubWebhookSubscription extends SubscriptionHandler {
     }
   }
 
+  /// Handles a GitHub webhook with the event type "merge_group".
+  ///
+  /// Unlike a pull request a merge group contains multiple commits created
+  /// from multiple squashed pull requests. A merge group is therefore not
```
jtmcdole, line 220: is this part true? the "branch" is actually a squash of all PRs at that point?  My understanding was there was a merge_group created and checks requested for each PR in the merge queue; with CI running for each request in the merge queue.

yjbanov: I'll reword it. Each PR is squashed individually and produces one commit, then that commit is stacked on top of other commits in the merge group. So a group will contain multiple commits, one per PR. There's no extra squashing done.

jtmcdole: But we fire of CI for each PR?

test_utilities/bin/prepare_environment.sh
```diff
@@ -9,5 +9,5 @@
 set -ex
 
 dart pub global activate tuneup
-flutter channel master
+flutter channel stable
```
jtmcdole, line 12: intentional?

jtmcdole (approved):

test_utilities/bin/prepare_environment.sh
yjbanov: Was trying to see if stable `dart analyze` would be less crashy, but it crashed in another way, so I reverted it.

PR #3983, opened 2024-10-24
Proper merge queue JSON model; add a dummy check

yjbanov:
Get onto the JsonSerializable bandwagon for the `merge_group` data. Add a dummy check so we can start testing MQ <=> Cocoon integration. Failing PRs can be artificially created by adding the string "MQ_FAIL" in the commit message.


app_dart/lib/src/model/github/checks.dart
```diff
@@ -52,3 +52,96 @@ class CheckRun {
...
+
+  /// The commit SHA generated for the candidate PR to be merged.
+  ///
+  /// This is the SHA that must be checked by the CI.
+  final String headSha;
+
+  /// The ref to the read-only branch generated by Github for the respective PR.
+  final String headRef;
+
+  /// The SHA of the parent commit of the [headSha] commit.
```
jtmcdole, line 108: Maybe:
```
///
/// This is the SHA head of a merge queue protect branch
```

app_dart/lib/src/request_handlers/github/webhook_subscription.dart
```diff
@@ -233,8 +234,10 @@ class GithubWebhookSubscription extends SubscriptionHandler {
       throw BadRequestException('Malformed merge_group request:\n$rawRequest');
     }
 
-    final eventAction = request['action'] as String;
-    final headSha = (request['merge_group'] as Map<String, Object?>)['head_sha'] as String;
+    final mergeGroupEvent = MergeGroupEvent.fromJson(request);
```
jtmcdole, line 237: fun:
  final MergeGroupEvent(:mergeGroup, :action, :headSha) = MergeGroupEvent.fromJson(request);

but you don't use eventAction?

jtmcdole (approved):

app_dart/lib/src/request_handlers/github/webhook_subscription.dart
jtmcdole: ignore the second part; eventAction is used in the zippie

yjbanov: Done.

app_dart/lib/src/model/github/checks.dart
yjbanov: Used github docs verbatim for now. We can clarify when we know more about what this is.

jtmcdole (approved):
still LGTM

PR #3984, opened 2024-10-24
explain Github token TTL value

yjbanov:
Without an explanation it's not clear why we pass 8 minutes right after mentioning 10 minutes in the comment.


jtmcdole (approved):
10 minutes feels short.

PR #3985, opened 2024-10-25
add a required "CI tasks" check that locks the merge queue

yjbanov:
Because the MQ only waits for "required" checks to pass before merging PRs and merge groups, at least one required check is needed that stays in pending state while all other (non-required) checks pass. Required checks cannot be added dynamically. They are pre-configured in the repositories settings. The "CI tasks" check is that one pre-configured required check whose status is used as a proxy for "the status of all other checks".

### DIFF
--- a/app_dart/lib/src/service/scheduler.dart
+++ b/app_dart/lib/src/service/scheduler.dart
@@ -463,9 +463,7 @@ class Scheduler {
       slug,
       headSha,
       lock,
-      conclusion == CheckRunConclusion.success
-        ? null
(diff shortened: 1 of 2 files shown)


jtmcdole:
From conversation:

codefu believes the automagically created `check_suite` will also automagically track `check_runs` based on commit sha and repo, so cocoon checks generate during `ci.yaml validation` _also_ block the merge queue like they would block a presubmit.

we need to solve the new repo showing up in cocoon first.

jtmcdole (approved):

PR #3993, opened 2024-10-28
Match macOS and Xcode versions in framework repo

jmagman:
Fixes https://github.com/flutter/flutter/issues/157739

The Ruby builds purposely run in the devicelab, so the macOS and Xcode versions should match the framework devicelab builders.
https://github.com/flutter/flutter/issues/157739#issuecomment-2442348324


jmagman:
```
Error: Xcode alone is not sufficient on Sonoma.
Install the Command Line Tools:
  xcode-select --install
```
https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8732809658654453665/+/u/build_package/stdout

I guess cocoon recipe is missing some work that was done to support Xcode 15 cipd packages?

christopherfujino (approved):
LGTM

christopherfujino:
Mac_arm64 ruby is failing for the same root cause as https://github.com/flutter/flutter/issues/157814

PR #3995, opened 2024-10-29
Add isFusion checks

jtmcdole:
Monorepo checks are if a ref or PR is based on a tree with DEPS and engine/src/.gn.


yjbanov (approved):
lgtm

PR #3996, opened 2024-10-30
Fusion: apply engine labels

jtmcdole:
* Make FusionTester class for easier faking and testing.
* Fetch file list ahead of time - we don't want to fetch these 2x
* Count type of files changed

### DIFF
--- a/app_dart/lib/src/foundation/utils.dart
+++ b/app_dart/lib/src/foundation/utils.dart
@@ -37,65 +37,72 @@ Duration twoSecondLinearBackoff(int attempt) {
   return const Duration(seconds: 2) * (attempt + 1);
 }
 
-/// Tests if the [pr] is in flutter/flutter and engine assets are available.
-Future<bool> isFusionPR(
-  PullRequest pr, {
-  required HttpClientProvider httpClientProvider,
-  Duration timeout = _githubTimeout,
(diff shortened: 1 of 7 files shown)


jtmcdole:
Fixed the nits from the analyzer. Will continue working on expanding this into other parts of the scheduler.

app_dart/lib/src/foundation/utils.dart
```diff
@@ -37,65 +37,72 @@ Duration twoSecondLinearBackoff(int attempt) {
...
+      }
 
-    log.fine('isFusionRef: fusion ref - ');
-    return true;
-  } catch (e) {
-    log.fine('isFusionRef: not a fusion ref - error: $e');
-    return false;
+      log.fine('isFusionRef: fusion ref - ');
+      return true;
+    } catch (e) {
```
christopherfujino, line 102: I worry we will catch errors caused by something other than that it's not a fusion PR here, such as the GitHub API is temporarily down. Maybe we should test that it was specifically 404 http status code?

jtmcdole: Yeah, sounds good.

jtmcdole: Done.

PR #3997, opened 2024-10-31
apply the "CI tasks" check to all repos

yjbanov:
It should be safe to apply the "CI tasks" to all repos.

In repos that are not yet configured to use MQ, it will simply appear as an extra check on every PR that completes together with "ci.yaml validation", which shouldn't be disruptive.

Repos that are configured to use MQ will have this check configured as required, and this code will simply complete it as normal both in the PR and in the merge queue.


jtmcdole:
I think there's some failing tests that make assumptions that no other tasks are scheduled?

yjbanov:
Yes, I need to update tests for non-flaux repos that currently do not expect to see this check.

jtmcdole (approved):

PR #4006, opened 2024-11-04
Scheduler ci yaml merge

jtmcdole:
CiYaml now tracks one or more ci.yaml files in a fusion tree

CiYamlInner is the old CiYaml. Instead of merging or updating protobufs to know which target are engine targets and which are framework targets; keep them apart. We'll have to use this information later when scheduling engine builds first and waiting before we can schedule the test runs.

The fusion-engine-ciyaml will need to be updated to include paths; that needs to be added to the merge script.

- Downloading a ci.yaml path should respect the path requested
- Only adding the ability to find targets; scheduling to follow this PR.

The `settings.json` file has only the dart line length specified; which should be a shared setting between all developers as it is enforced by CI.

### DIFF
--- a/app_dart/lib/src/foundation/utils.dart
+++ b/app_dart/lib/src/foundation/utils.dart
@@ -294,14 +294,13 @@ List<String> validateOwnership(
   final YamlMap? ciYaml = loadYaml(ciYamlContent) as YamlMap?;
   final pb.SchedulerConfig unCheckedSchedulerConfig = pb.SchedulerConfig()..mergeFromProto3Json(ciYaml);
 
-  final CiYamlInner ciYamlFromProto = CiYamlInner(
-    type: CiType.any,
+  final CiYaml ciYamlFromProto = CiYaml(
     slug: Config.flutterSlug,
     branch: Config.defaultBranch(Config.flutterSlug),
-    config: unCheckedSchedulerConfig,
+    yamls: {CiType.any: unCheckedSchedulerConfig},
   );
 
-  final pb.SchedulerConfig schedulerConfig = ciYamlFromProto.config;
+  final pb.SchedulerConfig schedulerConfig = ciYamlFromProto.configForInner(CiType.any);
 
   for (pb.Target target in schedulerConfig.targets) {
     final String builder = target.name;
--- a/app_dart/lib/src/foundation/utils.dart
+++ b/app_dart/lib/src/foundation/utils.dart
@@ -22,6 +22,7 @@ import '../request_handling/exceptions.dart';
 import '../service/logging.dart';
 
 const String kCiYamlPath = '.ci.yaml';
+const String kCiYamlFusionPath = 'engine/src/flutter/$kCiYamlPath';
 const String kTestOwnerPath = 'TESTOWNERS';
 
 /// Signature for a function that calculates the backoff duration to wait in
@@ -293,7 +294,8 @@ List<String> validateOwnership(
   final YamlMap? ciYaml = loadYaml(ciYamlContent) as YamlMap?;
   final pb.SchedulerConfig unCheckedSchedulerConfig = pb.SchedulerConfig()..mergeFromProto3Json(ciYaml);
 
-  final CiYaml ciYamlFromProto = CiYaml(
+  final CiYamlInner ciYamlFromProto = CiYamlInner(
+    type: CiType.any,
     slug: Config.flutterSlug,
     branch: Config.defaultBranch(Config.flutterSlug),
     config: unCheckedSchedulerConfig,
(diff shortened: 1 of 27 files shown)


app_dart/lib/src/model/ci_yaml/ci_yaml.dart
```diff
@@ -11,19 +11,130 @@ import 'package:github/github.dart';
...
+  /// Get's the [CiYamlInner] for the requested [type].
+  ///
+  /// The type is expected to exist and will fail otherwise.
+  CiYamlInner innerFor(CiType type) => configs[type]!;
+
+  /// The [RepositorySlug] that [config] is from.
+  final RepositorySlug slug;
+
+  /// The git branch currently being scheduled against.
+  final String branch;
```
yjbanov, line 69: Feels wrong for something that represents the ci.yaml config to be "scheduled against" some branch. Can we separate the scheduling concern from the configuration?

app_dart/lib/src/model/ci_yaml/ci_yaml.dart
```diff
@@ -11,19 +11,130 @@ import 'package:github/github.dart';
...
+  }) =>
+      configs[type]!.getFirstPostsubmitTarget(builderName);
+
+  /// List of target names used to filter target from release candidate branches
+  /// that were already removed from main.
+  List<String>? totTargetNames({CiType type = CiType.any}) => configs[type]!.totTargetNames;
+
+  /// List of postsubmit target names used to filter target from release candidate branches
+  /// that were already removed from main.
+
```
yjbanov, line 90: nit: unnecessary empty line

app_dart/lib/src/service/scheduler.dart
```diff
@@ -674,6 +699,8 @@ class Scheduler {
   /// Relevant APIs:
   ///   https://developer.github.com/v3/checks/runs/#check-runs-and-requested-actions
   Future<bool> processCheckRun(cocoon_checks.CheckRunEvent checkRunEvent) async {
+    /// Figure out if we're in fusion or not.
```
yjbanov, line 702: Should this be a TODO? Also, nit: probably not `///`

yjbanov (approved):
Just nitpicks. LGTM.

app_dart/lib/src/model/ci_yaml/ci_yaml.dart
jtmcdole: This is copy-pasta from the old CiYaml and I don't want to make big logical changes at the moment. My understanding is this is specifically for release branch handling - and it looks inside the ci yaml for special bits.

app_dart/lib/src/model/ci_yaml/ci_yaml.dart
jtmcdole: done

app_dart/lib/src/service/scheduler.dart
jtmcdole: done

jtmcdole (commented):
done

app_dart/lib/src/foundation/utils.dart
```diff
@@ -22,6 +22,7 @@ import '../request_handling/exceptions.dart';
 import '../service/logging.dart';
 
 const String kCiYamlPath = '.ci.yaml';
+const String kCiYamlFusionPath = 'engine/src/flutter/$kCiYamlPath';
```
christopherfujino, line 25: should this be `kCiYamlFusionEnginePath`?

app_dart/lib/src/model/ci_yaml/ci_yaml.dart
```diff
@@ -11,19 +11,129 @@ import 'package:github/github.dart';
...
+  /// Post-fusion: framework, etc
+  ///
+  /// NOTE: Required to be present
+  any,
+
+  /// Engine's .ci.yaml file located in the monorepo
+  fusionEngine,
+}
+
+/// Wrapper arround one or more [CiYaml]
```
christopherfujino, line 28: Can you clarify more? Why do we need an abstraction around multiple CiYamls?

Also, should maybe CiYamlInner be named CiYaml, and this be named something else, maybe CiYamlSet?

app_dart/lib/src/model/proto/internal/scheduler.proto
```diff
@@ -27,7 +27,7 @@ message SchedulerConfig {
 }
 
 // A unit of work for infrastructure to run.
-// Next ID: 17
+// Next ID: 18
```
christopherfujino, line 30: Should this not be 19? Also, why do we need this?

christopherfujino (approved):
LGTM

app_dart/lib/src/model/ci_yaml/ci_yaml.dart
jtmcdole: > Can you clarify more

Yes, I'll take a stab

> CiYaml vs CiYamlSet

I originally wanted a smaller diff-set and file changes.  If I change it to CiYamlSet, would folks expect it to actually be a Set<CiYaml>?  I'll give it a stab and see how it looks (at the call site, I'm sure it'll read better).

app_dart/lib/src/model/proto/internal/scheduler.proto
jtmcdole: Ah, it should be. This looks like a relic of ye'olden times.  I'll just back this out and we can have a single PR updating this.

PR #4017, opened 2024-11-08
[mq] do not enqueue already enqueued PRs

yjbanov:
A PR stays open while it's in the merge queue because the MQ can eject it back if it doesn't pass tests, so `PullRequestValidationService` can receive pings that would look like a request to validate and merge the PR (it will be all green, reviewed, and have an `autosubmit` label). However, if a PR is already in the merge queue, there's nothing to do. So short-circuit the validation process, do nothing other than print a log message that it happened.

### DIFF
--- a/auto_submit/lib/service/pull_request_validation_service.dart
+++ b/auto_submit/lib/service/pull_request_validation_service.dart
@@ -142,6 +142,20 @@ class PullRequestValidationService extends ValidationService {
       await githubService.createComment(slug, prNumber, message);
       log.info(message);
     } else {
+      // Remove the autosubmit label post enqueue/merge to avoid infinite loops.
+      // Here's an example of an infinite loop:
+      //
+      // 1. Autosubmit bot is notified that a PR is ready (reviewed, all green, has `autosubmit` label).
+      // 2. Autosubmit bot puts the PR onto the merge queue.
+      // 3. The PR fails some tests in the merge queue.
+      // 4. Github kicks the PR back, removing it fom the merge queue.
+      // 5. GOTO step 1.
+      //
+      // Removing the `autosubmit` label will prevent the autosubmit bot from
+      // repeating the process, until a human looks at the PR, decides that it's
+      // ready again, and manually adds the `autosubmit` label on it.
+      await githubService.removeLabel(slug, prNumber, Config.kAutosubmitLabel);
+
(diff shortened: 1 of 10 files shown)


auto_submit/lib/service/pull_request_validation_service.dart
```diff
@@ -58,6 +58,18 @@ class PullRequestValidationService extends ValidationService {
   }) async {
     final github.RepositorySlug slug = messagePullRequest.base!.repo!.slug();
     final int prNumber = messagePullRequest.number!;
+
+    // TODO(yjbanov): figure out how to determine if the repo is MQ-enabled.
```
jtmcdole, line 62: more specifically: how to tell if a repo+branch is MQ-enabled; some branches will not be (e.g. release branches)

auto_submit/lib/service/pull_request_validation_service.dart
```diff
@@ -130,6 +142,20 @@ class PullRequestValidationService extends ValidationService {
...
+      // Remove the autosubmit label post enqueue/merge to avoid infinite loops.
+      // Here's an example of an infinite loop:
+      //
+      // 1. Autosubmit bot is notified that a PR is ready (reviewed, all green, has `autosubmit` label).
+      // 2. Autosubmit bot puts the PR onto the merge queue.
+      // 3. The PR fails some tests in the merge queue.
+      // 4. Github kicks the PR back, removing it fom the merge queue.
+      // 5. GOTO step 1.
+      //
+      // Removing the `autosubmit` label will prevent the autosubmit bot from
```
jtmcdole, line 154: This sounds perfectly fine to me. Maybe we come up with more labels in the future to better describe / search what was happening ("autosubmit-attempted", "mq-failed previously") but that's MVP+1

jtmcdole (approved):
lgtm

auto_submit/lib/service/pull_request_validation_service.dart
yjbanov: Good point. I'll rephrase to "determine if the PR is MQ-enabled".

auto_submit/lib/service/pull_request_validation_service.dart
yjbanov: The default Github UI actually provides a pretty detailed audit trail. Take a look at this PR that has endured a lot of different situations:

https://github.com/flutter/flaux/pull/42

PR #4018, opened 2024-11-08
CiStaging Firestore model + test

jtmcdole:
Define the interface for "staging" documents in Cocoon. These are documents in firestore that will record different stages in a monorepo such as the engine-build stage (required before we can run tests).

Previously ci.yaml validation was a single stage: "can cocoon schedule all the build tasks"
In a monorepo we will break this up into multiple stages, with tasks filling in the fields of the firebase document.

Take note; this must be done with a transaction read/write. We will be told by github/luic about jobs being finished asynchronously. Two tasks completing at the same time would set the "remaining" work to the same value and break.

### DIFF
--- a/app_dart/lib/src/model/firestore/ci_staging.dart
+++ b/app_dart/lib/src/model/firestore/ci_staging.dart
@@ -1,0 +1,141 @@
+import 'package:cocoon_service/src/service/logging.dart';
+import 'package:github/github.dart';
+
+import '../../service/firestore.dart';
+import 'package:googleapis/firestore/v1.dart' hide Status;
+
+/// Representaion the current work scheduled for a given stage of monorepo check runs.
+///
+/// Staging is required so that engine build artifacts can be made available to any tests that
+/// depend on them.
+///
+/// This document layout is currently:
+///  /projects/flutter-dashboard/databases/cocoon/ciStaging/
+///     document: <owner>_<repo>_<sha>_<stage>
+///       total: int >= 0
+///       remaining: int >= 0
+///       [*fields]: string {scheduled, success, failure, skipped}
+class CiStaging extends Document {
+  /// Firestore collection for the staging documents.
+  static const kCollectionId = 'ciStaging';
+  static const kRemainingField = 'remaining';
+  static const kTotalField = 'total';
+  static const kEngineStage = 'engine';
+  static const kDefaultTaskStatus = 'scheduled';
+
+  /// Returns a firebase documentName used in [fromFirestore].
+  static String documentNameFor({required RepositorySlug slug, required String sha, required String stage}) {
+    // Document names cannot cannot have '/' in the document id.
+    final docId = '${slug.owner}_${slug.name}_${sha}_$stage';
+    return '$kDocumentParent/ciStaging/$docId';
+  }
+
+  /// Lookup [Commit] from Firestore.
+  ///
+  /// Use [documentNameFor] to get the correct [documentName]
+  static Future<CiStaging> fromFirestore({
+    required FirestoreService firestoreService,
+    required String documentName,
+  }) async {
+    final Document document = await firestoreService.getDocument(documentName);
+    return CiStaging.fromDocument(commitDocument: document);
+  }
+
+  /// Create [CiStaging] from a Commit Document.
+  static CiStaging fromDocument({
+    required Document commitDocument,
+  }) {
+    return CiStaging()
+      ..fields = commitDocument.fields!
+      ..name = commitDocument.name!;
+  }
+
+  /// The remaining number of checks in this staging.
+  int? get remaining => int.tryParse(fields![kRemainingField]!.integerValue ?? '');
+
+  /// The total number of checks in this staging.
+  int? get total => int.tryParse(fields![kTotalField]!.integerValue ?? '');
+
+  /// Mark a [checkRun] for a given [stage] with [conclusion].
+  ///
+  /// Returns a [StagingConclusion] record or throws. If the check_run was
+  /// both valid and recorded successfully, the record's `remaining` value
+  /// signals how many more tests are running. Returns the record (valid: false)
+  /// otherwise.
+  static Future<StagingConclusion> markConclusion({
+    required FirestoreService firestoreService,
+    // required String documentNameString,
+    required RepositorySlug slug,
+    required String sha,
+    required String stage,
+    required String checkRun,
+    required String conclusion,
+  }) async {
+    final logCrumb = 'markConclusion(${slug.owner}_${slug.name}_${sha}_$stage, $checkRun, $conclusion)';
+
+    // Marking needs to happen while in a transaction to ensure `remaining` is
+    // updated correctly. For that to happen correctly; we need to as perform a
+    // read of the document in the transaction as well. So start the transaction
+    // first thing.
+    final docRes = await firestoreService.documentResource();
+    final transactionResponse = await docRes.beginTransaction(
+      BeginTransactionRequest(
+        options: TransactionOptions(
+          readWrite: ReadWrite(),
+        ),
+      ),
+      kDatabase,
+    );
+    final transaction = transactionResponse.transaction;
+    if (transaction == null) {
+      log.warning('$logCrumb: transaction was null');
+      throw '$logCrumb: transaction was null';
+    }
+
+    // First: read the fields we want to change.
+    final documentPath = documentNameFor(sha: sha, slug: slug, stage: stage);
+    final doc = await docRes.get(documentPath, mask_fieldPaths: [checkRun, kRemainingField], transaction: transaction);
+
+    // Fields and remaining _must_ be present.
+    if (doc.fields == null ||
+        doc.fields![kRemainingField] == null ||
+        doc.fields![kRemainingField]!.integerValue == null) {
+      log.warning('$logCrumb: missing field for $transaction / ${doc.fields}');
+      throw '$logCrumb missing fields';
+    }
+
+    final remaining = int.parse(doc.fields![kRemainingField]!.integerValue!);
+    final newRemaining = remaining - 1;
+    // We will have check_runs scheduled after the engine was built successfully, so missing the checkRun field
+    // is an OK response to have. All fields should have been written at creation time.
+    if (doc.fields?[checkRun] == null || doc.fields?[checkRun]!.stringValue == null) {
+      log.warning('$logCrumb: $checkRun not present in doc; remaining=$remaining');
+      return (valid: false, remaining: remaining);
+    }
+
+    // Now we can modify the document the change in the conculsion
+    if (kDefaultTaskStatus != doc.fields![checkRun]!.stringValue) {
+      log.info('$logCrumb: setting remaining to $newRemaining and changing ${doc.fields![checkRun]!.stringValue}');
+      doc.fields![checkRun] = Value(stringValue: conclusion);
+      doc.fields![kRemainingField] = Value(integerValue: '$newRemaining');
+    } else {
+      log.info("$logCrumb: '$conclusion' already recorded? ${doc.fields![checkRun]!.stringValue}");
+      await docRes.rollback(RollbackRequest(transaction: transaction), kDatabase);
+      return (valid: false, remaining: remaining);
+    }
+
+    // Commit this write firebase and if no one else was writing at the same time, return success
+    final CommitRequest commitRequest =
+        CommitRequest(transaction: transaction, writes: documentsToWrites([doc], exists: true));
+
+    final response = await docRes.commit(commitRequest, kDatabase);
+    log.info('$logCrumb: results = ${response.writeResults?.map((e) => e.toJson())}');
+
+    return (valid: true, remaining: newRemaining);
+  }
+}
+
+/// Results from attempting to mark a staging task as completed.
+///
+/// See: [CiStaging.markConclusion]
+typedef StagingConclusion = ({bool valid, int remaining});
--- a/app_dart/lib/src/model/firestore/ci_staging.dart
+++ b/app_dart/lib/src/model/firestore/ci_staging.dart
@@ -38,16 +38,16 @@ class CiStaging extends Document {
     required String documentName,
   }) async {
     final Document document = await firestoreService.getDocument(documentName);
-    return CiStaging.fromDocument(commitDocument: document);
+    return CiStaging.fromDocument(ciStagingDocument: document);
   }
 
   /// Create [CiStaging] from a Commit Document.
   static CiStaging fromDocument({
-    required Document commitDocument,
+    required Document ciStagingDocument,
   }) {
     return CiStaging()
-      ..fields = commitDocument.fields!
-      ..name = commitDocument.name!;
+      ..fields = ciStagingDocument.fields!
+      ..name = ciStagingDocument.name!;
   }
 
   /// The remaining number of checks in this staging.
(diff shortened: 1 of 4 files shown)


app_dart/lib/src/model/firestore/ci_staging.dart
```diff
@@ -0,0 +1,157 @@
...
+
+      // Fields and remaining _must_ be present.
+      if (doc.fields == null ||
+          doc.fields![kRemainingField] == null ||
+          doc.fields![kRemainingField]!.integerValue == null) {
+        log.warning('$logCrumb: missing field for $transaction / ${doc.fields}');
+        throw '$logCrumb missing fields';
+      }
+
+      final remaining = int.parse(doc.fields![kRemainingField]!.integerValue!);
```
yjbanov, line 117: Can the above be shortened to this?

```dart
final remaining = int.tryParse(doc.fields?[kRemainingField]?.integerValue ?? '');
if (remaining == null) {
  log.warning('$logCrumb: missing field for $transaction / ${doc.fields}');
  throw '$logCrumb missing fields';
}
```

app_dart/lib/src/model/firestore/ci_staging.dart
```diff
@@ -0,0 +1,157 @@
...
+      newRemaining = remaining - 1;
+      // We will have check_runs scheduled after the engine was built successfully, so missing the checkRun field
+      // is an OK response to have. All fields should have been written at creation time.
+      if (doc.fields?[checkRun] == null || doc.fields?[checkRun]!.stringValue == null) {
+        log.info('$logCrumb: $checkRun not present in doc; remaining=$remaining');
+        await docRes.rollback(RollbackRequest(transaction: transaction), kDatabase);
+        return (valid: false, remaining: remaining);
+      }
+
+      // Now we can modify the document the change in the conculsion
```
yjbanov, line 127: Something's wrong with this sentence. I think it's trying to say something like "Now we can set the conclusion in the document".

app_dart/lib/src/model/firestore/ci_staging.dart
```diff
@@ -0,0 +1,157 @@
...
+      // Fields and remaining _must_ be present.
+      if (doc.fields == null ||
+          doc.fields![kRemainingField] == null ||
+          doc.fields![kRemainingField]!.integerValue == null) {
+        log.warning('$logCrumb: missing field for $transaction / ${doc.fields}');
+        throw '$logCrumb missing fields';
+      }
+
+      final remaining = int.parse(doc.fields![kRemainingField]!.integerValue!);
+      newRemaining = remaining - 1;
```
yjbanov, line 118: Should we have protections against this going negative?

app_dart/lib/src/model/firestore/ci_staging.dart
```diff
@@ -0,0 +1,157 @@
...
+
+    var newRemaining = -1;
+
+    late final Document doc;
+
+    // transaction block
+    try {
+      // First: read the fields we want to change.
+      final documentPath = documentNameFor(sha: sha, slug: slug, stage: stage);
+      doc = await docRes.get(documentPath, mask_fieldPaths: [kRemainingField, checkRun], transaction: transaction);
```
yjbanov, line 107: Maybe extract `doc.fields` and null-check it right away, so you can avoid all the `!` and `?` below? Or can the object backing `doc.fields` change as things evolve?

app_dart/lib/src/model/firestore/ci_staging.dart
```diff
@@ -0,0 +1,157 @@
...
+      }
+
+      // Now we can modify the document the change in the conculsion
+      if (kDefaultTaskStatus == doc.fields![checkRun]!.stringValue) {
+        log.info('$logCrumb: setting remaining to $newRemaining and changing ${doc.fields![checkRun]!.stringValue}');
+        doc.fields![checkRun] = Value(stringValue: conclusion);
+        doc.fields![kRemainingField] = Value(integerValue: '$newRemaining');
+      } else {
+        log.warning("$logCrumb: '$conclusion' already recorded? ${doc.fields![checkRun]!.stringValue}");
+        throw "$logCrumb: '$conclusion' already recorded? ${doc.fields![checkRun]!.stringValue}";
```
yjbanov, line 134: The message can be deduped.

app_dart/lib/src/model/firestore/ci_staging.dart
```diff
@@ -0,0 +1,157 @@
...
+      if (doc.fields?[checkRun] == null || doc.fields?[checkRun]!.stringValue == null) {
+        log.info('$logCrumb: $checkRun not present in doc; remaining=$remaining');
+        await docRes.rollback(RollbackRequest(transaction: transaction), kDatabase);
+        return (valid: false, remaining: remaining);
+      }
+
+      // Now we can modify the document the change in the conculsion
+      if (kDefaultTaskStatus == doc.fields![checkRun]!.stringValue) {
+        log.info('$logCrumb: setting remaining to $newRemaining and changing ${doc.fields![checkRun]!.stringValue}');
+        doc.fields![checkRun] = Value(stringValue: conclusion);
```
yjbanov, line 130: A little bit uneasy about storing check run as just conclusion string. It will be difficult to add more data later. Maybe wrap it in a field? Something like:

```dart
doc.fields![checkRun]!['conclusion'] = Value(stringValue: conclusion);
```

Then in the future it would trivial to add something like:

```dart
doc.fields![checkRun]!['finishDateTime'] = ...;
doc.fields![checkRun]!['isRequired'] = ...;
...
```

app_dart/lib/src/model/firestore/ci_staging.dart
```diff
@@ -0,0 +1,157 @@
...
+        throw "$logCrumb: '$conclusion' already recorded? ${doc.fields![checkRun]!.stringValue}";
+      }
+    } catch (e) {
+      await docRes.rollback(RollbackRequest(transaction: transaction), kDatabase);
+      rethrow;
+    }
+
+    // Commit this write firebase and if no one else was writing at the same time, return success.
+    // If this commit fails, that means someone else modified firestore and the caller should try again.
+    // We do not need to rollback the transaction; firebase documentation says a failed commit takes care of that.
```
yjbanov, line 143: Love the "We do not need to rollback...` explainer

app_dart/lib/src/model/firestore/ci_staging.dart
```diff
@@ -0,0 +1,157 @@
...
+    final CommitRequest commitRequest =
```
yjbanov, line 144: Remove the inferred type for consistency?

app_dart/test/model/firestore/ci_staging_test.dart
```diff
@@ -0,0 +1,326 @@
...
+
+import '../../src/utilities/mocks.dart';
+
+void main() {
+  group('CiStaging', () {
+    late MockFirestoreService firestoreService;
+
+    setUp(() {
+      firestoreService = MockFirestoreService();
+    });
```
yjbanov, line 20: nit: add one empty line

app_dart/test/model/firestore/ci_staging_test.dart
```diff
@@ -0,0 +1,326 @@
...
+                    t.writes!.first.update!.fields!.length == 2 &&
+                    t.writes!.first.update!.fields!['Linux build_test']!.stringValue == 'mulligan' &&
+                    t.writes!.first.update!.fields![CiStaging.kRemainingField]!.integerValue == '0';
+              }),
+            ),
+            kDatabase,
+          ),
+        ).called(1);
+        verifyNever(docRes.rollback(any, kDatabase));
+      });
```
yjbanov, line 272: nit: add one empty line

yjbanov (approved):
LGTM w/ a bunch of nits

app_dart/lib/src/model/firestore/ci_staging.dart
jtmcdole: What you're asking for is a map - I'd have to check to see how firestore actually send that down as data - and also how the filters work with that.

app_dart/lib/src/model/firestore/ci_staging.dart
```diff
@@ -0,0 +1,157 @@
...
+  /// Create [CiStaging] from a Commit Document.
+  static CiStaging fromDocument({
+    required Document ciStagingDocument,
+  }) {
+    return CiStaging()
+      ..fields = ciStagingDocument.fields!
+      ..name = ciStagingDocument.name!;
+  }
+
+  /// The remaining number of checks in this staging.
```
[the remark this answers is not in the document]
jtmcdole: I have changed it to non-null. It was a pattern in the other model files, but it's an error in my case.  I believe reading a document that doesn't exist is an error in Firebase so I'll double check that.

app_dart/lib/src/model/firestore/ci_staging.dart
```diff
@@ -0,0 +1,157 @@
...
+  int? get remaining => int.tryParse(fields![kRemainingField]!.integerValue ?? '');
+
+  /// The total number of checks in this staging.
```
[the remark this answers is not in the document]
jtmcdole: dito on non-null now.

app_dart/lib/src/model/firestore/ci_staging.dart
```diff
@@ -0,0 +1,157 @@
...
+
+  /// Mark a [checkRun] for a given [stage] with [conclusion].
+  ///
+  /// Returns a [StagingConclusion] record or throws. If the check_run was
+  /// both valid and recorded successfully, the record's `remaining` value
+  /// signals how many more tests are running. Returns the record (valid: false)
+  /// otherwise.
+  static Future<StagingConclusion> markConclusion({
+    required FirestoreService firestoreService,
+    // required String documentNameString,
```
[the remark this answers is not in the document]
jtmcdole: Indeed. I wanted to create this string internally.

app_dart/lib/src/model/firestore/ci_staging.dart
```diff
@@ -0,0 +1,157 @@
...
+      BeginTransactionRequest(
+        options: TransactionOptions(
+          readWrite: ReadWrite(),
+        ),
+      ),
+      kDatabase,
+    );
+    final transaction = transactionResponse.transaction;
+    if (transaction == null) {
+      log.warning('$logCrumb: transaction was null');
```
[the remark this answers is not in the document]
jtmcdole: `error` does not exist; switched to `severe`

app_dart/lib/src/model/firestore/ci_staging.dart
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    var newRemaining = -1;
+
+    late final Document doc;
+
+    // transaction block
+    try {
+      // First: read the fields we want to change.
+      final documentPath = documentNameFor(sha: sha, slug: slug, stage: stage);
```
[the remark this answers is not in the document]
jtmcdole: It's called `name` in the `docRes.get()`, so switching to that.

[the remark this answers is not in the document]
jtmcdole: Are you also suggesting the document name be that as well?

flutter_flaux_engine_<sha>?

When I look at the document list, I kinda want:  slug/sha/stage, in case we add more stages later.

Updated the ordering here.

app_dart/lib/src/model/firestore/ci_staging.dart
jtmcdole: I don't think it can / should change.  Not during a transaction.  Updated

app_dart/lib/src/model/firestore/ci_staging.dart
jtmcdole: It can.

app_dart/lib/src/model/firestore/ci_staging.dart
jtmcdole: Added a log / throw above. This would be a bad state.  The calling function should only look for valid == true and remaining == 0.

app_dart/lib/src/model/firestore/ci_staging.dart
jtmcdole: I've changed some things around thanks to this comment.

1) I do want to record the new conclusion so it matches github calls. "valid = false" for the response
2) I only want to update remaining - 1 if the original conclusion is `kDefaultTaskStatus`. valid = true for the response.

I've updated the comment to reflect this.

app_dart/lib/src/model/firestore/ci_staging.dart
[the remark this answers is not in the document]
christopherfujino: did you? :)

app_dart/lib/src/model/firestore/ci_staging.dart
```diff
@@ -0,0 +1,157 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:cocoon_service/src/service/logging.dart';
+import 'package:github/github.dart';
+
+import '../../service/firestore.dart';
+import 'package:googleapis/firestore/v1.dart' hide Status;
+
+/// Representaion the current work scheduled for a given stage of monorepo check runs.
```
christopherfujino, line 11:
```suggestion
/// Representation of the current work scheduled for a given stage of monorepo check runs.
```

app_dart/lib/src/model/firestore/ci_staging.dart
jtmcdole: technically; all of them can. the assumption is that the caller will handle printing the error.

app_dart/lib/src/model/firestore/ci_staging.dart
```diff
@@ -0,0 +1,157 @@
...
+///
+/// Staging is required so that engine build artifacts can be made available to any tests that
```
christopherfujino, line 13: Can you call out explicitly what "staging" means here, as elsewhere it usually means our non-prod environment.

christopherfujino (approved):
LGTM

app_dart/lib/src/model/firestore/ci_staging.dart
jtmcdole: Map value issue: 

```
/// The map keys, represented as UTF-8,
/// must not exceed 1,500 bytes and cannot be empty.
```
  
Assuming that comment is correct for MapValue (firebase maps!), I'm uncomfortable with adding this as a landmine for the future (I added a test and it exploded).  I view staging as "must build" already; but we could have an array later?

app_dart/lib/src/model/firestore/ci_staging.dart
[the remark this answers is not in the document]
yjbanov: It's just a habit of mine. I'm not sure if this can be applied consistently everywhere.

yjbanov (approved):
lgtm

app_dart/lib/src/model/firestore/ci_staging.dart
jtmcdole: Marking as resolved for now. We can make changes in the future; but staging fields are "required" right now.  I'm not sure how / why we'd want non-required fields, but we can make a new document to encode that in the future.

app_dart/lib/src/model/firestore/ci_staging.dart
jtmcdole: /// 'Staging' is the breaking up of the CI tasks such that some are performed before others.
/// This is required so that engine build artifacts can be made available to any tests that
/// depend on them.

PR #4028, opened 2024-11-11
[mq] enable merge queue guard check in all repos

yjbanov:
Enables the "Merge Queue Guard" check in all repos. No impact on contribution process is expected. This check should simply pass or fail in tandem with the "ci.yaml validation" check.

### DIFF
--- a/app_dart/test/service/scheduler_test.dart
+++ b/app_dart/test/service/scheduler_test.dart
@@ -275,7 +275,7 @@ void main() {
         final Commit commit = shaToCommit('1');
         db.values[commit.key] = commit;
 
-        db.onCommit = (List<gcloud_db.Model<dynamic>> inserts, List<gcloud_db.Key<dynamic>> deletes) {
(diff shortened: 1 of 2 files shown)


jtmcdole (approved):

app_dart/test/service/scheduler_test.dart
```diff
@@ -1024,7 +1024,12 @@ targets:
         expect(
           verify(mockGithubChecksUtil.createCheckRun(any, any, any, captureAny, output: captureAnyNamed('output')))
               .captured,
-          <dynamic>[
+          <Object?>[
+            Scheduler.kMergeQueueLockName,
+            const CheckRunOutput(
+              title: Scheduler.kMergeQueueLockName,
+              summary: Scheduler.kMergeQueueLockDescription,
+            ),
```
jtmcdole, line 1032: Nit / clarification: Does the order matter here?

PR #4030, opened 2024-11-14
Update CiStaging to handle flip-flop of test conclusions

jtmcdole:
This is specifically handling the ability of users to re-run tests (e.g. in case of flake or infra failure). We want to make it easier for folks to make progress on a PR by continue to build the engine assets and then running the framework / engine tests.

### DIFF
--- a/app_dart/lib/src/model/firestore/ci_staging.dart
+++ b/app_dart/lib/src/model/firestore/ci_staging.dart
@@ -162,8 +162,8 @@ class CiStaging extends Document {
       // "remaining" should only go down if the previous state was scheduled - this is the first state
       // that is written by the scheduler.
       // "failed_count" can go up or down depending on:
-      //   previous == failure && current == success: +1
-      //   previous != failure && current == failure: -1
+      //   recordedConclusion == failure && conclusion == success: down (-1)
+      //   recordedConclusion != failure && conclusion == failure: up (+1)
       // So if the test existed and either remaining or failed_count is changed; the response is valid.
       if (recordedConclusion == kScheduledValue && conclusion != kScheduledValue) {
         // Guard against going negative and log enough info so we can debug.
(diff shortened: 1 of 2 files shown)


jtmcdole:
I'm breaking up my PRs right now to make this more manageable. While this code isn't used yet - it will be.  Coverage is pretty high, but not 100.

app_dart/lib/src/model/firestore/ci_staging.dart
```diff
@@ -108,49 +121,109 @@ class CiStaging extends Document {
...
+      // GitHub sends us 3 "action" messages for check_runs: created, completed, or rerequested.
+      //   - We are responsible for the "created" messages.
+      //   - The user is responsible for "rerequested"
+      //   - LUCI is responsible for the completed.
+      // Completed messages are either success / failure.
+      // "remaining" should only go down if the previous state was scheduled - this is the first state
+      // that is written by the scheduler.
+      // "failed_count" can go up or down depending on:
+      //   previous == failure && current == success: +1
+      //   previous != failure && current == failure: -1
```
yjbanov, line 166: If "previous" refers to "recordedConclusion" and "current" refers to "conclusion", then shouldn't the +1 and -1 be swapped in the comment above?

app_dart/lib/src/model/firestore/ci_staging.dart
```diff
@@ -108,49 +121,109 @@ class CiStaging extends Document {
...
-        log.warning(
-          "$logCrumb: '$conclusion' already recorded for ${fields[checkRun]!.stringValue}, not updated $remaining",
-        );
-        newRemaining = remaining;
-        fields[checkRun] = Value(stringValue: conclusion);
-        valid = false;
+        // previous "conclusion" is either freshly scheduled or a success.
+        if (recordedConclusion == kSuccessValue) {
+          // The test succeeded, the user requested a rerun (why?), and now it fails.
+          log.info('$logCrumb: conclusion flipped to negative - assuming test was re-run');
```
yjbanov, line 192: How do we know this without checking the value of `conclusion` first?

app_dart/lib/src/model/firestore/ci_staging.dart
```diff
@@ -108,49 +121,109 @@ class CiStaging extends Document {
...
+        } else {
+          // The test is a new failure
+          log.info('$logCrumb: test failed');
```
yjbanov, line 195: `recordedConclusion` can be "scheduled", "success", or "failure", correct? Looking at the `if` branches that lead to this like, it seems `recordedConclusion` can only possibly be in "scheduled" state ("success" and "failure" lead to other branches). If so, how do we conclude that the test is a new failure at this point?

app_dart/lib/src/model/firestore/ci_staging.dart
jtmcdole: yes - I screwed this up when I was renaming variables. Thanks for the catch!

jtmcdole: done

app_dart/lib/src/model/firestore/ci_staging.dart
jtmcdole: good callout. I've removed these because they don't really add much value... instead I've made this block look more like the failure block above.

app_dart/lib/src/model/firestore/ci_staging.dart
jtmcdole: It is only "scheduled" when we create the document, that is not a status message that is sent to use from Github. We only get "failed" and "success" from github because those are the bits used by LUCI -> Cocoon -> GitHub.

As above; I've removed these logs.

jtmcdole:
Biggest change was restructuring those if blocks to make it clearer to the reader.

yjbanov (approved):
lgtm

PR #4032, opened 2024-11-15
remove unnecessary async/await

yjbanov:
The `_createPresubmitScheduleBuild` method does not perform any async work, so it just unnecessarily creates an async gap, and also confuses things by making it look like it does some network/database stuff when it doesn't.


jtmcdole:
>    info - lib/src/service/luci_build_service.dart:477:60 - Uses 'await' on an instance of 'ScheduleBuildRequest', which is not a subtype of 'Future'. Try removing the 'await' or changing the expression. - await_only_futures

yjbanov:
Done!

jtmcdole (approved):

PR #4033, opened 2024-11-16
Add test exemption for engine code under testing/*

cbracken:
The top-level `testing/` directory contains unit test support files. Most of these are test-specific OpenGL/Metal/Vulkan/Software context/surface code.

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/app_dart/lib/src/request_handlers/github/webhook_subscription.dart
+++ b/app_dart/lib/src/request_handlers/github/webhook_subscription.dart
@@ -465,6 +465,7 @@ class GithubWebhookSubscription extends SubscriptionHandler {
         filename.startsWith('${engineBasePath}impeller/playground') ||
         filename.startsWith('${engineBasePath}shell/platform/embedder/tests') ||
         filename.startsWith('${engineBasePath}shell/platform/embedder/fixtures') ||
+        filename.startsWith('${engineBasePath}testing') ||
         filename.startsWith('${engineBasePath}tools/clangd_check');
   }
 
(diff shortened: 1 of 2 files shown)


app_dart/lib/src/request_handlers/github/webhook_subscription.dart
```diff
@@ -465,6 +465,7 @@ class GithubWebhookSubscription extends SubscriptionHandler {
         filename.startsWith('${engineBasePath}impeller/playground') ||
         filename.startsWith('${engineBasePath}shell/platform/embedder/tests') ||
         filename.startsWith('${engineBasePath}shell/platform/embedder/fixtures') ||
+        filename.startsWith('${engineBasePath}testing') ||
```
jtmcdole, line 468: note to self: don't forget to set engineBasePath  later.

jtmcdole (approved):

app_dart/lib/src/request_handlers/github/webhook_subscription.dart
stuartmorgan, line 468: Shouldn't these really all end in `/`? It's probably never going to matter, but if the intent is to match a specific directory rather than all directories that start with that prefix, there should be a `/`.

stuartmorgan (approved):
LGTM

app_dart/lib/src/request_handlers/github/webhook_subscription.dart
cbracken: Yeah was wondering the same when I added this. git only deals in files so we'll never miss anything by adding the /. Didn't want to be the weirdo that's different from the rest in this patch, but I'll send a followup that fixes it.

cbracken: Done in https://github.com/flutter/cocoon/pull/4034

PR #4041, opened 2024-11-19
[mq] remove autosubmit label on dequeue only

yjbanov:
Only remove the autosubmit label if the PR is dequeued.

This is a speculative fix for https://github.com/flutter/flutter/issues/159139

### DIFF
--- a/app_dart/lib/src/request_handlers/github/webhook_subscription.dart
+++ b/app_dart/lib/src/request_handlers/github/webhook_subscription.dart
@@ -297,7 +297,7 @@ class GithubWebhookSubscription extends SubscriptionHandler {
     // Removing the `autosubmit` label will prevent the autosubmit bot from
     // repeating the process, until a human looks at the PR, decides that it's
     // ready again, and manually adds the `autosubmit` label on it.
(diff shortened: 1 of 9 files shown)


app_dart/lib/src/request_handlers/github/webhook_subscription.dart
```diff
@@ -269,6 +273,36 @@ class GithubWebhookSubscription extends SubscriptionHandler {
...
+    // 1. Autosubmit bot is notified that a PR is ready (reviewed, all green, has `autosubmit` label).
+    // 2. Autosubmit bot puts the PR onto the merge queue.
+    // 3. The PR fails some tests in the merge queue.
+    // 4. Github kicks the PR back, removing it fom the merge queue.
+    // 5. GOTO step 1.
+    //
+    // Removing the `autosubmit` label will prevent the autosubmit bot from
+    // repeating the process, until a human looks at the PR, decides that it's
+    // ready again, and manually adds the `autosubmit` label on it.
+    final bool hasAutosubmitLabel = pr.labels!.any((label) => label.name == Config.kAutosubmitLabel);
```
jtmcdole, line 300: Is labels always present (hence the !) or could we use `?.any() ?? false`

jtmcdole (approved):
LGTM

PR #4043, opened 2024-11-23
[mq] delete temp branches when PR is actually merged

yjbanov:
In the merge queue world it is incorrect to assume that a PR is immediately merged when it's submitted. The PR may be in the queue for a while. Deleting the source branch leads to PR being [removed from the queue and closed](https://github.com/flutter/flutter/issues/159365).

This PR moves the logic that deletes the branch from the revert validation hook to the `pull_request/closed` hook.

Fixes https://github.com/flutter/flutter/issues/159365


app_dart/lib/src/request_handlers/github/webhook_subscription.dart
```diff
@@ -355,6 +338,42 @@ class GithubWebhookSubscription extends SubscriptionHandler {
...
+    // don't care about it any more.
+    await scheduler.cancelPreSubmitTargets(
+      pullRequest: pr,
+      reason: (!pr.merged!) ? 'Pull request closed' : 'Pull request merged',
+    );
+
+    if (pr.merged!) {
+      log.fine('Pull request ${pr.number} was closed and merged.');
+
+      // To avoid polluting the repo with temporary revert branches, delete the
```
jtmcdole, line 357: You might want to note why we're doing this here wrt to other actors (bots vs github vs merge queue).

yjbanov: Done.

PR #4052, opened 2024-11-25
Query GitHub Less: pull request / check_run map in Firestore

jtmcdole:
Problem: we're hitting some quota on github, typically around check_runs Future problem: merging repos and having engine staging requires more check_runs mapping to PRs, which is a query, check suite look up, and comparison.

This PR setups a solution where were record a new document in firestore that can be queried at each check_run update when we need access to the pull request field.

```
/projects/flutter-dashboard/databases/cocoon/prCheckRuns/
   document: <firebase unique id>
     pullRequest: json string
     slug: json string
     sha: string
     [*fields]: <string check_run id>: empty string
```

### DIFF
--- a/app_dart/lib/src/model/firestore/pr_check_runs.dart
+++ b/app_dart/lib/src/model/firestore/pr_check_runs.dart
@@ -95,7 +95,7 @@ class PrCheckRuns extends Document {
     }
   }
 
-  /// Retrieve the [PullRequest] for a given [checkRun] or throw an error.  
+  /// Retrieve the [PullRequest] for a given [checkRun] or throw an error.
   static Future<PullRequest> findDocumentFor(
     FirestoreService firestoreService,
     CheckRun checkRun,
--- a/app_dart/lib/src/model/firestore/pr_check_runs.dart
+++ b/app_dart/lib/src/model/firestore/pr_check_runs.dart
@@ -30,9 +30,8 @@ import 'package:googleapis/firestore/v1.dart' hide Status;
 ///       pullRequest: json string
 ///       slug: json string
 ///       sha: string
-///       [*fields]: <string check_run id>: empty string
+///       [*fields]: "test_name": "check_run id"
 class PrCheckRuns extends Document {
-  /// Firestore collection for the staging documents.
   static const kCollectionId = 'prCheckRuns';
   static const kPullRequestField = 'pull_request';
   static const kSlugField = 'slug';
@@ -72,7 +71,7 @@ class PrCheckRuns extends Document {
       kPullRequestField: Value(stringValue: json.encode(pullRequest.toJson())),
       kSlugField: Value(stringValue: json.encode(pullRequest.head!.repo!.slug().toJson())),
       kShaField: Value(stringValue: pullRequest.head!.sha!),
-      for (final run in checks) '${run.id}': Value(stringValue: run.name),
+      for (final run in checks) run.name!: Value(stringValue: '${run.id}'),
     };
 
     final document = Document(fields: fields);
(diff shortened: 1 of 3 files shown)


jtmcdole:
If this looks good; the follow up PR will be creating a document every time to we schedule presubmits or the merge queue creation (basically anywhere we schedule with luci).

app_dart/lib/src/model/firestore/pr_check_runs.dart
```diff
@@ -0,0 +1,111 @@
...
+  ///
+  /// The list of tasks will be written as fields of a document with additional fields for tracking the total
+  /// number of tasks, remaining count. It is required to include [checkRunGuard] as a json encoded [CheckRun] as this
+  /// will be used to unlock any check runs blocking progress.
+  ///
+  /// Returns the created document or throws an error.
+  static Future<Document> initializeDocument({
+    required FirestoreService firestoreService,
+    required PullRequest pullRequest,
+    required List<CheckRun> checks,
```
yjbanov, line 66: We should maybe combine this PR with your previous one that ignores ci yaml and merge guard, and make sure that we do not add those to this list.

app_dart/lib/src/model/firestore/pr_check_runs.dart
```diff
@@ -0,0 +1,111 @@
...
+    }
+  }
+
+  /// Retrieve the [PullRequest] for a given [checkRun] or throw an error.
+  static Future<PullRequest> findDocumentFor(
+    FirestoreService firestoreService,
+    CheckRun checkRun,
+  ) async {
+    final filterMap = <String, Object>{
+      '${checkRun.id} =': checkRun.name!,
```
yjbanov, line 104: I admit, I don't understand how this filter map works. This will look like `{ 'abc123': 'Linux host_debug' }` or something, and I'm not sure how that tells firestore to query for the check run with that ID, and what it has to do with the check run's name. I expected to see something like `{ 'id': 'abc123' }`.

app_dart/lib/src/model/firestore/pr_check_runs.dart
jtmcdole: I don't - because those are created in Scheduler and the luci checks are created in luci.  Even if they were recorded, it doesn't matter since this is just a <check run> -> <pr> helper.

app_dart/lib/src/model/firestore/pr_check_runs.dart
jtmcdole: I have no clue what's going on here other than someone wrote a util on top of firestore's apis. I assume this works since its used elsewhere.  What I'm looking for is specificity - the id must exist as a field and the name must be the same as what it was created with (in case GitHub is lying about the uniqueness of check run ids). If we find a doc with that unique signature; then we're good to go with the PR.

I'm realizing now that firebase might not like this if we're creating single index keys https://firebase.google.com/docs/firestore/query-data/index-overview#single-field_indexes

I should flip this and have it as `"name": "id"` and it would mean the key space is ~150.

app_dart/lib/src/model/firestore/pr_check_runs.dart
```diff
@@ -0,0 +1,111 @@
...
+///
+/// This document layout is currently:
+///  /projects/flutter-dashboard/databases/cocoon/prCheckRuns/
+///     document: <firebase unique id>
+///       pullRequest: json string
+///       slug: json string
+///       sha: string
+///       [*fields]: <string check_run id>: empty string
+class PrCheckRuns extends Document {
+  /// Firestore collection for the staging documents.
```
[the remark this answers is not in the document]
jtmcdole: Its copy pasta!

app_dart/lib/src/model/firestore/pr_check_runs.dart
jtmcdole: To be clear; the util is parsing this map and making some newer filter thing that ultimiatly looks like "{field: fu, op: EQUALS, value: bar}".

```dart
  /// The [filterMap] follows format:
  ///   {
  ///     'fieldInt =': 1,
  ///     'fieldString =': 'string',
  ///     'fieldBool =': true,
  ///   }
  /// Note
  ///   1. the space in the key, which will be used to retrieve the field name and operator.
  ///   2. the value could be any type, like int, string, bool, etc.
  ```

yjbanov (approved):
lgtm

PR #4053, opened 2024-11-26
Don't process check_runs for synthetic blockers

jtmcdole:


app_dart/lib/src/service/scheduler.dart
```diff
@@ -763,6 +767,10 @@ class Scheduler {
 
     if (name == null || sha == null || slug == null || conclusion == null) return true;
 
+    if (kCheckRunsToIgnore.contains(name)) {
+      return true;
+    }
```
yjbanov, line 772: The above two ifs that short-circuit specific situations can move above `await config.createFirestoreService()` and maybe save us a couple RPCs.

yjbanov (approved):

app_dart/lib/src/service/scheduler.dart
jtmcdole: done

PR #4054, opened 2024-11-26
Record & Query PrCheckRuns in luci when scheduling

jtmcdole:
Reduces the need to search for PRs by check suite / check runs in a monorepo

### DIFF
--- a/app_dart/lib/src/service/scheduler.dart
+++ b/app_dart/lib/src/service/scheduler.dart
@@ -8,6 +8,7 @@ import 'dart:typed_data';
 
 import 'package:buildbucket/buildbucket_pb.dart' as bbv2;
 import 'package:cocoon_service/src/model/firestore/ci_staging.dart';
+import 'package:cocoon_service/src/model/firestore/pr_check_runs.dart';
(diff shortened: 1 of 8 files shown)


app_dart/lib/src/service/scheduler.dart
```diff
@@ -888,11 +896,25 @@ class Scheduler {
...
+
+    // We'va failed to find the pull request; try a reverse look it from the check suite.
+    if (pullRequest == null) {
+      final int checkSuiteId = checkRunEvent.checkRun!.checkSuite!.id!;
+      pullRequest = await githubChecksService.findMatchingPullRequest(slug, sha, checkSuiteId);
+    }
+
+    // We cannot make any forward progress. Abandon all hope, Check runs who enter here.
     if (pullRequest == null) {
       throw 'No PR found matching this check_run';
```
yjbanov, line 919: We could maybe leave the check run ID or some other breadcrumb?

yjbanov (approved):
lgtm

app_dart/lib/src/service/scheduler.dart
jtmcdole: Done.

PR #4069, opened 2024-12-03
Tweaks to get Fusion working:

jtmcdole:
1. Don't ask for masked fields - our tests names have spaces and potentially other "regex escaped fields" - instead just get the whole document.
2. Set the 'is_fusion' build property when re/scheduling presubmits

### DIFF
--- a/app_dart/bin/gae_server.dart
+++ b/app_dart/bin/gae_server.dart
@@ -23,12 +23,15 @@ Future<void> main() async {
       accessTokenService: AccessTokenService.defaultProvider(config),
     );
 
+    final fusionTester = FusionTester();
+
     /// LUCI service class to communicate with buildBucket service.
     final LuciBuildService luciBuildService = LuciBuildService(
       config: config,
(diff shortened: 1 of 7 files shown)


app_dart/bin/gae_server.dart
```diff
@@ -23,12 +23,15 @@ Future<void> main() async {
...
 
+    final fusionTester = FusionTester();
+
     /// LUCI service class to communicate with buildBucket service.
     final LuciBuildService luciBuildService = LuciBuildService(
       config: config,
       cache: cache,
       buildBucketClient: buildBucketClient,
       pubsub: const PubSub(),
+      fusionTester: fusionTester,
```
christopherfujino, line 34: any reason not to just invoke the constructor here on this line?

jtmcdole: Its a shared caching service; the Scheduler uses it as well.

christopherfujino (approved):
LGTM

PR #4073, opened 2024-12-04
CiStaging: record first failure

jtmcdole:
Tests didn't catch this because they started with a failure recorded.  Changing the test triggers the failure and then removing the bad code turns it green again.


yjbanov (approved):
lgtm

PR #4075, opened 2024-12-05
Fix bad logCrumb - slugs are from "head"

jtmcdole:
there was nothing wrong in the code other than this errant log.


christopherfujino (approved):
LGTM

PR #4087, opened 2024-12-12
[tech debt] apply and enforce directives_ordering lint

yjbanov:
Add the `directives_ordering` lint and use `dart fix` to reorder imports.


Piinks (approved):
Flutter_LGTM

PR #4101, opened 2024-12-18
Fix cipd exe

jtmcdole:


app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -1109,7 +1109,11 @@ class LuciBuildService {
 
     final processedProperties = target.getProperties().cast<String, Object?>();
     processedProperties['git_branch'] = commit.branch!;
-    final cipdExe = 'refs/heads/${commit.branch}';
+
+    final mqBranch = tryParseGitHubMergeQueueBranch(commit.branch!);
```
jason-simmons, line 1113: Does there need to be a fallback if the branch is not recognized as an MQ branch?

(`notGitHubMergeQueueBranch.branch` is an empty string)

jtmcdole: This function is `_createMergeGroupScheduleBuild` and should only be called in the merge queue. If that changes, far bigger problems will have happened.  I'm happy assuming this will be the case as GitHub's documentation says this is the private branch.

jtmcdole: If you want me to use the repo and get the default branch or just default to 'master', I can do that.

christopherfujino: yeah, since this is specifically for creating a merge group, i think null pointer exception here is prob better

yjbanov: Merge groups are only created for the special branches that _must_ parse as MQ branch. We don't need a fallback.

jason-simmons: No need - it looks like this code path is only used for the MQ

yjbanov (approved):
lgtm

app_dart/lib/src/service/luci_build_service.dart
jtmcdole: It won't be a null pointer, it will just be `refs/heads/`

jason-simmons (approved):

PR #4107, opened 2024-12-19
[monorepo] fix post-submit bringup targets

yjbanov:
Fixes https://github.com/flutter/flutter/issues/160620

### DIFF
--- a/app_dart/lib/src/model/ci_yaml/ci_yaml.dart
+++ b/app_dart/lib/src/model/ci_yaml/ci_yaml.dart
@@ -201,7 +201,7 @@ class CiYaml {
       enabledTargets = filterOutdatedTargets(slug, enabledTargets, totPostsubmitTargetNames);
     }
     // filter if release_build true if current branch is a release candidate branch, or a fusion tree.
-    enabledTargets = _filterReleaseBuildTargets(enabledTargets);
+    enabledTargets = _selectTargetsForBranch(enabledTargets);
     return enabledTargets;
   }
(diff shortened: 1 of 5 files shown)


app_dart/lib/src/model/ci_yaml/ci_yaml.dart
```diff
@@ -222,20 +222,29 @@ class CiYaml {
...
+      // targets because they are built outside of Cocoon. This applies in both
+      // fusion and non-fusion repos.
+      return [...targets.where((target) => !target.isReleaseBuildTarget && !target.isBringupTarget)];
+    } else {
+      // For non-release branches we also want to include bringup targets.
+      // However, there's a difference between fusion and non-fusion repos.
+      if (isFusion) {
+        // Fusion repos run release targets in the MQ, so we only need to
+        // schedule non-release targets in post-submit.
+        return targets.where((target) => !target.isReleaseBuildTarget).toList();
```
jtmcdole, line 241: [...targets.where((target) => !target.isReleaseBuildTarget)];

jtmcdole (approved):
More logging is good; we should start using log names and formatting them; but appengine has its hooks into the logger :'(. It's why I did the logCrumb stuff before.

lgtm

app_dart/lib/src/model/ci_yaml/ci_yaml.dart
yjbanov: Done.

PR #4108, opened 2024-12-19
[tech debt] remove the MQ simulation

yjbanov:
It was there for testing MQ functionality. We don't need it anymore.


jtmcdole (approved):
lgtm

PR #4109, opened 2024-12-19
[monorepo] cancel builds after merge group is destroyed

yjbanov:
When a PR gets kicked out of the queue, Cocoon will cancel any remaining LUCI builds to free up resources.

### DIFF
--- a/app_dart/lib/src/service/luci_build_service.dart
+++ b/app_dart/lib/src/service/luci_build_service.dart
@@ -366,8 +366,6 @@ class LuciBuildService {
   }
 
   /// Cancels all the current builds on [pullRequest] with [reason].
-  ///
-  /// Builds are queried based on the [RepositorySlug] and pull request number.
   Future<void> cancelBuilds({
     required github.PullRequest pullRequest,
     required String reason,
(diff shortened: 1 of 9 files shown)


app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -401,6 +407,44 @@ class LuciBuildService {
...
+  /// Cancels all the current builds against the give [sha] with [reason].
+  ///
+  /// Builds are queried based on the [RepositorySlug] and git SHA.
+  Future<void> cancelBuildsBySha({
+    required String sha,
+    required String reason,
+  }) async {
+    log.info('Attempting to cancel builds (v2) for git SHA $sha');
+
+    final Iterable<bbv2.Build> builds = await getProdBuilds(sha: sha);
```
jtmcdole, line 419: remove LHS types

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -401,6 +407,44 @@ class LuciBuildService {
...
+    log.info('Found ${builds.length} builds.');
```
jtmcdole, line 420: move down to below the isEmpty test

jtmcdole (changes requested):
Write a test for the new method that verifies when "get prod builds" returns 3 (complete, started, scheduled); the response is removing the two.

jtmcdole (approved):

jtmcdole (approved):

PR #4112, opened 2024-12-20
Add an optional third argument for a relative path from the root directory.

eyebrowsoffire:
This is to support .ci.yaml roller in the monorepo. This is to prepare for this recipe change: https://flutter-review.googlesource.com/c/recipes/+/61680

### DIFF
--- a/app_dart/test/generate_jspb_test.dart
+++ b/app_dart/test/generate_jspb_test.dart
@@ -8,8 +8,7 @@ import 'package:test/test.dart';
 
 import '../bin/generate_jspb.dart' as gen_jspb show main, debugHttpClientFactory;
 
-const String fakeCiYaml =
-'''
+const String fakeCiYaml = '''
 enabled_branches:
   - main
 ''';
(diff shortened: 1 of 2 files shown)


app_dart/test/generate_jspb_test.dart
```diff
@@ -0,0 +1,38 @@
...
+  setUp(() {
+    gen_jspb.debugHttpClientFactory = () => mockClient;
+  });
+
+  test('generate with two args', () async {
+    mockClient = MockClient((request) async {
+      expect(request.url, Uri.parse('https://raw.githubusercontent.com/flutter/flutter/flutter/abcde/.ci.yaml'));
+      return http.Response(fakeCiYaml, 200);
+    });
+    await gen_jspb.main(['flutter/flutter', 'abcde']);
```
christopherfujino, line 27:
```suggestion
    await gen_jspb.main(['flutter', 'abcde']);
```

I think the first arg should just be `flutter`, right? Also see the expected request url, which looks wrong.

christopherfujino: this is a nit, since a test.

christopherfujino (approved):
LGTM

app_dart/test/generate_jspb_test.dart
eyebrowsoffire: Good catch. I'll change it.

PR #4113, opened 2024-12-20
When scheduling merge queue builds; tag them as such.

jtmcdole:
The merge queue is expensive; and all tests must pass. We should retry failed tests if they were flaky.

This was very rough because the tests are very much not well.

### DIFF
--- a/app_dart/lib/src/request_handlers/presubmit_luci_subscription.dart
+++ b/app_dart/lib/src/request_handlers/presubmit_luci_subscription.dart
@@ -178,7 +178,9 @@ class PresubmitLuciSubscription extends SubscriptionHandler {
     final Map<String, Object> properties = target.getProperties();
     if (!properties.containsKey('presubmit_max_attempts')) {
       // Give any test in the merge queue another try... its expensive otherwise.
-      return tags.any((pair) => pair.key == LuciBuildService.kMergeQueueKey) ? 3 : 1;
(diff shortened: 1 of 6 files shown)


christopherfujino (approved):
LGTM

app_dart/lib/src/request_handlers/presubmit_luci_subscription.dart
```diff
@@ -175,7 +177,8 @@ class PresubmitLuciSubscription extends SubscriptionHandler {
     final target = targets.singleWhere((element) => element.value.name == builderName);
     final Map<String, Object> properties = target.getProperties();
     if (!properties.containsKey('presubmit_max_attempts')) {
-      return 1;
+      // Give any test in the merge queue another try... its expensive otherwise.
+      return tags.any((pair) => pair.key == LuciBuildService.kMergeQueueKey) ? 3 : 1;
```
zanderso, line 181: Overheard your discussion with Yegor, and given that the test is `<` above, I agree this should probably be 4 instead of 3 in order to get 3 attempts.

yjbanov (approved):
lgtm

app_dart/lib/src/request_handlers/presubmit_luci_subscription.dart
jtmcdole: Will do

PR #4114, opened 2024-12-20
shout less and explain better empty build list

yjbanov:
It's totally OK that when we attempt to cancel merge group builds, there aren't any. Some merge groups may still be waiting to be scheduled when the "destroy" message comes in, and there's nothing wrong with that. Warnings would make it seem like something is wrong. So I'm downgrading this to info.


jtmcdole (approved):
mirco-lgtm

PR #4137, opened 2024-12-23
Fix `.ci.yaml` validation for Fusion (the monorepo)

matanlurey:
Closes https://github.com/flutter/flutter/issues/160707.

Fixes a bug where `DEPS` was accidentally (or purposefully, there were no tests to verify) not checked for framework builds, and then expands on the validation by adding a check for `engine/**` for framework builds, adding a check nobody uses `runIfNot`, which is extremely broken and should not be used, and then lastly refactors the code and adds tests for the next shmuck to want to change this.

### DIFF
--- a/app_dart/lib/src/model/ci_yaml/ci_yaml.dart
+++ b/app_dart/lib/src/model/ci_yaml/ci_yaml.dart
@@ -76,12 +76,10 @@ class CiYamlSet {
   final String branch;
 
   /// Gets all [Target] that run on presubmit for this config.
-  List<Target> presubmitTargets({CiType type = CiType.any}) =>
-      configs[type]!.presubmitTargets;
+  List<Target> presubmitTargets({CiType type = CiType.any}) => configs[type]!.presubmitTargets;
 
   /// Gets all [Target] that run on postsubmit for this config.
-  List<Target> postsubmitTargets({CiType type = CiType.any}) =>
-      configs[type]!.postsubmitTargets;
+  List<Target> postsubmitTargets({CiType type = CiType.any}) => configs[type]!.postsubmitTargets;
 
   /// Gets the first [Target] matching [builderName] or null.
   Target? getFirstPostsubmitTarget(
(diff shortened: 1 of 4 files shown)


app_dart/lib/src/model/ci_yaml/ci_yaml.dart
```diff
@@ -429,6 +421,52 @@ class CiYaml {
...
+    // These serve mostly as documentation.
+    assert(!isFusion, 'Expected to be called in a non-monorepo');
+    assert(target.runIf.isNotEmpty, 'Expected to be called when non-empty');
+
+    // 1. Every target must depend on .ci.yaml at the root of the repo.
+    if (!target.runIf.contains('.ci.yaml')) {
+      exceptions.add('ERROR: ${target.name} is missing `.ci.yaml` in runIf');
+    }
+
+    // 2. The engine repo must additionally depend on DEPS.
```
yjbanov, line 434: Can you please leave a TODO with a link to an issue reminding us to clean this up once we're ready to archive the engine repo?

app_dart/lib/src/model/ci_yaml/ci_yaml.dart
```diff
@@ -429,6 +421,52 @@ class CiYaml {
...
+    assert(slug == Config.flutterSlug, 'Expected to be in the combined repo');
+
+    // 1. Every target must depend on .ci.yaml.
+    // The path depends on whether the framework or engine are being validated;
+    // while both belong in the same (mono)repo, they have separate .ci.yaml
+    // files located in different paths.
+    final ciYamlPath = switch (type) {
+      CiType.fusionEngine => 'engine/src/flutter/.ci.yaml',
+      _ => '.ci.yaml',
+    };
```
yjbanov, line 453: Can these be added automatically by the `.ci.yaml` parser (or by the logic that evaluates `runIf` conditions)? This way we don't have to specify anything in the yaml files. Seems like they add a bunch of noise. Totally fine we want to take this one step at a time though.

yjbanov (approved):

app_dart/lib/src/model/ci_yaml/ci_yaml.dart
matanlurey: Filed as https://github.com/flutter/flutter/issues/160874

app_dart/lib/src/model/ci_yaml/ci_yaml.dart
matanlurey: Filed as https://github.com/flutter/flutter/issues/160874

PR #4141, opened 2024-12-27
Allow showing, and hide `bringup` builds by default.

matanlurey:
Closes https://github.com/flutter/flutter/issues/160697.

I did test that it is possible to show bringup builds (and the query parameters works, for example).

---

There are a _lot_ of ways this code[base] could be better, but intentionally tried to limit the amount of changes in this PR.

(I would be happy to try and improve things in Q1 if we decide this is valuable to do)

/cc @jtmcdole

### DIFF
--- a/dashboard/lib/logic/qualified_task.dart
+++ b/dashboard/lib/logic/qualified_task.dart
@@ -28,14 +28,15 @@ class QualifiedTask {
   QualifiedTask.fromTask(Task task)
       : stage = task.stageName,
         task = task.builderName,
-        pool = task.isFlaky ? 'luci.flutter.staging' : 'luci.flutter.prod';
+        pool = task.isFlaky ? 'luci.flutter.staging' : 'luci.flutter.prod',
+        isBringup = task.isFlaky;
 
   final String? pool;
   final String? stage;
   final String? task;
 
   /// Whether this task originated as a `bringup: true` task.
-  bool get isBringUp => pool == 'luci.flutter.staging';
+  final bool isBringup;
 
   /// Get the URL for the configuration of this task.
   ///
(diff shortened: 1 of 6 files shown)


dashboard/lib/logic/task_grid_filter.dart
```diff
@@ -173,6 +183,10 @@ class TaskGridFilter extends FilterPropertySource {
       return false;
     }
 
+    if ((!_allProperties['showBringUp']?.value) && qualifiedTask.pool == 'luci.flutter.staging') {
```
yjbanov, line 186: Let's add `isFlaky` (or `isBringup`) to `QualifiedTask` instead of using a magic `pool` value as a proxy.

jmagman (approved):
An aside: I wonder why this isn't what the "Staging" filter does (the staging pool runs the post-submit bringup tests), filtering by name only doesn't seem very useful... #1817

dashboard/lib/logic/task_grid_filter.dart
matanlurey: Done!

matanlurey:
> An aside: I wonder why this isn't what the "Staging" filter does (the staging pool runs the post-submit bringup tests), filtering by name only doesn't seem very useful... #1817

Yeah, I have next to no clue. We have a single one in the current dashboard.

Filed an issue: https://github.com/flutter/flutter/issues/160927.

dashboard/lib/logic/qualified_task.dart
```diff
@@ -34,6 +34,9 @@ class QualifiedTask {
   final String? stage;
   final String? task;
 
+  /// Whether this task originated as a `bringup: true` task.
+  bool get isBringUp => pool == 'luci.flutter.staging';
```
yjbanov, line 38: Even better: make it `final bool isBringup` and then do `isBringup = task.isFlaky` in the constructor. No need to use the pool name as a proxy.

yjbanov, line 38: Also a nit: looks like we use `bringup` as one word, so presumably we should capitalize it as `Bringup` instead of `BringUp`.

yjbanov (approved):

matanlurey:
Filed https://github.com/flutter/flutter/issues/160931 to make changing goldens in the future not suck.

PR #4148, opened 2025-01-06
Remove Flaux

jtmcdole:
Flaux has served its purpose. Thanks, Flaux!

### DIFF
--- a/app_dart/test/foundation/utils_test.dart
+++ b/app_dart/test/foundation/utils_test.dart
@@ -534,16 +534,14 @@ void main() {
         },
       );
 
-      for (var request in [goodFlutterRef, goodFlutterRef]) {
-        final tester = FusionTester(httpClientProvider: () => branchHttpClient);
+      final tester = FusionTester(httpClientProvider: () => branchHttpClient);
 
-        final fusion = await tester.isFusionBasedRef(
-          request.slug,
-          request.sha,
-          retryOptions: noRetry,
-        );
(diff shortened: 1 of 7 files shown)


jtmcdole:
🖖🏻

app_dart/test/foundation/utils_test.dart
```diff
@@ -526,22 +525,16 @@ void main() {
...
-                'https://raw.githubusercontent.com/flutter/flaux/abcd/DEPS',
               )) {
             return http.Response('FUSION', HttpStatus.ok);
           }
           return http.Response('test', HttpStatus.ok);
         },
       );
 
-      for (var request in [goodFlutterRef, goodFlauxRef, goodFlutterRef, goodFlauxRef]) {
+      for (var request in [goodFlutterRef, goodFlutterRef]) {
```
christopherfujino, line 537: why does `goodFlutterRef` appear twice? are we explicitly testing we can handle multiple requests?

christopherfujino (approved):
LGTM

app_dart/test/foundation/utils_test.dart
jtmcdole: ... I don't know. I'll remove it though.

jtmcdole: Tested fine with just one, and maybe there was a copy-pasta error for what should have been `[goodFlutterRef, goodFlauxRef]`

PR #4154, opened 2025-01-10
[monorepo] add CI stage document for tests

yjbanov:
- When `fusionEngineBuild` completes, create a firestore document for the `fusionTests` CI stage tasks.
- When receiving a status update for a test check run, update the respective document.

This PR does not yet alter the MQ guard behavior. I'd like to debug the document behavior first.

Progress towards https://github.com/flutter/flutter/issues/159898

### DIFF
--- a/app_dart/lib/src/model/firestore/ci_staging.dart
+++ b/app_dart/lib/src/model/firestore/ci_staging.dart
@@ -153,7 +153,11 @@ class CiStaging extends Document {
         log.info('$logCrumb: $checkRun not present in doc for $transaction / $doc');
         await docRes.rollback(RollbackRequest(transaction: transaction), kDatabase);
         return StagingConclusion(
-            result: StagingConclusionResult.missing, remaining: remaining, checkRunGuard: null, failed: failed,);
+          result: StagingConclusionResult.missing,
+          remaining: remaining,
+          checkRunGuard: null,
+          failed: failed,
+        );
       }
 
       // GitHub sends us 3 "action" messages for check_runs: created, completed, or rerequested.
@@ -209,7 +213,11 @@ class CiStaging extends Document {
         log.info('$logCrumb: staging document not found for $transaction');
         await docRes.rollback(RollbackRequest(transaction: transaction), kDatabase);
         return StagingConclusion(
-            result: StagingConclusionResult.failed, remaining: -1, checkRunGuard: null, failed: failed,);
+          result: StagingConclusionResult.failed,
+          remaining: -1,
+          checkRunGuard: null,
+          failed: failed,
+        );
       }
       // All other errors should bubble up and be retried.
       await docRes.rollback(RollbackRequest(transaction: transaction), kDatabase);
--- a/app_dart/lib/src/model/firestore/ci_staging.dart
+++ b/app_dart/lib/src/model/firestore/ci_staging.dart
@@ -308,6 +308,8 @@ enum CiStage implements Comparable<CiStage> {
   String toString() => name;
 }
 
+/// Explains what happened when attempting to mark the conclusion of a check run
+/// using [CiStaging.markConclusion].
 enum StagingConclusionResult {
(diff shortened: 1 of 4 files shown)


app_dart/lib/src/model/firestore/ci_staging.dart
```diff
@@ -293,37 +308,52 @@ enum CiStage implements Comparable<CiStage> {
   String toString() => name;
 }
 
+enum StagingConclusionResult {
+  /// Check run update recorded successfully in the respective CI stage.
+  ok,
```
jtmcdole, line 315: Nit: maybe add:

/// It is OK to evaluate returned results for stage completeness

app_dart/lib/src/model/firestore/ci_staging.dart
```diff
@@ -293,37 +308,52 @@ enum CiStage implements Comparable<CiStage> {
...
+
+  /// The check run is not in the specified CI stage.
+  ///
+  /// Perhaps it's from a different CI stage.
+  missing,
+
+  /// The check run was found in the specified CI stage but the update failed.
+  failed,
```
jtmcdole, line 323: Higher up:
        log.info('$logCrumb: staging document not found for $transaction');
          result: StagingConclusionResult.failed,

So the check run wasn't found either; but there was an internal error. Do we want that visibility here?

app_dart/lib/src/service/scheduler.dart
```diff
@@ -936,6 +955,28 @@ class Scheduler {
...
+        return _closeSuccessfulEngineBuildStage(checkRunEvent, stagingConclusion, slug, sha, logCrumb);
+      case CiStage.fusionTests:
+        return _closeSuccessfulTestStage(logCrumb);
+    }
+  }
+
+  Future<bool> _closeSuccessfulEngineBuildStage(
+    cocoon_checks.CheckRunEvent checkRunEvent,
+    StagingConclusion stagingConclusion,
+    RepositorySlug slug,
```
jtmcdole, line 976: Do you need stagingConclusion if its "close successful"?

app_dart/lib/src/service/scheduler.dart
```diff
@@ -1091,31 +1153,26 @@ class Scheduler {
...
-      stagingConclusion = await r.retry(
-        () => markCheckRunConclusion(
+
+    // We're doing a transactional update, which could fail if multiple tasks are running at the same time; so retry
+    // a sane amount of times before giving up.
+    const RetryOptions r = RetryOptions(
+      maxAttempts: 3,
+      delayFactor: Duration(seconds: 2),
+    );
+
```
jtmcdole, line 1163: What should we do if this throws an error?

jtmcdole (changes requested):

app_dart/lib/src/service/scheduler.dart
yjbanov: I thought this was the case of "everything is on fire". Normally, we just let the error propagate and handle it at the request handler level. I thought the only reason we had a catch-all here is because we were cautious about this logic when introducing it into flutter/flutter.

app_dart/lib/src/model/firestore/ci_staging.dart
yjbanov: If the document wasn't found, how can we know whether the check run would be found in the doc or not?

jtmcdole: Either that or we have another conclusion.  The comment is "the check run was found" - it wasn't because we don't have the document.

jtmcdole (approved):

app_dart/lib/src/model/firestore/ci_staging.dart
yjbanov: Ah, gotcha. Changed it to `internalError` and updated the docs.

jtmcdole:
Still LGTM

PR #4161, opened 2025-01-16
[monorepo] do not fail merge queue guard on failed fusionTests stage

yjbanov:
Today if a test check run fails (i.e. one of the post engine build check runs), Cocoon also fails the merge queue guard. This is wrong according to the current semantics of the guard, which is only meant to watch engine build check runs.

Fixes: flutter/flutter#161779


jtmcdole (approved):
lgtm

jtmcdole:
Fixes: flutter/flutter#161779

PR #4170, opened 2025-01-22
AllowedList has been broken for 3 years...

jtmcdole:
The `required` parameter causes mirrors to fail to create the class, giving the user a 500 error. Fun!


yjbanov (approved):
lgtm

matanlurey (approved):

PR #4171, opened 2025-01-22
Break glass v2

yjbanov:
Implement the new break glass behavior for the monorepo world.

Fixes https://github.com/flutter/flutter/issues/159898
Fixes https://github.com/flutter/flutter/issues/132811

Below is an updated copy of https://github.com/flutter/flutter/issues/159898#issuecomment-2597209435:

# Pull request

## Two-state Merge Queue Guard

The guard can only be in two states:

- **Pending** (yellow): this is the initial state, and the guard stays in this state for as long as the pull request is not allowed to enter the merge queue.
- **Success** (green): the guard enters this state when the infra decides that the PR is allowed to enter the merge queue.

The state can only go from pending to success. There are no other state transitions.

## Deciding when the guard goes green

Going from pending to green is the only transition we need to worry about. This is how we decide it:

- **Normal case**: the normal workflow for landing a PR is for all the checks to go green. Once this happens, Cocoon closes the guard (by making it green). This will allow the `autosubmit` label to enqueue the PR, and it will allow the author to press "Merge When Ready".
- **Emergency case**: the a PR must be landed despite failing checks (typically on a red tree status). The author adds the `emergency` label. Cocoon immediately unlocks the guard, ignoring any pending or failed checks. In conjunction with the `autosubmit` label, Cocoon will also automatically enqueue the PR after all tests pass even if the tree status is red. With an `emergency` Cocoon will also let the PR to jump the queue. However, if the PR must be landed right away, the author can use the "Merge When Ready" button manually as soon as Cocoon unlocks the merge guard.

## Explainer

The above system makes everything simpler. There are fewer states (just pending and success), fewer transitions (just pending => success), and fewer situations to consider (normal and emergency). We don't need to do anything special w.r.t. the guard for the purposes of retrying failed flaky tests. It simply stays pending while the author is doing whatever is necessary to make the PR green enough to land: push fixes, retry check runs, rebase, get approvals, etc.

Importantly, it covers all situations we need to handle in PR management.

### Why not have a third "failed" state?

The guard's job is not to tell you whether any tests are failing. You can already tell which tests are failing by looking at the status of respective check runs (this may change in the future, but when that time comes we'll find a new visual cue). The guard's job is only to tell you whether your PR is allowed to be enqueued. Permission to proceed never "fails". It's either granted (green), or not yet granted (pending). Once granted the PR is typically enqueued or landed immediately, so performing any other state transitions after green is mostly meaningless. Green is the final state of the guard, that's all.

The other reason for keeping the guard in two states is for simpler state management. Once green, the PR can immediately be enqueued. There's no transition from pending to failed, from failed to pending, or from failed to green. We can remove some of the existing Cocoon code around this, and we don't have to add anything new to support normal and emergency PR landing.

# Merge group

There are two possible outcomes for a merge group:

- **Land**: all check runs are green => Cocoon completes the merge guard as success and GitHub lands the merge group.
- **Fail**: some check runs failed => Cocoon fails the merge guard and GitHub pulls the merge group (and the corresponding PR) out of the queue.

### DIFF
--- a/app_dart/lib/src/service/config.dart
+++ b/app_dart/lib/src/service/config.dart
@@ -32,8 +32,8 @@ const String kDefaultBranchName = 'master';
 class Config {
   Config(this._db, this._cache);
 
-  /// When present on a pull request, instructs Cocoon to land it automatically
-  /// as soon as all the required checks pass.
+  /// When present on a pull request, instructs Cocoon to submit it
+  /// automatically as soon as all the required checks pass.
   ///
   /// Keep this in sync with the similar `Config` class in `auto_submit`.
   static const String kAutosubmitLabel = 'autosubmit';
@@ -43,6 +43,25 @@ class Config {
   ///
   /// Keep this in sync with the similar `Config` class in `auto_submit`.
   static const String kEmergencyLabel = 'emergency';
+
+  /// Validates that CI tasks were successfully created from the .ci.yaml file.
+  ///
+  /// If this check fails, it means Cocoon failed to fully populate the list of
+  /// CI checks and the PR/commit should be treated as failing.
+  static const String kCiYamlCheckName = 'ci.yaml validation';
+
+  /// A required check that stays in pending state until a sufficient subset of
+  /// checks pass.
+  ///
+  /// This check is "required", meaning that it must pass before Github will
+  /// allow a PR to land in the merge queue, or a merge group to land on the
+  /// target branch (main or master).
+  ///
+  /// IMPORTANT: the name of this task - "Merge Queue Guard" - must strictly
+  /// match the name of the required check configured in the repo settings.
+  /// Changing the name here or in the settings alone will break the PR
+  /// workflow.
+  static const String kMergeQueueLockName = 'Merge Queue Guard';
 
   final DatastoreDB _db;
 
@@ -157,7 +176,6 @@ class Config {
 
   // GitHub App properties.
   Future<String> get githubPrivateKey => _getSingleValue('githubapp_private_pem');
-  Future<String> get overrideTreeStatusLabel => _getSingleValue('override_tree_status_label');
   Future<String> get githubPublicKey => _getSingleValue('githubapp_public_pem');
   Future<String> get githubAppId => _getSingleValue('githubapp_id');
   Future<Map<String, dynamic>> get githubAppInstallations async {
--- a/app_dart/lib/src/service/config.dart
+++ b/app_dart/lib/src/service/config.dart
@@ -32,10 +32,17 @@ const String kDefaultBranchName = 'master';
 class Config {
   Config(this._db, this._cache);
 
-  /// Labels autosubmit looks for on pull requests
+  /// When present on a pull request, instructs Cocoon to land it automatically
+  /// as soon as all the required checks pass.
   ///
   /// Keep this in sync with the similar `Config` class in `auto_submit`.
   static const String kAutosubmitLabel = 'autosubmit';
+
+  /// When present on a pull request, allows it to land without passing all the
+  /// checks, and jumps the merge queue.
+  ///
+  /// Keep this in sync with the similar `Config` class in `auto_submit`.
+  static const String kEmergencyLabel = 'emergency';
 
   final DatastoreDB _db;
 
--- a/app_dart/lib/src/service/scheduler.dart
+++ b/app_dart/lib/src/service/scheduler.dart
@@ -101,28 +101,9 @@ class Scheduler {
   /// Name of the subcache to store scheduler related values in redis.
   static const String subcacheName = 'scheduler';
 
-  /// Validates that CI tasks were successfully created from the .ci.yaml file.
-  ///
-  /// If this check fails, it means Cocoon failed to fully populate the list of
-  /// CI checks and the PR/commit should be treated as failing.
-  static const String kCiYamlCheckName = 'ci.yaml validation';
-
-  /// A required check that stays in pending state until a sufficient subset of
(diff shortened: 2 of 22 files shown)


app_dart/lib/src/service/config.dart
```diff
@@ -32,11 +32,18 @@ const String kDefaultBranchName = 'master';
 class Config {
   Config(this._db, this._cache);
 
-  /// Labels autosubmit looks for on pull requests
+  /// When present on a pull request, instructs Cocoon to land it automatically
+  /// as soon as all the required checks pass.
```
jtmcdole, line 36: instead of "landing" which is what GitHub does; enqueued? submit pr?

auto_submit/lib/service/config.dart
```diff
@@ -59,11 +59,37 @@ class Config {
...
+  /// allow a PR to land in the merge queue, or a merge group to land on the
+  /// target branch (main or master).
+  ///
+  /// IMPORTANT: the name of this task - "Merge Queue Guard" - must strictly
+  /// match the name of the required check configured in the repo settings.
+  /// Changing the name here or in the settings alone will break the PR
+  /// workflow.
+  static const String kMergeQueueLockName = 'Merge Queue Guard';
+
   /// GitHub check stale threshold.
```
jtmcdole, line 93: Nit; shouldn't these be in the same location as the shared logic PR you did?

auto_submit/lib/service/pull_request_validation_service.dart
```diff
@@ -104,14 +177,92 @@ class PullRequestValidationService extends ValidationService {
     for (MapEntry<String, ValidationResult> result in validationsMap.entries) {
```
jtmcdole, line 177: for (MapEntry(:key, :result) in validationsMap.entries) {
    // now you could use key and result thanks to patterns
}

auto_submit/lib/service/pull_request_validation_service.dart
```diff
@@ -104,14 +177,92 @@ class PullRequestValidationService extends ValidationService {
...
       }
     }
 
+    if (shouldReturn) {
+      log.info('The pr ${slug.fullName}/$prNumber is not eligible for emergency landing.');
+      return false;
+    }
+
+    // If PR has some failures to ignore temporarily do nothing and continue.
+    for (MapEntry<String, ValidationResult> result in validationsMap.entries) {
```
jtmcdole, line 195: Ditto on using the pattern / deconstruction

auto_submit/lib/service/validation_service.dart
```diff
@@ -288,4 +282,14 @@ extension PullRequestExtension on github.PullRequest {
     final slug = base!.repo!.slug();
     return mqEnabledRepos.contains(slug.fullName);
   }
+
+  /// Extracts label names from the `IssueLabel` list [labels].
+  List<String> get labelNames {
+    final labels = this.labels;
+    if (labels == null) {
+      return const <String>[];
+    }
```
jtmcdole, line 291: nit:

```dart
if (labels == null) return const <String>[];
return labels.map<String>((github.IssueLabel labelMap) => labelMap.name).toList();
```

or 

```dart
return labels?.map<String>((github.IssueLabel labelMap) => labelMap.name).toList() ?? <String> [];
```

auto_submit/lib/validations/validation_filter.dart
```diff
@@ -58,6 +60,22 @@ class PullRequestValidationFilter implements ValidationFilter {
...
+
+  final Config config;
+  final RepositoryConfiguration repositoryConfiguration;
+
+  @override
+  Set<Validation> getValidations() {
+    final Set<Validation> validationsToRun = {};
+    validationsToRun.add(Approval(config: config));
+    validationsToRun.add(Mergeable(config: config));
+    return validationsToRun;
```
jtmcdole, line 75: return {
    Approval(config: config),
    Mergeable(config: config),
};

app_dart/lib/src/service/scheduler.dart
```diff
@@ -709,51 +690,46 @@ class Scheduler {
...
+  /// This removes the merge group from the merge queue without landing it. The
+  /// corresponding pull request will have to be fixed and re-enqueued again.
+  Future<void> failGuardForMergeGroup(RepositorySlug slug, String headSha, CheckRun lock) async {
+    log.info('Failing merge group guard for merge group $headSha in $slug');
+    await githubChecksService.githubChecksUtil.updateCheckRun(
+      config,
+      slug,
+      lock,
+      status: CheckRunStatus.completed,
+      conclusion: CheckRunConclusion.failure,
```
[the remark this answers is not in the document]
yjbanov: Added some basics. We can expand more in the future.

auto_submit/lib/service/config.dart
yjbanov: Yep, eventually. I haven't unified the `Config` classes yet.

auto_submit/lib/service/validation_service.dart
yjbanov: The compiler cannot promote `this.labels` to non-nullable type without a local variable. The `foo?...... ?? defaultValue` trick works and is clever, but I'm not sure if it's readable 🤔

jtmcdole (approved):
still lgtm

auto_submit/lib/service/validation_service.dart
jtmcdole: sgtm

yjbanov:
Reason for revert: the new code seems to be landing on red check runs

PR #4172, opened 2025-01-22
Add new gold shard names to Flutter Gold.

matanlurey:
Shard was renamed in https://github.com/flutter/flutter/pull/162020.


Piinks (commented):
Should this have a test?

matanlurey:
To be honest, I don't think so. A test would have not caught this issue, and adding a test would not catch a new issue.

That is, there is sufficient coverage that this list of strings work for the intended purpose, and I don't think there is value to adding additional tests for different strings.

jtmcdole (approved):

Piinks:
Ok. The rest have tests. So if someone alters this list of strings and it gets dropped, I guess it gets dropped. 🤷‍♀️

PR #4173, opened 2025-01-23
fail merge queue guard when failing to schedule merge group

yjbanov:
1. Remove the fake "Merge queue check" that was used for testing flutter/flaux.
2. Fail the "Merge Queue Guard" when failing to schedule LUCI builds for the merge group (instead of failing the fake).


jtmcdole (approved):
LGTM

PR #4176, opened 2025-01-24
Break glass v2 (attempt 2)

yjbanov:
This relands commit b9bd02b1370bb10704ecd904a78a82cd229d0904.

The previous attempt was reverted in https://github.com/flutter/cocoon/pull/4175 because of the suspicion that it was landing PRs with failed checks. However, that suspicion seems to have been wrong. This is what actually happened:

* 12:25PM: @polina-c creates the [PR](https://github.com/flutter/flutter/pull/162106); Cocoon starts testing it and unlocks the Merge Queue Guard after all engine builds are done (old behavior).
* 1:02PM: I deploy the new Cocoon with the new behavior.
* 1:10PM: @polina-c hits the "Merge When Ready" button, and GitHub allows the PR to enter the queue.

GitHub allowed that last action because the guard was already unlocked by the old behavior. It wasn't the new code misbehaving. In fact, this is exactly the problem that the new code is aiming to fix.

### DIFF
--- a/app_dart/lib/src/model/firestore/ci_staging.dart
+++ b/app_dart/lib/src/model/firestore/ci_staging.dart
@@ -88,7 +88,8 @@ class CiStaging extends Document {
     required String checkRun,
     required String conclusion,
   }) async {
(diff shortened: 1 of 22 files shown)


jtmcdole (approved):
lgtm

PR #4192, opened 2025-01-28
Use the correct `flutter_realm` for retries.

matanlurey:
_Let's test in prod!_

### DIFF
--- a/app_dart/lib/src/service/luci_build_service.dart
+++ b/app_dart/lib/src/service/luci_build_service.dart
@@ -595,9 +595,7 @@ class LuciBuildService {
     final isFusion = await fusionTester.isFusionBasedRef(slug, sha);
     if (isFusion) {
       properties['is_fusion'] = 'true';
-      if (properties.containsKey('flutter_prebuilt_engine_version')) {
-        properties['flutter_realm'] = '';
-      } else {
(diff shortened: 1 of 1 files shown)


app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -595,9 +595,13 @@ class LuciBuildService {
     final isFusion = await fusionTester.isFusionBasedRef(slug, sha);
     if (isFusion) {
       properties['is_fusion'] = 'true';
-      // When in fusion, we want the recipes to override the engine.realm
-      // if some environment variable (FLUTTER_REALM) is set.
-      properties['flutter_realm'] = 'flutter_archives_v2';
+      if (properties.containsKey('flutter_prebuilt_engine_version')) {
```
jtmcdole, line 598: Maybe:
```
      if (!properties.containsKey('flutter_prebuilt_engine_version')) {
        // When in fusion, we want the recipes to override the engine.realm
        // if some environment variable (FLUTTER_REALM) is set.
        properties['flutter_realm'] = 'flutter_archives_v2';
      }
      ```

jtmcdole (approved):

app_dart/lib/src/service/luci_build_service.dart
matanlurey: Done!

PR #4193, opened 2025-01-28
use head SHA when looking for merge guard

yjbanov:
Fixes https://github.com/flutter/flutter/issues/162288


matanlurey (approved):
Fun! I've made this mistake twice already.

PR #4196, opened 2025-01-28
Add merge queue test

matanlurey:
Implements https://github.com/flutter/flutter/issues/162329 behind a "flag" that is turned on by default.

If a builder has the property `runs_in_merge_queue: "true"`, then schedule it during the merge queue stage.

/cc @yjbanov

### DIFF
--- a/app_dart/bin/gae_server.dart
+++ b/app_dart/bin/gae_server.dart
@@ -54,6 +54,7 @@ Future<void> main() async {
       getFilesChanged: GithubApiGetFilesChanged(config),
       luciBuildService: luciBuildService,
       fusionTester: fusionTester,
+      experimentalOptInGitHubUsernames: {'matanlurey'},
     );
 
     final BranchService branchService = BranchService(
--- a/app_dart/lib/src/service/config.dart
+++ b/app_dart/lib/src/service/config.dart
@@ -214,6 +214,9 @@ class Config {
   /// our API failure when many issues/PRs are created in a short time.
   int get issueAndPRLimit => 2;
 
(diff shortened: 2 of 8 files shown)


christopherfujino (approved):
LGTM

matanlurey:
I am going to merge this as an experiment, and if need be will revert (it should only impact me).

matanlurey:
Update: There was no way to flag it per user, so guess it's getting turned on for everyone.

I'll revert/re-deploy if needed.

matanlurey:
@jtmcdole I need a second review from you.

Specifically, and here is the question I can't answer.

Let's say we took the [`Linux analyze`](https://github.com/flutter/flutter/blob/18077040090a5c2218fa02b9dfe68a7af93b7410/.ci.yaml#L349) target, and ran it during the merge queue. How would this run? To [download the Dart SDK](https://github.com/flutter/flutter/blob/main/bin/internal/update_dart_sdk.sh#L164) we'd need the engine to be built, right, because `FLUTTER_PREBUILT_ENGINE_VERSION` is not set.

I also am not sure I'd want to set it, because now we'd run the risk of using an older Dart SDK than the one being updated in the merge group itself. In other words, this seems like a variant of the problem described in https://github.com/flutter/flutter/issues/161843.

jtmcdole (commented):
We chatted offline; we could hack around this by waiting for the linux_host_test to complete and upload artifacts, but that would be in an odd state for the tool.  We could try to parse out the dart sdk from the DEPs file, but that would also require tuning some start up scripts.

The best thing to do is build a multi-phase build plan and run the test after the artifacts were enabled.

PR #4199, opened 2025-01-29
Test for master branch when applying optimization.

matanlurey:
Unblocks presubmits such as https://github.com/flutter/flutter/pull/162383 without hacks required.

I want to fix this logic properly, but will do that in a follow-up.

### DIFF
--- a/app_dart/lib/src/service/scheduler.dart
+++ b/app_dart/lib/src/service/scheduler.dart
@@ -436,7 +436,7 @@ class Scheduler {
         // Feature request: skip engine builds and tests in monorepo if the PR only contains framework related
         // files.
         // NOTE: This creates an empty staging doc for the engine builds as staging is handled on check_run completion
-        //       events from GitHub. Engine Tests are also skipped, and the base.ref is passed to LUCI to use prod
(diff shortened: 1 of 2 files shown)


christopherfujino (approved):
ship it

reidbaker (approved):

app_dart/lib/src/service/scheduler.dart
```diff
@@ -508,7 +508,11 @@ class Scheduler {
 
   /// Checks if the framework only path is enabled
   static bool _isFrameworkOnlyAllowed(PullRequest pullRequest) {
-    // An allow-list or flag checking could be done here.
+    // TODO(matanlurey): Debug further why this doesn't work on release branches.
+    if (pullRequest.base?.ref case final branch? when branch != 'master') {
```
jtmcdole, line 512: This line is 100% unreadable.

if (Config.defaultBranch(commit.slug) != pullRequest.base?.ref) {
  // refused.
}

matanlurey: K will fix

PR #4203, opened 2025-01-30
move emergency label code to app_dart

yjbanov:
`auto_submit` cannot complete the "Merge Queue Guard" check because GitHub requires that the app that's completing a check run is the same as the one that created it. Because `app_dart` creates the check it must be the one completing it.


app_dart/lib/src/request_handlers/github/webhook_subscription.dart
```diff
@@ -840,3 +870,122 @@ class GithubWebhookSubscription extends SubscriptionHandler {
...
   }
 }
+
+typedef PullRequestLabelProcessorProvider = PullRequestLabelProcessor Function({
+  required Config config,
+  required GithubService githubService,
+  required PullRequest pullRequest,
+});
+
+class PullRequestLabelProcessor {
```
matanlurey, line 880: I realize this class is intended to be package-private, but it would be nice to hide (make private) the methods and members private (`_private`) so that it's more clear what is intended to be used and what isn't (like, I wouldn't want to be calling `processEmergencyLabel()` instead of `processLabels()`)

app_dart/lib/src/request_handlers/github/webhook_subscription.dart
```diff
@@ -840,3 +870,122 @@ class GithubWebhookSubscription extends SubscriptionHandler {
...
+      logInfo('no emergency label; moving on.');
+    }
+  }
+
+  Future<void> processEmergencyLabel() async {
+    final hasApprovals = await computeTeamMemberApproval();
+
+    if (!hasApprovals) {
+      // Do not apply emergency label without an approval from a team member.
+      logInfo('No team member approval yet.');
```
matanlurey, line 918: It would be nice if we removed the label and added a message why we removed it.

I understand that might not be feasible though, but I can definitely see this causing "I added the label but don't see it working @ YEGOR HELP.

app_dart/test/request_handlers/github/webhook_subscription_test.dart
```diff
@@ -2587,6 +2598,255 @@ void foo() {
         await tester.post(webhook);
       });
     });
+
```
matanlurey, line 2601: Add a test that if:

- emergency is added
- there are no approvals
- nothing happens

... but then if an aprpoval happens (and it's a flutter-hacker)

the emergency label is processed as expected (we talked about this offline)

app_dart/lib/src/service/github_service.dart
```diff
@@ -3,8 +3,9 @@
 // found in the LICENSE file.
 
 import 'dart:convert';
-import 'dart:math';
+import 'dart:math' as math;
```
matanlurey, line 6: YES

app_dart/lib/src/service/github_service.dart
```diff
@@ -159,6 +160,64 @@ class GithubService {
     );
   }
 
+  Future<List<PullRequestReview>> listPullRequestReviews(RepositorySlug slug, int prNumber) async {
+    return github.pullRequests.listReviews(slug, prNumber).toList();
+  }
+
+  /// Retrieves check runs with the ref.
+  Future<List<CheckRun>> getCheckRuns(
```
matanlurey, line 168: I'm confused on this and the APIs added below - didn't we already have methods for doing this we were already using?

matanlurey (commented):
Overall LG, a few questions/comments, as well as a request for a workflow (add -> approve -> still works) explained in the test.

app_dart/lib/src/service/github_service.dart
matanlurey: Got it, that makes more sense. Thank you!

matanlurey (approved):
LGTM.

app_dart/lib/src/request_handlers/github/webhook_subscription.dart
```diff
@@ -182,6 +183,7 @@ class GithubWebhookSubscription extends SubscriptionHandler {
         break;
       case 'labeled':
         log.info('$crumb: PR labels = [${pr.labels?.map((label) => '"${label.name}"').join(', ')}]');
+        await _processLabels(pullRequestEvent.pullRequest!);
```
jtmcdole, line 186: the pull_request "labeled" event will have "emergency" as the `pullRequestEvent.label.name` field. So you could skip "give me the whole pull request"

jtmcdole (approved):

app_dart/lib/src/request_handlers/github/webhook_subscription.dart
yjbanov: Yep. `pullRequestEvent.pullRequest.labels` also has the label, so I rewrote it to just use the PR object we're already passing.

PR #4218, opened 2025-02-07
Use GraphQL to get the commit timestamp and parent sha 

jtmcdole:
Use graphql to figure out the actual commit time + parent sha. 

Towards flutter/flutter#162830


app_dart/lib/src/service/scheduler.dart
```diff
@@ -143,12 +146,70 @@ class Scheduler {
...
+
+    // In non-merge queue cases, the `mergedAt` time for closed pull request was
+    // enough to correctly identify the order of linear history.
+    //
+    // However, in a merge queue, commits can get held up by slower running
+    // tests on commits ahead in the queue. When this occurs, all commits waiting
+    // are merged at the exact same time. This will confuse the dashboard and
+    // backfill scheduler.
+    //
+    // Solution in both cases: use graphql to query the time `committedDate`.
```
yjbanov, line 165: If multiple commits are held up in the queue then landed simultaneously with identical timestamps, could we have multiple PRs enter the queue simultaneously with the same `committedDate`? Say your queue limit is 5, and GitHub lands a batch of 3 commit on master. There are now 3 free slots in the queue. GitHub will now go ahead an add 3 PRs to the queue simultaneously, creating commits for them at the same time, with the same `commttedDate` (according to the interwebs git commit date precision is 1 second, which is way too coarse-grained).

matanlurey (changes requested):
+1 @yjbanov's comments; I think_a_ problem is we cannot depend on the committed date _or_ `mergedAt`. We need to know the order in which the PRs _physically_ landed in the queue, and use those as a tie-breaker for `mergedAt` collisions.

jtmcdole:
I believe you are wrong about committed date for the reasons:

1. Each PR that is added (actively) to the merge queue gets a new magic-branch at head (gh-readonly-queue).
2. In order for N+1 commits to be added to the queue, they must wait for the first one to be created and pushed to.

From the GitHub [documentation on merge queues](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#triggering-merge-group-checks-with-third-party-ci-providers):

> With third-party CI providers, you will need to update your CI configuration to run when a branch that begins with the special prefix gh-readonly-queue/{base_branch} is pushed to. These are the temporary branches that are created on your behalf by a merge queue and **contain a different sha from the pull request**.

I posit it takes more than 1 second to make a new sha and push it to a branch, and so the committed times will be different. We could look back at all the commits for the last year and see if any had the exact same times if you like.

That said; this PR also adds the parent sha which can be used to disambiguate if needed.

matanlurey:
> I posit it takes more than 1 second to make a new sha and push it to a branch, and so the committed times will be different. We could look back at all the commits for the last year and see if any had the exact same times if you like.
> 
> That said; this PR also adds the parent sha which can be used to disambiguate if needed.

This seems reasonable to me, but if spot-checking a few recent commits (post-merge queue) is easy enough I'd prefer we do that.

yjbanov:
@jtmcdole 

According to `git log | grep "Date: " | sort | uniq -c | grep '2 Date:'`, you are right. We do have some duplicate timestamps, but they are all from 2023 and older. Nothing since we introduced the MQ.

yjbanov (approved):
lgtm

matanlurey (approved):

jtmcdole:
> git log | grep "Date: " |

git log shows the author date. for instance, 78e5d7995cdb16e0b8a3eb12bb62ff325e8c601d : 

```
AuthorDate: Mon Feb 10 18:39:06 2025 +0100
CommitDate: Mon Feb 10 17:39:06 2025 +0000
```

Or in the case of the crashing commits on the dashboard last week:

```
AuthorDate: Fri Feb 7 00:37:01 2025 +0530
CommitDate: Thu Feb 6 19:07:01 2025 +0000

AuthorDate: Thu Feb 6 10:40:25 2025 -0800
CommitDate: Thu Feb 6 18:40:25 2025 +0000
```

Instead; if you want to just get 2024+2025:

```shell
git log '--pretty=%ct' | sort -nr | uniq -c | grep -E " [^1] 17"

   2 1738788338
   2 1738693926
   2 1738627119
   2 1738349888
   2 1738291906
   2 1738116290
   2 1736812144
   3 1736482883
   2 1736482814
   2 1736273168
   2 1736272565
   2 1736193072
   2 1734378540
   2 1733944605
   2 1733862765
   2 1728860846
   2 1723850234
   2 1722357848
   2 1719513844
   2 1719278429
   2 1714596714
   2 1713910250
   2 1712091894
   2 1711588855
   2 1710282871
   2 1710198827
   2 1708503001
   2 1705603680
   2 1702512692
```

jtmcdole:
Using that, I can see two hashes that conflict during the merge queue.

5ce5251e400889e17ada92d816d0437e5003ba41
0781d2efc859b68f0c0843464c62c427d77d9f2f

CommitDate: Wed Feb 5 20:45:38 2025 +0000
CommitDate: Wed Feb 5 20:45:38 2025 +0000

So I'm wrong and I need to figure out why

matanlurey:
> So I'm wrong and I need to figure out why

We love you anyway (particularly for double checking).

matanlurey (changes requested):
no-LGTM (mostly as a reminder to John this needs work).

jtmcdole:
Looking at graphql; the mergedAt times for the two sha's mentioned above are different - but that supports my original statement that these were not blocked by them.

So; I can change this PR To just have the base sha and we can either:

* Every 15 seconds someone wants to see the status; do a graphql for the sha's and then look them up in datastore
* Or: query like we do now, sort by time and disambiguate as needed.

The dashboard sort order also needs to follow this.  Thoughts?

app_dart/lib/src/service/scheduler.dart
```diff
@@ -143,12 +146,67 @@ class Scheduler {
...
+        parents(first: 1) {  # Get the first parent (usually the only one)
+          nodes {
+              oid
+          }
+        }
+      }
+    }
+  }
+}''';
+    final result = await graphql.query(
```
matanlurey, line 188: Not worth blocking this PR, but this does introduce an additional place where we can hit a burst quota limit and grind everything to a halt. We need a strategy to deal with this, I'll file a bug.

matanlurey (approved):

app_dart/lib/src/service/scheduler.dart
jtmcdole: I believe GraphQL falls in another [API bucket](https://docs.github.com/en/graphql/overview/rate-limits-and-node-limits-for-the-graphql-api#calculating-points-for-the-secondary-rate-limit): 

> No more than 900 points per minute are allowed for REST API endpoints, and no more than 2,000 points per minute are allowed for the GraphQL API endpoint

jtmcdole:
Reason for revert: failing to find the mergedAtSha (500 server error) does not retry and fails to schedule post submit

PR #4242, opened 2025-02-22
Fail unscheduled CheckRuns

jtmcdole:
When LUCI complains there isn't a builder, instead of leaving dangling check runs - fail them.

Fixes flutter/flutter#163869

### DIFF
--- a/app_dart/lib/src/request_handlers/scheduler/scheduler_request_subscription.dart
+++ b/app_dart/lib/src/request_handlers/scheduler/scheduler_request_subscription.dart
@@ -7,6 +7,8 @@ import 'dart:convert';
 import 'package:buildbucket/buildbucket_pb.dart' as bbv2;
 import 'package:cocoon_server/logging.dart';
 import 'package:cocoon_service/src/request_handling/subscription_handler.dart';
+import 'package:collection/collection.dart';
+import 'package:github/github.dart';
 import 'package:meta/meta.dart';
 import 'package:retry/retry.dart';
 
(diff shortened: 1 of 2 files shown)


app_dart/lib/src/request_handlers/scheduler/scheduler_request_subscription.dart
```diff
@@ -59,38 +61,90 @@ class SchedulerRequestSubscription extends SubscriptionHandler {
...
       await retryOptions.retry(
         () async {
-          final List<bbv2.BatchRequest_Request> requestsToRetry = await _sendBatchRequest(batchRequest);
+          final requestsToRetry = await _sendBatchRequest(batchRequest);
 
           // Make a copy of the requests that are passed in as if simply access the list
           // we make changes for all instances.
-          final List<bbv2.BatchRequest_Request> requestListCopy = [];
-          requestListCopy.addAll(requestsToRetry);
+          final requestListCopy = [...requestsToRetry.retries];
```
matanlurey, line 75: It's not clear what `requestListCopy` does here, I think we could just use `requestsToRetry.retries` on line 77 and 81 (unless something mutates it between those two, but that is terrifying)

app_dart/lib/src/request_handlers/scheduler/scheduler_request_subscription.dart
```diff
@@ -59,38 +61,90 @@ class SchedulerRequestSubscription extends SubscriptionHandler {
...
+        checkRun: checkRun,
+        status: CheckRunStatus.completed,
+        conclusion: CheckRunConclusion.failure,
+        output: CheckRunOutput(
+          title: 'Failed to schedule build',
+          summary: '''
+Failed to schedule `$builder`:
+
+```
+${errors.isEmpty ? 'unknown' : errors}
```
matanlurey, line 132: errors.join('\n') or something might read cleaner

app_dart/lib/src/request_handlers/scheduler/scheduler_request_subscription.dart
```diff
@@ -108,16 +162,19 @@ class SchedulerRequestSubscription extends SubscriptionHandler {
 
     for (bbv2.BatchResponse_Response batchResponseResponse in response.responses) {
       if (batchResponseResponse.hasScheduleBuild()) {
-        retry.removeWhere((element) => batchResponseResponse.scheduleBuild.builder == element.scheduleBuild.builder);
+        retry.remove(
```
matanlurey, line 165: Just checking: It's intentional you changed this from "remove all builders with a matching builder, to the first one"? Is it worth doing `.singleWhere` to bail if we misunderstood something? (Could also be an assert, but won't catch things in prod)

app_dart/lib/src/request_handlers/scheduler/scheduler_request_subscription.dart
```diff
@@ -108,16 +162,19 @@ class SchedulerRequestSubscription extends SubscriptionHandler {
...
+          retry.firstWhere((element) => batchResponseResponse.scheduleBuild.builder == element.scheduleBuild.builder),
+        );
       } else {
         log.warning('Response does not have schedule build: $batchResponseResponse');
+        errors.add('$batchResponseResponse');
```
matanlurey, line 170: Are the toString()'s of these objects readable?

jtmcdole: Define readable... They are strings. The errors I've seen are 'builder does not exist' which is certainly helpful.

matanlurey: Ah ok, I just meant they aren't like `Instance of Foo`. Sg.

app_dart/lib/src/request_handlers/scheduler/scheduler_request_subscription.dart
jtmcdole: This is just compressing the previous code's `addAll()`. I can just back it out if you want; but `_sendBatchRequest`  does mutate `batchRequest.requests`, which is done in a loop - and this does get called multiple times.. e.g.

1. Schedule 5 tests
2. 3 schedule 2 fail - return 2
3. loop: schedule 2 tests

Honestly; having written the above; my head hurts. The proto that is mutated would just be passed right back in after having the elements removed. 

I test on Monday and see what is really going on.

app_dart/lib/src/request_handlers/scheduler/scheduler_request_subscription.dart
jtmcdole: Fun fact, builder names (`Linux A`) must be unique across **ALL** ci yamls in all repositories (the builders on LUCI are setup this way). So functionally it has to be identical.

I will change it back though - I was originally caching the firstWhere and using it for something else.

app_dart/lib/src/request_handlers/scheduler/scheduler_request_subscription.dart
jtmcdole: It already outputs with a newline (see tests) - but having section headers might be nice.  Will do on Monday.

app_dart/lib/src/request_handlers/scheduler/scheduler_request_subscription.dart
jtmcdole: I actually changed this to the function creating its own list rather than mutating the containing list. Take another look?

app_dart/lib/src/request_handlers/scheduler/scheduler_request_subscription.dart
jtmcdole: I actually changed this to the function creating its own list rather than mutating the containing list. Take another look?

matanlurey (approved):

PR #4262, opened 2025-02-25
Fail checkruns created when retrying (if the need fail'n)

jtmcdole:
towards flutter/flutter#164152


matanlurey (approved):

jtmcdole:
Maybe I don't do this and instead try to schedule a new build. No sense in trying to create a checkrun, only to fail it right away.

jtmcdole:
Dropping for #4264

PR #4264, opened 2025-02-26
Fix retrying builds from GitHub

jtmcdole:
1. Removing the `reschedulePresubmitBuildUsingCheckRunEvent` function, which is largely a copy of scheduleTryBuilds with some lookup.
2. Call scheduleTryBuilds after looking up the PR by sha
3. Fix up tests.


matanlurey (approved):
Beautiful!

jtmcdole:
resolve merge conflict (which wasn't over the same lines?)

PR #4265, opened 2025-02-26
Deobsfucate what a `BuildTag` is by using concrete types (part 1).

matanlurey:
Towards https://github.com/flutter/flutter/issues/164242.

After this PR merges, I'll do a follow-up that changes `luci_build_service.dart`, but didn't want to stomp on your changes.

### DIFF
--- a/app_dart/lib/src/service/luci_build_service/build_tags.dart
+++ b/app_dart/lib/src/service/luci_build_service/build_tags.dart
@@ -1,0 +1,364 @@
+// Copyright 2020 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:buildbucket/buildbucket_pb.dart' as bbv2;
+import 'package:collection/collection.dart';
+import 'package:meta/meta.dart';
+
+/// Valid tags for [bbv2.ScheduleBuildRequest.tags].
+///
+/// Tags are indexed arbitrary string key-value pairs, defined by the user that
+/// has scheduled the build. This class exists in order to ensure we don't
+/// "fat finger" the wrong name, and know exactly what tags we are (and aren't)
+/// sending, statically.
+///
+/// See go/buildbucket#concepts for more details.
+@immutable
+sealed class BuildTag {
+  /// Parses and recognizes expected [BuildTag]s from their string-pair equivalent.
+  factory BuildTag.from(bbv2.StringPair pair) {
+    switch (pair.key) {
+      case UserAgentBuildTag._keyName:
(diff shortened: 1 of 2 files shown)


app_dart/lib/src/service/luci_build_service/build_tags.dart
```diff
@@ -0,0 +1,364 @@
...
+/// has scheduled the build. This class exists in order to ensure we don't
+/// "fat finger" the wrong name, and know exactly what tags we are (and aren't)
+/// sending, statically.
+///
+/// See go/buildbucket#concepts for more details.
+@immutable
+sealed class BuildTag {
+  /// Parses and recognizes expected [BuildTag]s from their string-pair equivalent.
+  factory BuildTag.from(bbv2.StringPair pair) {
+    switch (pair.key) {
```
jtmcdole, line 21:
```dart
    final bbv2.StringPair(:key, :value) = pair;
```
and then just use key/value?

app_dart/lib/src/service/luci_build_service/build_tags.dart
```diff
@@ -0,0 +1,364 @@
...
+
+  /// Which repository in `flutter.googlesource.com/mirrors` this commit is for.
+  final String slugName;
+
+  @override
+  String get _value => 'commit/gitiles/flutter.googlesource.com/mirrors/$slugName/+/$commitSha';
+}
+
+/// A link back to the GitHub PR for this build.
+final class GitHubPullRequestBuildTag extends BuildTag {
```
jtmcdole, line 220: Instead of having overrides for every _key/_value pair, can we:

```dart
@immutable
sealed class BuildTag {

...
BuildTag(this.key, this.value);

final String key;
final String value;
}
```

Which on one of the harder ones (after its parsed):

```dart
final class GitHubPullRequestBuildTag extends BuildTag {
  GitHubPullRequestBuildTag({
    required this.slugOwner,
    required this.slugName,
    required this.pullRequestNumber,
  }) : super._('github_link', 'https://github.com/$slugOwner/$slugName/pull/$pullRequestNumber');

  /// Which repository in `https://github.com/{owner}`.
  final String slugOwner;

  /// Which repository in `https://github.com/{owner}/{slugName}`.
  final String slugName;

  /// Pull request number.
  final int pullRequestNumber;
}
```

jtmcdole (changes requested):

app_dart/lib/src/service/luci_build_service/build_tags.dart
matanlurey: Done!

jtmcdole (approved):
lgtm

PR #4270, opened 2025-02-27
Mitigate a crash caused by not finding a presubmit target on a checkRun

matanlurey:
Mitigates https://github.com/flutter/flutter/issues/164342.


jtmcdole (approved):

jtmcdole:
presubmit targets might not exist if ToT has them removed.

matanlurey:
Guess we'll find out!

PR #4278, opened 2025-03-01
Finish moving the remaining `bbv2.StringPair` literals to `BuildTag`s.

matanlurey:
Closes https://github.com/flutter/flutter/issues/164242.

### DIFF
--- a/app_dart/lib/src/service/luci_build_service/build_tags.dart
+++ b/app_dart/lib/src/service/luci_build_service/build_tags.dart
@@ -46,16 +46,15 @@ final class BuildTags {
 
   /// Replaces an existing build tag with a matching key with [buildTag].
   ///
-  /// Returns the existing tag that was replaced, or `null` if not replaced.
-  T? replaceIfExisting<T extends BuildTag>(T buildTag) {
+  /// Returns `true` if there was an existing value with a matching key.
+  bool addOrReplace(BuildTag buildTag) {
     final index = _buildTags.indexWhere((e) => e._key == buildTag._key);
     if (index == -1) {
       _buildTags.add(buildTag);
-      return null;
+      return false;
     }
-    final existing = _buildTags[index];
     _buildTags[index] = buildTag;
-    return existing as T;
+    return true;
   }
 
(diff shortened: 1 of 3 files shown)


app_dart/lib/src/service/luci_build_service/build_tags.dart
```diff
@@ -27,11 +27,37 @@ final class BuildTags {
...
+
   /// Adds multiple [buildTags].
   void addAll(Iterable<BuildTag> buildTags) {
     _buildTags.addAll(buildTags);
   }
 
+  /// Replaces an existing build tag with a matching key with [buildTag].
+  ///
+  /// Returns the existing tag that was replaced, or `null` if not replaced.
+  T? replaceIfExisting<T extends BuildTag>(T buildTag) {
```
jtmcdole, line 50: nit: name could be `addOrReplace` - which would match the name above `addIfAbsent`.   Map uses putIfAbsent - but we don't have to agree with them.

app_dart/lib/src/service/luci_build_service/build_tags.dart
```diff
@@ -27,11 +27,37 @@ final class BuildTags {
...
+  }
+
   /// Adds multiple [buildTags].
   void addAll(Iterable<BuildTag> buildTags) {
     _buildTags.addAll(buildTags);
   }
 
+  /// Replaces an existing build tag with a matching key with [buildTag].
+  ///
+  /// Returns the existing tag that was replaced, or `null` if not replaced.
```
jtmcdole, line 49: Why return the old value at all? Just a boolean for "existed" would be good enough.

jtmcdole (approved):
LGTM with nits

app_dart/lib/src/service/luci_build_service/build_tags.dart
matanlurey: Done.

app_dart/lib/src/service/luci_build_service/build_tags.dart
matanlurey: Done.

PR #4286, opened 2025-03-04
Add `POST /api/schedule-postsubmits-for-commit`

matanlurey:
Closes https://github.com/flutter/flutter/issues/164032.

This needs more work that I want to pair with @jtmcdole on, namely:

1. Would `addPullRequest` be where we skip (merged) PRs on a release branch?
    **Yes**. 
1. If we skip, doesn't that mean `flutter-dashboard` would be missing the commit, and there would not be a commit to click?
    **Yes. We need to add the commit to datastore with no tasks.**. 
1. Does skipping mean, "add the commit, but schedule no builds.
    **Yes.**
1. If (3), then this would fail, as it expects the commit to not already exist. What should we do instead?
    **We need to make this function re-entrant, that starts with a 0-task commit.**

### DIFF
--- a/app_dart/lib/src/model/appengine/commit.g.dart
+++ b/app_dart/lib/src/model/appengine/commit.g.dart
@@ -8,7 +8,8 @@ part of 'commit.dart';
 // JsonSerializableGenerator
 // **************************************************************************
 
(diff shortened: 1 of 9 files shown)


matanlurey:
Superseded (mostly) by https://github.com/flutter/cocoon/pull/4328.

Tracking additional work: https://github.com/flutter/flutter/issues/163896.

PR #4289, opened 2025-03-04
Be more explicit about required parameters, `userData`, and correct tests.

matanlurey:
Before:

> Be me
> Pass a `Map<String, Object?>` around
> Check a key, and just `return false` with a log message if it wouldn't work, even though it was documented as required

After:

- Pass around exactly what is needed from `userData`
- Don't silently fail if we require data
- Update the tests that were "benefiting" from this by omitting required data

I don't _imagine_ this is a breaking change, but I'll monitor error logs tomorrow I guess?


jtmcdole (approved):
good cleanup; waiting on the next one that turns `Map<Trust, Me> bro` into a real parsed value.

PR #4291, opened 2025-03-05
Fix a bug in how we fallback to default recipe bundles.

matanlurey:
Closes https://github.com/flutter/flutter/issues/164592.

### DIFF
--- a/app_dart/test/src/service/fake_gerrit_service.dart
+++ b/app_dart/test/src/service/fake_gerrit_service.dart
@@ -24,7 +24,7 @@ class FakeGerritService extends GerritService {
         );
 
   List<String> branchesValue;
-  static const List<String> _defaultBranches = <String>['main'];
+  static const List<String> _defaultBranches = <String>['refs/heads/master'];
 
   List<GerritCommit>? commitsValue;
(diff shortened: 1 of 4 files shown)


app_dart/test/src/service/fake_gerrit_service.dart
```diff
@@ -24,7 +24,7 @@ class FakeGerritService extends GerritService {
         );
 
   List<String> branchesValue;
-  static const List<String> _defaultBranches = <String>['main'];
+  static const List<String> _defaultBranches = <String>['refs/heads/master'];
```
jtmcdole, line 27: I know this is only in the Fake service, but was this suppose to be a git ref?

jtmcdole (changes requested):

app_dart/test/src/service/fake_gerrit_service.dart
matanlurey: Yes, this would always hit the "not found" scenario in a test (if we had checked the logs)

See https://flutter-review.googlesource.com/projects/recipes/branches for example.

jtmcdole (approved):

PR #4292, opened 2025-03-05
Do not treat `no builds were found` as a warning case, we cancel builds aggressively

matanlurey:
I.e. often before triggering new builds

This reduces warning+ level spam, such as [here](https://pantheon.corp.google.com/logs/query;cursorTimestamp=2025-03-05T01:31:50.869Z;duration=PT30H;query=resource.type%3D%22gae_app%22%0Aresource.labels.module_id%3D%22default%22%0Aseverity%3E%3DWARNING%0Atimestamp%3D%222025-03-05T01:31:52.349Z%22%0AinsertId%3D%22kmk1sf5kqd0p%22?project=flutter-dashboard).


jtmcdole (approved):
I hated how it would switch all logs after this to "warning". Thanks!

PR #4303, opened 2025-03-06
Migrate everything related to `scheduler/policy` from datastore to firestore

matanlurey:

This is intended to be a 1:1, that queries Firestore for `queryRecentTasksByName` instead of Datastore.


app_dart/lib/src/service/scheduler.dart
```diff
@@ -224,11 +224,13 @@ class Scheduler {
       var policy = target.schedulerPolicy;
       // Release branches should run every task
       if (Config.defaultBranch(commit.slug) != commit.branch) {
-        policy = GuaranteedPolicy();
+        policy = const GuaranteedPolicy();
       }
       final priority = await policy.triggerPriority(
-        task: task,
-        datastore: datastore,
+        task.name!,
```
jtmcdole, line 230: I had to check out the PR to see which Task this was and that 'Name'->'name' is the same as `Mac_mokey run_debug_test_android`

Can't wait to be done with the migration.

jtmcdole (changes requested):

jtmcdole (changes requested):
One bug to look at; but generally lgtm

jtmcdole (approved):
lgtm

PR #4316, opened 2025-03-08
Stop ignoring argument type and invalid type error types.

matanlurey:
More cleanup post unifying formatting and analysis options.

No behavioral changes, all of the `as ___` casts were implicit before and are explicit now.


jtmcdole (approved):
image

PR #4324, opened 2025-03-11
Copy `../packages` so Docker can resolve it for `app_dart`.

matanlurey:
Closes https://github.com/flutter/flutter/issues/165011.


jtmcdole (changes requested):
Woopsie

jtmcdole (approved):

PR #4329, opened 2025-03-12
Don't treat `.ci.yaml` validation errors as an `InternalServerError`

matanlurey:
Closes https://github.com/flutter/flutter/issues/165018.

This splits the possible result states of `processCheckRuns` into:

- Success (previously: `true`)
- User error (previously: `false`), where we don't log an exception/error but return a `400` to the user
- Internal error (previously: `false`), where we log an exception/stack trace and return a `500` to the user

### DIFF
--- a/app_dart/lib/src/service/scheduler.dart
+++ b/app_dart/lib/src/service/scheduler.dart
@@ -1587,11 +1587,9 @@ $stacktrace
                 checkRunEvent.checkRun!.headSha!,
               );
               if (pullRequest == null) {
-                log.warning(
+                return ProcessCheckRunResult.userError(
                   'Asked to reschedule presubmits for unknown sha/PR: ${checkRunEvent.checkRun!.headSha!}',
                 );
-                // TODO(matanlurey): Follow up with jtmcdole on why this was/is a success case.
-                return const ProcessCheckRunResult.success();
               }
 
               final isFusion = await fusionTester.isFusionBasedRef(slug, sha);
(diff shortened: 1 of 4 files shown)


app_dart/lib/src/service/scheduler.dart
```diff
@@ -1588,7 +1590,8 @@ $stacktrace
                 log.warning(
                   'Asked to reschedule presubmits for unknown sha/PR: ${checkRunEvent.checkRun!.headSha!}',
                 );
-                return true;
+                // TODO(matanlurey): Follow up with jtmcdole on why this was/is a success case.
```
jtmcdole, line 1593: follow up: from first glance; I have no clue. *maybe* related to "ack: please don't retry"

jtmcdole (approved):
Just to double check; we're not going to get repeat pings for some of these failures?

app_dart/lib/src/service/scheduler.dart
matanlurey: Should this be a 4XX (user error) class?

matanlurey:
> Just to double check; we're not going to get repeat pings for some of these failures?

What do you mean?

app_dart/lib/src/service/scheduler.dart
jtmcdole: I think it's safe to make it a user error. WE should _not_ get unknown sha / PRs on cocoon.

jtmcdole:
> > Just to double check; we're not going to get repeat pings for some of these failures?
> 
> What do you mean?

In the handlers, there's some retry-logic on ISEs

```dart
      throw InternalServerError(
        '${pr.mergeCommitSha!} was not found on GoB. Failing so this event can be retried...',
```

jtmcdole (approved):

PR #4333, opened 2025-03-13
Refactor out logic that fetches/assembles a `CiYamlSet` -> `CiYamlFetcher`.

matanlurey:
Makes it much easier to (incrementally) finish https://github.com/flutter/flutter/issues/142951.

In general, this PR does one thing: it takes all of the methods (and logic) buried in `Scheduler` around fetching/assembling a `CiYamlSet` from a commit, and encapsulates in nicely in a `CiYamlFetcher` class with the same contract, but a bit more flexibility (for example, it allows fetching from _both_ datastore _and_ firestore commits, which will make an incremental migration of services and handlers that use the former to the latter much easier).

You'll notice this PR is a tad bigger than _just_ that refactor; the main reason is that by having a better fake implementation (it's much easier to create a good `FakeCiYamlFetcher` than what `FakeScheduler` did before), it exposed a number of tests that actually couldn't have worked before under the prod constraints.

Most examples included relatively simple things like running on commits that weren't enabled as a branch.

I also ended up adding `log`-based debugging, because lots of the errors are hidden by mocks failing in a test; I think we can do this much better (https://github.com/flutter/flutter/issues/164652), but that deserves its own PR and not buried in this already big one.

---

I'd recommend reviewing this PR in 3 batches:

1. `CiYamlFetcher`: Do you like how it works/is tested as a standalone piece of code?
1. The changes to the prod code - does they seem 1:1 with the previous behavior (this is the place the least changed)
1. The changes to the _existing_ tests - do you want any comments or clarification on changes?


app_dart/lib/src/foundation/utils.dart
```diff
@@ -186,7 +186,14 @@ Future<String> githubFileContent(
           ),
       retryIf: (Exception e) => e is HttpException || e is NotFoundException,
     );
-  } catch (e) {
+  } on Exception catch (e) {
+    // A logical error (i.e. something like an ArgumentError) should not be
+    // used as a retry signal. For example, 'TestFailure' (package:test) is an
+    // exception but is not recoverable.
```
jtmcdole, line 192: this was one of my "oh, my passing test were actually failing but the error expect() was swallowed" for me.

app_dart/lib/src/service/scheduler.dart
```diff
@@ -885,12 +787,10 @@ $stackTrace
...
-    late CiYamlSet ciYaml;
-    if (commit.branch == Config.defaultBranch(commit.slug)) {
-      ciYaml = await getCiYaml(commit, validate: true);
-    } else {
-      ciYaml = await getCiYaml(commit);
-    }
+    final ciYaml = await _ciYamlFetcher.getCiYamlByDatastoreCommit(
+      commit,
+      validate: commit.branch == Config.defaultBranch(commit.slug),
+    );
```
jtmcdole, line 793: nit: Are these the only places we call validate: true? Maybe we just move that logic into the getCiYamlByDatastoreCommit()?


I think I see one place where we don't - but that feels like the exception.
```dart
// process checkrun
final ciYaml = await _ciYamlFetcher.getCiYamlByDatastoreCommit(
commit,
);
```

app_dart/test/request_handlers/postsubmit_luci_subscription_test.dart
```diff
@@ -108,23 +120,27 @@ void main() {
...
       scheduler: scheduler,
+      ciYamlFetcher: ciYamlFetcher,
     );
     request = FakeHttpRequest();
-
     tester = SubscriptionTester(request: request);
   });
 
+  tearDown(() {
+    printOnFailure('LOGGER BUFFER:\n${logs.join('\n')}');
```
jtmcdole, line 141: Neat. Why worry about an array and just use a single StringBuffer?

app_dart/test/service/scheduler_test.dart
```diff
@@ -154,6 +154,30 @@ void main() {
     );
   }
 
+  setUp(() {
+    logs = [];
+    log = Logger.detached('scheduler_test');
```
jtmcdole, line 159: Same comment; if we only use this to print when there's an error; why not just StringBuffer() and print later?

jtmcdole (commented):
LGTM w

app_dart/lib/src/service/scheduler.dart
matanlurey: Will revisit in a follow-up PR.

app_dart/test/request_handlers/postsubmit_luci_subscription_test.dart
matanlurey: I think this goes away with https://github.com/flutter/cocoon/pull/4334 assuming we go in that direction.

jtmcdole (approved):

PR #4334, opened 2025-03-14
Add a `packages/cocoon_*` set of packages, including logging, re-organize `cocoon_server`.

matanlurey:
Towards https://github.com/flutter/flutter/issues/164652.

In general though, this is just some repository hygiene. It:

- Moves all non-app packages imported by our apps into `packages/`
- Moves the test-only functionality in `package:cocoon_server` (which imports mockito/test) into `cocoon_server_test`
- Adds a `package:cocoon_common[_test]` package, which includes a full-featured logger interface
- Adds a `BufferedLogger` and matchers for nicely asserting against the state of the logger/log messages

This change has no behavioral impact to any application (i.e. the logger is unused outside of tests).

I want to brainstorm a bit on @jtmcdole (in a future PR, if possible) on:

- **Hierarchical logging**: ?. i.e. child loggers with names, i.e. `log.child('Scheduler').log('I scheduled my pants')`
- **Structured logging**: Supported by [app engine](https://cloud.google.com/appengine/docs/standard/writing-application-logs?tab=python#stdout_stderr), we just don't use it (yet).

The code to directly map `LogSink` to `package:appengine` is [~50 LoC](https://github.com/dart-lang/labs/blob/main/pkgs/appengine/lib/src/server/logging_package_adaptor.dart), so that is trivial.

Have a nice weekend!


matanlurey:
Talked to John, structured logging > hierarchical logging.

packages/cocoon_common/test/buffered_logger_test.dart
```diff
@@ -0,0 +1,52 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:cocoon_common/cocoon_common.dart';
+import 'package:cocoon_common_test/cocoon_common_test.dart';
+import 'package:test/test.dart';
+
+void main() {
+  test('uses the provided DateTime for recordedAt', () {
+    final moment = DateTime.now();
```
jtmcdole, line 11: we sure we don't just want to use Clock?

jtmcdole (approved):

packages/cocoon_common/test/buffered_logger_test.dart
matanlurey: Go clock yourself.

PR #4344, opened 2025-03-17
Use `dart_flutter_team_lints: 3.5.0`.

matanlurey:
And apply ignores/simple fixes where needed.

Closes https://github.com/flutter/cocoon/pull/4340.
Closes https://github.com/flutter/cocoon/pull/4338.
Closes https://github.com/flutter/cocoon/pull/4337.

### DIFF
--- a/dashboard/lib/service/google_authentication.dart
+++ b/dashboard/lib/service/google_authentication.dart
@@ -2,6 +2,8 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'dart:async';
+
 import 'package:flutter/foundation.dart';
 import 'package:flutter/services.dart';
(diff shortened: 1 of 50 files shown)


dashboard/lib/service/google_authentication.dart
```diff
@@ -26,7 +28,7 @@ class GoogleSignInService extends ChangeNotifier {
     });
 
     try {
-      _googleSignIn.signInSilently();
+      unawaited(_googleSignIn.signInSilently());
```
jtmcdole, line 31: wait, why is there a try/catch that doesn't actually catch anything?

jtmcdole (approved):

dashboard/lib/service/google_authentication.dart
matanlurey: That is a good question, for another time.

PR #4347, opened 2025-03-18
Use the new `LogSink` interface (as a shim to `Logger` for now)

matanlurey:
Towards https://github.com/flutter/flutter/issues/164652.

**No behavioral change**, though I did cleanup some calls to provide error and stack trace when I did.

Next step will be test by test, add to `initTestLogger()` to each `main()` in tests.

### DIFF
--- a/app_dart/lib/src/foundation/github_checks_util.dart
+++ b/app_dart/lib/src/foundation/github_checks_util.dart
@@ -136,11 +136,13 @@ class GithubChecksUtil {
           conclusion: conclusion,
         );
       },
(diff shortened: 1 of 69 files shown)


Piinks (approved):
LGTM!

As my own sanity check, this is the mapping I observed for the log methods:

| Log | Log2 |
|---|---|
| info | info|
| fine or finest | debug |
| warning | warn or warnJson |
| severe | error|

Thanks! 👍

matanlurey:
TBR @jtmcdole (we talked about this offline, and it's internal only so easy to change after the fact).

PR #4348, opened 2025-03-18
Replace `build_status_response.proto` with `build_status_response.dart`.

matanlurey:


jtmcdole (approved):
If you keep going, we won't have protofu anymore...

A+

matanlurey:
This broke the dashboard tree status, as the response is in proto JSON format (tags instead of fields).

We can fix this, but it requires both a server _and_ client change.

reason for revert: Broke dashboard.

PR #4350, opened 2025-03-18
Remove unused files, add `.gitattributes` to make code review better.

matanlurey:


.gitattributes
```diff
@@ -0,0 +1 @@
+**/*.g.dart linguist-generated=true
```
jtmcdole, line 1: Interesting!
alt text

jtmcdole (approved):

PR #4351, opened 2025-03-18
Use `Config.defaultBranch` for tip-of-tree `.ci.yaml`.

matanlurey:
Closes https://github.com/flutter/flutter/issues/165433.

I believe this restores the [logic we had before](https://github.com/flutter/cocoon/pull/4333/files#diff-44d26c271d034096ac3f0506b7400724ed6dd135aaff3b443ef628880a21457e) that I accidentally broke.


jtmcdole (commented):
Were we going to add the validate: false in this PR?

matanlurey:
It is already validate false I'm just dumb

Matan Lurey | Software Engineer | ***@***.***

jtmcdole (approved):

PR #4356, opened 2025-03-19
Remove unused `get-status-firestore`.

matanlurey:
Invalidated through the work (so far) in https://github.com/flutter/flutter/issues/165025.

Our strategy now is "the API returns a backend-agnostic model", so there is no reason to have this.


jonahwilliams (approved):
LGTM

PR #4360, opened 2025-03-20
Use a `ProgressButton` for run all tasks, because slow.

matanlurey:
Otherwise the button stays enabled for 30s+ which can lead to people spamming the button.


jtmcdole (approved):
can we prevent spamming on the server as well? 
kahn-star-trek

PR #4362, opened 2025-03-21
Run `/api/vacuum-github-commits` at most once at a time

matanlurey:
🚫 Blocked. Needs Redis transaction support (https://github.com/flutter/flutter/issues/165687).

---

Attempt on mitigating https://github.com/flutter/flutter/issues/165612.

If we receive an additional request while one is already running, we can ignore subsequent requests.

It's not clear this solves https://github.com/flutter/flutter/issues/165612, so if it doesn't we can either adjust it (like, "run this at most ___ times in __" as part of the logic, instead of pending tasks), or remove the `SingleExecutionRequestHandler` part and just keep the (small) refactors.

(If we like this, we could make it the base class for the other Cron-type APIs)

### DIFF
--- a/app_dart/lib/src/request_handling/single_execution_request_handler.dart
+++ b/app_dart/lib/src/request_handling/single_execution_request_handler.dart
@@ -1,0 +1,113 @@
+// Copyright 2019 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:io';
+import 'dart:typed_data';
+
+import 'package:cocoon_server/logging.dart';
+import 'package:meta/meta.dart';
+
+import '../service/cache_service.dart';
+import 'api_request_handler.dart';
+import 'body.dart';
+
+/// A class that services an HTTP `GET` request that discards new requests.
+///
+/// [Scheduling jobs with `cron.yaml`][1] reads:
+/// > Only a single instance of a job should run at any time. The Cron service
+/// > is designed to provide "at least once" delivery; that is, if a job is
+/// > scheduled, App Engine sends the job request at least one time. In some
+/// > rare circumstances, it is possible for multiple instances of the same job
+/// > to be requested, therefore, your request handler should be idempotent, and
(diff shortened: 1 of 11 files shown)


app_dart/lib/src/request_handling/single_execution_request_handler.dart
```diff
@@ -0,0 +1,113 @@
...
+
+  const SingleExecutionRequestHandler({
+    required super.config,
+    required super.authenticationProvider,
+    required CacheService cache,
+    @visibleForTesting DateTime Function() now = DateTime.now,
+  }) : _cache = cache,
+       _now = now;
+
+  final CacheService _cache;
```
jtmcdole, line 42: had to double check this was redis backed (it is).

app_dart/lib/src/request_handling/single_execution_request_handler.dart
```diff
@@ -0,0 +1,113 @@
...
+    } else {
+      // Mark the task as running.
+      try {
+        final now = Uint64List(1)..[0] = _now().millisecondsSinceEpoch;
+        await _cache.setWithLocking(
+          subCacheName,
+          cacheKey,
+          now.buffer.asUint8List(),
+          ttl: maxExecutionTime,
+        );
```
jtmcdole, line 100: So setWithLocking isn't transactional? If two calls to this request handler come in at the same time to two different services, they could both pass line 77 and come down to this call which will succeed for both?  If this were firestore, it would be (start transaction; read from key; maybe write key; complete transaction).

jtmcdole (changes requested):

app_dart/lib/src/request_handling/single_execution_request_handler.dart
matanlurey: That is a good point, maybe this doesn't actually WAI.

Redis does support transactions:
https://redis.io/docs/latest/develop/interact/transactions/

... but the Redis client (neat_cache) does not support it.

Moving this back to draft, and I'll file a bug for Redis transactional support.

PR #4363, opened 2025-03-21
Delete unused `/api/update-task-status`.

matanlurey:
Towards https://github.com/flutter/flutter/issues/165618.

/cc @yjbanov Thanks for the suggestion!


jtmcdole (approved):

jtmcdole (commented):
Could we just return 200 for now and log if we see anyone making this API call?

matanlurey:
I'll add something if something breaks :)

PR #4364, opened 2025-03-21
Leave a comment instead of check run.

matanlurey:
Closes https://github.com/flutter/flutter/issues/165662.

The check run was cute but it now means we need to filter out re-run attempts on it.

Given we'll probably move away from check runs (and use a dashboard), this seems fine?


jtmcdole (approved):
https://github.com/user-attachments/assets/b056fb4d-f501-402f-8fe6-2903a22c9147

PR #4365, opened 2025-03-21
Remove remaining datastore keys from the "public" API

matanlurey:

This PR refactors the last occurrence of `__key` leaking into the public API, which was via `/api/public/get-build-status?lastCommitKey=<...>`, by replacing it with a semantically equivalent `/api/public/get-build-status?lastCommitSha=<...>`, and then removing the concept of keys as part of the data model across both the frontend and backend.

To make future migrations easier, also added/refactored a `firestore.Task.fromFirestoreBySha`.


app_dart/lib/src/request_handlers/get_status.dart
```diff
@@ -4,40 +4,40 @@
...
-class GetStatus extends RequestHandler<Body> {
+final class GetStatus extends RequestHandler<Body> {
   const GetStatus({
     required super.config,
     @visibleForTesting
     this.datastoreProvider = DatastoreService.defaultProvider,
     @visibleForTesting
     this.buildStatusProvider = BuildStatusService.defaultProvider,
-  });
+    @visibleForTesting DateTime Function() now = DateTime.now,
```
jtmcdole, line 26: A broken `package:clock/clock.dart` is right more than 2 times per day...

jtmcdole (approved):

PR #4367, opened 2025-03-22
Remove datastore from all read paths related to build/commit/task status

matanlurey:
This PR removes datastore, and related datastore only helper classes (`Stage[Builder]`, `CommitStatus`).

It does so by rewriting `GetStatus` (and associated code, i.e. `BuildStatusService`), which are _read-only_ paths, to exclusively use Firestore, and no longer use (or convert from) Datastore. One hack can to be introduced to get this into a good state, https://github.com/flutter/flutter/issues/165718, which IMHO is well worth the overall delta removed.

Summary of changes:

- Deletes `Stage` and `StageBuilder`, which were unused (legacy) on the frontend, and do not exist in Firestore
- Change `BuildStatusService` to read from Firestore, and _only_ Firestore, removing dependencies on Datastore
- Keep compatibility with most of the RPC JSON model, but remove fields that were not used on the frontend


jonahwilliams (approved):
LGTM

jtmcdole (approved):
wish vscode had an imagediff on goldens :(

PR #4368, opened 2025-03-22
Raise limit to `1000` framework-only files and modern release branches.

matanlurey:
Closes https://github.com/flutter/flutter/issues/161462.

It turns out @jtmcdole  had _accidentally_ fixes fixed by using our `GithubService` wrapper while refactoring, meaning `PaginationHelper` has been used this _entire_ time, and we have been getting the full list of files (since https://github.com/flutter/cocoon/pull/4191 at least), so this can actually be lifted.

In addition, we can just gate the optimization on a specific release branch (to speed up beta/new stables).

### DIFF
--- a/app_dart/lib/src/service/config.dart
+++ b/app_dart/lib/src/service/config.dart
@@ -353,7 +353,7 @@ class Config {
   };
 
   /// The maximum [gh.PullRequest.changedFilesCount] to consider a PR for a "framework-only" CI optimization.
-  int get maxFilesChangedForSkippingEnginePhase => 29;
+  int get maxFilesChangedForSkippingEnginePhase => 1000;
 
   /// Max retries for Luci builder with infra failure.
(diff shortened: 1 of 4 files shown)


jtmcdole (commented):
There still is a API limit right? 1000 files max with pagination?

matanlurey:
Ah:

> Responses include a maximum of 3000 files.

Ok, we can limit to 1000 files just in case.

jtmcdole:
> Ah:
> 
> > Responses include a maximum of 3000 files.
> 
> Ok, we can limit to 1000 files just in case.

Yeah, just got on my PC to check it; looks like that + we could be asking for 100 files `per_page` if we added the query parameter.

If we just get the PullRequest, it should tell us `changed_files`:

```json
    "changed_files": {
      "type": "integer",
      "examples": [
        5
      ]
    }
```

jtmcdole (approved):

PR #4371, opened 2025-03-22
Remove more datastore references across Cocoon

matanlurey:
Additional (small) changes to reduce the dependencies on Datastore:

- Add (and use) `tryFromFirestoreBySha` to check if a commit already exists;
- Move `_commitExistsInDatastore` to use Firestore as the source of truth;
- Change calls from `.getCiYamlByDatastoreCommit` to `.getCiYaml`;


app_dart/lib/src/service/scheduler.dart
```diff
@@ -916,18 +911,17 @@ $s
...
 
-    final ciYaml = await _ciYamlFetcher.getCiYamlByDatastoreCommit(
-      commit,
-      validate: commit.branch == Config.defaultBranch(commit.slug),
+    final branch = pullRequest.base!.ref!;
+    final slug = pullRequest.base!.repo!.slug();
+    final ciYaml = await _ciYamlFetcher.getCiYaml(
+      commitBranch: branch,
+      commitSha: pullRequest.head!.sha!,
+      slug: pullRequest.base!.repo!.slug(),
```
jtmcdole, line 923:
```suggestion
      slug: slug,
```

jtmcdole (changes requested):

jtmcdole (approved):

PR #4372, opened 2025-03-23
Move `SecretManager` to `package/cocoon_server[_test]`.

matanlurey:

Datastore is currently used (incorrectly) to store secrets in `app_dart`. This refactor moves (a refactored version of) the `SecretManager` in `auto_submit` to `cocoon_server[_test]`, in preparation for it to be used in `app_dart` in a follow-up PR (and after moving secrets).


packages/cocoon_server/lib/secret_manager.dart
```diff
@@ -0,0 +1,121 @@
...
+
+  /// Returns the latest secret with the provided [name] as a UTF-16 string.
+  ///
+  /// If the secret could not be found, `null` is returned.
+  Future<String?> tryGetString(String name) async {
+    final bytes = await tryGetBytes(name);
+    if (bytes == null) {
+      return null;
+    }
+    return String.fromCharCodes(bytes);
```
jtmcdole, line 80: Should this be utf8.decode(bytes)? Is the secrets manager actually storing UTF-16?

jtmcdole (approved):

PR #4374, opened 2025-03-23
Exclusively write/read `Github(Build|Gold)Status` to Firestore

matanlurey:
Removes all code related to:

- `GithubBuildStatus` 
- `GithubGoldStatus`

... that was used for Datastore.

If this PR needs to be reverted we will have data loss (Datastore).


app_dart/lib/src/service/scheduler.dart
```diff
@@ -785,15 +783,12 @@ $s
...
-      sha: headSha,
-    );
-
-    final ciYaml = await _ciYamlFetcher.getCiYamlByDatastoreCommit(
-      commit,
-      validate: commit.branch == Config.defaultBranch(commit.slug),
+    final branch = baseRef.substring('refs/heads/'.length);
+    final ciYaml = await _ciYamlFetcher.getCiYaml(
+      slug: slug,
+      commitBranch: baseRef.substring('refs/heads/'.length),
```
jtmcdole, line 789:
```suggestion
      commitBranch: branch,
```

app_dart/lib/src/service/scheduler.dart
```diff
@@ -916,18 +911,17 @@ $s
...
 
-    final ciYaml = await _ciYamlFetcher.getCiYamlByDatastoreCommit(
-      commit,
-      validate: commit.branch == Config.defaultBranch(commit.slug),
+    final branch = pullRequest.base!.ref!;
+    final slug = pullRequest.base!.repo!.slug();
+    final ciYaml = await _ciYamlFetcher.getCiYaml(
+      commitBranch: branch,
+      commitSha: pullRequest.head!.sha!,
+      slug: slug,
```
jtmcdole, line 923:
```suggestion
      slug: slug,
```

jtmcdole (changes requested):
The absolute safest route would be to keep writing to two locations and switch the reads first; then kill the writes.

jtmcdole (approved):
lgtm-beavis

PR #4376, opened 2025-03-23
Ask LUCI for updated prod statuses for in-progress tasks.

matanlurey:
This is based off a discussion between me and @jtmcdole that I thought I filed a bug for, but I guess I did not. Basically, we can sometimes have "In Progress" tasks where LUCI had sent as an updated build status, and for one reason or another, we dropped it on the floor.

This re-uses the existing checks and also asks LUCI if we seem to have dropped something.


jtmcdole (approved):
image

PR #4390, opened 2025-03-24
Collate tasks correctly for `/api/get-status`.

matanlurey:
Closes https://github.com/flutter/flutter/issues/165828.

I assumed `CommitTasksStatus.collateTasks()` worked, but it never did 😢 .

Fixed it, and added some trivial tests to catch the obvious null-check bugs that were crashing.

### DIFF
--- a/app_dart/lib/src/model/firestore/commit_tasks_status.dart
+++ b/app_dart/lib/src/model/firestore/commit_tasks_status.dart
@@ -58,7 +58,7 @@ class CommitTasksStatus {
     }
 
     // Sort build numbers, and return.
-    final fullTasks = fullTasksMap.values.toList();
+    final fullTasks = [...fullTasksMap.values];
     for (final fullTask in fullTasks) {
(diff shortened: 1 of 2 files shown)


app_dart/lib/src/model/firestore/commit_tasks_status.dart
```diff
@@ -30,21 +30,39 @@ class CommitTasksStatus {
...
+      if (task.buildNumber case final buildNumber?) {
+        taskToAddBuildNumber.buildList.add(buildNumber);
+      }
+
+      // And store it for the next loop
+      fullTasksMap[task.taskName!] = taskToAddBuildNumber;
+    }
+
+    // Sort build numbers, and return.
+    final fullTasks = fullTasksMap.values.toList();
```
jtmcdole, line 61:
```suggestion
    final fullTasks = [...fullTasksMap.values];
```

jtmcdole (approved):

PR #4391, opened 2025-03-24
Use updated `index.html`.

matanlurey:
Cribbed from [`flutter_tools/templates/app/web/index.html.tmpl`](https://github.com/flutter/flutter/blob/master/packages/flutter_tools/templates/app/web/index.html.tmpl).


jtmcdole (approved):
lgtm

PR #4396, opened 2025-03-25
Stop using the (external) field names of `Commit` outside of the `Commit` class

matanlurey:


app_dart/lib/src/model/firestore/base.dart
```diff
@@ -0,0 +1,36 @@
...
+    // give beautiful toString() representations for logs and testing, so you'll
+    // let it slide.
+    //
+    // Basically, toJson() does: {
+    //   if (isString) 'stringValue': stringValue,
+    //   if (isDouble) 'doubleValue': doubleValue,
+    // }
+    //
+    // So instead of copying that, we'll just use what they do.
+    return value.toJson().values.first;
```
jtmcdole: Maybe, just in case, on the slightest chance, that the if-chain block returns an empty map...

```suggestion
    return value.toJson().values.firstOrNull;
```

matanlurey:
I made it 10% more pretty.

jtmcdole (approved):

PR #4398, opened 2025-03-26
New Dart interface for Firestore + full in-memory fake

matanlurey:
This PR adds a new, more typed, and easier to mock/fake/assert against `Firestore` API that wraps `g.FirestoreApi`, and provides a full in-memory fake implementation that simulates the real Firestore API, within reason, with tests, and then starts to adopt the API in `app_dart`.

Specifically, `fs.batchWriteDocuments([singleDoc])` has been converted to `fs.insert(singleDoc)`.

I could imagine a better design than this, but a main requirement I kept in mind was everything being incremental, and providing immediate value (i.e., makes testing easier, so we have less of a chance of regressions or fat-finger issues on future refactors).


app_dart/lib/src/model/firestore/base.dart
```diff
@@ -6,14 +6,48 @@ import 'dart:convert';
...
+    required String Function(T) documentName,
+    required T Function(g.Document) fromDocument,
+  }) : _collectionId = collectionId,
+       _documentName = documentName,
+       _fromDocument = fromDocument;
+
+  /// Returns the relative path of [document] within Firestore.
+  String relativePath(T document) {
+    return p.posix.join(_collectionId, _documentName(document));
+  }
```
jtmcdole, line 25:
```suggestion
  /// Returns the relative path of [document] within Firestore.
  String relativePath(T document) {
    return p.posix.join(_collectionId, 'fu', _documentName(document));
  }
```

jtmcdole: This commit suggestion was a demo in a meeting and meaningless to this review.

jtmcdole (approved):

PR #4401, opened 2025-03-26
Fix and add regression tests for `override: ` label handeling

matanlurey:
Closes https://github.com/flutter/flutter/issues/166014.
Works around https://github.com/flutter/flutter/issues/166022.

### DIFF
--- a/app_dart/lib/src/model/firestore/pr_check_runs.dart
+++ b/app_dart/lib/src/model/firestore/pr_check_runs.dart
@@ -47,10 +47,20 @@ class PrCheckRuns extends Document {
   }
 
   /// The json string of the pullrequest belonging to this document.
-  PullRequest get pullRequest => PullRequest.fromJson(
-    json.decode(fields![kPullRequestField]!.stringValue!)
-        as Map<String, Object?>,
-  );
+  PullRequest get pullRequest {
+    final jsonData =
+        jsonDecode(fields![kPullRequestField]!.stringValue!)
+            as Map<String, Object?>;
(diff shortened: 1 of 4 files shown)


app_dart/lib/src/model/firestore/pr_check_runs.dart
```diff
@@ -47,10 +47,20 @@ class PrCheckRuns extends Document {
...
-        as Map<String, Object?>,
-  );
+  PullRequest get pullRequest {
+    final jsonData =
+        jsonDecode(fields![kPullRequestField]!.stringValue!)
+            as Map<String, Object?>;
+    final result = PullRequest.fromJson(jsonData);
+
+    // Workaround for https://github.com/flutter/flutter/issues/166022.
+    if (jsonData['labels'] case final List<Object?> labelData) {
```
jtmcdole, line 57: I don't like the external library. It's not maintained (last commit was 11 months ago and we had to do another workaround). Sigh.

app_dart/lib/src/model/firestore/pr_check_runs.dart
```diff
@@ -47,10 +47,20 @@ class PrCheckRuns extends Document {
...
+      result.labels = [
+        ...labelData.cast<Map<String, Object?>>().map(IssueLabel.fromJson),
```
jtmcdole, line 59: If we wanted to be really defensive about this?:

...labelData.whereType<Map>().cast<Map<String, Object?>>().map(IssueLabel.fromJson)

jtmcdole (approved):

PR #4402, opened 2025-03-27
Logging tweaks

jtmcdole:
- silence noise neat_cache
- convert iterables to lists (avoids '...' data loss)


matanlurey (approved):
Lee-gee-tee-em

PR #4403, opened 2025-03-27
Avoid retrying deleted branches in the Merge Queue handler

matanlurey:
Closes https://github.com/flutter/flutter/issues/166078.

Part 2 of https://github.com/flutter/cocoon/pull/4393.

Before we were returning `5xx` class errors to Cloud PubSub, which means "retry". After this PR, for PRs that are >= 15 minutes old, we check and see "was the corresponding branch deleted", and if it was, we return a `4xx` class error, which will discard the Cloud PubSub message instead.

This reduces the amount of indefinite hops that happens from Cocoon <> Cloud PubSub.


matanlurey:
Added R=@jonahwilliams as trivial review (nice to get in to avoid merge conflicts).

jonahwilliams (approved):
LGTM

PR #4406, opened 2025-03-28
Disable the file-based check for whether a flutter/flutter commit is a Fusion/monorepo commit

jason-simmons:
This check recently had errors due to a GitHub outage (see https://github.com/flutter/flutter/issues/166110).  This caused Cocoon's isFusion cache to be populated with incorrect results for various commits.

The check is now obsolete because there will not be any more release branches that use the legacy pre-monorepo layout.


yjbanov (approved):
lgtm

PR #4417, opened 2025-03-31
Handle duplicate gold patchsets.

jonahwilliams:
Gold can return multiple patchsets for the same commit. We need to sum the untriaged images of all of them, otherwise based on the exact ordering we can miss golden diffs.

### DIFF
--- a/app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
+++ b/app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
@@ -320,12 +320,13 @@ class PushGoldStatusToGithub extends ApiRequestHandler<Body> {
         );
       }
       final patchsets = jsonResponseTriage['patchsets'] as List<dynamic>;
+      // Note: there can be multiple patchsets with the same id, ensure that
+      // all are collected.
       var untriaged = 0;
       for (var i = 0; i < patchsets.length; i++) {
         final patchset = patchsets[i] as Map<String, dynamic>;
         if (patchset['patchset_id'] == pr.head!.sha) {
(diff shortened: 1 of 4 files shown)


matanlurey (approved):

app_dart/lib/src/request_handlers/push_gold_status_to_github.dart
```diff
@@ -320,12 +320,13 @@ class PushGoldStatusToGithub extends ApiRequestHandler<Body> {
...
       final patchsets = jsonResponseTriage['patchsets'] as List<dynamic>;
+      // Note: there can be multiple patchsets with the same id, ensure that
+      // all are collected.
       var untriaged = 0;
       for (var i = 0; i < patchsets.length; i++) {
         final patchset = patchsets[i] as Map<String, dynamic>;
         if (patchset['patchset_id'] == pr.head!.sha) {
-          untriaged = patchset['new_untriaged_images'] as int;
-          break;
+          untriaged += patchset['new_untriaged_images'] as int;
```
Piinks, line 329: I wonder if this might create a situation where,

```
patchset 1 has 5 untriaged images

:make some code changes, fix regression whatever:

patchset 2 has 0 untriaged images

:flutter-gold reports images changes anyways:
```

I don't know if Gold reconciles that from patchset to patchset. I think the report back we get is just, this is what Gold saw from change to change. So I guess maybe we'll just see?

jonahwilliams: but we're filtering down by patchset id = latest sha anyway. So this should only change the results if there are multiple patchsets with the same id and that Id is the head commit

PR #4419, opened 2025-03-31
Move `commit_tasks_status`, explain `documentId` path Firestore models

matanlurey:
Towards https://github.com/flutter/flutter/issues/166229.

This makes it easier to remember where these models are stored in a NoSQL path than looking it up each time on Firebase. Also moved a "model" that was not actually serialized and stored in Firestore out of the `model` directory.


jonahwilliams (approved):
LGTM

PR #4420, opened 2025-04-01
Finish adopting `AppDocument<T>` across Firestore models

matanlurey:
Towards https://github.com/flutter/flutter/issues/165931.

No real functional change, this just has all of the `model/firestore/*.dart` files implement `AppDocument<T>`.

For consistency:
- Add `documentIdFor` for each
- Each has a `.fromDocument` constructor

This is all to reduce the API surface of `FirestoreService` so that all reads/writes flow through an API that can be adequately tested, since `DatastoreService` had an in-memory implementation, and `FirestoreService` _originally_ did not, but one is being added as part of this work.

A follow-up PR will do more cleanup, but this is was enough to justify a PR.


jonahwilliams (approved):
LGTM

matanlurey:
reason for revert: Caused https://github.com/flutter/flutter/issues/166463.

PR #4421, opened 2025-04-01
Set `hierarchicalLoggingEnabled` to disable `neat_cache` logs properly.

matanlurey:
Closes https://github.com/flutter/flutter/issues/166345.

Without this change, the server fails to startup.


loic-sharma (approved):

jtmcdole:
Thank you - I forgot about the global `hierarchicalLoggingEnabled` that came with package logging.

PR #4422, opened 2025-04-01
Replace `Task()..property = ...` with `generateTaskForTest(...)` or similar

matanlurey:

When we remove `task.proto`, it will be a nightmare to change all mutable properties to a constructor call.

Nearly all of these happen in tests, so let's just change them now.


jonahwilliams (approved):
LGTM

PR #4423, opened 2025-04-01
Ignore `<Document>.name` when inserting a new document.

matanlurey:
Fixes a regression due to https://github.com/flutter/cocoon/pull/4398, which is over a week ago and does not revert cleanly.

The underlying issue is https://github.com/flutter/flutter/issues/166229, where we are trying to derive fields from `document.name` (such as `<Task>.attempts`), which in turn causes contortions for insertion logic (we want to _start_ with an existing Task, and then edit the attempt number, and then insert the new task).


jonahwilliams (approved):
LGTM

PR #4425, opened 2025-04-01
Finish removing `lib/model` and `package:proto`

matanlurey:
Closes https://github.com/flutter/flutter/issues/165025.

There is a case for follow-up changes, i.e. moving `rpc_model` to `package:cocoon_common`, and guaranteeing protocol compatibility, but that can be follow-up work (https://github.com/flutter/flutter/issues/166376).


dashboard/lib/service/dev_cocoon.dart
```diff
@@ -417,13 +415,13 @@ class DevelopmentCocoonService implements CocoonService {
...
-      endTimestamp: Int64(
-        commitTimestamp + (index * 1000 * 60) + (index * 1000 * 60),
-      ),
+      createTimestamp: commitTimestamp + index,
+      startTimestamp: commitTimestamp + (index * 1000 * 60),
+      endTimestamp: commitTimestamp + (index * 1000 * 60) + (index * 1000 * 60),
+
       builderName: 'Linux_android $index',
       attempts: attempts,
+      buildNumberList: List.generate(attempts, (i) => i).join(','),
```
jtmcdole, line 424:
```dart
[for (var i = 0; i < attempts; i++) i].join(',')
```
/itsjustaprank
/ducks

dashboard/lib/src/rpc_model/base.dart
```diff
@@ -12,13 +13,20 @@ import 'package:meta/meta.dart';
 @immutable
 @internal
 abstract base mixin class Model {
-  @mustBeOverridden
+  static const _deepEquality = DeepCollectionEquality();
```
jtmcdole, line 16: do you want unordered given a sub-list could be [1,2,3] vs [2,1,3]?

My guess is no and that this is fine because _listEq in `build_status_response.dart` does ListEquality, requiring they be in-order.

dashboard/protofu.yaml
```diff
@@ -1,10 +0,0 @@
-compiler:
```
jtmcdole, line 1: o7 goodby protofu!

jtmcdole (approved):

dashboard/protofu.yaml
matanlurey: 🫡

PR #4426, opened 2025-04-02
Add synthetic fields that are missing from Firestore.

matanlurey:
Towards https://github.com/flutter/flutter/issues/166229.

After this PR, _new_ documents inserted store each field explicitly, instead of relying on the hack-y "store the field delimited by `_` in `name` and parse it".  This allows us to do queries, both for administrative purposes, and for business logic, and, once we backfill, simplify some of our logic (for example, we won't need special keys, and can use the default Firestore unique keys).

(We should probably do the backfill on like, the weekend)


matanlurey:
https://github.com/flutter/cocoon/pull/4433 is a prototype of what a batch migration API could do.

jtmcdole (approved):

PR #4428, opened 2025-04-02
Stop automatically signing out on non-401 errors.

matanlurey:
It's really annoying to get signed out because of a server 500, particularly as a developer testing.


jtmcdole (approved):
lgtm-cat-shake

PR #4435, opened 2025-04-03
Add new (unused) APIs to the shared Firestore layer.

matanlurey:
Towards https://github.com/flutter/flutter/issues/165931.

This is _part of_ https://github.com/flutter/cocoon/pull/4407, which needs to get split and landed in a few smaller PRs (the last couple of days have taught me that we just really can't rely on our own test coverage and the PRs need to be more targeted/smaller, even if it means landing unused code first, and then change the prod code second, like in this PR).


matanlurey:
I want to make some tweaks here, moving back to draft.

PR #4437, opened 2025-04-03
Revert: New Dart interface for Firestore.

matanlurey:
Reverts https://github.com/flutter/cocoon/pull/4398.

This is a _very_ ambitious project, as I've seen in the last couple of days, understanding exactly what Firestore does (and doesn't do) is a job in itself. Reverts back to just using the existing `FirestoreService` API and `googleapis.DocumentsResource`.

We can continue to improve the models in `lib/model/firestore/**`, and add more helper methods to `FirestoreService` to encapsulate more of a [repository-like pattern](https://martinfowler.com/eaaCatalog/repository.html) to make things like testing easier, but we should do that critically as part of https://github.com/flutter/flutter/issues/142951 and not as a side-effort.


jtmcdole:
kahn-star-trek

PR #4438, opened 2025-04-03
Try parsing a `bbv2.BuildV2PubSub` instead of `json[][][]`.

matanlurey:
Towards https://github.com/flutter/flutter/issues/166535.

I'd like to know what the type is instead of guessing, so we'll test it in prod[^1]

[^1]: With log statements

### DIFF
--- a/app_dart/lib/src/request_handlers/dart_internal_subscription.dart
+++ b/app_dart/lib/src/request_handlers/dart_internal_subscription.dart
@@ -8,6 +8,7 @@ import 'package:buildbucket/buildbucket_pb.dart' as bbv2;
 import 'package:cocoon_common/is_dart_internal.dart';
 import 'package:cocoon_server/logging.dart';
 import 'package:fixnum/fixnum.dart';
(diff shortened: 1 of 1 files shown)


app_dart/lib/src/request_handlers/dart_internal_subscription.dart
```diff
@@ -64,6 +63,33 @@ class DartInternalSubscription extends SubscriptionHandler {
...
+    // See https://github.com/flutter/flutter/issues/166535.
+    try {
+      final data = bbv2.BuildsV2PubSub.fromJson(message.data!);
+      var mismatch = false;
+      if (data.build.builder.project != project) {
+        log.debug('[dart_internal_166535] mismatch: project=$project');
+        mismatch = true;
+      }
+      if (data.build.builder.bucket != bucket) {
+        log.debug('[dart_internal_166535] mismatch: bucket=$bucket');
```
jtmcdole, line 76: `mismatch = true` here and below.

jtmcdole (approved):

PR #4443, opened 2025-04-04
Implement fakes of `FirestoreService` except `documentsResource()`

matanlurey:
Closes https://github.com/flutter/flutter/issues/165931.

A basic in-memory implementation of Firestore in-so-far it supports `FirestoreService`.

For more advanced use-cases, i.e. `documentsResource().any.possible.api()`, mocking is still required, but for much of our API surface, it now _really_ reads, writes, and updates to an in-memory database that supports queries, filtering, ordering, and a (very basic) transaction.

We can move more code from `.documentsResource()` to `service.something()`, and ensure those are supported by the fake. In a follow-up PR we can start using this in our own tests where appropriate.

No production code changes.

### DIFF
--- a/app_dart/lib/src/service/firestore.dart
+++ b/app_dart/lib/src/service/firestore.dart
@@ -217,7 +217,7 @@ mixin FirestoreQueries {
 }
 
 class FirestoreService with FirestoreQueries {
-  /// Creates a [BigqueryService] using Google API authentication.
+  /// Creates a [FirestoreService] using Google API authentication.
   static Future<FirestoreService> from(GoogleAuthProvider authProvider) async {
     final client = await authProvider.createClient(
       scopes: const [FirestoreApi.datastoreScope],
--- a/app_dart/lib/src/service/firestore.dart
+++ b/app_dart/lib/src/service/firestore.dart
@@ -9,6 +9,7 @@ import 'package:cocoon_server/google_auth_provider.dart';
 import 'package:github/github.dart';
 import 'package:googleapis/firestore/v1.dart';
 import 'package:http/http.dart' as http;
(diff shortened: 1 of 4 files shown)


app_dart/lib/src/service/firestore.dart
```diff
@@ -250,9 +214,68 @@ class FirestoreService {
       return githubBuildStatuses.single;
     }
   }
+}
+
+class FirestoreService with FirestoreQueries {
+  /// Creates a [BigqueryService] using Google API authentication.
```
jtmcdole, line 220: s/BigqueryService/FirestoreService

app_dart/test/src/service/fake_firestore_service.dart
```diff
@@ -0,0 +1,424 @@
...
+    }
+
+    final result = buffer.toString();
+    if (_documents.containsKey(result)) {
+      return _generateUniqueId();
+    }
+
+    return result;
+  }
+
```
jtmcdole, line 65:
```suggestion
  final alphabet = ('ABCDEFGHIJKLMNOPQRSTUVWXYZ'
              'abcdefghijklmnopqrstuvwxyz'
              '0123456789' *
          32)
      .split('');

  String _generateUniqueId() {
    final result = (alphabet..shuffle()).take(20).join('');

    if (_documents.containsKey(result)) {
      return _generateUniqueId();
    }

    return result;
  }

```

jtmcdole (changes requested):

jtmcdole (approved):
https://github.com/user-attachments/assets/929621c5-82fa-4536-9e5e-c507bebf1450

PR #4444, opened 2025-04-04
Log `message.data` for `dart-internal-build-results-sub`.

matanlurey:
See https://github.com/flutter/flutter/issues/166535#issuecomment-2779593051.


jonahwilliams (approved):
LGTM

PR #4448, opened 2025-04-04
Simplify `src/model/firestore`, `AppDocument` is-a `Document`.

matanlurey:
This mostly removes a several dozen `fields!` across the codebase.

Also update unnecessary `\w+\?\s+get` getters.


jtmcdole (approved):
I hated all the ! everywhere. A+. Would aprove a second time.

PR #4451, opened 2025-04-07
Split Firestore from Datastore in `CommitService`

matanlurey:
Now converting and saving to both stores are distinct operations.

### DIFF
--- a/app_dart/lib/src/service/commit_service.dart
+++ b/app_dart/lib/src/service/commit_service.dart
@@ -5,130 +5,189 @@
 import 'dart:async';
 
 import 'package:cocoon_server/logging.dart';
-import 'package:gcloud/db.dart';
-import 'package:github/github.dart';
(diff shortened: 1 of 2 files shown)


app_dart/lib/src/service/commit_service.dart
```diff
@@ -5,130 +5,189 @@
...
+       _config = config,
+       _now = now;
 
-  final Config config;
-  final DatastoreServiceProvider datastoreProvider;
+  final Config _config;
+  final DatastoreServiceProvider _datastoreProvider;
+  final DateTime Function() _now;
 
   /// Add a commit based on a [CreateEvent] to the Datastore.
```
jtmcdole, line 37: /// Add a commit based on a [CreateEvent] to the backing stores.

jtmcdole (approved):

PR #4464, opened 2025-04-09
Refactor `luci_build_service_test.dart` into multiple distinct test files.

matanlurey:
Splits a ~2.5K LoC test into 6 separate files that include the specific functionality being tested.

In addition, migrated the tests to use `FakeFirestoreService` (off `MockFirestoreService`), which will make it easier to migrate this class to entirely be Firestore-first, as we won't need to deal with Firestore mocking.

Also it turns out if you didn't provide a fake `PubSub`, we used a real one that really worked... in a test.
 
Closes https://github.com/flutter/flutter/issues/166872.


app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -614,7 +594,7 @@ class LuciBuildService {
   }) async {
     final cacheValue = await cache.getOrCreate(
       subCacheName,
-      'builderlist',
+      'builderlist/$project/$bucket',
```
jtmcdole, line 597: The only side effect of this being flutter/prod vs flutter/try having a different builder list?

matanlurey: Correct.

This has 0 impact in today's use of this method, but if you were to provide a different bucket you would use the wrong cache (bad) or fill the wrong cache (worse).

app_dart/lib/src/model/firestore/pr_check_runs.dart
```diff
@@ -75,10 +75,20 @@ final class PrCheckRuns extends AppDocument<PrCheckRuns> {
...
   String get sha => fields[kShaField]!.stringValue!;
 
+  /// The repository slug associated with the pull request.
   RepositorySlug get slug => RepositorySlug.fromJson(
     json.decode(fields[kSlugField]!.stringValue!) as Map<String, Object?>,
   );
 
+  /// The recorded check-runs, a map of "test_name": "check_run id".
+  Map<String, String> get checkRuns {
+    final fields = this.fields.map((k, v) => MapEntry(k, v.stringValue!));
```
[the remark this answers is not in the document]
matanlurey: I'll avoid burning LUCI resources to rerun just for this change, but happy to make an adjustment in a future PR.

jtmcdole (approved):
lgtm-beavis

PR #4467, opened 2025-04-09
Calls to build-status can also trigger updated to discord

jtmcdole:
We lost the tree-status-bot (flutter/flutter/issues/161983)

Long live the tree status bot.

Note: this is guarded by the 15 second poll timer on build-status checks.


app_dart/lib/src/request_handlers/get_build_status.dart
```diff
@@ -40,4 +55,52 @@ base class GetBuildStatus extends RequestHandler<Body> {
       failingTasks: status.failedTasks,
     );
   }
+
+  /// Sends tree status updates to discord when they change from what was
+  /// last sent.
+  Future<void> maybeNotifyDiscord(BuildStatus status) async {
```
matanlurey, line 61: Can you move the bulk of this into a `DiscordService` (or just `Discord`)?

Ideally that class would not know anything about the rest of Cocoon (I.e. not import or use Firestore, or anything like that). In fact I'd imagine the public API is limited to:

```dart
interface class Discord {
  Future<DiscordResponse> postTreeStatusMessage(String message);
}
```

app_dart/lib/src/request_handlers/get_build_status.dart
```diff
@@ -32,6 +44,9 @@ base class GetBuildStatus extends RequestHandler<Body> {
     final repoName = request!.uri.queryParameters[_kRepoParam] ?? 'flutter';
     final slug = RepositorySlug('flutter', repoName);
     final status = (await _buildStatusService.calculateCumulativeStatus(slug))!;
+
+    await maybeNotifyDiscord(status);
```
matanlurey, line 48: So this service is called a *lot*, like, it's called every single time someone requests a badge of the build status, every time the dashboard refreshes, etc.

A better implementation would be a cron-based implementation (every 5m) that is something like `TreeStatusMessageUpdater extends GetBuildStatus`.

See the request handler in this same directory that does that for SVG badge generator for inspiration. It also means we would not have to fake out/provide an HTTP client to every `GetBuildStatus`.

app_dart/test/request_handlers/get_build_status_test.dart
matanlurey, line 1: Ditto above, this should all go into a `post_tree_status_message_test.dart` when done.

matanlurey (changes requested):

app_dart/lib/src/request_handlers/get_build_status.dart
jtmcdole: Sure, that part is easy.

app_dart/lib/src/request_handlers/get_build_status.dart
jtmcdole: This is already called a lot - but it's wrapped in a cache service and rate limited to once every 15 seconds, so on the grand-scale of things, not really that often (0.07 qps).   The http client is only provided if the status has changed.

jtmcdole: Moved

app_dart/lib/src/request_handlers/update_discord_status.dart
```diff
@@ -0,0 +1,80 @@
...
+
+import 'package:cocoon_common/rpc_model.dart' as rpc_model;
+import 'package:googleapis/firestore/v1.dart' show Document, Value;
+import 'package:meta/meta.dart';
+
+import '../../cocoon_service.dart';
+import '../service/discord_service.dart' show DiscordService;
+
+@immutable
+base class UpdateDiscordStatus extends GetBuildStatus {
```
matanlurey, line 15: nit: You can make this `final class`.

(The reason the linter told you to add base or final is because GetBuildStatus is base)

app_dart/lib/src/request_handlers/update_discord_status.dart
```diff
@@ -0,0 +1,80 @@
...
+  const UpdateDiscordStatus({
+    required super.config,
+    required super.buildStatusService,
+    required DiscordService discord,
+    @visibleForTesting DateTime Function() now = DateTime.now,
```
matanlurey, line 20: One of us, one of us.

app_dart/test/request_handlers/update_discord_status_test.dart
```diff
@@ -0,0 +1,252 @@
...
+import '../src/request_handling/request_handler_tester.dart';
+import '../src/utilities/entity_generators.dart';
+import '../src/utilities/mocks.mocks.dart';
+
+void main() {
+  useTestLoggerPerTest();
+
+  late RequestHandlerTester tester;
+  late UpdateDiscordStatus handler;
+  late MockFirestoreService mockFirestoreService;
```
matanlurey, line 26: Can you use `FakeFirestoreService` instead of `MockFirestoreService`? I'm trying to actively get rid of the latter, which does not need to be used anymore because similar to Datastore we have a full in-memory Fake (with real filter validation, etc).

app_dart/lib/src/request_handlers/update_discord_status.dart
jtmcdole: I feel dirty

app_dart/lib/src/request_handlers/update_discord_status.dart
jtmcdole: final it is.

app_dart/test/request_handlers/update_discord_status_test.dart
jtmcdole: I should have said "in a future pr" - but here I am..

matanlurey (approved):

PR #4468, opened 2025-04-10
Refactor `GerritService`, use retries for `GET`s, fix FIXME

matanlurey:
This PR removes a FIXME, lints for FIXMEs in the future, and fixes the test.

The original test did nothing, because if you write:
```dart
try {
  doSomethingThatNeverThrows();
} catch (e) {
  expect(e, 'You can literally write anything here');
}
```

The test will always pass. It turns out `GerritService` never retried (or threw) on `.get()` requests.

Fixing this and using a retry-client will also reduce the amount of pub-sub messages we use just to check Gerrit when waiting a few seconds in thread would have been sufficient. In addition did some TLC to `GerritService` and `FakeGerritService` across the repo.

### DIFF
--- a/app_dart/bin/gae_server.dart
+++ b/app_dart/bin/gae_server.dart
@@ -47,19 +47,23 @@ Future<void> main() async {
       accessTokenService: AccessTokenService.defaultProvider(config),
     );
 
+    // Gerrit service class to communicate with GoB.
+    final gerritService = GerritService(
+      config: config,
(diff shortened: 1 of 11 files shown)


jtmcdole (approved):
Would like to have hooks that give instant feedback on "FIXME" or other failures. `dart test test` doesn't really inspire confidence my PR will acutally pass :-/

PR #4470, opened 2025-04-10
Remove `\t` from query - update FakeFirstoreService

jtmcdole:
The regex for firestore is:

> Unquoted property paths must match regex ([a-zA-Z_][a-zA-Z_0-9]*), and quoted property paths must match regex (`(?:[^`\\]|(?:\\.))+`)

We appear to only have unquoted property paths.


matanlurey (approved):

jtmcdole:
Also towards https://github.com/flutter/flutter/issues/161983

PR #4474, opened 2025-04-10
Start a `/v2/...` dashboard for an improved Cocoon UI

matanlurey:
Imagine the following:

```sh
open https://flutter-dashboard.appspot.com/flutter/flutter/1234/presubmit
```

Ok this doesn't do that yet, but it does create a placeholder widget for it at `v2/<owner>/<repo>`.

Also disables Firebase Crashlytics in debug mode, it doesn't work and spams the console.

### DIFF
--- a/dashboard/lib/main.dart
+++ b/dashboard/lib/main.dart
@@ -14,6 +14,7 @@ import 'build_dashboard_page.dart';
 import 'firebase_options.dart';
 import 'service/cocoon.dart';
 import 'service/google_authentication.dart';
+import 'src/pages/v2_landing_page.dart';
 import 'state/build.dart';
 import 'widgets/now.dart';
 import 'widgets/state_provider.dart';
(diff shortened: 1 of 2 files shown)


dashboard/lib/main.dart
```diff
@@ -91,26 +96,68 @@ class MyApp extends StatelessWidget {
...
+    final [repoOwner, ...] = v2PathSegments;
+    if (v2PathSegments.length == 1) {
+      return (_) => V2LandingPage(repoOwner);
+    }
+
+    final [_, repoName, ...] = v2PathSegments;
+    if (v2PathSegments.length == 2) {
+      return (_) => V2LandingPage(repoOwner, repoName);
+    }
+
```
jtmcdole, line 160: This got me hacking; not a nit or anything else:

```dart
   switch (v2PathSegments) {
     case [var repoOwner, var repoName, ...]:
         print('repoOwner/Name: $repoOwner $repoName');
     case [var repoOwner, ...]:
         print('repoOwner: $repoOwner');
     default:
         print('empty');
   }
```

Without the whole switch/case:

```dart
  if (v2PathSegments case [var repoOwner, var repoName, ...]) {
    print('repoOwner/Name: $repoOwner $repoName');
    return;
  }
  if (v2PathSegments case [var repoOwner, ...]) {
      print('repoOwner: $repoOwner');
    return;
  }  
  print('empty');
  ```

jtmcdole (approved):
🤘 yes 🤘

dashboard/lib/main.dart
matanlurey: Done.

PR #4476, opened 2025-04-11
Tweak `UpdateDiscordStatus` a bit

matanlurey:
**Give ourselves a model** (I know, I know, but like, I'd prefer to know what's in the database - at some point someone will want to change something or something will break and purely for consistency it's nice to have all of the documents in a single place) and **make the Discord message include more details** (and diff between that and the previous status).

In addition:
1. Stores the failing tasks instead of the raw Discord message, in Firestore.
1. Reset the collection name so that the schemas are not incompatible (can delete the old one after).

Links would be nice, but I'd prefer to just building out a `/v2/flutter/flutter` page that is nice to look at instead.

### DIFF
--- a/app_dart/lib/src/request_handlers/update_discord_status.dart
+++ b/app_dart/lib/src/request_handlers/update_discord_status.dart
@@ -6,10 +6,11 @@ import 'dart:async';
 
 import 'package:cocoon_common/rpc_model.dart' as rpc_model;
 import 'package:cocoon_server/logging.dart';
-import 'package:googleapis/firestore/v1.dart' show Document, Value;
 import 'package:meta/meta.dart';
 
 import '../../cocoon_service.dart';
+import '../model/firestore/build_status_snapshot.dart';
+import '../request_handling/exceptions.dart';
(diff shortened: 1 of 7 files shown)


app_dart/lib/src/request_handlers/update_discord_status.dart
```diff
@@ -31,7 +32,7 @@ final class UpdateDiscordStatus extends GetBuildStatus {
   Future<Body> get() async {
     // For now, limit these to flutter/flutter only
     if (request!.uri.queryParameters[_kRepoParam] != 'flutter') {
-      return Body.forJson(const <String, dynamic>{});
+      throw const BadRequestException('Only ?repo=flutter is supported');
     }
```
jtmcdole, line 36: I think we can default this to "flutter" since the super class defaults.

```suggestion
  if (request!.uri.queryParameters[_kRepoParam] case final repo?) {
    if (repo != 'flutter') {
        throw const BadRequestException('Only ?repo=flutter is supported');
    }
  }
```

jtmcdole (approved):

PR #4483, opened 2025-04-11
Remove `DatastoreServceProvider` where not needed in services.

matanlurey:
All of these classes already can just self-wrap `config.db` if needed.

This reduces the amount of work required to rip out Datastore entirely when the time (soon) comes.


jtmcdole (approved):
lgtm

PR #4512, opened 2025-04-14
Use `FLUTTER_PREBUILT_ENGINE_VERSION` in post-submit.

matanlurey:
Closes https://github.com/flutter/flutter/issues/167010.

/cc @reidbaker @eyebrowsoffire @camsim99 for FYI.


app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -884,6 +884,16 @@ class LuciBuildService {
 
     final isFusion = commit.slug == Config.flutterSlug;
     if (isFusion) {
+      processedProperties.addAll({
+        'is_fusion': 'true',
+
+        // Always provide an engine version, just like we do in presubmit.
+        // See https://github.com/flutter/flutter/issues/167010.
+        'flutter_prebuilt_engine_version': commit.sha,
```
jtmcdole, line 892: Thoughts:
- For release+prod builds, I think we should add this.
- For master+prod builds I think we should not provide it, otherwise how will we know if any soup-de-hash isn't broken?

matanlurey: update_version_test?

I just think it's less complex to avoid gating on specific branches, but I'm not out right opposed to saying "master prod implicitly tests finding the HEAD hash"

jtmcdole: > update_version_test?

That just unit tests the hash is produced when you call it, but its not a test that flutter-tool's bat/sh files call it.

If cocoon has a checkout of the tree and produce the hash the same for each condition (try/prod, master/release).  With CAH work going on, Cocoon gets the hash from an external source.

Maybe that's good enough, but my gut says have the tests run as if they are real for prod.

matanlurey: Done.

jtmcdole (approved):

PR #4514, opened 2025-04-14
Move `ignoreCheck` logic into `PostsubmitLuciSubscription`.

matanlurey:
Closes https://github.com/flutter/flutter/issues/166858.

Deletes the most confusing boolean ever created in favor of inlining the logic into where it is used.


app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
```diff
@@ -189,8 +187,41 @@ final class PostsubmitLuciSubscription extends SubscriptionHandler {
...
+      limit: 1,
+      slug: currentCommit.slug,
+      branch: currentCommit.branch,
+    );
+    final latestCommit = commitList.single;
+
+    if (latestCommit.sha != currentCommit.sha) {
+      log.info('Not tip of tree: ${currentCommit.sha}');
+      return false;
+    }
```
jtmcdole, line 224: Can we add a note here that the Merge Queue uses PresubmitLuciSubscription and that this should be safe?

jtmcdole (approved):

app_dart/lib/src/request_handlers/postsubmit_luci_subscription.dart
matanlurey: Done.

PR #4516, opened 2025-04-15
Stop using `Target.value.*` to access `pb.Target` internals.

matanlurey:
Closes https://github.com/flutter/flutter/issues/166793.

I think this might have actually caught a bug or two!


jtmcdole:
what am I, the LGTM mill?

puppy_lgtm3

jtmcdole (approved):

PR #4522, opened 2025-04-15
Migrate from Google SI -> Firebase Auth

jtmcdole:
Fixes flutter/flutter#167006 by moving to firebase_auth. This has the benefit of storing credentials in the browser.


matanlurey (approved):

jtmcdole:
Reason for revert: client id is wrong. still debugging, but revert to unblock infra.

PR #4539, opened 2025-04-17
[Reland] Migrate from Google SI -> Firebase Auth

jtmcdole:
Switch Dashboard to Firebase Auth; now that cocoon supports both auths (temporarily), and the `client_id` is correct; this can reland.

Tested with chrome-devtools overrides to main.dart.js and index.html.


matanlurey (approved):
Re-land LGTM.

jtmcdole:
Reason for revert: wasm compilation fails to sign in; dart2js doesn't have a problem. UI widgets from firebase_ui_auth are using deprecated (and broken?) methods.

PR #4549, opened 2025-04-18
Add a config-guarded flag for Datastore usage.

matanlurey:

### DIFF
--- a/app_dart/lib/src/request_handlers/vacuum_github_commits.dart
+++ b/app_dart/lib/src/request_handlers/vacuum_github_commits.dart
@@ -12,25 +12,19 @@ import '../model/firestore/commit.dart' as fs;
 import '../request_handling/api_request_handler.dart';
 import '../request_handling/body.dart';
 import '../service/config.dart';
-import '../service/datastore.dart';
 import '../service/github_service.dart';
 import '../service/scheduler.dart';
 
-/// Query GitHub for commits from the past day and ensure they exist in datastore.
+/// Ensures Github commits from the past day exist in Firestore.
 @immutable
-class VacuumGithubCommits extends ApiRequestHandler<Body> {
+final class VacuumGithubCommits extends ApiRequestHandler<Body> {
   const VacuumGithubCommits({
     required super.config,
     required super.authenticationProvider,
-    required this.scheduler,
-    @visibleForTesting
-    this.datastoreProvider = DatastoreService.defaultProvider,
-  });
+    required Scheduler scheduler,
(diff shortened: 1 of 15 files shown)


app_dart/lib/src/request_handlers/vacuum_github_commits.dart
```diff
@@ -35,13 +35,11 @@ class VacuumGithubCommits extends ApiRequestHandler<Body> {
 
   @override
   Future<Body> get() async {
```
jtmcdole, line 37: go ahead and remove the other floating datastore references in the file?  Its not used by _vacuumRepository and _vacuumBranch

app_dart/lib/src/service/config.dart
```diff
@@ -184,7 +184,33 @@ class Config {
...
+  ///   final db = config.db;
+  /// }
+  ///
+  /// // OK
+  /// if (!config.useLegacyDatastore) {
+  ///   return;
+  /// }
+  /// final db = config.db;
+  /// ```
+  bool get useLegacyDatastore => true;
```
jtmcdole, line 203: Could change to _getSingleValue('COCOON_USE_LEGACY_DATASTORE') - we'd use it just for a short while, but it would give us the "flip the bit" faster than rolling?

jtmcdole (approved):

app_dart/lib/src/service/config.dart
matanlurey: Sounds good. Done!

app_dart/lib/src/service/config.dart
```diff
@@ -184,7 +195,41 @@ class Config {
...
       'bin/internal/release-candidate-branch.version';
 
-  DatastoreDB get db => _db;
+  /// Whether to read and write to Datastore.
+  ///
+  /// If `false`, the [db] will throw `UnsupportedError`, and as such usages
+  /// of Datastore should be guarded with if-statements and early terminations:
+  /// ```dart
+  /// // OK
+  /// if (config.useLegacyDatastore) {
```
jtmcdole, line 204: if (await config.useLegacyDatastore)

jtmcdole (approved):

app_dart/lib/src/request_handlers/vacuum_github_commits.dart
matanlurey: Done

app_dart/lib/src/service/config.dart
matanlurey: Done.

PR #4552, opened 2025-04-18
Add another inert schema field.

matanlurey:
Still crashes on commit without these (otherwise unused) fields.


jesswrd (approved):

jonahwilliams (approved):
LGTM

PR #4557, opened 2025-04-20
Remove all references to (now unused) Datastore

matanlurey:
Removes all references to Datastore.

Happy to hold off on this for a few days if we want to investigate anything.


app_dart/lib/src/model/appengine/key_helper.dart
```diff
@@ -1,160 +0,0 @@
-// Copyright 2019 The Flutter Authors. All rights reserved.
-// Use of this source code is governed by a BSD-style license that can be
-// found in the LICENSE file.
-
-import 'dart:convert';
-import 'dart:mirrors';
```
jtmcdole, line 6: NO MORE MIRRORS?!!?! :1st_place_medal: 🎉

jtmcdole (approved):
https://github.com/user-attachments/assets/f20a0612-2772-4514-8268-4f04f7c4e551

PR #4564, opened 2025-04-21
Refactor AppDocument to rename `name` field to `documentName`.

vikdubey:
Refactor AppDocument to rename `name` field to `documentName`.

https://github.com/flutter/flutter/issues/167426

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/app_dart/lib/src/model/firestore/base.dart
+++ b/app_dart/lib/src/model/firestore/base.dart
@@ -98,7 +98,7 @@ final class AppDocumentMetadata<T extends AppDocument<T>> {
 abstract class AppDocument<T extends AppDocument<T>> implements g.Document {
   AppDocument([g.Document? from])
     : _fields = from?.fields ?? {},
-      name = from?.name,
+      documentName = from?.name,
       createTime = from?.createTime,
       updateTime = from?.updateTime;
 
@@ -116,7 +116,15 @@ abstract class AppDocument<T extends AppDocument<T>> implements g.Document {
   final Map<String, g.Value> _fields;
 
   @override
-  String? name;
+  String? get name => documentName;
+
+  @override
+  set name(String? value) {
+    documentName = value;
+  }
+
+  /// The name of the document corresponding to `name` in [g.Document].
+  String? documentName;
 
   @override
   String? updateTime;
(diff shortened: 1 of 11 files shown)


app_dart/lib/src/model/firestore/base.dart
```diff
@@ -116,7 +116,15 @@ abstract class AppDocument<T extends AppDocument<T>> implements g.Document {
   final Map<String, g.Value> _fields;
 
   @override
-  String? name;
```
matanlurey, line 119: Unfortunately this doesn't really solve the underlying problem, now there is a `name` and `documentName`, but you can still use say, `task.name` and accidentally be referring to the full `database/path/to/document` instead of `task.taskName`.

I appreciate the contribution but this isn't enough to warrant merging.

matanlurey (changes requested):
See comment.

If you'd like to keep iterating on this I can give more pointers.

app_dart/lib/src/model/firestore/base.dart
vikdubey: Thanks @matanlurey for the prompt review and feedback. 

I understand the issue with making both `name` and `documentName` accessible. What about we make the `name` field unaccessible like below,

@override
  String? get name {
    throw UnsupportedError(
        'Accessing `name` is not allowed. Use `documentName` instead.');
  }

  @override
  set name(String? value) {
    throw UnsupportedError(
        'Setting `name` is not allowed. Use `documentName` instead.');
  }

Please let me know if this would work, I can make code changes accordingly. If not, could you please provide some pointers on potential design. Thanks.

matanlurey: Unfortunately that won't be a safe change to make - `<Document>.name` is used by the underlying (generated) `googeapis.v1.firebase` package, so it needs to exist, we just don't want to use it (directly). There are a few things we could do, in order of easiest to most difficult:

1. Mark `String get/set name` as `@protected`. Firebase APIs can still use, we won't (at least not by accident)
1. Make `AppDocument` no longer implement/extend `Document`, and instead using composition

Given the work you already did on this PR, I'd be OK with going with (1) as a stop-gap (it makes the current situation better), and leaving a bug open for (2). What do you think?

PR #4567, opened 2025-04-21
Use `mockito: ^5.4.6` to fix tip-of-tree analyzer breakages.

matanlurey:
Closes https://github.com/flutter/flutter/issues/167229.


analyze/pubspec.yaml
```diff
@@ -11,5 +11,5 @@ dependencies:
   platform: 3.1.6
 
 dev_dependencies:
-  mockito: 5.4.4
```
srawlins, line 14: Looks like this one might need to stay at 5.4.4 for... old build package or something? https://ci.chromium.org/ui/p/flutter/builders/try/Linux%20Cocoon/6890/overview

srawlins (approved):

analyze/pubspec.yaml
matanlurey: Let me try bumping!

PR #4577, opened 2025-04-22
Fix `rerun-all` logic with cases involving `flutter_release_builder`.

matanlurey:
Closes https://github.com/flutter/flutter/issues/167600.

Prior to this PR, there were multiple bugs:

- Most "recent" task was computed incorrectly (regression test added)
- It wasn't clear how we handled `flutter_release_builder` (made explicit ignore/error cases now)


jonahwilliams (approved):
LGTM

PR #4580, opened 2025-04-23
Backfill release candidates commits (otherwise are ignored)

matanlurey:
Closes https://github.com/flutter/flutter/issues/167675.

I think using `githubFileContent` (the same technique used by `CiYamlFetcher`) is the best way to reduce the GitHub API cost of this added operation, but I could be talked into either just letting it stand, or going for another option (caching, batch updates, etc).

### DIFF
--- a/app_dart/lib/server.dart
+++ b/app_dart/lib/server.dart
@@ -127,12 +127,14 @@ Server createServer({
       ciYamlFetcher: ciYamlFetcher,
       luciBuildService: luciBuildService,
       firestore: firestore,
+      branchService: branchService,
     ),
     '/api/v2/scheduler/batch-backfiller': BatchBackfiller(
       config: config,
       ciYamlFetcher: ciYamlFetcher,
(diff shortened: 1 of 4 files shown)


matanlurey:
Actually, maybe this isn't required in practice and is just a consequence of this particular process being long, since we'd normally merge PRs into the branch, and they would be considered new-ish. I'm trying that out by manually updating the [`createTimestamp` of `SHA=a2e3722fc63c5e44ff70c67053142f4bd5a67564`](https://console.cloud.google.com/firestore/databases/cocoon/data/panel/commits/a2e3722fc63c5e44ff70c67053142f4bd5a67564?e=-13802955&mods=logs_tg_prod&project=flutter-dashboard), and if that works maybe we can abandon this.

matanlurey:
No, this is required:
```dart
// firestore.dart#queryRecentCommits
branch ??= Config.defaultBranch(slug);
```

jtmcdole (approved):

PR #4581, opened 2025-04-23
`BranchService`: Use `githubFileContent` (CDN) instead of `.getFileContents` (API).

matanlurey:
Mitigates the API cost in https://github.com/flutter/cocoon/pull/4580.


jtmcdole (commented):
Only comment: there might be a unauthed limit to reading these files - but if this is every 5 minutes, I'd be surprised if we hit it.

PR #4589, opened 2025-04-24
Mark tasks that are not eligible for backfill as skipped

matanlurey:
Closes https://github.com/flutter/flutter/issues/167756.
Starts https://github.com/flutter/flutter/issues/167755.


app_dart/lib/src/request_handlers/scheduler/backfill_grid.dart
```diff
@@ -118,9 +118,45 @@ final class BackfillGrid {
...
+        for (final task in column) {
+          final commit = _commitsBySha[task.commitSha];
+          if (commit == null) {
+            throw StateError(
+              'A commit ("${task.commitSha}") should have existed in the grid',
+            );
+          }
+          yield SkippableTask._from(task, target: target, commit: commit);
+        }
+      }
```
jtmcdole, line 149:
```suggestion
      if (target.backfill) continue;
      for (final task in column) {
        final commit = _commitsBySha[task.commitSha];
        if (commit == null) {
          throw StateError(
              'A commit ("${task.commitSha}") should have existed in the grid',
          );
        }
        yield SkippableTask._from(task, target: target, commit: commit);
      }
```

jtmcdole (approved):

PR #4591, opened 2025-04-24
Fix `get-green-commits`, use `bringup` instead of `flaky`.

matanlurey:
Closes https://github.com/flutter/flutter/issues/167750.

There is always at least one failing task, which is bringup, so as a result we never returned any green commits.


jtmcdole (approved):
kahn-star-trek

PR #4602, opened 2025-04-25
fix: have all instances create keystore from cache

jtmcdole:
the first instance always had the correct keystore, the remaining keystores were empty.

added a test.


matanlurey (approved):
Nice! Thank you!

PR #4603, opened 2025-04-25
Simplify and fix build status provider.

matanlurey:
Closes https://github.com/flutter/flutter/issues/167846.

### DIFF
--- a/app_dart/lib/src/service/build_status_provider/commit_tasks_status.dart
+++ b/app_dart/lib/src/service/build_status_provider/commit_tasks_status.dart
@@ -50,11 +50,25 @@ final class CommitTasksStatus {
           didAtLeastOneFailureOccur: tasks.any(
             (t) => Task.taskFailStatusSet.contains(t.status),
           ),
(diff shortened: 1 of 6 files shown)


app_dart/lib/src/service/build_status_provider/commit_tasks_status.dart
```diff
@@ -50,12 +50,26 @@ final class CommitTasksStatus {
           didAtLeastOneFailureOccur: tasks.any(
             (t) => Task.taskFailStatusSet.contains(t.status),
           ),
-          lastCompletedAttemptWasFailure:
-              tasks.length > 1 &&
-              Task.taskFailStatusSet.contains(tasks[1].status),
+          lastCompletedAttemptWasFailure: _lastCompletedAttemptWasFailure(
```
jtmcdole, line 53: not that we should do this know, but this could be an enum for "lastCompletedStatus {success, failure, unknown}

jtmcdole (approved):

PR #4604, opened 2025-04-25
Support non-default branches with `build_status` APIs.

matanlurey:
... and rename `provider` -> `service`.


app_dart/lib/src/request_handlers/get_build_status.dart
```diff
@@ -30,8 +31,14 @@ base class GetBuildStatus extends RequestHandler<Body> {
   @protected
   Future<rpc_model.BuildStatusResponse> createResponse() async {
     final repoName = request!.uri.queryParameters[_kRepoParam] ?? 'flutter';
+    final repoBranch = request!.uri.queryParameters[_kBranchParam];
```
jtmcdole, line 34: Why not move a line down and include  `request!.uri.queryParameters[_kBranchParam] ?? Config.defaultBranch(slug)` - it would remove the null.

jtmcdole (approved):

PR #4612, opened 2025-04-28
Consolidate to one library for JWT / JWS

jtmcdole:
- Removes `corsac_jwt` dependency (which is using an older jose package).
- Moves nearly identical Json Web Signature code to shared folder + adds tests
- Behind the scenes: update secrets manager with the correctly formatted PEM.

### DIFF
--- a/packages/cocoon_server/test/generate_github_jws_test.dart
+++ b/packages/cocoon_server/test/generate_github_jws_test.dart
@@ -1,3 +1,7 @@
+// Copyright 2025 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
 import 'package:cocoon_server/generate_github_jws.dart';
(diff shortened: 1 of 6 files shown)


packages/cocoon_server/test/generate_github_jws_test.dart
```diff
@@ -0,0 +1,58 @@
...
+
+    final jwt = await JsonWebToken.decodeAndVerify(tokenString, keyStore);
+    final claims = jwt.claims;
+    expect(claims.issuedAt, now);
+    expect(claims.expiry, now.add(const Duration(minutes: 10)));
+    expect('${claims.issuer}', 'abc123');
+  });
+}
+
+const testKeyPem = '''
```
jtmcdole, line 29: this is a randomly generated rse key - NOT a key that is in secrets manager.

matanlurey (approved):

PR #4613, opened 2025-04-29
Process (write) content hash to firestore only

jtmcdole:
Towards: flutter/flutter#167778

Currently no action is taken and every merge group entry builds engine artifacts. This will start tracking (but not updating the status of) builds that could have been hashed / skipped.

  - ContentAwareHashBuild now addressed by the hash (no query)
  - Updates fake firestore to support ArrayValue FieldTransform
 
Future improvements:
  - Update the build status once the artifacts are built.
  - Passing the hash along to the FIRST builder to start uploading to commit sha and hash.
  - Adding a flag to enable waiting for artifacts.
  - Change the flutter tooling to look for the new hash.

### DIFF
--- a/app_dart/lib/src/service/content_aware_hash_service.dart
+++ b/app_dart/lib/src/service/content_aware_hash_service.dart
@@ -6,8 +6,13 @@ import 'dart:convert';
 
 import 'package:cocoon_server/logging.dart';
 import 'package:github/github.dart';
+import 'package:googleapis/firestore/v1.dart';
+import 'package:meta/meta.dart';
+import 'package:retry/retry.dart';
 
 import '../../cocoon_service.dart';
+import '../model/firestore/content_aware_hash_builds.dart'
+    show BuildStatus, ContentAwareHashBuilds;
 import '../model/github/annotations.dart';
 import '../model/github/workflow_job.dart';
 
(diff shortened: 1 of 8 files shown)


app_dart/lib/src/service/content_aware_hash_service.dart
```diff
@@ -112,4 +127,102 @@ interface class ContentAwareHashService {
...
+
+  /// Finds the hash status for [job] and updates any tracking docs.
+  Future<MergeQueueHashStatus> processWorkflowJob(WorkflowJobEvent job) async {
+    final hash = await hashFromWorkflowJobEvent(job);
+    if (hash == null) return MergeQueueHashStatus.unknown;
+
+    final headSha = job.workflowJob!.headSha!;
+
+    final r = const RetryOptions(
+      maxAttempts: 5, // number of entries in the merge group?
```
matanlurey, line 139: In other places in app_dart we use 3 retries, so I'd be curious why we think we need 5 (or more)

app_dart/lib/src/service/content_aware_hash_service.dart
```diff
@@ -112,4 +127,102 @@ interface class ContentAwareHashService {
     // Fail
     return null;
   }
+
+  /// Finds the hash status for [job] and updates any tracking docs.
+  Future<MergeQueueHashStatus> processWorkflowJob(WorkflowJobEvent job) async {
+    final hash = await hashFromWorkflowJobEvent(job);
+    if (hash == null) return MergeQueueHashStatus.unknown;
```
matanlurey, line 134: Do we expect this to happen as part of normal workflows?

If no, let's add a log.warn I think.

matanlurey (approved):

app_dart/lib/src/service/content_aware_hash_service.dart
jtmcdole: `hashFromWorkflowJobEvent()` is thrown _all workflow jobs_; including non-content-aware-hash, queueing, etc.  I will change this to "ignoreJob" and then add an error with a warn so we can monitor the logs.

```dart
      case 'workflow_job':
        try {
          final job = workflow_job.WorkflowJobEvent.fromJson(
            json.decode(webhook.payload) as Map<String, Object?>,
          );
          log.debug('workflow_job: $job');
          await scheduler.processWorkflowJob(job);
```

PR #4617, opened 2025-05-01
Add a hook to complete artifacts in the CAH service

jtmcdole:
> [!NOTE]  
> This does not currently complete any check runs, its only to record the data.

Towards flutter/flutter#167778

Add the ability to mark the content aware documents as completed from the scheduler.  This flows from:

```mermaid
sequenceDiagram
    participant S as Scheduler
    participant CAHS as ContentAwareHashService

    S->>S: processCheckRunCompletion(checkRunEvent)
    activate S

    alt Staging Conclusion is Error AND Is Merge Group
        S->>S: failGuardForMergeGroup(...)
        S->>CAHS: completeArtifacts(commitSha: sha, failed)
        activate CAHS
        CAHS-->>S: List<String> completedShas
        deactivate CAHS
        S-->>GitHub: complete_check_runs(failure)
    else Staging Conclusion is Success AND Is Merge Group
        S->>S: _closeSuccessfulEngineBuildStage(...)
        S->>CAHS: completeArtifacts(commitSha: sha, success)
        activate CAHS
        CAHS-->>S: List<String> completedShas
        deactivate CAHS
        S-->>GitHub: complete_check_runs(success)
    end
    deactivate S
```

Minor updates:
 - Performs a shallow copy for `fromDocument()` calls to `AppDocument<T>`. Without this, "commits" that fail in testing still have the document modified in memory.

### DIFF
--- a/app_dart/lib/src/model/firestore/base.dart
+++ b/app_dart/lib/src/model/firestore/base.dart
@@ -96,8 +96,10 @@ final class AppDocumentMetadata<T extends AppDocument<T>> {
 /// Provides methods across [g.Document] sub-types in `model/firestore/*.dart`.
 @internal
 abstract class AppDocument<T extends AppDocument<T>> implements g.Document {
+  /// Makes a new app document with potentially a shallow clone of the [from]
+  /// document.
   AppDocument([g.Document? from])
-    : _fields = from?.fields ?? {},
+    : _fields = {...?from?.fields},
       name = from?.name,
       createTime = from?.createTime,
       updateTime = from?.updateTime;
--- a/app_dart/test/service/content_aware_hash_service_test.dart
+++ b/app_dart/test/service/content_aware_hash_service_test.dart
@@ -249,6 +249,79 @@ void main() {
       expect(result, MergeQueueHashStatus.error);
     });
   });
+
+  group('markArtifactsAsComplete', () {
+    setUp(() {
+      firestoreService.putDocument(
+        ContentAwareHashBuilds(
+          createdOn: DateTime.now(),
+          contentHash: '1' * 40,
+          commitSha: 'a' * 40,
+          buildStatus: BuildStatus.inProgress,
+          waitingShas: ['b' * 40, 'c' * 40, 'd' * 40],
+        ),
+      );
+    });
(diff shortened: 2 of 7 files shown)


app_dart/lib/src/model/firestore/base.dart
```diff
@@ -96,8 +96,10 @@ final class AppDocumentMetadata<T extends AppDocument<T>> {
 /// Provides methods across [g.Document] sub-types in `model/firestore/*.dart`.
 @internal
 abstract class AppDocument<T extends AppDocument<T>> implements g.Document {
+  /// Makes a new app document with potentially a shallow clone of the [from]
+  /// document.
   AppDocument([g.Document? from])
-    : _fields = from?.fields ?? {},
+    : _fields = {...?from?.fields},
```
matanlurey, line 102: Nioce

app_dart/test/service/content_aware_hash_service_test.dart
```diff
@@ -249,6 +249,84 @@ void main() {
...
+          waitingShas: ['b' * 40, 'c' * 40, 'd' * 40],
+        ),
+      );
+      final shas = await cahs.completeArtifacts(
+        commitSha: 'a' * 40,
+        successful: true,
+        maxAttempts: 1,
+      );
+      expect(shas, isEmpty);
+      expect(firestoreService.rollbacks, hasLength(1));
```
matanlurey, line 299: Just a thought, does not have to be in this PR:

What if instead we threw an error if a transaction was ever _not_ comitted and not rolled back?

matanlurey (approved):

app_dart/test/service/content_aware_hash_service_test.dart
jtmcdole: So instead of `try/catch{rollback}`? It does throw if the retries fail 5 times in a row since those would not get committed or rolled back.

Or - are you asking that fake firestore has a teardown handler that sees if there are any transactions that have not been committed rolled back?

matanlurey: The latter.

PR #4618, opened 2025-05-01
Allow rerunning `Linux flutter_release_builder` through Cocoon

matanlurey:
Towards https://github.com/flutter/flutter/issues/168090.

I have not added tests, as I (regrettably) need to test this in production before knowing it actually works.

/cc @reidbaker @camsim99 @justinmc

### DIFF
--- a/app_dart/lib/src/service/luci_build_service.dart
+++ b/app_dart/lib/src/service/luci_build_service.dart
@@ -1083,11 +1083,12 @@ class LuciBuildService {
     // https://flutter.googlesource.com/recipes/+/refs/heads/main/recipes/release/release_builder.py#162
     final search = await _buildBucketClient.searchBuilds(
       bbv2.SearchBuildsRequest(
-        predicate: bbv2.BuildPredicate(descendantOf: Int64(buildNumber)),
+        predicate: bbv2.BuildPredicate(childOf: Int64(buildNumber)),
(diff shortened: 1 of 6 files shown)


app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -1060,4 +1061,85 @@ class LuciBuildService {
...
+      'rerunDartInternalReleaseBuilder(buildNumber=$buildNumber for $commit)',
+    );
+
+    // Because this is a large orchestrator build (a build that schedules many other sub-builds), and it is
+    // unlikely that every single build failed, we want to take advantage of the "retry_override_list" optional
+    // property, if able:
+    // https://flutter.googlesource.com/recipes/+/refs/heads/main/recipes/release/release_builder.py#162
+    final search = await _buildBucketClient.searchBuilds(
+      bbv2.SearchBuildsRequest(
+        predicate: bbv2.BuildPredicate(descendantOf: Int64(buildNumber)),
```
jtmcdole, line 1086: convo: does this include all descendants or just the immediate children?

jtmcdole (approved):
It looks right to me. On a tangent: it would be good to think about how we can take over some of the sub builds in a more optimum way rather than re-running whole orchestrators.

app_dart/lib/src/service/luci_build_service.dart
matanlurey: Fixed.

PR #4633, opened 2025-05-05
Move `CommitRef` and `TaskRef`, use `toRef()`.

matanlurey:
The previous setup made sense as a platform for migrating off Datastore, but these classes are actually really nice to use across scheduling APIs (and such), since they aren't connected to a mutable class, and are cheap to pass around.


app_dart/lib/src/model/commit_ref.dart
```diff
@@ -0,0 +1,43 @@
+// Copyright 2019 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:github/github.dart';
+import 'package:meta/meta.dart';
+
+/// Represents components of a GitHub commit without specifying the backend.
+@immutable
+final class CommitRef {
```
jtmcdole, line 10: Nit / curious: instead of importing a upper level path and adding the "toRef()" methods on the other objects; why not have a named constructor (CommitRef.fromCommit()) or an extension?  The extension would keep the other changes in the PR the same.

jtmcdole (approved):

app_dart/lib/src/model/commit_ref.dart
matanlurey: This PR gets rid of `.fromCommit` so you don't need an import something in order to convert:
image

I don't see the value in the extension, we own the type.

PR #4634, opened 2025-05-05
Start a shared `TaskStatus` enum-_lite_.

matanlurey:
Towards https://github.com/flutter/flutter/issues/167284.

This change is non-breaking, because `extension type implements String` _is_ a `String`, and will be represented using the same exact (backing) implementation. The next step(s) would be to remove all manual string references (i.e. to `'New'`), and start using the getters on `TaskStatus` instead of things like `failedStatusSet.contains(...)` across the codebase.

### DIFF
--- a/packages/cocoon_common/lib/task_status.dart
+++ b/packages/cocoon_common/lib/task_status.dart
@@ -1,0 +1,83 @@
+// Copyright 2019 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:collection/collection.dart';
+
+/// Represents differerent states of a task, or an execution of a build target.
+///
+/// TODO(matanlurey): Finish migrating (https://github.com/flutter/flutter/issues/167284):
+/// - [ ] All usages of `Task.status` should be of type `TaskStatus`
+/// - [ ] Stop implementing `String` and handle conversion elsewherer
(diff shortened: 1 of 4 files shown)


packages/cocoon_common/lib/task_status.dart
```diff
@@ -0,0 +1,83 @@
...
+
+import 'package:collection/collection.dart';
+
+/// Represents differerent states of a task, or an execution of a build target.
+///
+/// TODO(matanlurey): Finish migrating (https://github.com/flutter/flutter/issues/167284):
+/// - [ ] All usages of `Task.status` should be of type `TaskStatus`
+/// - [ ] Stop implementing `String` and handle conversion elsewherer
+/// - [ ] Replace extension type with an `enum`
+extension type const TaskStatus._(String _schemaValue) implements String {
```
jtmcdole, line 13: That's a mighty fancy enum you got there... I'm guessing its mostly because you want it to a) be String-like and b) access the String like value in a const fashion?

I ever tell you how much I hate dart enums since the language first started?

matanlurey: I actually am fine with using an `enum` (and want to end up there) but it's a huge breaking change so it's easier to start with an extension type :)

jtmcdole (approved):
I swear I hit "approve"!

kahn-star-trek

PR #4635, opened 2025-05-05
More wide-spread use of `CommitRef` instead of 3-4 params.

matanlurey:
More work needed to remove/consolidate, but this migrates some code.


jtmcdole (approved):
lgtm-beavis

PR #4636, opened 2025-05-05
Stop using Zones for `Request`/`requestData`, which was barely used.

matanlurey:
Towards https://github.com/flutter/flutter/issues/168333.

Removes an over-engineered use of zones, and pre-reading data just to stick the data and request on `RequestHandler.this.` - there was exactly _one_ class that ever used the Zone-based features of `RequestHandler` (`file_flaky_issue_and_pr.dart`), and it was trivially changed to a named argument instead of reading directly from `this.request`.


jtmcdole (approved):
https://github.com/user-attachments/assets/b7d69513-51f5-4d94-8d2d-7f8334a1a5e6

PR #4640, opened 2025-05-06
Remove `no_auth_request_handler`, which is unused.

matanlurey:

PR #4641, opened 2025-05-06
Remove `<T extends Body>` generics as unused

matanlurey:
Should submit https://github.com/flutter/cocoon/pull/4640 first to make this cleaner.


jtmcdole (approved):
lgtm-cat-shake

PR #4645, opened 2025-05-06
feat(config): support dynamic config for cocoon

jtmcdole:
Today we have to update secrets in cloud console; but it would be better to track some configs in git. The goal is to fetch (from HEAD) a checked in "config.json" and cache that between the services.

Using json serializable allows us to sanity check the changes.


app_dart/lib/src/service/config.dart
```diff
@@ -9,21 +9,25 @@ import 'package:cocoon_server/generate_github_jws.dart';
...
 
+part 'config.g.dart';
+
 /// Name of the default git branch.
 const String kDefaultBranchName = 'master';
 
 interface class Config {
   /// Creates and returns a [Config] instance.
-  Config(this._cache, this._secrets);
+  Config(this._cache, this._secrets, {DynamicConfig? dynamicConfig})
```
matanlurey, line 29: It would be nice to not accidentally initialize this to the default state unless that is expected.

In other words, something like:
```dart
Config(this._cache, this._secrets, {required DynamicConfig initialConfig}) : dynamicConfig = initialConfig;
Config.useRepoForInitialConfig(this._cache, this._secrets) : dynamicConfig = DynamicConfig.fromJson({});
```

Or if you prefer doing:
```dart
// Make it required to explicitly write `config: null` if we don't want to use the fetched config.
Config(this._cache, this._secrets, {required DynamicConfig? config})
```

I personally like (1) but (2) is also fine.

app_dart/lib/src/service/config.dart
```diff
@@ -59,6 +63,8 @@ interface class Config {
   final CacheService _cache;
   final SecretManager _secrets;
 
+  DynamicConfig dynamicConfig;
```
matanlurey, line 66: Can we make this private? I wouldn't want to see anyone _but_ `Config` read from this directly.

app_dart/lib/src/service/config.dart
```diff
@@ -489,3 +495,27 @@ interface class Config {
     return GithubService(github);
   }
 }
+
+/// Flags for the service that can be updated dynamically with out a restart.
+///
+/// Should be read from git/HEAD/app_dart/config.yaml and cached between
+/// services.
+@JsonSerializable()
+class DynamicConfig {
```
matanlurey, line 504: nit: `final` and `@immutable`.

app_dart/lib/src/service/config.dart
```diff
@@ -489,3 +495,27 @@ interface class Config {
...
+  /// This limits the number of commits to be checked to backfill. When bots
+  /// are idle, we hope to scan as many commit rows as possible.
+  @JsonKey(defaultValue: 50)
+  final int backfillerCommitLimit;
+
+  DynamicConfig({required this.backfillerCommitLimit});
+
+  /// Connect the generated [_$DynamicConfigFromJson] function to the `fromJson`
+  /// factory.
+  factory DynamicConfig.fromJson(Map<String, dynamic> json) =>
```
matanlurey, line 516: nit: Use `Map<String, Object?>` instead of `Map<String, dynamic>`.

It's the same backend type representation, but avoids accidental dynamic access.

app_dart/lib/src/service/dynamic_config_updater.dart
```diff
@@ -0,0 +1,85 @@
...
+    @visibleForTesting
+    RetryOptions retryOptions = const RetryOptions(
+      maxAttempts: 3,
+      delayFactor: Duration(seconds: 3),
+    ),
+  }) : _random = random ?? Random(),
+       _httpClient = httpClient ?? Client(),
+       _retryOptions = retryOptions;
+
+  final Duration delay;
```
matanlurey, line 30: nit: `_delay`

app_dart/lib/src/service/dynamic_config_updater.dart
```diff
@@ -0,0 +1,85 @@
...
+import 'package:cocoon_server/logging.dart';
+import 'package:http/http.dart';
+import 'package:meta/meta.dart';
+import 'package:retry/retry.dart';
+import 'package:yaml/yaml.dart' show YamlMap, loadYaml;
+
+import '../foundation/utils.dart';
+import 'config.dart';
+
+class DynamicConfigUpdater {
```
matanlurey, line 16: Can you add  ~1 line of doc comments?

If this is not intended to be extended, can you also add `interface class`?

app_dart/lib/src/service/dynamic_config_updater.dart
```diff
@@ -0,0 +1,85 @@
...
+
+  /// Fetches and parses the `config.yaml` from HEAD `flutter/cocoon/app_dart/`.
+  Future<DynamicConfig> fetchDynamicConfig() async {
+    final file = await githubFileContent(
+      Config.cocoonSlug,
+      'app_dart/config.yaml',
+      httpClientProvider: () => _httpClient,
+      retryOptions: _retryOptions,
+    );
+    final configYaml = loadYaml(file) as YamlMap;
```
matanlurey, line 43: I think you can do `as YamlMap<String, Object?>` (instead of below)

app_dart/lib/src/service/dynamic_config_updater.dart
```diff
@@ -0,0 +1,85 @@
...
+
+  UpdaterStatus _status = UpdaterStatus.stopped;
+
+  void stopUpdateLoop() {
+    if (_status != UpdaterStatus.running) return;
+    log.info('Stopping config update loop...');
+    _status = UpdaterStatus.stopping;
+  }
+
+  void startUpdateLoop(Config config) async {
```
matanlurey, line 55: It might be nice to reflect here and above that `stop` is a terminal state.

matanlurey (approved):
LGTM. A few comments.

app_dart/lib/src/service/config.dart
jtmcdole: I'll just fix FakeConfig - it should be required to have dynamic config first.

app_dart/lib/src/service/dynamic_config_updater.dart
jtmcdole: I have to cast it this way. `YamlMap` doesn't have generic types.

app_dart/lib/src/service/dynamic_config_updater.dart
jtmcdole: Its restartable

app_dart/lib/src/service/config.dart
jtmcdole: Done

app_dart/lib/src/service/config.dart
jtmcdole: Talked off line - done by pulling the updater into here so it has access.

PR #4652, opened 2025-05-07
Support 95% of tests in workflow (not dashboard)

jtmcdole:
Most of LUCI is just: 
 - pub get
 - test test
 - test integration

If we support flutter in the tool cache, we could get dashboard as well.


jtmcdole:
works on windows+docker desktop, old linux laptop+podman desktop (flatpak).

Maybe we revisit having a setup-flutter github action...

PR #4658, opened 2025-05-08
Remove TODOs and update goldens for filter defaults.

matanlurey:
We now have a script that makes this easy-ish.

Though if we end up only using GitHub actions, the script will need to get updated!


jtmcdole (approved):
`git@github.com:ewanmellor/git-diff-image.git` was helpful

PR #4663, opened 2025-05-08
Add and use extensions for dealing with common `Uint8List`-things.

matanlurey:
Towards https://github.com/flutter/flutter/issues/168333.

Reducing the reliance on reading `response!` from `Zone` values, so that `RequestHandler` is a pure function/class.


packages/cocoon_common/lib/core_extensions.dart
```diff
@@ -0,0 +1,65 @@
...
+
+  @override
+  Object? convert(List<int> input) {
+    return _utf8JsonDecoder.convert(input);
+  }
+}
+
+extension BytesExtension on Uint8List {
+  /// Parses a collection of bytes as a UTF-8 encoded JSON value.
+  T parseAsJson<T extends Object?>() {
```
jtmcdole, line 52: What about allowing `Converter<List<int>, Object?>` to be passed in, but optionally set to jsonUtf8Decoder?

jtmcdole (approved):

packages/cocoon_common/lib/core_extensions.dart
matanlurey: Sure we can add it when we need it :)

PR #4680, opened 2025-05-13
feat: diff the configs and output them

jtmcdole:
Side effect: don't set them if they aren't changing... shouldn't really be a side effect since we're single threaded.


matanlurey (approved):

matanlurey (approved):
LGTM

PR #4684, opened 2025-05-14
Use `GITHUB_ACTIONS` to determine a CI run.

matanlurey:
Closes https://github.com/flutter/flutter/issues/168859.

### DIFF
--- a/dashboard/test/utils/golden.dart
+++ b/dashboard/test/utils/golden.dart
@@ -50,8 +50,14 @@ class CocoonFileComparator extends LocalFileComparator {
     return super.getGoldenBytes(golden);
   }
 
-  static bool get _isLuciCi {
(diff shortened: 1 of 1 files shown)


jtmcdole (approved):
kahn-star-trek

dashboard/test/utils/golden.dart
```diff
@@ -50,8 +50,14 @@ class CocoonFileComparator extends LocalFileComparator {
     return super.getGoldenBytes(golden);
   }
 
-  static bool get _isLuciCi {
-    return !kIsWeb && Platform.environment.containsKey('LUCI_CONTEXT');
+  static bool get _isRunningOnCi {
```
jtmcdole, line 53: should we log that we're running on CI so.. you know... we see it?

actions are setup to run on ubuntu, so it should be the same locally or on github as well.

PR #4707, opened 2025-05-19
Fix dynamic parsing

jtmcdole:
Required since `YamlMap.cast()` does not deeply cast.

Work around till: https://github.com/dart-lang/tools/issues/2094


jtmcdole:
Followed up with a bug: https://github.com/dart-lang/tools/issues/2094

matanlurey (approved):

PR #4709, opened 2025-05-19
Simplify `Target.isReleaseBuild`, with otherwise the same output.

matanlurey:
Note that in `getProperties()` there is a pre-step that already converts the string `"true"` to the boolean `true`, which is confusing, but go figure.


jtmcdole (approved):
I added the "parse doubles" to that code as well :)

PR #4710, opened 2025-05-19
Build engine binaries in post-submit on experimental branches.

matanlurey:
Closes https://github.com/flutter/flutter/issues/169088.

### DIFF
--- a/app_dart/lib/src/model/ci_yaml/ci_yaml.dart
+++ b/app_dart/lib/src/model/ci_yaml/ci_yaml.dart
@@ -4,6 +4,7 @@
 
 import 'dart:convert';
 
+import 'package:cocoon_common/is_release_branch.dart';
 import 'package:collection/collection.dart';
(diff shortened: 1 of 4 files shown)


app_dart/lib/src/model/ci_yaml/ci_yaml.dart
```diff
@@ -220,29 +221,26 @@ class CiYaml {
...
+      // bringup: true         --> Always omitted.
+      // release_build: "true" --> Built by Linux flutter_release_builder.
       return [
         ...targets.where(
           (target) => !target.isReleaseBuild && !target.isBringup,
         ),
       ];
+    } else if (branch.contains('experimental')) {
+      // For experimental branches:
+      // Run all targets in post-submit.
```
jtmcdole, line 238: Might be good to note that this will produce binaries and upload them to gcp buckets.

jtmcdole (approved):

PR #4713, opened 2025-05-19
Removes unused feature `target.dependencies` and `ciYaml.getInitialTargets()`.

matanlurey:
Note that `target.properties.dependencies`, something entirely different, is not affected.

Closes https://github.com/flutter/flutter/issues/169103.


app_dart/lib/src/model/proto/internal/scheduler.proto
```diff
@@ -32,7 +32,7 @@ message Target {
     // Unique, human readable identifier.
     optional string name = 1;
     // Names of other targets required to succeed before triggering this target.
-    repeated string dependencies = 2;
```
jtmcdole, line 35: So this was future planning for some "build plan" esque thing? huh

jtmcdole (approved):

PR #4714, opened 2025-05-19
chore: turn on `waitOnContentHash` flag

jtmcdole:
Round 2: roll-forward after a new build that supports this.


matanlurey (approved):

jtmcdole:
I'll squash tomorrow

jtmcdole:
The services we care about picked this up. The older services that are still running are not happy, but we can stop them for now to prevent log pollution. 

image

matanlurey:
reason for revert: Not working as expected.

PR #4715, opened 2025-05-20
Opt out `bin/internal/engine.version` from most presubmits.

matanlurey:
Closes https://github.com/flutter/flutter/issues/169139.


app_dart/lib/src/service/scheduler/files_changed_optimization.dart
```diff
@@ -66,25 +66,33 @@ final class FilesChangedOptimizer {
...
           if (file == 'DEPS' || file.startsWith('engine/')) {
             log.info(
               '$refusePrefix: Engine sources changed.\n${filesChanged.join('\n')}',
             );
             return FilesChangedOptimization.none;
           }
-          if (markdownOnly && p.posix.extension(file) != '.md') {
-            markdownOnly = false;
+          if (noSourceImpact &&
+              p.posix.extension(file) != '.md' &&
```
jtmcdole, line 78: at some point, not today, we'll probably want to refactor this to a chain filter or something.

PR #4718, opened 2025-05-20
fix: `waitOnContentHash` should also trigger on completed hash

jtmcdole:
Prior to flipping completely over; we need to trigger tests for all shas, even if the content hash was generated.


matanlurey (approved):

jtmcdole:
I'll try the flags tomorrow - AFTER  LUCI recovers.

PR #4719, opened 2025-05-21
Add and use a `TimeRange` class to represent before/after timestamp.

matanlurey:
I started to work on https://github.com/flutter/flutter/issues/169234 and realized the current way timestamp filtering is done with Firebase is inflexible (and personally, confusing). This adds a small-ish `TimeRange` class in `cocoon_commons` and uses it instead of `int? timestamp` to determine which commits to find.

(Even if we want a different approach than https://github.com/flutter/flutter/issues/169234, this seems worth it?)

### DIFF
--- a/packages/cocoon_common/lib/src/time_range.dart
+++ b/packages/cocoon_common/lib/src/time_range.dart
@@ -1,0 +1,161 @@
+// Copyright 2019 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:meta/meta.dart';
+
+/// A definition of a time range, which can be specific or indefinite.
+@immutable
+sealed class TimeRange {
+  const TimeRange();
+
+  /// Matches any time.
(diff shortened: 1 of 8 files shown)


jtmcdole (commented):
What is inflexible about `timestamp`?

packages/cocoon_common/lib/src/time_range.dart
```diff
@@ -0,0 +1,161 @@
...
+  /// Creates a range matching time between [start] and [end].
+  ///
+  /// If [exclusive] is set, the specific moments described in [start] and [end]
+  /// are excluded.
+  ///
+  /// [end] must be after [start].
+  factory TimeRange.between(
+    DateTime start, //
+    DateTime end, {
+    bool exclusive,
```
jtmcdole, line 24: Is this either: 
`(start, end)` or `[start, end]`
and never `(start, end]` or `[start, end)`

jtmcdole (approved):
Looked at the code and you are filtering, so it looks like a nice adapter on top.

PR #4731, opened 2025-05-28
Add a server flag `onlyUseTipOfTreeTargetsExistenceToFilterTargets`.

matanlurey:
Towards https://github.com/flutter/flutter/issues/169625.

### DIFF
--- a/app_dart/lib/src/service/config.dart
+++ b/app_dart/lib/src/service/config.dart
@@ -2,6 +2,9 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+/// @docImport '../model/ci_yaml/ci_yaml.dart';
+library;
+
 import 'dart:convert';
 import 'dart:math' show Random;
 import 'dart:typed_data';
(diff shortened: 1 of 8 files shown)


app_dart/lib/src/service/config.dart
```diff
@@ -31,12 +34,18 @@ const String kDefaultBranchName = 'master';
...
+    this._cache, //
+    this._secrets, {
+    required DynamicConfig dynamicConfig,
+  }) : _dynamicConfig = dynamicConfig;
 
   /// Access dynamically configured flags.
   DynamicConfig get flags => _dynamicConfig;
 
+  /// See [CiYaml.onlyUseTipOfTreeTargetsExistenceToFilterTargets].
+  static const bool kOnlyUseTipOfTreeTargetsExistenceToFilterTargets = false;
```
jtmcdole, line 47: Rather than exposing the default and maybe hiding places that do not pass this flag (error prone); make people refer to `config.onlyUseTipOfTreeTargetsExistenceToFilterTargets`.

jtmcdole (changes requested):

PR #4732, opened 2025-05-28
Refactor and split up `DynamicConfig*`, prepare for more flag groups.

matanlurey:
Doing as pre-work before requested changes on https://github.com/flutter/cocoon/pull/4731.

I ran into a number of cases where changes were needed, and figured a separate PR was cleaner:

- Moved all of the bits out of `config.dart` and into `flags/_*.dart` files to reduce file size

- We have command-line tools in `app_dart/bin` that are _not_ a server (i.e. `verify_test_ownership.dart`) that create `CiYaml` objects, and also need `CiYamlFlags`. Instead of hacking something together, I added `DynamicConfig.fromRemoteLatest` and `DynamicConfig.fromLocalRepo` to use, respectively, in these command-line tools (which are used in CI production, not unused tools).

- I made all of the objects `const`-able, and made it easy to create defaults, even if they are not `const` themselves, by using the `factory` constructor pattern. That also gets rid of the kludge we had with `contentAwareHashing` default values.

### DIFF
--- a/app_dart/lib/src/service/config.dart
+++ b/app_dart/lib/src/service/config.dart
@@ -3,7 +3,6 @@
 // found in the LICENSE file.
 
 import 'dart:convert';
-import 'dart:math' show Random;
 import 'dart:typed_data';
 
 import 'package:cocoon_server/generate_github_jws.dart';
@@ -12,30 +11,20 @@ import 'package:cocoon_server/secret_manager.dart';
 import 'package:github/github.dart' as gh;
 import 'package:graphql/client.dart' hide JsonSerializable;
 import 'package:http/http.dart' as http;
-import 'package:json_annotation/json_annotation.dart';
 import 'package:meta/meta.dart';
 import 'package:retry/retry.dart';
-import 'package:yaml/yaml.dart' show YamlList, YamlMap, loadYaml;
 
 import '../../cocoon_service.dart';
-import '../foundation/providers.dart' show Providers;
-import '../foundation/typedefs.dart' show HttpClientProvider;
-import 'flags/content_aware_hashing_flags.dart';
+import 'flags/dynamic_config_updater.dart';
 import 'github_service.dart';
 import 'luci_build_service/cipd_version.dart';
-
-part 'config.g.dart';
 
 /// Name of the default git branch.
 const String kDefaultBranchName = 'master';
 
-interface class Config {
+interface class Config extends DynamicallyUpdatedConfig {
   /// Creates and returns a [Config] instance.
-  Config(this._cache, this._secrets, {required DynamicConfig dynamicConfig})
-    : _dynamicConfig = dynamicConfig;
-
-  /// Access dynamically configured flags.
-  DynamicConfig get flags => _dynamicConfig;
+  Config(this._cache, this._secrets, {required super.initialConfig});
 
   /// When present on a pull request, instructs Cocoon to submit it
   /// automatically as soon as all the required checks pass.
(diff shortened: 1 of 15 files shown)


app_dart/lib/src/service/config.dart
```diff
@@ -12,30 +11,20 @@ import 'package:cocoon_server/secret_manager.dart';
...
 import 'github_service.dart';
 import 'luci_build_service/cipd_version.dart';
 
-part 'config.g.dart';
-
 /// Name of the default git branch.
 const String kDefaultBranchName = 'master';
 
-interface class Config {
+interface class Config extends DynamicallyUpdatedConfig {
```
jtmcdole, line 25: 👍🏻

app_dart/lib/src/service/flags/dynamic_config.dart
```diff
@@ -0,0 +1,139 @@
...
+
+  /// Creates [DynamicConfig] flags from a [yaml] object.
+  factory DynamicConfig.fromYaml(YamlMap? yaml) {
+    return DynamicConfig.fromJson(yaml?.asMap);
+  }
+
+  /// Returns the latest copy of [DynamicConfig] fetched from tip-of-tree.
+  ///
+  /// Equivalent to a single call to [DynamicConfigUpdater.fetchDynamicConfig].
+  static Future<DynamicConfig> fromRemoteLatest() async {
```
jtmcdole, line 86: nit: don't need `async` if you don't await on anything.

app_dart/lib/src/service/flags/dynamic_config.dart
```diff
@@ -0,0 +1,139 @@
...
+    return DynamicConfigUpdater().fetchDynamicConfig();
+  }
+
+  /// Returns the latest copy of [DynamicConfig] fetched from the repository.
```
jtmcdole, line 90: fromLocalFilesystem?

app_dart/lib/src/service/flags/dynamic_config.dart
```diff
@@ -0,0 +1,139 @@
...
+      dir = dir.parent;
+    }
+
+    final configPath = io.File(p.join(dir.path, 'app_dart', 'config.yaml'));
+    if (!await configPath.exists()) {
+      throw StateError('Could not find config.yaml at ${configPath.path}');
+    }
+
+    final yaml = loadYaml(await configPath.readAsString()) as YamlMap;
+    return DynamicConfig.fromJson(yaml.cast());
```
jtmcdole, line 110: this will fail because cast<>() doesn't do a deep copy, you'll just hit another YamlMap under the hood as a runtime failure. Call fromYaml()?

jtmcdole (changes requested):

app_dart/lib/src/service/flags/dynamic_config.dart
matanlurey: Done!

app_dart/lib/src/service/flags/dynamic_config.dart
matanlurey: Done!

app_dart/lib/src/service/flags/dynamic_config.dart
matanlurey: Doh! I wrote it specifically to avoid doing this and then ... didn't use it. Done!

jtmcdole (approved):
https://github.com/user-attachments/assets/f7306818-d993-4b99-bab1-001c9cedb516

PR #4752, opened 2025-06-02
Update Docker SDK images to latest `stable`.

matanlurey:
... instead of manual revs.

We can always pin to an older version if that ends up being needed, but this will be in line with how most of us do development (using the latest version of `dart` that ships with the master channel of `flutter`).


jtmcdole (approved):
printer2

PR #4759, opened 2025-06-03
Add `/api/update-tree-status`, which is currently write-only

matanlurey:
Towards https://github.com/flutter/flutter/issues/74529.

Inserts documents into the collection `tree_status_change`, with the following "schema":
```dart
static Future<TreeStatusChange> create(
  FirestoreService firestore, {
  required DateTime createdOn,
  required TreeStatus status,
  required String authoredBy,
  required RepositorySlug repository,
});
```

To determine if the current tree should be (manually) failing, we'd run:
```dart
static Future<TreeStatusChange?> getLatest(
  FirestoreService firestore, {
  required RepositorySlug repository,
}) async {
  final docs = await firestore.query(
    metadata.collectionId,
    {'$_fieldRepository =': repository.fullName},
    limit: 1,
    orderMap: {_fieldCreateTimestamp: kQueryOrderDescending},
  );
  return docs.isEmpty ? null : TreeStatusChange.fromDocument(docs.first);
}
```

... however, that is not hooked up yet (we also need to create indexes first, etc).

### DIFF
--- a/app_dart/lib/src/model/firestore/tree_status_change.dart
+++ b/app_dart/lib/src/model/firestore/tree_status_change.dart
@@ -1,0 +1,78 @@
+// Copyright 2019 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:googleapis/firestore/v1.dart' as g;
+
+import '../../service/firestore.dart';
+import 'base.dart';
+
+/// A row for each tree status change.
+final class TreeStatusChange extends AppDocument<TreeStatusChange> {
+  /// Returns the latest [TreeStatusChange].
+  ///
+  /// If no changes exist, returns `null`.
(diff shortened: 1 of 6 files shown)


app_dart/lib/src/model/firestore/tree_status_change.dart
```diff
@@ -0,0 +1,89 @@
...
+    return DateTime.parse(fields[_fieldCreateTimestamp]!.timestampValue!);
+  }
+
+  TreeStatus get status {
+    return TreeStatus.values.byName(fields[_fieldStatus]!.stringValue!);
+  }
+
+  String get authoredBy {
+    return fields[_fieldAuthoredBy]!.stringValue!;
+  }
```
jtmcdole, line 84: Can we have an optional String reason?

jtmcdole (approved):

PR #4761, opened 2025-06-04
Delete the V2 experiment, add a `TreeStatusPage`.

matanlurey:
Creates the UI-facing bits of https://github.com/flutter/flutter/issues/74529.

This is "hidden" behind the page URL not being part of the navigation.

I ended up EOL'ing the `V2` experiment, it's too much work to adopt a new navigation model at the moment. Instead, I'm pulling out a reusable `CocoonScaffold` which allows new pages to be added without much work. Here's an example if you're too lazy to run this locally:

Screenshot 2025-06-03 at 5 25 53 PM

Screenshot 2025-06-03 at 5 26 07 PM


dashboard/lib/service/appengine_cocoon.dart
```diff
@@ -188,6 +188,63 @@ class AppEngineCocoonService implements CocoonService {
...
+    String? reason,
+  }) async {
+    final updateTreeStatusUrl = apiEndpoint('/api/public/update-tree-status');
+    final response = await _client.post(
+      updateTreeStatusUrl,
+      body: jsonEncode({
+        'repo': repo,
+        'status': status.name,
+        if (reason != null) 'reason': reason,
+      }),
```
jtmcdole, line 237: We don't want a shared JsonSerializable for this? `TreeStatusChange` has repo/status/reason.

packages/cocoon_common/lib/src/rpc_model/tree_status_change.dart
```diff
@@ -0,0 +1,45 @@
...
+import 'package:json_annotation/json_annotation.dart';
+import 'package:meta/meta.dart';
+
+import 'base.dart';
+
+part 'tree_status_change.g.dart';
+
+@JsonSerializable(checked: true)
+@immutable
+final class TreeStatusChange extends Model {
```
jtmcdole, line 14: Same as the firebase TreeStatusChange?

matanlurey: Yeah that's intentional - the frontend uses this as well.

dashboard/lib/service/appengine_cocoon.dart
matanlurey: Mental TODO, yeah this should just be using TreeStatusChange.toJson().

PR #4762, opened 2025-06-04
Remove completed flag migrations for CiYaml.

matanlurey:
Closes https://github.com/flutter/flutter/issues/169625.

Both:
- `onlyUseTipOfTreeTargetsExistenceToFilterTargets`
- `targetEnabledBranchesOverridesTipOfTreeTargetExistence`

... have been enabled for a few days without incident. Let's clean up the tech-debt.


app_dart/config.yaml
```diff
@@ -6,8 +6,12 @@
 backfillerCommitLimit: 50
 
 ciYaml:
+  # ----------------------------------------------------------------------------
+  # TODO(matanlurey): These are inert in the latest builds and can be removed
+  # a few days after https://github.com/flutter/cocoon/pull/4762 is merged.
```
jtmcdole, line 11: 14 days if the "old images are removed after 14 days" rule is being followed.

PR #4766, opened 2025-06-04
Behind a flag, `allowManualTreeClosures`, allow closing the tree.

matanlurey:
Towards https://github.com/flutter/flutter/issues/74529.

### DIFF
--- a/app_dart/lib/src/service/build_status_service.dart
+++ b/app_dart/lib/src/service/build_status_service.dart
@@ -12,12 +12,18 @@ import 'package:meta/meta.dart';
 
 import '../../cocoon_service.dart';
 import '../model/firestore/github_build_status.dart';
+import '../model/firestore/tree_status_change.dart';
 import 'build_status_provider/commit_tasks_status.dart';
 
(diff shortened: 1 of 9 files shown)


app_dart/lib/src/service/build_status_service.dart
```diff
@@ -77,6 +83,18 @@ interface class BuildStatusService {
       }
     }
 
+    if (failingTasks.isEmpty && _config.flags.allowManualTreeClosures) {
```
jtmcdole: I'd want to see Manual Closure in front of any failing tasks if its enabled (so its less of a surprise on the dashboard). So consider moving this aboveline 73 and always adding to the failingTasks list if manual closed. The list should then look like:

['Manual Closure: blah', 'Linux tool_tests_general']

jtmcdole (changes requested):

app_dart/lib/src/service/build_status_service.dart
matanlurey: Donezuo.

PR #4769, opened 2025-06-05
Pass an authenticated `idToken` to `*TreeStatus` API calls.

matanlurey:


dashboard/lib/service/dev_cocoon.dart
```diff
@@ -120,6 +121,7 @@ class DevelopmentCocoonService implements CocoonService {
 
   @override
   Future<CocoonResponse<void>> updateTreeStatus({
+    required String idToken,
```
jtmcdole, line 124: Do something with this token?

matanlurey: I did

PR #4785, opened 2025-06-16
Use pinned `engine.version` for release candidate prod tests.

matanlurey:
Closes https://github.com/flutter/flutter/issues/170568.

### DIFF
--- a/app_dart/test/service/luci_build_service/schedule_prod_builds_test.dart
+++ b/app_dart/test/service/luci_build_service/schedule_prod_builds_test.dart
@@ -447,7 +447,7 @@ void main() {
     );
   });
 
-  // Regression test for https://github.com/flutter/flutter/issues/167010.
+  // Regression test for https://github.com/flutter/flutter/issues/170568.
   test('schedules a post-submit release candidate build', () async {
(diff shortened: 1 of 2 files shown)


app_dart/test/service/luci_build_service/schedule_prod_builds_test.dart
```diff
@@ -533,7 +533,7 @@ void main() {
       'os': bbv2.Value(stringValue: 'debian-10.12'),
       'recipe': bbv2.Value(stringValue: 'devicelab/devicelab'),
       'is_fusion': bbv2.Value(stringValue: 'true'),
-      'flutter_prebuilt_engine_version': bbv2.Value(stringValue: commit.sha),
+      // Intentionally omitted: flutter_prebuilt_engine_version
```
jtmcdole, line 536: Do you want to check that `flutter_prebuilt_engine_version` is not actually present; i.e. "is expect matching "fields" doing deep, === equality?"

matanlurey: It does do deep equality checks.

PR #4786, opened 2025-06-17
When a PR is `labeled`, store the (fresh) PR for later same-commit re-runs

matanlurey:
Closes https://github.com/flutter/flutter/issues/170300.


app_dart/lib/src/request_handlers/github/webhook_subscription.dart
```diff
@@ -1036,6 +1046,8 @@ The "Merge" button is also unlocked. To bypass presubmits as well as the tree st
     } else {
       logInfo('no emergency label; moving on.');
     }
+
+    // Store the pull request.
```
jtmcdole: Remove?

matanlurey: Done.

PR #4791, opened 2025-06-24
Do not misreport breakages on a non-default branch as a tree status failure

matanlurey:
Closes https://github.com/flutter/flutter/issues/171104.

### DIFF
--- a/app_dart/lib/src/service/build_status_service.dart
+++ b/app_dart/lib/src/service/build_status_service.dart
@@ -48,7 +48,7 @@ interface class BuildStatusService {
   /// - Task `B` fails becuse its last known _completed_ status was failing
   Future<BuildStatus> calculateCumulativeStatus(
     RepositorySlug slug, {
-    String? branch,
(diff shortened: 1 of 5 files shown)


app_dart/lib/src/service/build_status_service.dart
```diff
@@ -110,9 +110,10 @@ interface class BuildStatusService {
   Future<List<CommitTasksStatus>> retrieveCommitStatusFirestore({
     required RepositorySlug slug,
     required int limit,
+    required String branch,
     TimeRange? created,
-    String? branch,
   }) async {
+    // branch ??= Config.defaultBranch(slug);
```
jtmcdole: remove?

jtmcdole (approved):
kahn-star-trek

app_dart/lib/src/service/build_status_service.dart
matanlurey: Donezuo

PR #4792, opened 2025-06-25
fix: update tree-status should fetch the entire PR list

jtmcdole:
Avoid enumerating multiple pages of PRs over a period of time in which the list can be mutated by outside factors.

potentially fixes #171179


matanlurey (approved):
LGTM.

PR #4800, opened 2025-07-02
fix: signin api breakage

jtmcdole:
Removes calling the plugin directly. Uses the `authenticationEvents` stream, and the web-only exposed render button.

fixes flutter/flutter#171479

### DIFF
--- a/dashboard/lib/widgets/sign_in_button/sign_in_button_web.dart
+++ b/dashboard/lib/widgets/sign_in_button/sign_in_button_web.dart
@@ -36,11 +36,13 @@ class _SignInButtonState extends State<SignInButton> {
           GoogleSignInAuthenticationEventSignIn() => data.user,
           GoogleSignInAuthenticationEventSignOut() => null,
         };
+        final fireAuth = FirebaseAuth.instance;
+
         if (user == null) {
+          await fireAuth.signOut();
           print('signed out');
           return;
         }
-        final fireAuth = FirebaseAuth.instance;
         final credential = GoogleAuthProvider.credential(
           idToken: user.authentication.idToken,
         );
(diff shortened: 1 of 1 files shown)


dashboard/lib/widgets/sign_in_button/sign_in_button_web.dart
```diff
@@ -5,8 +5,14 @@
 // import 'package:firebase_ui_oauth_google/firebase_ui_oauth_google.dart';
 import 'package:firebase_auth/firebase_auth.dart';
 import 'package:flutter/material.dart';
-import 'package:google_sign_in_platform_interface/google_sign_in_platform_interface.dart';
-import 'package:google_sign_in_web/google_sign_in_web.dart';
+
+import 'package:google_sign_in/google_sign_in.dart'
+    show
```
stuartmorgan-g, line 10: Why the restrictions on the import, out of curiosity?

dashboard/lib/widgets/sign_in_button/sign_in_button_web.dart
```diff
@@ -17,29 +23,33 @@ class SignInButton extends StatefulWidget {
...
-        final credential = GoogleAuthProvider.credential(idToken: data.idToken);
-        try {
-          final cred = await auth.signInWithCredential(credential);
-          print('${cred.user?.email} signed in');
-        } catch (e) {
-          print('error signing into firebase');
+
+    _initializing = GoogleSignIn.instance.initialize();
+
+    GoogleSignIn.instance.authenticationEvents.listen((data) async {
```
stuartmorgan-g, line 33: You probably want an `onError` that at least logs to console or something, for debugging issues. IIRC the old stream didn't really get errors, but the new one does (usually it's a `GoogleSignInException` that has useful information).

dashboard/lib/widgets/sign_in_button/sign_in_button_web.dart
jtmcdole: You mean "show"? No reason other than reflex.

dashboard/lib/widgets/sign_in_button/sign_in_button_web.dart
jtmcdole: Will look

PR #4830, opened 2025-08-15
feat: allow for array contains (@>) and OR filter on firestore

jtmcdole:
Firestore supports this natively; updating our implementation to offer it for next feature.

### DIFF
--- a/app_dart/test/src/service/fake_firestore_service.dart
+++ b/app_dart/test/src/service/fake_firestore_service.dart
@@ -502,11 +502,20 @@ abstract base class _FakeInMemoryFirestoreService
       return collectionId == collection;
     });
 
-    results = results.where((document) {
-      return _matchesFilter(document.fields!, filterMap);
-    });
-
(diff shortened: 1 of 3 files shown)


app_dart/test/src/service/fake_firestore_service.dart
```diff
@@ -502,11 +502,20 @@ abstract base class _FakeInMemoryFirestoreService
...
-      return _matchesFilter(document.fields!, filterMap);
-    });
-
-    if (compositeFilterOp != kCompositeFilterOpAnd) {
+    if (compositeFilterOp == kCompositeFilterOpAnd) {
+      results = results.where((document) {
+        return _matchesFilter(document.fields!, filterMap);
+      });
+    } else if (compositeFilterOp == kCompositeFilterOpOr) {
+      final listFilters = [
```
jtmcdole, line 510: I'm going to ignore the part where Firebase allows for nested filters, so you could do nested and/or.

PR #4832, opened 2025-08-18
moved processCheckRunCompleted call to LUCI presubmit subscription

ievdokdm:
Process to the check-run status in the merge queue document during the LUCI callback.

Issue: [166466](https://github.com/flutter/flutter/issues/166466)

### DIFF
--- a/app_dart/lib/src/model/proto/internal/github_webhook.pb.dart
+++ b/app_dart/lib/src/model/proto/internal/github_webhook.pb.dart
@@ -30,41 +30,53 @@ class GithubWebhookMessage extends $pb.GeneratedMessage {
     return $result;
   }
   GithubWebhookMessage._() : super();
-  factory GithubWebhookMessage.fromBuffer($core.List<$core.int> i, [$pb.ExtensionRegistry r = $pb.ExtensionRegistry.EMPTY]) => create()..mergeFromBuffer(i, r);
-  factory GithubWebhookMessage.fromJson($core.String i, [$pb.ExtensionRegistry r = $pb.ExtensionRegistry.EMPTY]) => create()..mergeFromJson(i, r);
+  factory GithubWebhookMessage.fromBuffer($core.List<$core.int> i,
+          [$pb.ExtensionRegistry r = $pb.ExtensionRegistry.EMPTY]) =>
+      create()..mergeFromBuffer(i, r);
+  factory GithubWebhookMessage.fromJson($core.String i,
+          [$pb.ExtensionRegistry r = $pb.ExtensionRegistry.EMPTY]) =>
+      create()..mergeFromJson(i, r);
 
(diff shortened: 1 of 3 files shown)


app_dart/README.md
```diff
@@ -46,6 +46,13 @@ $ dart run build_runner build
 
 Any updates should be checked into source control.
 
+#### Mockito
```
matanlurey, line 49: 👍🏼

app_dart/config.yaml
matanlurey, line 1: If this will immediately take affect, I'd prefer to start it as `false` (so the commit trail is cleaner).

app_dart/config.yaml
```diff
@@ -14,3 +14,7 @@ contentAwareHashing:
   # This will cause PRs that enter the merge queue to wait on CAH hashing
   # to happen before scheduling builds.
   waitOnContentHash: true
+
+# Whether to close the MQ guard right after LUCI presubmit compleated
```
matanlurey: s/compleated/completed

app_dart/lib/src/request_handling/subscription_handler.dart
```diff
@@ -50,6 +50,7 @@ abstract base class SubscriptionHandler extends RequestHandler {
       getValue<AuthenticatedContext>(ApiKey.authContext)!;
 
   /// The [PushMessage] from this [HttpRequest].
+  // TODO(dmgr): Message should be either input param of a post or a method.
```
matanlurey, line 53: Agreed. The current `context` implementation is ... bad (precedes all of us).

app_dart/lib/src/request_handlers/presubmit_luci_subscription.dart
```diff
@@ -121,6 +127,29 @@ final class PresubmitLuciSubscription extends SubscriptionHandler {
...
+      // the LUCI callback.
+      final conclusion = _githubChecksService.conclusionForResult(build.status);
+      await _scheduler.processCheckRunCompleted(
+        cocoon_checks.CheckRun(
+          id: userData.checkRunId,
+          name: builderName,
+          headSha: userData.commit.sha,
+          conclusion: '$conclusion',
+          checkSuite: CheckSuite(
+            id: userData.checkSuiteId,
```
jtmcdole, line 142: Just note that `check_suite_id` can be null when the flag is first flipped; there will be outstanding checkruns for ~50 minutes depending on the time of day we flip the flag.

I _think_ that will still work as the update checkrun REST endpoint doesn't appear to require the check_suite_id:
https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28#update-a-check-run

app_dart/config.yaml
jtmcdole: Looks like this was done - agreed on timing, I like to see the code + flag off, and then a different commit with the feature turned on. Makes for easier rollbacks.

jtmcdole (approved):
Looks good to me!

PR #4842, opened 2025-08-22
Provide a `'content_hash'` LUCI property for engine try-builds

matanlurey:
Towards https://github.com/flutter/flutter/issues/174244.

I was trying to debug why, on presubmit (example: https://github.com/flutter/flutter/pull/174236#issuecomment-3212328062), we were not uploading engine artifacts by content-hash (meaning my changes to a tool to download artifacts from a content-hash instead of a git commit SHA were not working).

This is my best attempt to, if available, provide a `'content_hash'` property, similar to what is being done in the [release builder](https://flutter.googlesource.com/recipes/+/55c85973bd371f314a4f7cebc3f18bdf3db6f8a6/recipes/release/release_builder.py#234) and [in the merge queue ](https://github.com/flutter/cocoon/blob/1bf9fb9d226c79655388ef9d6ee7ce372ceea0d4/app_dart/lib/src/service/scheduler.dart#L705-L707) (though, the merge queue implementation also seems broken). It should start providing a property, which [the `recipes/engine_v2/builder` recipe will use to upload it to GCS](https://flutter.googlesource.com/recipes/+/55c85973bd371f314a4f7cebc3f18bdf3db6f8a6/recipes/engine_v2/builder.py#289).

Added minimal tests - it's so hard to incrementally test `Scheduler`, we should refactor this file at some point 😢

### DIFF
--- a/app_dart/lib/src/service/scheduler.dart
+++ b/app_dart/lib/src/service/scheduler.dart
@@ -403,6 +403,7 @@ class Scheduler {
         // When running presubmits for a fusion PR; create a new staging document to track tasks needed
         // to complete before we can schedule more tests (i.e. build engine artifacts before testing against them).
         final EngineArtifacts engineArtifacts;
+        final String? contentHash;
         if (isFusion) {
           await CiStaging.initializeDocument(
             firestoreService: _firestore,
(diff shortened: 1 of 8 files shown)


app_dart/lib/src/service/scheduler.dart
```diff
@@ -417,15 +418,18 @@ class Scheduler {
           // release candidate build, where the engine artifacts are built
           // via the dart-internal builder.
           engineArtifacts = const EngineArtifacts.usingExistingEngine();
+          contentHash = await _contentAwareHash.getHashByCommitSha(sha);
```
jtmcdole, line 421: Are you trying to look up the engine artifacts for the sha at build time? I don' tthink this will work because the sha here is the PR sha which hasn't been built yet, so it won't exist in firestore.

matanlurey:
This is not a safe approach. Closing.

PR #4844, opened 2025-08-22
Do not provide FPEV for release-branch presubmits that do not touch engine

matanlurey:
I believe this will fix cases such as https://github.com/flutter/flutter/pull/174306 where:

(a) The engine is not touched
(b) We expect the `flutter` tool to resolve an engine for us, i.e. using a content-aware hash

PR #4845, opened 2025-08-22
Always provide a git SHA of an engine to use if the try-bots built the engine for the commit

matanlurey:
https://github.com/flutter/cocoon/pull/4844 was partially correct.

However, PRs such as https://github.com/flutter/flutter/pull/174310 are still blocked because they touch the engine, and try-bots (as of currently) only upload engines to a git SHA, not a content hash. This PR relaxes the try-bot logic to continue to provide the current SHA as "this is the engine" if the engine is being built from source.

Implicitly, it will no longer provide `FLUTTER_PREBUILT_ENGINE_VERSION=` for the `master` branch, if and only if the engine is not being touched. That means `flutter <any command>` will start to download the engine using the content hash, which should be fine.


harryterkelsen (approved):
LGTM

PR #4846, opened 2025-08-25
Let the flutter tool resolve engines on post-submit

matanlurey:
We weren't sure why we are still doing this post content-aware hasing.

### DIFF
--- a/app_dart/lib/src/service/luci_build_service.dart
+++ b/app_dart/lib/src/service/luci_build_service.dart
@@ -875,10 +875,6 @@ class LuciBuildService {
       processedProperties['is_fusion'] = 'true';
       if (commit.branch != Config.defaultBranch(Config.flutterSlug)) {
         processedProperties.addAll({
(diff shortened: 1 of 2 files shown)


matanlurey:
Ping @jtmcdole.

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -875,10 +874,6 @@ class LuciBuildService {
       processedProperties['is_fusion'] = 'true';
       if (commit.branch != Config.defaultBranch(Config.flutterSlug)) {
         processedProperties.addAll({
-          // For release candidates, let the flutter tool pick the right engine.
```
jtmcdole, line 878: OH - This was "not master"; so did we do this to enable experimental branches as we don't have a merge queue? In which case, content hashing would solve this so long as they don't make engine builds...

matanlurey: Hmm.

Experimental branches _do_ make engine builds though. What should we do instead?

PR #4853, opened 2025-09-15
Change PR validation to verify PR base expiration

ievdokdm:
Change PR validation to verify that PR base is no more than X days old, not the head

Issue: [170476](https://github.com/flutter/flutter/issues/170476)


matanlurey:
So, this might be OK, but let's talk it out loud for a moment:

This will "fix" the problem where the following occurs:
```txt
abc123 HEAD 2025-09-01
abc122
abc121
```

And someone opens up a PR, on 2025-09-15, _based_ on `abc123` (2025-09-01). This is likely to occur on a release branch (days or weeks _can_ go by without any changes), and unlikely to occur on `master`, so I think that's generally fine. I can't think of a scenario where the PR is <14 days old, but we'd still _want_ to block it.

matanlurey (approved):

PR #4858, opened 2025-09-30
feat: do NOT skip presubmit optimizations for content hashing changes

jtmcdole:
this isn't an engine change, but we do need to build and test engines because the hash just isn't there.  this is a rare condition.

### DIFF
--- a/dashboard/lib/views/build_dashboard_page.dart
+++ b/dashboard/lib/views/build_dashboard_page.dart
@@ -136,58 +136,63 @@ class BuildDashboardPageState extends State<BuildDashboardPage> {
         child: Material(
           color: Colors.transparent,
           child: FocusTraversalGroup(
-            child: Column(
-              mainAxisSize: MainAxisSize.min,
-              children: <Widget>[
-                if (_smallScreen)
-                  ..._buildRepositorySelectionWidgets(context, buildState),
-                AnimatedBuilder(
-                  animation: buildState,
-                  builder: (context, child) {
-                    final isAuthenticated =
-                        buildState.authService.isAuthenticated;
-                    return TextButton(
-                      onPressed: isAuthenticated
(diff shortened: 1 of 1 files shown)


app_dart/lib/src/service/scheduler/files_changed_optimization.dart
```diff
@@ -102,6 +102,8 @@ final class FilesChangedOptimizer {
     'release-candidate-branch.version',
   );
 
+  static final _engineFilePaths = RegExp(r'^(DEPS|engine/.*|bin/internal/content_aware_hash\.(ps1|sh))$');
```
zanderso, line 105: This might be a bit overly paranoid, but if the name of the script is every changed, how will we remember to update it here?

jtmcdole: If the name of the script is ever changed:
1. It will be a HUGE pain in the keister across multiple tools
2. That PR would only land if someone was also changing the engine in that PR - for the reason I need this to land here (optimization would  be triggered and pre submits would fail to run).
3. If not # 2 (because DEPS or something else was changed) - the very next time someone changes the script it will fail for the same reasons as # 2 - so still safe.

I'm honestly not concerned with this.

jtmcdole:
dangit. running dart analyze locally has no problem, its a breaking change somewhere else.

app_dart/lib/src/service/scheduler/files_changed_optimization.dart
ievdokdm, line 105: Do we follow a common order of class members where fields are defined before methods?

jtmcdole: Not according to this code - the statics are here at the bottom.

PR #5015, opened 2026-04-03
Gracefully handle failure to acknowledge a pubsub message.

eyebrowsoffire:
If the pubsub message has expired for whatever reason, we should catch the exception and continue with the rest of the code path.

### DIFF
--- a/auto_submit/lib/request_handling/pubsub.dart
+++ b/auto_submit/lib/request_handling/pubsub.dart
@@ -59,15 +59,21 @@ class PubSub {
   ///
   /// The PubSub system can remove the relevant messages from the subscription.
   Future<void> acknowledge(String subscription, String ackId) async {
-    final Client httpClient = await clientViaApplicationDefaultCredentials(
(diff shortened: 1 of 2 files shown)


auto_submit/lib/request_handling/pubsub.dart
```diff
@@ -59,15 +59,21 @@ class PubSub {
...
+      );
+      final pubsubApi = pubsub.PubsubApi(httpClient);
+      final ackIds = <String>[ackId];
+      final acknowledgeRequest = pubsub.AcknowledgeRequest(ackIds: ackIds);
+      await pubsubApi.projects.subscriptions.acknowledge(
+        acknowledgeRequest,
+        '${Config.pubsubSubscriptionsPrefix}/$subscription',
+      );
+    } catch (e) {
+      log.warning(
```
ievdokdm, line 74: I think log.error is better when we suppress any exception

ievdokdm (approved):
LGTM

PR #5022, opened 2026-04-15
Create a separate "Flutter Presubmits" checkrun for the unified checkrun flow.

eyebrowsoffire:
This allows us to add "Flutter Presubmits" to the required checks for release candidate branches as well, and have a less confusing name (since release candidate branches don't use the merge queue).

Merge Queue Guard is still created but immediately completed when unified checkruns are created, because this is a required check.

Similarly, "Flutter Presubmits" is always created, and if the unified checkruns are disabled, we immediately complete it as well. This allows us to make this a required check for the `master` branch and the release candidate branches as well.

### DIFF
--- a/app_dart/lib/src/service/scheduler.dart
+++ b/app_dart/lib/src/service/scheduler.dart
@@ -117,6 +117,10 @@ class Scheduler {
       'If you need to merge your PR without tests (a rare situation, typically '
       'an emergency), then you can use the `emergency` label.';
 
+  /// Briefly describes what the "Flutter Presubmits" check is for.
+  static const String kFlutterPresubmitsDescription =
+      'Flutter Presubmits unified check run.';
+
   /// Ensure [commits] exist in Cocoon.
   ///
   /// If the commit already exists, it is ignored.
@@ -332,6 +336,10 @@ class Scheduler {
 
     final slug = pullRequest.base!.repo!.slug();
 
+    final isUnifiedCheckRun = _config.flags.isUnifiedCheckRunFlowEnabledForUser(
+      pullRequest.user!.login!,
+    );
+
     // The MQ only waits for "required status checks" before deciding whether to
     // merge the PR into the target branch. This required check added to both
     // the PR and to the merge group, and so it must be completed in both cases.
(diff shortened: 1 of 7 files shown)


app_dart/lib/src/service/scheduler.dart
```diff
@@ -837,8 +848,43 @@ $s
...
+          _config,
+          slug,
+          headSha,
+          Config.kFlutterPresubmitsName,
+          output: const CheckRunOutput(
+            title: Config.kFlutterPresubmitsName,
+            summary: kFlutterPresubmitsDescription,
+          ),
+          detailsUrl: isUnifiedCheckRun ? detailsUrl : null,
+        );
```
jtmcdole, line 865: Do we need to create both?  Is the idea that we'll have "MQG" and "PSG"?

app_dart/lib/src/service/scheduler.dart
```diff
@@ -837,8 +848,43 @@ $s
...
+
+    if (isUnifiedCheckRun) {
+      // Skip MQ Guard
+      await _githubChecksService.githubChecksUtil.updateCheckRun(
+        _config,
+        slug,
+        mqGuard,
+        status: CheckRunStatus.completed,
+        conclusion: CheckRunConclusion.success,
```
jtmcdole, line 874: Prefer using neutral or skipped?

app_dart/lib/src/service/scheduler.dart
```diff
@@ -837,8 +848,43 @@ $s
...
+      );
+      return flutterPresubmits;
+    } else {
+      // Skip Flutter Presubmits
+      await _githubChecksService.githubChecksUtil.updateCheckRun(
+        _config,
+        slug,
+        flutterPresubmits,
+        status: CheckRunStatus.completed,
+        conclusion: CheckRunConclusion.success,
```
jtmcdole, line 884: ditto neutral/skipped

jtmcdole (approved):

jtmcdole (commented):
Left some comments - mostly nits.

jtmcdole (approved):

app_dart/lib/src/service/scheduler.dart
eyebrowsoffire: The reason we need to create both is that both will be required checks for the master branch.

app_dart/lib/src/service/scheduler.dart
eyebrowsoffire: I'm not sure how the required checks are configured. Will they still allow the PR to merge if the status is neutral or skipped?

eyebrowsoffire: It seems like the answer is "sometimes": https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks

Whether the check blocks when it is skipped depends on where it comes from, but it actually doesn't say anything about when its status is set programatically through the GH API. I think I'm just going to leave these as is, so that we don't break anything for anyone.

PR #5025, opened 2026-04-18
Made it so dismissed approvals don't block the autosubmit bot.

### LINKED ISSUE #185224: `autosubmit` blocks merging on stale "request changes" pr
## example

https://github.com/flutter/flutter/pull/184984

## reproduction steps

1) X authors PR
1) Y reviews PR with request changes
1) X makes changes
1) Y approves changes
1) X fixes merge conflicts (which invalidates all previous approvals)
1) Z approves changes

In this state the PR just has one review that is an approval, but it will fail to merge the PR with the following message:

```
autosubmit label was removed for flutter/flutter/184984, because This PR has not met approval requirements for merging. Changes were requested by {Y}, please make the needed changes and resubmit this PR.
The PR author is a member of flutter-hackers and needs 1 more review(s) in order to merge this PR.

- Merge guidelines: A PR needs at least one approved review if the author is already part of flutter-hackers or two member reviews if the author is not a member of flutter-hackers before re-applying the autosubmit label. Reviewers: If you left a comment approving, please use the "approve" review action instead.
```

## observed behavior

The PR is blocked from merging, despite the PR having only one active review which is approval.

## expected behavior

Since Y's last review was approval, and the PR has an active review from Z, it should be merged.

gaaclarke:
fixes https://github.com/flutter/flutter/issues/185224

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/auto_submit/lib/validations/approval.dart
+++ b/auto_submit/lib/validations/approval.dart
@@ -188,7 +188,8 @@ class Approver {
           _remainingReviews++;
         }
         _reviewAuthors.add(authorLogin);
-      } else if (state == DISMISSED_STATE && !_reviewAuthors.contains(authorLogin)) {
+      } else if (state == DISMISSED_STATE &&
+          !_reviewAuthors.contains(authorLogin)) {
         _reviewAuthors.add(authorLogin);
       }
     }
--- a/auto_submit/lib/validations/approval.dart
+++ b/auto_submit/lib/validations/approval.dart
@@ -180,15 +180,17 @@ class Approver {
         if (_remainingReviews > 0) {
           _remainingReviews--;
         }
+        _reviewAuthors.add(authorLogin);
       } else if (state == CHANGES_REQUESTED_STATE &&
           !_reviewAuthors.contains(authorLogin)) {
         _changeRequestAuthors.add(authorLogin);
         if (_remainingReviews < targetReviewCount) {
           _remainingReviews++;
         }
+        _reviewAuthors.add(authorLogin);
+      } else if (state == DISMISSED_STATE && !_reviewAuthors.contains(authorLogin)) {
+        _reviewAuthors.add(authorLogin);
       }
-
-      _reviewAuthors.add(authorLogin);
     }
 
     _approved =
--- a/auto_submit/test/validations/approval_test.dart
+++ b/auto_submit/test/validations/approval_test.dart
@@ -397,7 +397,7 @@ void main() {
         githubService.isTeamMemberMockMap['jmagman'] = true;
         githubService.isTeamMemberMockMap['Z'] = true;
 
-        const String bugScenario = '''
+        const bugScenario = '''
       {
         "repository": {
           "pullRequest": {
@@ -459,7 +459,7 @@ void main() {
         githubService.isTeamMemberMockMap['jmagman'] = true;
         githubService.isTeamMemberMockMap['Z'] = true;
 
-        const String scenario = '''
+        const scenario = '''
       {
         "repository": {
           "pullRequest": {
--- a/auto_submit/test/validations/approval_test.dart
+++ b/auto_submit/test/validations/approval_test.dart
@@ -390,12 +390,14 @@ void main() {
       );
     });
 
-    test('DISMISSED review should override previous CHANGES_REQUESTED', () async {
-      githubService.isTeamMemberMockMap['author1'] = true;
-      githubService.isTeamMemberMockMap['jmagman'] = true;
-      githubService.isTeamMemberMockMap['Z'] = true;
+    test(
+      'DISMISSED review should override previous CHANGES_REQUESTED',
+      () async {
+        githubService.isTeamMemberMockMap['author1'] = true;
+        githubService.isTeamMemberMockMap['jmagman'] = true;
+        githubService.isTeamMemberMockMap['Z'] = true;
 
-      const String bugScenario = '''
+        const String bugScenario = '''
       {
         "repository": {
           "pullRequest": {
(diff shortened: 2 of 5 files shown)


auto_submit/lib/validations/approval.dart
```diff
@@ -180,15 +180,17 @@ class Approver {
...
         _changeRequestAuthors.add(authorLogin);
         if (_remainingReviews < targetReviewCount) {
           _remainingReviews++;
         }
+        _reviewAuthors.add(authorLogin);
+      } else if (state == DISMISSED_STATE && !_reviewAuthors.contains(authorLogin)) {
+        _reviewAuthors.add(authorLogin);
       }
-
-      _reviewAuthors.add(authorLogin);
```
b-luk, line 191: With the existing code, `_reviewAuthors` is updated outside of the if/else for APPROVED or CHANGES_REQUESTED.

So for the DISMISSED case, `_reviewAuthors` is updated and would override a previous CHANGES_REQUESTED already, without your change. Right?

If so, I don't see how this would behave any differently for the case we see in https://github.com/flutter/flutter/pull/185226. It seems like the existing code already lets DISMISSED override a previous CHANGES_REQUESTED.

Does the new "DISMISSED review should override previous CHANGES_REQUESTED" test that you added pass even when running without the changes here?

Maybe the issue is something else, like DISMISSED reviews getting filtered out before getting to this code?

b-luk: This code
https://github.com/flutter/cocoon/blob/710bc917c7dc1c6af48a0954533fd88a2b825c44/auto_submit/lib/requests/graphql_queries.dart#L70
seems like something that is requesting only APPROVED and CHANGES_REQUESTED reviews, and wouldn't return DISMISSED ones. Maybe this is related.

Caveat: I have no context for any of this code, so it's possible this is something totally unrelated.

b-luk:
Also, not directly related to this change, but I notice that the UI lets me dismiss a REQUESTED_CHANGES review:
image

Does this mean that a REQUESTED_CHANGES review doesn't actually block an author from submitting, because they can always just dismiss that review? So REQUESTED_CHANGES is more of a suggestion than something that is enforceed?

Or is this UI something most people don't have, and I just see it because I have elevated permissions?

gaaclarke:
> Does this mean that a REQUESTED_CHANGES review doesn't actually block an author from submitting, because they can always just dismiss that review? So REQUESTED_CHANGES is more of a suggestion than something that is enforceed?

The requirement is that you have approval, not that there is no requests changes.  Dismissing requests changes should be used respectfully obviously.

auto_submit/lib/validations/approval.dart
gaaclarke: Yea, after I put this into review too I thought it odd too.  I had convinced myself because the test looks exactly like our case.  I'm not sure it would fail without the change though.  I'll look through this again.

gaaclarke: Okay, looked at it.  I think you were right about graphql_queries.  Done.

b-luk: So we no longer even need the changes in this file, right? Because the graphql query ensures that the only review states we get here are APPROVED, REQUESTED_CHANGES, and DISMISSED. And with the original code, the behavior for all 3 of these cases is identical to the behavior when your changes are included.

Even if the changes in this file are optional, I think it's a little clearer to explicitly handle all 3 possible cases. So this seems fine to me.

auto_submit/test/validations/approval_test.dart
```diff
@@ -389,6 +389,113 @@ void main() {
...
+        }
+      }
+      ''';
+
+      final result = await computeValidationResult(bugScenario);
+
+      expect(result.result, isTrue);
+    });
+
+    test('COMMENTED review should NOT clear previous CHANGES_REQUESTED', () async {
```
b-luk: This is testing an impossible scenario, because our graphql query won't return a COMMENTED review. Maybe we can remove this test because it can be misleading.

b-luk (approved):
This looks good to me, but I have no experience with this code or with how it works. So please also get approval from someone else who is more familiar with this.

gaaclarke:
friendly ping @jtmcdole

auto_submit/lib/validations/approval.dart
```diff
@@ -180,15 +180,18 @@ class Approver {
         if (_remainingReviews > 0) {
           _remainingReviews--;
         }
+        _reviewAuthors.add(authorLogin);
       } else if (state == CHANGES_REQUESTED_STATE &&
           !_reviewAuthors.contains(authorLogin)) {
         _changeRequestAuthors.add(authorLogin);
         if (_remainingReviews < targetReviewCount) {
           _remainingReviews++;
         }
```
jtmcdole, line 189: Nit: Interesting: if you get 3 approvals, you can ignore the the changes requested status. I think that's fine, but it would be nice to have a strong DO NOT SUBMIT or weigh Owners vs random teammates trying to be helpful.

jtmcdole (approved):

gaaclarke:
I tried merging main because `ci.yaml validation` was stuck.  Now things that were passing are failing.  I think cocoon ci is sick.

auto_submit/lib/validations/approval.dart
b-luk: 3 approvals does not cause a changes requested status to be ignored.

The code still calls `_changeRequestAuthors.add(authorLogin)`, and the `_approved =` check at the end requires `_changeRequestAuthors` to be empty.

I do think this code is pretty confusing though.

jtmcdole:
> I tried merging main because `ci.yaml validation` was stuck. Now things that were passing are failing. I think cocoon ci is sick.

I suspect GitHub was sick.

PR #5027, opened 2026-04-21
feat: add app lifecycle observer to pause and resume presubmit timer

### LINKED ISSUE #184529: Presubmit checks page should refresh upon reactivation
Right now, the presubmit checks page refreshes the status of tasks every 30 seconds. If the tab becomes inactive, the polling stops. The polling should restart when the tab is made inactive again.

ievdokdm:
add app lifecycle observer to pause and resume presubmit timer
fix: https://github.com/flutter/flutter/issues/184529


eyebrowsoffire (approved):
LGTM

PR #5028, opened 2026-04-22
Overhaul CICD label management for presubmits

eyebrowsoffire:
This change overhauls the state management for the `CICD` label to ensure trusted users can run CI seamlessly and external contributors require explicit approval.

- PR Opened: Privileged users (non-draft) auto-trigger CI and get the label. Non-privileged or draft PRs go to awaiting state.
- Changes Pushed (Synchronize): Privileged users trigger CI if label is present. Non-privileged users have the label removed and go to awaiting state.
- Edits: Removed automatic label adding on base branch changes.
- Consolidated usage of `_scheduleIfMergeable` to ensure consistent conflict checking and label processing.
- Added flowchart documentation in `app_dart/docs/cicd_flowchart.md`.

Once this is merged and deployed, we can remove the "remove CICD label" workflows from the other repos since that's all handled in cocoon now.

### DIFF
--- a/app_dart/docs/cicd_flowchart.md
+++ b/app_dart/docs/cicd_flowchart.md
@@ -1,0 +1,33 @@
+# CICD Label Flowchart
+
+This flowchart describes the logic for managing the `CICD` label in Cocoon for running presubmit CI.
+
+```mermaid
+flowchart TD
+    PR_Opened([Event: PR Opened]) --> Is_Draft{Is Draft?}
+    
+    Is_Draft -- No --> Is_Privileged_Open{Is Privileged?}
+    Is_Draft -- Yes --> Create_Awaiting[Create Awaiting Check Run]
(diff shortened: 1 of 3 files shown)


eyebrowsoffire:
Also, when looking a the `cicd_flowchart.md` file in GitHub, you can click the "show rich diff" button and you'll be able to actually see the fancy mermaid diagram rendered.

gmackall (approved):
LGTM, though as I mentioned offline, I think the call to `githubService.isTeamMember` that `_isPrivilegedUser` depends upon is currently not working because (I believe, based on inspecting the logs I added in https://github.com/flutter/cocoon/pull/5023) the token we are using doesn't have permission to read org membership. The autosubmit label is doing this successfully, so I think we just need to look at the permissions diff and iron that out

PR #5035, opened 2026-04-28
Refactor CICD label management to use per-PR sequential queue

eyebrowsoffire:
This change overhauls the management of the `CICD` label in Cocoon to address race conditions and non-idempotent operations that caused issues in the previous attempt.

Key changes:
- Introduced `PullRequestManager` to handle events for a single PR sequentially via an operation queue.
- Moved PR-specific logic from `webhook_subscription.dart` to `PullRequestManager`, thinning out the webhook handler.
- Implemented new state machine logic for `opened` and `synchronize` events to handle privileged users and draft PRs correctly.
- Made `_scheduleIfMergeable` idempotent by tracking `scheduledSha` in Firestore to prevent duplicate triggering of presubmits for the same SHA.
- Refactored `PullRequestManager` creation to be synchronous in the cache, queueing the asynchronous initialization to avoid creation race conditions.
- Restored `cicd_flowchart.md` from a previous commit to document the flow.
- Fixed all static analysis issues and ensured all tests pass.


app_dart/docs/cicd_flowchart.md
```diff
@@ -0,0 +1,33 @@
+# CICD Label Flowchart
+
+This flowchart describes the logic for managing the `CICD` label in Cocoon for running presubmit CI.
+
+```mermaid
```
jtmcdole, line 5: (nit / resolved) for others wanting a state machine:

```plantuml
@startuml
state PR {
  state Waiting
  Waiting : entry / Create Awaiting Check Run
  Waiting : exit / resolve(awaiting check)

  state Running
  Running: entry/ Start Presubmits
  
  state if_priv2 <<choice>>
  Running --> if_priv2: onPushed
  if_priv2 --> Running: isPrivileged && labeled(CICD)
  if_priv2 --> Waiting : default: remove(CICD)

  state if_draft <<choice>>
  PR --> if_draft: openned

  if_draft --> Waiting : isDraft
  if_draft --> Running: isPrivileged
  if_draft --> Waiting : default

  Waiting --> Running: labeled(CICD)
}
@enduml
```

app_dart/lib/src/model/firestore/pull_request_state.dart
```diff
@@ -0,0 +1,85 @@
...
+  /// The PR number.
+  int get number => int.parse(fields[kNumberField]!.integerValue!);
+
+  set number(int value) {
+    fields[kNumberField] = Value(integerValue: '$value');
+  }
+
+  /// Generates the document ID for a PR.
+  static String getDocumentId(RepositorySlug slug, int number) {
+    return '${slug.owner}_${slug.name}_$number';
```
jtmcdole: Use Unicode "Information Separator" (US) https://www.compart.com/en/unicode/U+001F instead of `_`. Highly unlikely owner / name will have ␟ in the name.
Or optionally Unicode Character `␟` (U+241F)


```suggestion
    return '${slug.owner}␟${slug.name}␟$number';
```

app_dart/lib/src/model/firestore/pull_request_state.dart
```diff
@@ -0,0 +1,85 @@
...
+      fields[kIsPrivilegedField] = Value(booleanValue: value);
+    }
+  }
+
+  /// The latest SHA that we have processed for this PR.
+  String? get latestSha => fields[kLatestShaField]?.stringValue;
+
+  set latestSha(String? value) {
+    if (value != null) {
+      fields[kLatestShaField] = Value(stringValue: value);
```
jtmcdole: firestore_extensions has "toValue" for most data types.  So `value.toValue()`

```suggestion
      fields[kLatestShaField] = value.toValue();
```

app_dart/lib/src/model/firestore/pull_request_state.dart
```diff
@@ -0,0 +1,85 @@
...
+      ..fields = doc.fields!
+      ..name = doc.name!;
+  }
+
+  /// Whether the PR author is a privileged user (roller or flutter-hacker).
+  bool? get isPrivileged => fields[kIsPrivilegedField]?.booleanValue;
+
+  set isPrivileged(bool? value) {
+    if (value != null) {
+      fields[kIsPrivilegedField] = Value(booleanValue: value);
```
jtmcdole:
```suggestion
      fields[kIsPrivilegedField] = value.toValue();
```

app_dart/lib/src/model/firestore/pull_request_state.dart
```diff
@@ -0,0 +1,85 @@
...
+      fields[kLatestShaField] = Value(stringValue: value);
+    }
+  }
+
+  /// The SHA for which we have scheduled presubmits.
+  String? get scheduledSha => fields[kScheduledShaField]?.stringValue;
+
+  set scheduledSha(String? value) {
+    if (value != null) {
+      fields[kScheduledShaField] = Value(stringValue: value);
```
jtmcdole:
```suggestion
      fields[kScheduledShaField] = value.toValue();
```

app_dart/lib/src/model/firestore/pull_request_state.dart
```diff
@@ -0,0 +1,85 @@
...
+    }
+  }
+
+  /// The repository slug associated with the pull request.
+  RepositorySlug get slug => RepositorySlug.fromJson(
+    json.decode(fields[kSlugField]!.stringValue!) as Map<String, Object?>,
+  );
+
+  set slug(RepositorySlug value) {
+    fields[kSlugField] = json.encode(value.toJson()).toValue();
```
jtmcdole, line 77: Firebase does have a map value; this could just be `value.toJson().toValue()` - or I think we handle this generally everywhere else as `slug.fullName` and `RepositorySlug.full(fullNameString)`. Up to you.

app_dart/lib/src/model/firestore/pull_request_state.dart
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  /// The PR number.
+  int get number => int.parse(fields[kNumberField]!.integerValue!);
+
+  set number(int value) {
+    fields[kNumberField] = Value(integerValue: '$value');
```
jtmcdole:
```suggestion
    fields[kNumberField] = value.toValue();
```

app_dart/lib/src/service/pull_request_manager.dart
```diff
@@ -0,0 +1,1011 @@
...
+      }
+    });
+  }
+
+  bool get isQueueEmpty => _queue.isEmpty;
+
+  /// Adds an event to the queue to be processed sequentially.
+  Future<void> _addEvent(Future<void> Function() operation) {
+    return _queue.add(operation);
+  }
```
jtmcdole: I guess nit: why wasn't this just `_queue.add(op)` everywhere else if there's nothing special going on in this function?  If you keep it, consider putting it above its first use for readability.

app_dart/lib/src/service/pull_request_manager.dart
```diff
@@ -0,0 +1,1011 @@
...
+    required this.firestore,
+    required this.config,
+    required this.scheduler,
+    required this.gerritService,
+    required this.pullRequestLabelProcessorProvider,
+  });
+
+  /// Kicks off asynchronous initialization in the queue.
+  void _initialize(PullRequestEvent event) {
+    _addEvent(() async {
```
jtmcdole: I think this queue turns the entire instance into a serial state with Firestore. Dart might be single threaded in execution, but the second any of the `await` calls block, the AppEngine instance can handle another webhook from GitHub.  Granted the current "version" has 2 to 6 instances depending on needs.

My concerns
1. GitHub events being delivered to two different instances are going to race on data - this queue only "queues" events to the same instance, not all instances. For that you'd need to use memcache or redis.
2. AppEngine might timeout pending requests if they are at the end of the queue waiting for scheduling.

app_dart/lib/src/service/pull_request_manager.dart
```diff
@@ -0,0 +1,1011 @@
...
+        completer.completeError(e, s);
+      } finally {
+        _pendingOperations--;
+      }
+    });
+    return completer.future;
+  }
+}
+
+/// LRU Cache for [PullRequestManager] instances.
```
jtmcdole:
```suggestion
/// LRU Cache for [PullRequestManager] instances keyed by RepositorySlug and PR Number
```

app_dart/lib/src/service/pull_request_manager.dart
```diff
@@ -0,0 +1,1011 @@
...
+  PullRequestManagerCache({
+    required this.capacity,
+    required this.firestore,
+    required this.config,
+    required this.scheduler,
+    required this.gerritService,
+    required this.pullRequestLabelProcessorProvider,
+  });
+
+  PullRequestManager getOrCreate(PullRequestEvent event) {
```
jtmcdole: Dartdoc for global method. Reading this partway through the review had me originally believing `PullRequestManager` interleaved all requests.

app_dart/lib/src/service/pull_request_manager.dart
jtmcdole: Hmm, looking up cloud docs, it might be 1 hour for response. Luckily at this point we've already handed the original GitHub request to pubsub so we can work with it on our own time (unless pubsub has limits I'm unaware of). Also considering this "manager" is keyed to slug+prnumber, there's less contention.

app_dart/lib/src/service/pull_request_manager.dart
```diff
@@ -0,0 +1,1011 @@
...
+
+  /// The SHA for which we have scheduled presubmits.
+  String? get scheduledSha => _scheduledSha;
+  set scheduledSha(String? value) {
+    if (_scheduledSha != value) {
+      _scheduledSha = value;
+      _addEvent(persist);
+    }
+  }
+
```
jtmcdole, line 370: 1: Should the setters of this class be public members?
2: You've got two writes to the scheduled for handleOpened and handleSynchronized - but the first write _might_ win and have both latestSha and scheduledSha in the same write (second write a noop)

It might be way more complicated to address 2 and moot if triggerPresubmitTargets takes a sufficiently long time and since this is all in the slug-pr-event-queue.

jtmcdole (changes requested):
I am still working through this, but wanted to post some early comments.

app_dart/docs/cicd_flowchart.md
jtmcdole: image

app_dart/lib/src/service/pull_request_manager.dart
eyebrowsoffire: The idea is that each PR handles its events serially. Even if its event handling logic is asynchronous, we won't get interleaved logic from different event handlers.

As you pointed out, this is not universally true for all event handling globally. Event handlers from different PRs might interleave their logic. However, I think that is okay. The goal here is not to serialize *all* interactions with firestore at a global level. The PR manager only reads and mutates state for its particular PR, so it's only important that that particular PR's state is consistent at the beginning and the end of each handled event.

app_dart/lib/src/service/pull_request_manager.dart
eyebrowsoffire: Good points. Answer to (1) I think is no, I should probably change them not to be public setters. In fact, probably no public getters either, since you don't really have a guarantee externally that you're going to be able to get a view of the state that is consistent. Answer to (2), persist is on the serial queue, so it's probably consistent as it is, but you're right that it might double up on writes. Probably simple enough to have a dirty bit and then have the operation queue flush the changes at the end of each event handler.

app_dart/docs/cicd_flowchart.md
eyebrowsoffire: Thanks. Maybe I'll just keep both diagrams in the doc since the mermaid one can be rendered inline by GitHub natively.

app_dart/lib/src/service/pull_request_manager.dart
jtmcdole: Could keep track of "last write" time and "update scheduled time" and if the latter > former, just know it will get persisted when update happens?

app_dart/lib/src/service/pull_request_manager.dart
jtmcdole: (1) is still a concern, The PR could still be handled by two instances depending on which appengine instance gets an event, e.g.:

PubSub([
  openned
  edited
  synchronized
]) -> AppEngine#1, AppEngine#2, AppEngine#3

They are different processes and thus would have 3 different Managers with different states depending on where they are in the process.

eyebrowsoffire: Oh, I see what you're saying... we're not directly handling the HTTP requests from the webhook, there is a layer that bounces it through pubsub first and then different instances might pick up the different pubsub events. *sigh*

app_dart/lib/src/service/pull_request_manager.dart
eyebrowsoffire: I killed this LRU cache when I reworked this to work with multiple instances.

app_dart/lib/src/service/pull_request_manager.dart
eyebrowsoffire: This also is gone now.

app_dart/lib/src/service/pull_request_manager.dart
eyebrowsoffire: This was removed.

app_dart/lib/src/service/pull_request_manager.dart
eyebrowsoffire: I ended up fixing this by using our redis cache to ensure serial processing of PR events across instances. In order to avoid in-memory consistency issues, I removed the LRU cache and now have it load the document from firestore when processing each event.

app_dart/lib/src/model/firestore/pull_request_state.dart
```diff
@@ -0,0 +1,85 @@
...
+      ..name = doc.name!;
+  }
+
+  /// Whether the PR author is a privileged user (roller or flutter-hacker).
+  bool? get isPrivileged => fields[kIsPrivilegedField]?.booleanValue;
+
+  set isPrivileged(bool? value) {
+    if (value != null) {
+      fields[kIsPrivilegedField] = value.toValue();
+    }
```
jtmcdole, line 46: I don't know when we'd use it, but would you want:

```dart
        } else {
          fields.remove(kIsPrivilegedField);
        }
```

We could do that for kLatestShaField and kScheduledShaField as well - again, not sure when we'd use it, but maybe worth having?

app_dart/lib/src/service/pull_request_manager.dart
```diff
@@ -0,0 +1,1132 @@
...
+    required Future<T> Function() action,
+  }) async {
+    final pr = event.pullRequest!;
+    final slug = event.repository!.slug();
+    final prNumber = pr.number!;
+    final lockKey = 'pr_lock_${slug.owner}_${slug.name}_$prNumber';
+    final lockValue = Uint8List.fromList('l'.codeUnits);
+
+    // Attempt to acquire lock
+    final existingLock = await cache.getOrCreate(
```
jtmcdole: I had to go read `getOrCreateWithLocking()` to realize it is only "locked" for the local instance, and not across multiple instances.

The dashboard relies on uuid; so you could:

```dart
//  uuid: ^4.5.3
final uuid = Uuid();

final value = Uuid.parseAsByteList((uuid.v4()));
final existing = await cache.getOrCreate('pr_locks', lockKey, createFn: () => value, ttl: const Duration(minutes: 5));

/// I DO NOT KNOW IF THIS IS TRUE: set() doesn't have good dartdoc and I don't know if the value that is set is returned.
if (existingLock != value) {
       throw const ServiceUnavailable('PR is locked by another instance'); 
}

// no need to call "set()"

```

jtmcdole (changes requested):
Really just one change, the UUID get/set in one step.  LGTM otherwise

jtmcdole:
Looking at this again right now

app_dart/lib/src/service/pull_request_manager.dart
jtmcdole: I think this will still hit a snag if two webhooks come in at the same time, even if we use Uuid - both instances will see null, both will then write the uuid, and both will get back the Uuid that they wrote.  `neat_cache` is just really... poor for this.

do we want to consider using `package:redis` or `package:redis_dart_client`? They both appear to have "setex()" which is exactly what we're trying to do here? Or do we want to ask the dart team if they will add setex to neat_cache?

app_dart/lib/src/service/pull_request_manager.dart
```diff
@@ -0,0 +1,1131 @@
...
+    );
+
+    if (!lockAcquired) {
+      throw const ServiceUnavailable('PR is locked by another instance');
+    }
+
+    try {
+      return await action();
+    } finally {
+      await cache.purge('pr_locks', lockKey);
```
jtmcdole: If `action()` takes longer than 5 minutes, the lock is already lost because of the TTL. Another webhook will get the lock and it will have it purged by this call.

Use the UUID I mentioned in a previous comment. Then you can read it here and if it's the same as what you locked above, you can purge it.  If not, we should totally log a warning.

jtmcdole (changes requested):
I'm still concerned with "locking" atomicity. We could argue that two webhooks won't come in at the same time - but I believe we we get many hooks delivered in quick succession:

Same PR, messages at the same time:
*  aaf18850-6e79-11f1-8fbc-2d43f66bf3b9 pull_request.opened 2026-06-22 13:33:47
*  ab005560-6e79-11f1-84ca-8bb5774e4c64 pull_request.labeled 2026-06-22 13:33:47


```
SEARCH("`pull_request` 188347")
protoPayload.resource = "/api/github-webhook-pullrequest"
```
why labeled came in before create? No clue! But it looks like they got handled by different `instanceId`'s

app_dart/lib/src/service/cache_service.dart
```diff
@@ -2,198 +2,449 @@
...
+      return false;
     }
 
-    return value;
+    try {
+      await block();
+      return true;
+    } finally {
+      await _releaseLock(lockKey, token);
+    }
```
jtmcdole, line 116: This isn't angry at you for not return false at the bottom?  e.g. block() throws, you are returning what?

jtmcdole (approved):

app_dart/lib/src/service/cache_service.dart
eyebrowsoffire: I think the answer is that you aren't returning anything. It's not a catch block, so the exception is still going to propagate.

PR #5043, opened 2026-05-04
[Gemini Log Analyzer] Part 3: Backend Core (Analyze Logs API & Service)

ievdokdm:
This is Part 3 of the stacked Pull Request series implementing the Gemini Log Analyzer feature.

**Part 1 & Part 2 Backend Infrastructure have been merged!** This PR now contains only the core API implementation on top of main.

This PR implements the core backend service and API request handlers for analyzing logs:
- Instantiates the Genkit client and configures plugins (using `googleAI`) in `gae_server.dart`.
- Implements `LogAnalyzer` and its Genkit integration to fetch and parse LUCI build logs.
- Adds the `analyze_logs` request handler and exposes it as the `/api/analyze-logs` endpoint.

### Stacked PR Series:
- Part 1: Backend Foundation (Models & Setup) (Merged)
- Part 2: Backend Infrastructure (Config & Database Data Flow) (Merged)
- **Part 3: Backend Core API (Genkit Logic & Handlers)** (This PR)
- Part 4: Frontend Dashboard UI

PR #5044, opened 2026-05-04
[Gemini Log Analyzer] Part 4: Frontend Integration (Dashboard UI)

ievdokdm:
This is Part 4 of the stacked Pull Request series implementing the Gemini Log Analyzer feature.

**Depends on Part 3:** #5043 (Note: Targets `main` due to GitHub cross-repo limits, but chains to #5043 via description).

This PR integrates the feature into the Cocoon Dashboard UI:
- Adds "Log Analysis" tab to display the Gemini log analyzer results when available.
- Adds "Analyze Logs with Gemini" action button to trigger the API from the presubmit dashboard.
- Implements spellchecking, permission-checking, and error-handling for the trigger action.

### Stacked PR Series:
- Part 1: Backend Foundation (Models & Setup) (Merged)
- Part 2: Backend Infrastructure (Config & Database Data Flow) (#5045)
- Part 3: Backend Core API (Genkit Logic & Handlers) (#5043)
- **Part 4: Frontend Dashboard UI** (This PR)

### DIFF
--- a/app_dart/lib/src/service/flags/dynamic_config.dart
+++ b/app_dart/lib/src/service/flags/dynamic_config.dart
@@ -37,6 +37,7 @@ final class DynamicConfig {
     ciYaml: CiYamlFlags.defaultInstance,
     contentAwareHashing: ContentAwareHashing.defaultInstance,
     closeMqGuardAfterPresubmit: false,
+    enableGeminiLogAnalysis: false,
     unifiedCheckRunFlow: UnifiedCheckRunFlow.defaultInstance,
     dynamicTestSuppression: false,
     geminiModel: 'gemini-3-flash-preview',
@@ -62,6 +63,10 @@ final class DynamicConfig {
   @JsonKey()
   final bool closeMqGuardAfterPresubmit;
 
+  /// Whether to enable Gemini log analysis feature.
+  @JsonKey()
+  final bool enableGeminiLogAnalysis;
+
   /// Flags related tp unified check-run flow configuration.
   @JsonKey()
   final UnifiedCheckRunFlow unifiedCheckRunFlow;
@@ -79,6 +84,7 @@ final class DynamicConfig {
     required this.ciYaml,
     required this.contentAwareHashing,
     required this.closeMqGuardAfterPresubmit,
+    required this.enableGeminiLogAnalysis,
     required this.unifiedCheckRunFlow,
     required this.dynamicTestSuppression,
     required this.geminiModel,
@@ -92,6 +98,7 @@ final class DynamicConfig {
     CiYamlFlags? ciYaml,
     ContentAwareHashing? contentAwareHashing,
     bool? closeMqGuardAfterPresubmit,
+    bool? enableGeminiLogAnalysis,
     UnifiedCheckRunFlow? unifiedCheckRunFlow,
     bool? dynamicTestSuppression,
     String? geminiModel,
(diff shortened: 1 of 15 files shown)


dashboard/lib/state/presubmit.dart
```diff
@@ -69,6 +69,9 @@ class PresubmitState extends ChangeNotifier {
   String? get jobNameFilter => _jobNameFilter;
   String? _jobNameFilter;
 
+  String get insuficientPermissionMessage =>
```
jtmcdole: insufficientPermissionMessage

jtmcdole:
No clue why there is a discrepancy between VSCode checkout of your branch and github's render:

 
presubmit_job.dart line 251: 
q: Should you use int.tryParse just to avoid throwing if the field is unparsable?
you could also cheat and just do: `int.tryParse(fields[fieldBuildId]?.integerValue ?? '')`

app_dart/lib/src/model/firestore/presubmit_job.dart: line 314:
```dart
fields[fieldBuildId] = build.id.toValue();
```

app_dart/lib/src/request_handlers/analyze_logs.dart: line 59:
Can we validate this slug is in the configured repos (there is a set in Config)? I get UnifiedCheckRun.getLatestPresubmitGuardForPrNum might also check, but it would be nice to have this explicit.

app_dart/tool/local_server.dart: line 26:
you could allow someone working locally to test with their own key:
```dart
  secretManager.putString('APP_DART_GEMINI_LOG_ANALYZER_KEY', Platform.environment['APP_DART_GEMINI_LOG_ANALYZER_KEY'] ?? dummy_key);
```


packages/cocoon_integration_test/lib/src/fakes/fake_config.dart: line 183:
GenKit supports web (according to pub.dev) so you could use the following. Note, this is different from Platform and is instead a dart compile option (-DAPP_DART_GEMINI_LOG_ANALYZER_KEY="$APP_DART_GEMINI_LOG_ANALYZER_KEY")

```dart
      geminiLogAnalyzerKeyValue ?? 
      String.fromEnvironment('APP_DART_GEMINI_LOG_ANALYZER_KEY', defaultValue: 'fake-gemini-key');
```

dashboard/lib/state/presubmit.dart
ievdokdm: fixed

app_dart/lib/src/service/flags/dynamic_config.dart
```diff
@@ -62,6 +63,10 @@ final class DynamicConfig {
   @JsonKey()
   final bool closeMqGuardAfterPresubmit;
 
+  /// Whether to enable Gemini log analysis feature.
+  @JsonKey()
```
jtmcdole, line 67: Note to self: I don't think we need @JsonKey on every field if we're not overriding the defaults.

dashboard/lib/service/appengine_cocoon.dart
```diff
@@ -553,6 +553,43 @@ class AppEngineCocoonService implements CocoonService {
...
 
+  @override
+  Future<CocoonResponse<void>> analyzeLogs({
+    required String? idToken,
+    required String repo,
+    required int pr,
+    required int buildId,
+    String owner = 'flutter',
+  }) async {
+    if (idToken == null || idToken.isEmpty) {
```
jtmcdole: nifty pattern style catch:

```suggestion
    if (idToken case null || '') {
```

jtmcdole (approved):

PR #5049, opened 2026-05-06
fix: make some APIs cron-only

jtmcdole:
The biggest part of this change is trying to keep DashboardAuthentication somewhat flexible. You can extract a "cron only" authentication from the chain of responsibility.


jason-simmons (approved):

app_dart/bin/gae_server.dart
```diff
@@ -146,6 +146,7 @@ Future<void> main() async {
       bigQuery: bigQuery,
       cache: cache,
       dashboardAuthProvider: dashboardAuthProvider,
+      cronAuthProvider: dashboardAuthProvider.onlyCron,
```
ievdokdm: can it be simply initialized with the instance of `DashboardCronAuthentication` ?

ievdokdm: like `cronAuthProvider: const DashboardCronAuthentication(),`

jtmcdole: Changed to ChainOfAuth and I think its clearer.

jtmcdole:
cc: @straka

PR #5050, opened 2026-05-06
When properties are extracted from a target configuration, exclude properties that affect source locations used by LUCI recipes

jason-simmons:
These properties should only be set by Cocoon's build scheduler.

See http://b/509842706


jtmcdole (approved):
I could go 50/50 on this.

jtmcdole:
cc: @straka

PR #5052, opened 2026-05-14
Added Evan to unified checkrun beta testers

ievdokdm:
Added Evan to unified check run beta testers

### DIFF
--- a/app_dart/config.yaml
+++ b/app_dart/config.yaml
@@ -36,3 +36,4 @@ unifiedCheckRunFlow:
     - eyebrowsoffire
     - andywolff
     - camsim99
+    - ewall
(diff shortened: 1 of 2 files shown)


app_dart/config.yaml
```diff
@@ -36,3 +36,4 @@ unifiedCheckRunFlow:
     - eyebrowsoffire
     - andywolff
     - camsim99
+    - ewall
```
jtmcdole: This file is not being updated: https://github.com/flutter/cocoon/blob/main/app_dart/lib/src/generated_config.dart

jtmcdole (changes requested):

app_dart/config.yaml
ievdokdm: updated

walley892 (approved):
Very excited

jtmcdole (approved):

PR #5064, opened 2026-06-16
store state entirely in query parameters

### LINKED ISSUE #186357: Changing pr or sha outside of current pr doesnt work
**Steps:** 

- open dashboard for pr or sha or both like: https://flutter-dashboard.appspot.com/#/presubmit?repo=flutter&pr=183913&sha=d390280b57e155a9fe383a2fef40212a1507ca73
- try to change url to https://flutter-dashboard.appspot.com/#/presubmit?pr=181051&repo=flutter or https://flutter-dashboard.appspot.com/#/presubmit?repo=flutter&sha=1685ecc1d6296511722c3c0fa93eaa748f48f57a

**Actual result:** page stay on https://flutter-dashboard.appspot.com/#/presubmit?repo=flutter&pr=183913&sha=d390280b57e155a9fe383a2fef40212a1507ca73

**Expected result:** page switched to https://flutter-dashboard.appspot.com/#/presubmit?pr=181051&repo=flutter&sha=1685ecc1d6296511722c3c0fa93eaa748f48f57a

ievdokdm:
fix: https://github.com/flutter/flutter/issues/186357

### DIFF
--- a/dashboard/lib/state/presubmit.dart
+++ b/dashboard/lib/state/presubmit.dart
@@ -18,10 +18,7 @@ import '../service/firebase_auth.dart';
 /// This state manages the data for a specific Pull Request (PR) or commit SHA,
 /// including available SHAs, guard status, and individual job results.
 class PresubmitState extends ChangeNotifier {
-  PresubmitState({
-    required this.cocoonService,
-    required this.authService,
-  }) {
+  PresubmitState({required this.cocoonService, required this.authService}) {
     authService.addListener(onAuthChanged);
   }
 
--- a/dashboard/lib/state/presubmit.dart
+++ b/dashboard/lib/state/presubmit.dart
@@ -21,9 +21,11 @@ class PresubmitState extends ChangeNotifier {
   PresubmitState({
     required this.cocoonService,
     required this.authService,
-    this.pr,
-    this.sha,
+    String? pr,
+    String? sha,
   }) {
+    if (pr != null) _queryParams['pr'] = pr;
+    if (sha != null) _queryParams['sha'] = sha;
     authService.addListener(onAuthChanged);
     if (pr != null || sha != null) {
       unawaited(fetchIfNeeded());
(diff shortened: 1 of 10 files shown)


dashboard/lib/state/presubmit.dart
```diff
@@ -58,17 +102,14 @@ class PresubmitState extends ChangeNotifier {
   bool get isRerunningAll => _isRerunningAll;
   bool _isRerunningAll = false;
 
-  /// The currently selected task statuses for filtering.
```
jtmcdole, line 61: nit: interesting that it removed the dartdoc. Add back?

dashboard/lib/state/presubmit.dart
```diff
@@ -18,29 +18,73 @@ import '../service/firebase_auth.dart';
...
   }
 
   final CocoonService cocoonService;
   final FirebaseAuthService authService;
 
-  /// The repository name (e.g., 'flutter', 'engine').
-  String repo = 'flutter';
+  Map<String, String> get queryParameters {
+    final params = <String, String>{};
+    params['repo'] = _repo;
```
jtmcdole: we could extract magic strings to const values.

As long as we're not too crazy, we could also use JsonSerializable to describe the parameters as an object and then use toJson() to set the Query Parameters.

jtmcdole: hack concept - but it could let us make for better RPC handling as well.

```dart
import 'dart:convert';

extension JsonQueryEncoder on Map<String, dynamic> {
  /// Converts a JSON Map into a URL-safe query string (e.g., ?key=value&key2=value2)
  String toQueryString() {
    if (isEmpty) return '';

    final queryParts = <String>[];

    for (final entry in entries) {
      // Safely encode the key
      final key = Uri.encodeQueryComponent(entry.key);
      final value = entry.value;

      // Skip nulls to keep the query clean
      if (value == null) continue;

      // Handle lists (e.g., ?tags=a&tags=b)
      if (value is Iterable) {
        for (final item in value) {
          queryParts.add('$key=${Uri.encodeQueryComponent(item.toString())}');
        }
      } else {
        queryParts.add('$key=${Uri.encodeQueryComponent(value.toString())}');
      }
    }

    return '?${queryParts.join('&')}';
  }
}

extension JsonQueryDecoder on String {
  /// Converts a query string back into a Map<String, dynamic> for fromJson()
  Map<String, dynamic> toQueryMap() {
    if (trim().isEmpty) return {};

    // Remove the leading '?' if it exists
    final cleanString = startsWith('?') ? substring(1) : this;

    // splitQueryString automatically URL-decodes keys and values!
    final decodedMap = Uri.splitQueryString(cleanString);

    // Optional: Query params are inherently Strings. If your JsonSerializable 
    // model expects ints or booleans, you need to cast them back from Strings.
    return decodedMap.map((key, value) => MapEntry(key, _parseValue(value)));
  }

  /// Attempts to restore basic types lost in the query string conversion
  dynamic _parseValue(String value) {
    if (value.toLowerCase() == 'true') return true;
    if (value.toLowerCase() == 'false') return false;
    
    final intValue = int.tryParse(value);
    if (intValue != null) return intValue;
    
    final doubleValue = double.tryParse(value);
    if (doubleValue != null) return doubleValue;
    
    return value; // Leave as string
  }
}
```

ievdokdm: added comments, moved strings to constants and enums, realigned properties

dashboard/lib/state/presubmit.dart
ievdokdm: added comments

jtmcdole (approved):

PR #5068, opened 2026-06-16
Change Docker build context to `cloud_build/flutter_agy_docker`

jtmcdole:
This works if you run the cloud build command in the folder, but Cloud Console needs to know the directory though

After this change we can setup a schedule.


ievdokdm (approved):
LGTM

PR #5070, opened 2026-06-16
Rename “Flutter Presubmits” check to “Dashboard Checks” check

ievdokdm:
Rename “Flutter Presubmits” check to “Dashboard Checks” check and add some description


jtmcdole (approved):
I think you sill have some comments with "Unified Check Run" - but other than that, LGTM

PR #5088, opened 2026-06-25
fix: add install of jq to orchestrator

jtmcdole:
Two AI reviews suggested "jq" was included; it was in fact a lie.

Ran the cloud-sdk image locally in docker with this and it was able to work. Triggered remotely and waiting for confirmation


ievdokdm (approved):
never always the better!

PR #5091, opened 2026-06-26
Remove revert bot from autosubmit.

eyebrowsoffire:
This functionality has been replaced by a GitHub action, so we're going to remove this feature from cocoon.


app_dart/lib/src/service/config.dart
```diff
@@ -152,13 +152,6 @@ interface class Config extends DynamicallyUpdatedConfig {
     return defaultBranches[slug] ?? kDefaultBranchName;
   }
 
-  // The name of the bot that generates automated revert requests.
-  String get autosubmitBot => 'auto-submit[bot]';
```
jtmcdole, line 156: Odd that the name of the variable and the description are different. Was this used anywhere else other than revert checking?

auto_submit/bin/server.dart
```diff
@@ -40,10 +39,6 @@ Future<void> main() async {
         '/check-pull-request',
         CheckPullRequest(config: config, cronAuthProvider: authProvider).run,
       )
-      ..get(
```
jtmcdole, line 43: You need to remove this as well: https://github.com/flutter/cocoon/blob/994995d408ef4d6a45d001b4b7fa48ae0d67fbe6/cron.yaml#L67-L70

auto_submit/test/revert/revert_discord_message_test.dart
```diff
@@ -1,140 +0,0 @@
-// Copyright 2024 The Flutter Authors. All rights reserved.
-// Use of this source code is governed by a BSD-style license that can be
```
jtmcdole, line 2: Didn't realize we had discord messages for reverts. We do have a discord bot that we could use to publish messages if we see them, but I don't know what their value is.

jtmcdole (changes requested):
one change: make sure to remove the cron scheduler that calls the soon to be missing API.  I guess we should time this landing since that will take immediate effect on Cloud.

app_dart/lib/src/service/config.dart
eyebrowsoffire: My understanding is that `auto-submit[bot]` is the name of the app associated with the key that the auto submit bot uses to interact with GitHub. The revert bot was implemented a while ago, and I think the original plan was to add more functionality (like cherry-picks) to the bot so the "autosubmit bot" would open PRs of multiple different kinds (although to be honest even this is kind of confusing, and I'm not sure why the decision was made to put this functionality into the auto submit as opposed to the dashboard service or something). However, I think the infra team at the time discovered that it was easier to implement the cherry-pick stuff through github actions, so they did so, and reverts remained the only PRs created by the auto submit bot. Now that we're removing that functionality, I think there really aren't any PRs that will be opened by `auto-submit[bot]` anymore, and automated pull requests are either created by GitHub workflows (which use the `flutteractionsbot` account) or the skia autoroller (which uses an account called `engine-flutter-autroll` I think).

auto_submit/test/revert/revert_discord_message_test.dart
eyebrowsoffire: Hmm yeah. I'm sure I could add this functionality into the GitHub workflow if we really cared to, but maybe we should just wait and see if anyone actually cares.

jtmcdole (commented):
I missed this in the initial review:

* RevertPullRequestMutation  is still present in  auto_submit/lib/requests/graphql_queries.dart. [auto_submit/lib/requests/graphql_queries.dart](https://github.com/flutter/cocoon/blob/994995d408ef4d6a45d001b4b7fa48ae0d67fbe6/auto_submit/lib/requests/graphql_queries.dart#L117)
* Which I think means `RevertPullRequest` and `RevertPullRequestData` still
* `createPullRequestCommentMock` has `  "body": "A reason for requesting a revert of flutter/cocoon/0 could not be found or the reason was not properly formatted. Begin a comment with 'Revert reason:' to tell the bot why this issue is being reverted."
`
* Look for `reverts label` comment: https://github.com/flutter/cocoon/blob/994995d408ef4d6a45d001b4b7fa48ae0d67fbe6/auto_submit/lib/service/pull_request_validation_service.dart#L25 and https://github.com/flutter/cocoon/blob/994995d408ef4d6a45d001b4b7fa48ae0d67fbe6/auto_submit/lib/service/revert_request_validation_service.dart#L38

jtmcdole (approved):

PR #5103, opened 2026-07-10
Implement lock-free versioned (revisionId) and Set-based Redis caching for /tasks

eyebrowsoffire:
To eliminate read bottlenecks against the Firestore /tasks (kTaskCollectionId = 'tasks') collection without distributed locking (tryLock), this change introduces an optimistic, multi-tier caching layer across Cocoon (Task, FirestoreQueries, and CacheService):

1. Lock-Free Versioned (revisionId) Payload Caching (tasks subcache):
   - Every Task document contains a monotonically increasing revisionId integer.
   - CacheService.insertVersioned(subcacheName, entries) executes an atomic check-and-set via Redis Lua script (EVAL) verifying entry.revisionId > cachedRevisionId before updating payloads.
   - Chunked batching (batchSize = 20) guarantees <0.1ms Lua execution without starving concurrent readers (MGET).

2. Native Set-Based Commit Task Indexing (tasks_by_commit_ids):
   - Replaces monolithic JSON array strings with native Redis Sets (SMEMBERS, SADD) via getSet, updateSet, and addToSetIfExists.
   - Leverages two domain invariants:
     1. Immutable commitSha: Task mutations (updateCacheForTaskMutations) never change a commit's task list membership and do not touch or invalidate tasks_by_commit_ids.
     2. Monotonically Increasing Set: Task attempts only grow over time, so unioning IDs via SADD (updateCacheForCreatedTasks) safely converges without lock coordination.

3. Partial Cache Recovery (_queryTasksByCommitCached):
   - When reading commit tasks, _queryTasksByCommitCached retrieves cached task IDs (SMEMBERS) and performs a batch lookup (MGET).
   - If individual payloads are expired or missing (missingDocIds), _queryTasksByCommitCached selectively queries getDocument only for missing entries and merges them with foundTasks, eliminating redundant full-commit queries.

4. Modular & Explicit Cache Handlers (FirestoreQueries):
   - _cacheTaskDocuments(tasks): Reusable helper for versioned Task payload insertions.
   - _fetchAndCacheCommitTasks(commitSha): Reusable slow-path query helper shared across read-miss and task-creation-miss paths.
   - updateCacheForCreatedTasks(tasks) and updateCacheForTaskMutations(writes): Explicit domain handlers replacing generic ad-hoc cache invalidations.
   - _subcacheRecentTasksIds and _queryRecentTasksByNameCached removed in favor of _cacheTaskDocuments warming on query execution (~65 lines simplified).


app_dart/lib/src/service/cache_service.dart
```diff
@@ -183,6 +225,162 @@ class RedisCacheService extends CacheService {
...
+      local numKeys = tonumber(ARGV[1])
+      for i = 1, numKeys do
+        local key = KEYS[i]
+        local val = ARGV[1 + (i - 1) * 3 + 1]
+        local rev = tonumber(ARGV[1 + (i - 1) * 3 + 2])
+        local ttl = tonumber(ARGV[1 + (i - 1) * 3 + 3])
+
+        local revKey = "revisions/" .. key
+        local existingRev = tonumber(redis.call("get", revKey) or 0)
+        if rev > existingRev or (not redis.call("exists", key)) then
```
jtmcdole: Gemini suggests a truthy take for "if the incoming revision is newer, OR if the main key is missing entirely, write it.":

I don't know REDIS or Lua all that well, but it suggests that "key exists == 1" and that "not 0" is false.

```dart
        if rev > existingRev or redis.call("exists", key) == 0 then
          redis.call("set", key, val, "PX", ttl)
          redis.call("set", revKey, rev, "PX", ttl)
        end
```

app_dart/lib/src/service/firestore.dart
```diff
@@ -49,7 +52,216 @@ final kFieldMapRegExp = RegExp(
...
+///      ([CacheService.updateSet], [CacheService.addToSetIfExists]) safely converges to database state without locks.
+///
+/// ### 3. Partial Cache Recovery (`_queryTasksByCommitCached`)
+/// - When querying all tasks for a commit, we retrieve the cached Set of document IDs (`SMEMBERS`) and perform a batch
+///   payload lookup across `tasks` (`MGET`).
+/// - If any individual task entries have expired or are missing from `tasks`, successfully retrieved cached tasks
+///   ([foundTasks]) are preserved immediately. Missing document IDs (`missingDocIds`) are queried individually via
+///   [batchGetDocuments], inserted into `tasks` via `insertVersioned`, and combined with [foundTasks]—avoiding a full table query.
+/// - If `tasks_by_commit_ids` is missing entirely (`docIds.isEmpty`), we execute a full query (`_fetchAndCacheCommitTasks`),
+///   populate `tasks` for all items, and initialize the commit Set via [CacheService.updateSet].
```
jtmcdole: I think this could have used a mermaidjs sequence diagram or something. It was a lot to read and I'm fairly certain I understood about 25% of it via reading redis.io docs.

app_dart/lib/src/service/firestore.dart
```diff
@@ -49,7 +52,216 @@ final kFieldMapRegExp = RegExp(
...
 mixin FirestoreQueries {
+  CacheService? get cache => null;
+  Future<Document> getDocument(String name, {Transaction? transaction});
+  Future<List<Document>> batchGetDocuments(List<String> names);
+
+  static const String _subcacheTasksByCommitIds = 'tasks_by_commit_ids';
+
+  static Uint8List _serializeTask(Task task) {
+    return Uint8List.fromList(
+      utf8.encode(
```
jtmcdole: utf8.encode already returns a Uint8List: https://api.dart.dev/dart-convert/Utf8Codec/encode.html

app_dart/lib/src/service/firestore.dart
```diff
@@ -49,7 +52,216 @@ final kFieldMapRegExp = RegExp(
...
+        json.encode(Document(name: task.name, fields: task.fields).toJson()),
+      ),
+    );
+  }
+
+  static Task _deserializeTask(Uint8List data) {
+    final jsonMap = json.decode(utf8.decode(data)) as Map<String, dynamic>;
```
jtmcdole: inner matan: `Map<String, Object?>`

app_dart/lib/src/service/firestore.dart
```diff
@@ -49,7 +52,216 @@ final kFieldMapRegExp = RegExp(
...
+    return Task.fromDocument(Document.fromJson(jsonMap));
+  }
+
+  /// Caches the individual [Task] payloads into the `tasks` subcache.
+  Future<void> _cacheTaskDocuments(
+    List<Task> tasks, {
+    Duration ttl = const Duration(hours: 12),
```
jtmcdole: is `const Duration(hours: 12)` a number we every think we'll want to tweak or change on the fly? If so, consider adding it to the dynamic config with a default of 12 hours.

app_dart/lib/src/service/firestore.dart
```diff
@@ -49,7 +52,216 @@ final kFieldMapRegExp = RegExp(
...
+    final entries = tasks
+        .map(
+          (t) => VersionedCacheEntry(
+            key: p.basename(t.name!),
+            value: _serializeTask(t),
+            revisionId: t.revisionId,
+            ttl: ttl,
+          ),
+        )
+        .toList();
```
jtmcdole:
```suggestion
    final entries = [
        for (final task in tasks)
          VersionedCacheEntry(
            key: p.basename(task.name!),
            value: _serializeTask(task),
            revisionId: task.revisionId,
            ttl: ttl,
          ),
    ];
```

app_dart/lib/src/service/firestore.dart
```diff
@@ -49,7 +52,216 @@ final kFieldMapRegExp = RegExp(
...
+    final docIds = tasks.map((t) => p.basename(t.name!)).toSet();
+    if (docIds.isNotEmpty && cache != null) {
+      await cache!.updateSet(
+        _subcacheTasksByCommitIds,
+        commitSha,
+        docIds,
+        ttl: const Duration(hours: 12),
+      );
+    }
+    return tasks;
```
jtmcdole:
```suggestion
    final docIds = {
        for (final task in tasks)
          p.basename(task.name!)
    };
    // parallel await.
    await [
        _cacheTaskDocuments(tasks),
        if (docIds.isNotEmpty && cache != null)
            cache!.updateSet(
                _subcacheTasksByCommitIds,
                commitSha,
                docIds,
                ttl: const Duration(hours: 12),
            ),
    ].wait;
    return tasks;
```

app_dart/lib/src/service/firestore.dart
```diff
@@ -49,7 +52,216 @@ final kFieldMapRegExp = RegExp(
...
+  }
+
+  /// Explicitly updates the cache when new [Task]s are created.
+  Future<void> updateCacheForCreatedTasks(List<Task> tasks) async {
+    if (cache == null || tasks.isEmpty) return;
+    await _cacheTaskDocuments(tasks);
+    for (final task in tasks) {
+      if (task.commitSha.isNotEmpty) {
+        final docId = p.basename(task.name!);
+        final added = await cache!.addToSetIfExists(
```
jtmcdole: how many tasks in this list? Could this be parallelized?  If its too many, I don't want to risk an API failure though.

app_dart/lib/src/service/firestore.dart
```diff
@@ -49,7 +52,216 @@ final kFieldMapRegExp = RegExp(
...
+        continue;
+      }
+
+      if (write.update != null &&
+          write.updateMask == null &&
+          write.update!.fields != null &&
+          write.update!.fields!.containsKey(Task.fieldCreateTimestamp)) {
+        try {
+          mutatedTasks.add(Task.fromDocument(write.update!));
+        } catch (_) {}
```
jtmcdole: empty try/catch{} - do we need a log? What would go wrogn from mutatedTasks.add?

app_dart/lib/src/service/firestore.dart
```diff
@@ -49,7 +52,216 @@ final kFieldMapRegExp = RegExp(
...
+            if (write.updateTransforms != null) {
+              for (final transform in write.updateTransforms!) {
+                if (transform.fieldPath == Task.fieldRevisionId &&
+                    transform.increment != null) {
+                  cachedTask.incrementRevisionId();
+                }
+              }
+            } else {
+              cachedTask.incrementRevisionId();
+            }
```
jtmcdole: Gemini claims there's a bug here.

  • The Bug: In updateCacheForTaskMutations, if a write has no updateTransforms (i.e., a standard partial write), you unconditionally fall into the else block and call cachedTask.incrementRevisionId(). However, if the client performed a partial write without a field transform increment on Firestore, Firestore's revisionId was not incremented. This leaves the cache with a higher revisionId than the database. Any subsequent cache invalidation/eviction followed by a database read will result in a stale/older version in Firestore that cannot update the cache because newRevisionId > cachedRevisionId will be false.
  • The Fix: Only increment the cache's revisionId if you are certain a write-transform occurred or if the write payload explicitly contains the newly updated revisionId within update.fields.

app_dart/lib/src/service/firestore.dart
```diff
@@ -49,7 +52,216 @@ final kFieldMapRegExp = RegExp(
 );
 
 @visibleForTesting
+/// Mixin that encapsulates Firestore query and caching operations across Cocoon.
```
jtmcdole, line 54: This huge block is focused on Task cachiing, but its in this mixing. Can we extract some of these queries to extensions or separate classes to keep API concerns distinct?

app_dart/lib/src/service/firestore.dart
```diff
@@ -153,31 +364,143 @@ mixin FirestoreQueries {
...
     final orderMap = {Task.fieldCreateTimestamp: kQueryOrderDescending};
     final documents = await query(
       kTaskCollectionId,
       filterMap,
       orderMap: orderMap,
+      limit: limit,
+      transaction: transaction,
+    );
+    final tasks = documents.map(Task.fromDocument).toList();
+    if (cacheResults && transaction == null && cache != null) {
```
jtmcdole: I want a dynamic flag that overrides this and keeps the old path. That way we're talking about ~5 minutes to rollout or rollback.

app_dart/lib/src/service/firestore.dart
```diff
@@ -153,31 +364,143 @@ mixin FirestoreQueries {
...
+
+  Future<List<Task>> _queryTasksByCommit({
+    required String commitSha,
+    String? name,
+    TaskStatus? status,
+    int? limit,
+    Transaction? transaction,
+  }) async {
+    if (transaction == null && cache != null) {
+      return await _queryTasksByCommitCached(
```
jtmcdole, line 344: same here; caching shoudl be flagged in case we need to turn it off - we should set a date on when to remove the flag.

app_dart/lib/src/service/cache_service.dart
```diff
@@ -328,6 +530,119 @@ class InMemoryCacheService extends CacheService {
...
+      for (final entry in entries) {
+        final cacheKey = '$subcacheName/${entry.key}';
+        final existing = _entries[cacheKey];
+        if (existing == null ||
+            existing.isExpired ||
+            entry.revisionId > existing.revisionId) {
+          if (_entries.length >= maxEntries &&
+              !_entries.containsKey(cacheKey)) {
+            _entries.remove(_entries.keys.first);
+          }
```
jtmcdole: this makes the third copy of this code; move to a new function and update the other two identical versions of this

app_dart/lib/src/service/cache_service.dart
```diff
@@ -328,6 +530,119 @@ class InMemoryCacheService extends CacheService {
...
+          _entries[cacheKey] = _InMemoryCacheEntry(
+            entry.value,
+            DateTime.now().add(entry.ttl),
+            revisionId: entry.revisionId,
+          );
```
jtmcdole, line 611: Do you know if this cache service was trying to do an LRU cache? If so, shouldn't we remove and re-insert when read/writing to the cache so it updates in insertion order?

app_dart/lib/src/model/firestore/task.dart
```diff
@@ -136,17 +140,52 @@ final class Task extends AppDocument<Task> {
...
+        json.encode(Document(name: task.name, fields: task.fields).toJson()),
+      ),
+    );
+  }
+
+  static Task _deserializeTask(Uint8List data) {
+    final jsonMap = json.decode(utf8.decode(data)) as Map<String, dynamic>;
+    return Task.fromDocument(Document.fromJson(jsonMap));
+  }
+
```
jtmcdole: These are duplicates in firestore.dart.

jtmcdole (changes requested):

app_dart/lib/src/service/firestore.dart
eyebrowsoffire: You're right, we should parallelize. But also, I think we can do better than that. We should pregroup the tasks by their commit sha and then do one API call/database fallback per commit sha rather than per-task.

app_dart/lib/src/service/firestore.dart
eyebrowsoffire: I missed this... Gemini is lazy. I will fix this.

app_dart/lib/src/service/firestore.dart
eyebrowsoffire: Yeah actually I think this logic is definitely wrong. I don't think ever should be just incrementing the revisionId in the cache itself, we need to actually use the database as a source of truth for what the revisionId is after the write. I'll try reworking this section.

jtmcdole:
re-looking at this now.

app_dart/lib/src/service/cache_service.dart
```diff
@@ -183,6 +225,162 @@ class RedisCacheService extends CacheService {
...
+    }
+  }
+
+  @override
+  Future<void> insertVersioned(
+    String subcacheName,
+    List<VersionedCacheEntry> entries,
+  ) async {
+    if (entries.isEmpty) return;
+    const insertVersionedScript = '''
```
jtmcdole, line 262: I am not a LUA person, so I'm totally falling back on Gemini here:  keys (tasks/$docId) are passed as the KEYS arguments to EVAL. However, inside insertVersionedScript, the script dynamically constructs and accesses revision keys via Lua string concatenation.

> Redis specification dictates that ALL keys accessed by a Lua script MUST be explicitly declared in the KEYS array.

I **think** we ok if we have only one redis instance? But the robot says "just pass revKey along with key in KEYS, and then use `{tasks/$docId}`.

Again, I'm not the expert here.

app_dart/lib/src/service/cache_service.dart
```diff
@@ -183,6 +225,162 @@ class RedisCacheService extends CacheService {
...
+      final values = await _runCommand(
+        (client) => client.send_object(['MGET', ...redisKeys]),
+      );
+      if (values is! List) {
+        return List.filled(keys.length, null);
+      }
+      return values.map((value) {
+        if (value == null) return null;
+        return base64.decode(value as String);
+      }).toList();
```
jtmcdole:
```suggestion
      return [
        for (final value in values)
          value == null ? null : base64.decode(value as String)
      ];
```

app_dart/lib/src/service/cache_service.dart
```diff
@@ -183,6 +225,162 @@ class RedisCacheService extends CacheService {
...
+  Future<void> insertVersioned(
+    String subcacheName,
+    List<VersionedCacheEntry> entries,
+  ) async {
+    if (entries.isEmpty) return;
+    const insertVersionedScript = '''
+      local numKeys = tonumber(ARGV[1])
+      for i = 1, numKeys do
+        local key = KEYS[i]
+        local val = ARGV[1 + (i - 1) * 3 + 1]
```
jtmcdole: agian, not a lua person, but should `(i - 1) * 3` be pulled out as a local index?

app_dart/lib/src/service/cache_service.dart
```diff
@@ -183,6 +225,162 @@ class RedisCacheService extends CacheService {
...
+  Future<Set<String>> getSet(String subcacheName, String key) async {
+    final redisKey = '$subcacheName/$key';
+    try {
+      final values = await _runCommand(
+        (client) => client.send_object(['SMEMBERS', redisKey]),
+      );
+      if (values is! List || values.isEmpty) {
+        return const {};
+      }
+      return values.map((e) => e.toString()).toSet();
```
jtmcdole:
```suggestion
      return {for (var e in values) '$e'};
```

app_dart/lib/src/service/firestore/task_cache_service.dart
```diff
@@ -0,0 +1,214 @@
...
+          ttl: effectiveTtl,
+        ),
+    ];
+    await cache.insertVersioned('tasks', entries);
+  }
+
+  /// Evicts a deleted task payload from Redis.
+  Future<void> evictTaskPayload(String docId) async {
+    if (!isEnabled) return;
+    await cache.purge('tasks', docId);
```
jtmcdole: Gemini suggests this poisons the cache because:

> When a task document is deleted from Firestore, `evictTaskPayload(docId)` purges `tasks/$docId` and `revisions/tasks/$docId` from Redis. However, **it does not remove `docId` from `tasks_by_commit_ids/$commitSha`**.

app_dart/lib/src/service/firestore/task_cache_service.dart
```diff
@@ -0,0 +1,214 @@
...
+  /// Returns `true` if the set existed and task IDs were added; `false` if the set key is missing.
+  Future<bool> addTasksToCommitSet(
+    String commitSha,
+    Iterable<String> taskIds, {
+    Duration? ttl,
+  }) async {
+    if (taskIds.isEmpty || !isEnabled) return false;
+    var allAdded = true;
+    for (final taskId in taskIds) {
+      final added = await cache.addToSetIfExists(
```
jtmcdole: we easily have tasks.length > 100; so we're calling redis 100 times?

The current "addToSetIfExists" only returns true if the set exists, not if the key was added (missing), so you could do something like:

```lua
  if redis.call("exists", KEYS[1]) == 1 then
    redis.call("sadd", KEYS[1], unpack(ARGV, 2))
    return 1
  end
  return 0
```

app_dart/lib/src/service/cache_service.dart
```diff
@@ -183,6 +225,162 @@ class RedisCacheService extends CacheService {
...
+        ];
+        await _runCommand(
+          (client) => client.send_object([
+            'EVAL',
+            insertVersionedScript,
+            keys.length.toString(),
+            ...keys,
+            ...args,
+          ]),
+        );
```
jtmcdole: Chunking good: +1

you should add this future to "List<Future<void>> chunks" and then use one `await chunks.wait;`

app_dart/lib/src/service/firestore.dart
```diff
@@ -153,31 +315,130 @@ mixin FirestoreQueries {
...
+    tasks ??= await _fetchAndCacheCommitTasks(commitSha);
+
+    var result = tasks.toList();
+    if (name != null) {
+      result = result.where((t) => t.taskName == name).toList();
+    }
+    if (status != null) {
+      result = result.where((t) => t.status == status).toList();
+    }
+    result.sort((a, b) => b.createTimestamp.compareTo(a.createTimestamp));
```
jtmcdole: if the timestamp is the same, maybe also compare against currentAttempt?

jtmcdole (changes requested):
This change is still virtually too larger for me to review alone (~1500 lines).  I'm not a redis or lua expert, so I had to fall back on some gemini prompts to try and tease out.

app_dart/lib/src/service/firestore/task_cache_service.dart
eyebrowsoffire: Honestly, tasks are never deleted and this function is never called in production (it's just in a unit test), so I should just delete it.

app_dart/lib/src/service/cache_service.dart
```diff
@@ -48,6 +65,35 @@ abstract class CacheService {
...
   });
 
+  /// Atomically inserts multiple [entries] into [subcacheName] in a single batch API call
+  /// if and only if their [VersionedCacheEntry.revisionId] is strictly greater than any
+  /// existing cached revision for that key.
+  Future<void> insertVersioned(
+    String subcacheName,
+    List<VersionedCacheEntry> entries,
+  );
+
```
jtmcdole, line 75: Question on the description: is it insert All or Nothing if any extra's revision id is not greater than?

app_dart/lib/src/service/flags/dynamic_config.dart
```diff
@@ -43,6 +43,8 @@ final class DynamicConfig {
     orderedPresubmit: OrderedPresubmit.defaultInstance,
     dynamicTestSuppression: false,
     geminiModel: 'gemini-3-flash-preview',
+    taskCachingEnabled: true,
```
jtmcdole, line 46: As a matter of hygiene (cocoon roles out early in the AM) - I think this flag should be turned off by default, then enabled in the config.yaml when someone is online.  this includes all the places where its defaulted to "true".

app_dart/lib/src/service/firestore.dart
```diff
@@ -49,7 +51,169 @@ final kFieldMapRegExp = RegExp(
 );
 
 @visibleForTesting
+/// Mixin that encapsulates Firestore query and caching operations across Cocoon.
+///
+/// Task caching orchestration is delegated to [TaskCacheService].
+/// See [TaskCacheService] for full architectural documentation and caching invariant flowcharts.
 mixin FirestoreQueries {
+  CacheService? get cache => null;
+  Config? get config => null;
```
jtmcdole, line 60: config should probably not be null - everywhere else it appears to be set. Maybe this is just a firestore.dart quirk?

app_dart/lib/src/service/firestore.dart
```diff
@@ -153,31 +316,139 @@ mixin FirestoreQueries {
       );
     }
 
-    // For tasks, therer is no reason to _not_ order this way.
```
jtmcdole, line 156: does your agent hate comments or something? Why was this removed?

jtmcdole: (hint: my agent hates comments it didn't write)

app_dart/lib/src/service/firestore.dart
```diff
@@ -153,31 +316,139 @@ mixin FirestoreQueries {
...
+  Future<List<Task>> _queryTasksByCommitCached({
+    required String commitSha,
+    String? name,
+    TaskStatus? status,
+    int? limit,
+  }) async {
+    List<Task>? tasks;
+    final docIds = await _taskCache!.getTaskIdsForCommit(commitSha);
+    if (docIds != null && docIds.isNotEmpty) {
+      final lookupResult = await _taskCache!.getTaskPayloads(docIds);
```
jtmcdole, line 370: this can and will return null.  null! is not good.

packages/cocoon_integration_test/lib/src/fakes/fake_firestore_service.dart
```diff
@@ -195,36 +203,31 @@ abstract base class _FakeInMemoryFirestoreService
...
           );
-          continue;
         }
-        // this is most certainly wrong as the real firestore probably(?) updates
-        // the field. We're not using it that way in the tests, so if you find
-        // yourself getting this error - congrats and welcome to the team.
+      } else if (transform.increment case final inc?) {
+        if (field?.integerValue case final oldVal?) {
+          final newVal = int.parse(oldVal) + int.parse(inc.integerValue!);
+          fields[transform.fieldPath!] = Value(integerValue: newVal.toString());
```
jtmcdole, line 220: newVal.toValue()

jtmcdole (changes requested):

app_dart/lib/src/service/luci_build_service.dart
```diff
@@ -1213,10 +1213,7 @@ class LuciBuildService {
     task.resetAsRetry();
```
jtmcdole, line 1213: this changes the document name (_attempt_2?).

Not sure how valid this is, but agent says: "The new retry document ID is never added to tasks_by_commit_ids/$commitSha in Redis. When firestore.dart:354-420 is called, it reads the commit set from Redis (which only contains attempt 1), finds the cached payload, and completely ignores attempt 2 until the 12-hour TTL expire"

vs:
```dart
   if (createdTasks.isNotEmpty) {
      await _firestore.updateCacheForCreatedTasks(createdTasks);
    }
```

app_dart/bin/gae_server.dart
```diff
@@ -43,7 +43,10 @@ Future<void> main() async {
     }
 
     final cache = CacheService.redis();
-    final firestore = await FirestoreService.from(const GoogleAuthProvider());
+    final firestore = await FirestoreService.from(
+      const GoogleAuthProvider(),
+      cache: cache,
```
jtmcdole, line 48: We should probably pass config here?

app_dart/lib/src/request_handlers/rerun_prod_task.dart
```diff
@@ -208,32 +207,28 @@ final class RerunProdTask extends ApiRequestHandler {
...
 
+    final writes = documentsToWrites(tasksToUpdate);
     await Future.wait([
       cancelRunningTasks,
-      _firestore.commit(transaction, documentWrites),
+      _firestore.commit(transaction, writes),
     ]);
+    await _firestore.cacheTaskPayloads(tasksToUpdate);
+    if (createdTasks.isNotEmpty) {
+      await _firestore.updateCacheForCreatedTasks(createdTasks);
```
jtmcdole, line 230: just passing through: "Because createdTasks is a subset of tasksToUpdate, and updateCacheForCreatedTasks internally calls cacheTaskPayloads, those created tasks' payloads are serialized and sent to Redis twice in immediate succession."

app_dart/bin/gae_server.dart
eyebrowsoffire: Can you be more specific about what you mean exactly? Which config?

app_dart/lib/src/service/cache_service.dart
eyebrowsoffire: I updated the comment for clarity. It is not all-or-nothing.

app_dart/lib/src/service/firestore.dart
eyebrowsoffire: Hah, yeah I added it back.

app_dart/lib/src/service/firestore.dart
eyebrowsoffire: I think in practice this only ever gets called when there is an outer check for the task cache, but I reworked it to be safer.

app_dart/lib/src/request_handlers/rerun_prod_task.dart
eyebrowsoffire: Reworked this to avoid the redundancy.

app_dart/lib/src/service/luci_build_service.dart
eyebrowsoffire: Fixed

PR #5107, opened 2026-07-17
fix(auto_submit): Pass requestSha to GitHub merge API

jtmcdole:
Defense in depth - prevent autosubmit from trying to land a pushed (but not approved) pr. 

This is not a real issue since we have branch protection rules, but it's good to have.


yjbanov (approved):
lgtm

PR #5118, opened 2026-08-03
Eliminate redundant `getDocument` errors in Firestore.

eyebrowsoffire:
Looking through our GCP logs, a large portion of our errors are coming from calls to get documents from firestore where the document is not found. In many cases, we actually handle this 404 error gracefully and proceed without issue. However, this interaction still causes Firestore itself to log an error, even though cocoon handles it just fine.

This change adds a convenience method `getDocumentOrNull` which uses the batch document fetch API instead, which always returns a 200, and if the document is not found, just returns an empty list. I went through and found all the places we were catching the 404 error from fetching a document and used this method instead. This should eliminate a lot of these errors in our GCP logs.

### DIFF
--- a/app_dart/lib/src/model/firestore/account.dart
+++ b/app_dart/lib/src/model/firestore/account.dart
@@ -5,7 +5,6 @@
 import 'package:googleapis/firestore/v1.dart' as g;
 import 'package:path/path.dart' as p;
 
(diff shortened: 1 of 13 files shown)


ievdokdm (approved):
LGTM

PR #5142, opened 2026-08-31
do not reRequest CheckRun

ievdokdm:
attempt to solve mysterious merge without unlocked merge queue guard
https://github.com/flutter/flutter/issues/191610


Piinks:
Hey @ievdokdm does this need a review? Would this fix the issues we've been having in flutter/packages?

PR #5143, opened 2026-09-04
Decouple bringup tasks from staging bucket and run in prod

eyebrowsoffire:
Previously, tasks in .ci.yaml marked with bringup: true were scheduled to run in LUCI's staging bucket against staging builders and the staging pool of bots.

Because bringup tasks do not run in presubmit, the only places where they are scheduled are postsubmit and merge queue builds. Running them in the staging bucket is unnecessary; they should run against the standard prod bots and builders like any other task while maintaining their non-blocking behavior (no postsubmit check runs, no gating on tree status).

Staging builders and the staging pool are reserved for manual testing (e.g., via led) of experimental bot configurations and should not be automatically used for normal CI tasks.

This addresses https://github.com/flutter/flutter/issues/192311

### DIFF
--- a/app_dart/lib/src/model/ci_yaml/target.dart
+++ b/app_dart/lib/src/model/ci_yaml/target.dart
@@ -283,11 +283,6 @@ final class Target {
     return _value.name.split(' ').first.toLowerCase();
   }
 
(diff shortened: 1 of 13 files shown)


ievdokdm (approved):
LGTM

PR #5154, opened 2026-09-21
Don't mark dashboard checks as neutral on rerun.

eyebrowsoffire:
This is causing some PRs to autosubmit with failing checks. Ideally, we'd actually rerun failed checks when this is clicked. For now, we should make this a no-op.

This should address the main issue with  https://github.com/flutter/flutter/issues/191610 for now.


ievdokdm:
you also have to remove https://github.com/flutter/cocoon/blob/main/app_dart/lib/src/service/luci_build_service.dart#L448-L483

eyebrowsoffire:
> you also have to remove https://github.com/flutter/cocoon/blob/main/app_dart/lib/src/service/luci_build_service.dart#L448-L483

Is this code path just failing currently (and being caught by the try catch)?

eyebrowsoffire:
After discussing offline: the code path in question just kicks off a `rerequested` event, which triggers the code path I'm deleting. The `rerequested` event doesn't do anything else, so we can get rid of the source of the event.

ievdokdm (approved):
LGTM
with that change developers no longer would be able to add `autosubmit` label for PR where all failed check-runs were re-run requested in https://github.com/flutter/flutter/issues/191005

PR #5155, opened 2026-09-21
[ci] Build ruby CIPD packages with Xcode 17c52

okorohelijah:
Both ruby builders have been failing at `build package` because Homebrew requires Xcode >= 16.0 on macOS 15 and they request Xcode 15a240d:

```
Error: Your Xcode (15.0 => /opt/flutter/xcode/15a240d/XCode.app/Contents/Developer) at /Applications/Xcode.app is too outdated.
Please update to Xcode 26.3 (or delete it).
```

Move both to 17c52 (Xcode 26.2), the version every other Flutter builder uses, so mac-amd64 and mac-arm64 are built with the same toolchain. This also keeps them aligned with the Intel Xcode pin in https://flutter-review.googlesource.com/c/recipes/+/83760, which would otherwise change only the amd64 builder.

Examples: https://ci.chromium.org/b/8670377120581308353 (Mac ruby), https://ci.chromium.org/b/8670377021801718401 (Mac_arm64 ruby).

### What this does and does not fix

Homebrew's `fatal_build_from_source_checks` are fatal (exit 1) for `brew reinstall --build-from-source`, which `cipd_packages/ruby/tools/build.sh` uses; on Intel Homebrew no longer ships bottles at all, so even `brew install curl` builds from source. This PR clears one of those checks:

- [x] `check_xcode_minimum_version` (Xcode >= 16.0 on macOS 15) — fixed here.
- [ ] `check_if_xcode_needs_clt_installed` — `/Library/Developer/CommandLineTools` is not present on the devicelab hosts (`Error: Xcode alone is not sufficient on Sequoia`). Host provisioning, not fixable from `.ci.yaml`. Tracked in https://github.com/flutter/flutter/issues/162344.
- [ ] Once ruby compiles, `pod --version` failed through mid-2025 with `uninitialized constant ActiveSupport::LoggerThreadSafeLevel::Logger` (cocoapods 1.16.2 + activesupport 7.0.8). Tracked in https://github.com/flutter/flutter/issues/162341. Likely to reappear.

So these builders will stay red after this lands; this is a prerequisite, not the fix. Umbrella: https://github.com/flutter/flutter/issues/164665.

Note on `os: "Mac-15.5|Mac-15.7"`: 17c52 requires macOS 15.6, but there are currently 0 macOS 15.5 bots in `luci.flutter.prod` (150 Mac bots, all 15.7.x), so the dimension is left as is.

Consumers are unaffected either way: every `ruby` dependency in flutter/flutter (80) and flutter/packages (1) pins CIPD ref `ruby_3.1-pod_1.13` (registered 2023-10-23); these builders only re-register `latest`.


okorohelijah:
Proposal: rather than land this, we could delete both ruby targets (and `cipd_packages/ruby`) instead.

This PR only clears the first of three blockers (see description). Getting these builders green also needs Command Line Tools on the devicelab image (flutter/flutter#162344) and a CocoaPods/ActiveSupport dependency fix in the build script (flutter/flutter#162341). That's real work across two teams for a package that, as far as I can tell, has no remaining reason to exist:

- **CocoaPods is being deprecated.** SwiftPM is the default now, and flutter/flutter#168015 tracks dropping CocoaPods support entirely; flutter/flutter#183649 is already removing it from integration tests. The only purpose of `flutter/ruby/*` is to ship a Ruby with CocoaPods preinstalled.
- **Nobody consumes new builds.** All 80 `ruby` dependencies in flutter/flutter and the 1 in flutter/packages pin `ruby_3.1-pod_1.13`, an immutable CIPD ref from 2023-10-23. These builders only re-register `latest`, which nothing uses; they last succeeded ~Jan 2025 (amd64) / ~Sep 2024 (arm64) and no one noticed.
- **Intel is Tier 3 in Homebrew** (no bottles since 2025-08), so `Mac ruby` builds everything from source and will only get more fragile; Intel hosts are being retired anyway.

@vashworth wdyt?

vashworth (approved):

vashworth:
> Proposal: rather than land this, we could delete both ruby targets (and `cipd_packages/ruby`) instead.
> 
> This PR only clears the first of three blockers (see description). Getting these builders green also needs Command Line Tools on the devicelab image ([flutter/flutter#162344](https://github.com/flutter/flutter/issues/162344)) and a CocoaPods/ActiveSupport dependency fix in the build script ([flutter/flutter#162341](https://github.com/flutter/flutter/issues/162341)). That's real work across two teams for a package that, as far as I can tell, has no remaining reason to exist:
> 
> * **CocoaPods is being deprecated.** SwiftPM is the default now, and [[SwiftPM] Deprecate CocoaPods flutter#168015](https://github.com/flutter/flutter/issues/168015) tracks dropping CocoaPods support entirely; [Remove CocoaPods from integration tests flutter#183649](https://github.com/flutter/flutter/issues/183649) is already removing it from integration tests. The only purpose of `flutter/ruby/*` is to ship a Ruby with CocoaPods preinstalled.
> * **Nobody consumes new builds.** All 80 `ruby` dependencies in flutter/flutter and the 1 in flutter/packages pin `ruby_3.1-pod_1.13`, an immutable CIPD ref from 2023-10-23. These builders only re-register `latest`, which nothing uses; they last succeeded ~Jan 2025 (amd64) / ~Sep 2024 (arm64) and no one noticed.
> * **Intel is Tier 3 in Homebrew** (no bottles since 2025-08), so `Mac ruby` builds everything from source and will only get more fragile; Intel hosts are being retired anyway.
> 
> @vashworth wdyt?

Technically we still use CocoaPods in CI so we may have future use for these builders. Currently, they fail consistently and have for a while (they are in bringup), but I don't want to be blocked by figuring out the ramifications of removing these entirely. Let's file an issue to track removing these and we can follow up when appropriate.

